PR #10, opened 2016-03-04
First cut at a more idiomatic scala fastq reader.

tfenne:
@nh13 This isn't quite read for prime time, because it won't automatically handle `.gz` fastq file which is obviously needed.  We either need to implement `.gz` auto-handling in dagr-commons `Io`, or do something different in fgbio.  Thoughts?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
@@ -1,0 +1,112 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.fulcrumgenomics.fastq
+
+import java.io._
+
+import com.fulcrumgenomics.util.Io
+import dagr.commons.CommonsDef.{PathToFastq, yieldAndThen}
+
+import scala.io.Source
+
+/**
+  * Provides factory methods for creating a `FastqSource` from multiple types.
+  */
+object FastqSource {
+  /** Creates a new fastq source from a sequence of lines. */
+  def apply(lines: Seq[String]): FastqSource = new FastqSource(lines.iterator)
+
+  /** Creates a new fastq source from an iterator of lines. */
+  def apply(lines: Iterator[String]): FastqSource = new FastqSource(lines)
+
+  /** Creates a new fastq source from an input stream. */
+  def apply(stream: InputStream): FastqSource = new FastqSource(Source.fromInputStream(stream).getLines(), Some(stream))
+
+  /** Creates a new fastq source from an input stream. */
+  def apply(source: Source): FastqSource = new FastqSource(source.getLines(), Some(source))
+
+  /** Creates a new fastq source from a File. */
+  def apply(file: File): FastqSource = apply(path=file.toPath)
+
+  /** Creates a new fastq source from a Path. */
+  def apply(path: PathToFastq): FastqSource = apply(Io.toInputStream(path))
+}
+
+
+/**
+  * Reads fastq records from any text based source via a reader. Ensures that lines come in
+  * the expected groupings of four lines with the correct headers, and that bases and qualities
+  * are of the same length.  The underlying reader is closed automatically when EOF is reached.
+  */
+class FastqSource private(val lines: Iterator[String],
+                          private[this] val source: Option[{ def close(): Unit }] = None)
+  extends Iterator[FastqRecord] with Closeable {
+  
+  private var nextRecord: Option[FastqRecord] = fetchNextRecord()
+
+  /** True if calling `next()` will yield another record, false otherwise. */
+  override def hasNext: Boolean = this.nextRecord.isDefined
+
+  /** Returns the next record if available, or throws an exception if none is available. */
+  override def next(): FastqRecord =  yieldAndThen(nextRecord.get) { this.nextRecord = fetchNextRecord() }
+
+  /** Short hand to throw an illegal state exception. */
+  private def illegal(msg: String): Nothing = throw new IllegalStateException(msg)
+
+  /** Sets the current record to None, and then attempts to read the next record from the input. */
+  private def fetchNextRecord() : Option[FastqRecord] = {
+    this.lines.take(4).toList match {
+      case Nil =>
+        None
+      case header :: seq :: qheader :: quals :: Nil =>
+        if (!header.startsWith("@"))    illegal(s"Fastq sequence header must start with @: ${header}")
+        if (!qheader.startsWith("+"))   illegal(s"Fastq quality header must start with +: ${qheader}")
+        if (seq.length != quals.length) illegal(s"Sequence and qualities not same length for record: ${header}")
+
+        // Destructure the header line into name, read number and comment
+        val (fullName, comment) = header.indexWhere(_.isWhitespace) match {
+          case -1 => (header.drop(1), None)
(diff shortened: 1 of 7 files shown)


src/test/scala/com/fulcrumgenomics/util/IoTest.scala
```diff
@@ -0,0 +1,57 @@
...
+  "Io" should "round trip data to a gzipped file if it ends with .gz" in {
+    val text = "This is a stupid little text fragment for compression. Yay compression!"
+    val in   = Seq(text, text, text)
+    val f = makeTempFile("test", ".gz")
+    Io.writeLines(f, in)
+    val out = Io.readLines(f).toSeq
+    out shouldBe in
+  }
+
+
```
tfenne: @nh13: I'm saving this space for more tests :wink:

tfenne:
Will require https://github.com/fulcrumgenomics/dagr/pull/129 to be merged first.

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -0,0 +1,113 @@
...
+  override def hasNext: Boolean = this.nextRecord.isDefined
+
+  /** Returns the next record if available, or throws an exception if none is available. */
+  override def next(): FastqRecord =  yieldAndThen(nextRecord.get) { queueNextRecord() }
+
+  /** Short hand to throw an illegal state exception. */
+  private def illegal(msg: String): Nothing = throw new IllegalStateException(msg)
+
+  /** Sets the current record to None, and then attempts to read the next record from the input. */
+  private def queueNextRecord() : Unit = {
```
nh13: why not have `queueNextRecord` return `Option[FastqRecord]` so you can have `private var nextRecord: Option[FastqRecord]` above ?

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private var nextRecord: Option[FastqRecord] = None
+  queueNextRecord()
+
+  /** True if calling `next()` will yield another record, false otherwise. */
+  override def hasNext: Boolean = this.nextRecord.isDefined
+
+  /** Returns the next record if available, or throws an exception if none is available. */
+  override def next(): FastqRecord =  yieldAndThen(nextRecord.get) { queueNextRecord() }
```
nh13: should you match on `nextRecord` to throw a `NoSuchElementException`?

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  /** Short hand to throw an illegal state exception. */
+  private def illegal(msg: String): Nothing = throw new IllegalStateException(msg)
+
+  /** Sets the current record to None, and then attempts to read the next record from the input. */
+  private def queueNextRecord() : Unit = {
+    this.nextRecord = None
+    val header  = reader.readLine()
+    val seq     = reader.readLine()
+    val qHeader = reader.readLine()
```
nh13: it's a comment, not a header.

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -0,0 +1,113 @@
...
+    val quals   = reader.readLine()
+
+    if (header == null) {
+      this.reader.close()
+    }
+    else {
+      // Assert that we read four lines or no lines
+      if (header != null && (seq == null || qHeader == null || quals == null)) {
```
nh13: this may be a good case for using `null`, but really, scala tries to avoid it. Any suggestions on alternatives to having `null`?

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -0,0 +1,113 @@
...
+        illegal(s"Fastq source terminated mid-record at ${header}")
+      }
+
+      if (!header.startsWith("@"))  illegal(s"Fastq sequence header must start with @: ${header}")
+      if (!qHeader.startsWith("+")) illegal(s"Fastq quality header must start with @: ${qHeader}")
+      if (seq.length != quals.length) illegal(s"Sequence and qualities not same length for record: ${header}")
```
nh13: align the `illegal` or don't

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
nh13: I see what you did below, that's fine, perhaps we should check if it matches the description in the header?

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -0,0 +1,113 @@
...
+
+      // Destructure the header line into name, read number and comment
+      val (fullName, comment) = header.indexWhere(_.isWhitespace) match {
```
nh13: why not just 
```
header.split("whitespaceregex", 2) match { ...
```

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -0,0 +1,113 @@
...
+        case -1 => (header.drop(1), None)
+        case i  => (header.substring(1, i), Some(header.substring(i+1)))
+      }
+
+      val (name, readNumber) = fullName.endsWith("/1") || fullName.endsWith("/2") match {
+        case true  => (fullName.substring(0, fullName.length-2), Some(fullName.last.asDigit))
```
nh13: `fullName.substring(0, fullName.length-2)` -> `fullName.dropRight(2)`

nh13: This one may be less clear, but `fullName.last.asDigit` -> `fullName.takeRight(1).asDigit`

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -0,0 +1,113 @@
...
+        case false => (fullName, None)
+      }
+
+      this.nextRecord = Some(FastqRecord(
+        name = name,
+        bases = seq,
+        quals = quals,
+        comment = comment,
+        readNumber = readNumber
```
nh13: needs more alignment

src/main/scala/com/fulcrumgenomics/util/Io.scala
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  /** Adds the automatic handling of gzipped files when opening files for writing. */
+  override def toOutputStream(path: Path): OutputStream = {
+    PathUtil.extensionOf(path) match {
+      case Some(".gz") => new GZIPOutputStream(Files.newOutputStream(path), BufferSize)
+      case _           => super.toOutputStream(path)
+    }
+  }
+
```
nh13: image

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -0,0 +1,113 @@
...
+      ))
+    }
+  }
```
nh13: should we have any validation that the bases are bases (if we assume fastqs only handle bases), and that the base string is the same length as the quality string (not so true for SOLiD, but hey)?

nh13: Also, should we allow empty bases?

src/test/scala/com/fulcrumgenomics/testing/UnitSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+import java.nio.file.{Path, Files}
+
+import org.scalatest.{FlatSpec, Matchers}
+
+/** Base class for unit and integration testing */
+trait UnitSpec extends FlatSpec with Matchers {
+  /** Creates a new temp file for use in testing that will be deleted when the VM exits. */
+  protected def makeTempFile(prefix: String, suffix: String) : Path = {
+    val path = Files.createTempFile(prefix, suffix)
+    path.toFile.deleteOnExit()
```
nh13: `yieldAndThen`?

src/test/scala/com/fulcrumgenomics/util/IoTest.scala
nh13: I was about to ask

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
tfenne: `nextRecord.get` will throw a `NoSuchElementException` if it's `None`, so I don't really see the point.  There is a test that asserts calling `next` when there is no `next` throws the right exception, so we'll catch it if we accidentally change it at some point.

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
tfenne: I think the only suggestion is to write out own wrapper around BufferedReader for doing IO that internally catches the `null` and wraps it up in an `Option` or perhaps an `Either[String,EOF]` etc.  But I really don't want to take that on right now.

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
tfenne: I did think about that, but decided that the above would work and should be significantly more efficient that relying on a regex.

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
tfenne: Good questions - I'm not sure!  I guess I would rather leave those out for now as you never know what kind of junk you'll find in a fastq file that you might need to deal with, and add more validations in later if we see common errors that should be rejected.

tfenne:
@nh13 I've addressed many of your comments, and taken a shot at providing a `null`-free implementation.  Take a look and see what you think.

Before too long I think we should look very closely at how we do file IO in `fgbio` and agree on one _right_ way to do things.  It seems complicated though as using `java.io` directly is nasty, using `java.nio` directly is better but still non-idiomatic from scala, and using `scala.io.Source` seems to be somewhat frowned upon as it's not the greatest.

PR #12, opened 2016-03-21
Added a FastqWriter and TrimFastq...

tfenne:
...a program to trim sets of fastq file to a given length and optionally discard any beyond that length.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/TrimFastq.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/TrimFastq.scala
@@ -1,0 +1,81 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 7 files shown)


src/main/scala/com/fulcrumgenomics/fastq/FastqRecord.scala
```diff
@@ -26,4 +26,20 @@ package com.fulcrumgenomics.fastq
 /**
   * Represents a record that can be read from or written to a fastq file.
   */
-case class FastqRecord(name: String, bases: String, quals: String, comment: Option[String], readNumber: Option[Int])
+case class FastqRecord(name: String, bases: String, quals: String, comment: Option[String], readNumber: Option[Int]) {
+  /** Constructs the header line from the name, read number and comment. */
+  def header: String = name + readNumber.map("/" + _).getOrElse("") + comment.map(" " + _).getOrElse("")
```
nh13: `readNumber.map("/" + _).getOrElse("") ` could be `readNumber.fold("") { "/" + _ }`; I am not sure which I like better

src/main/scala/com/fulcrumgenomics/fastq/TrimFastq.scala
```diff
@@ -0,0 +1,80 @@
...
+)
+class TrimFastq
+(   @arg(flag="i", doc="One or more input fastq files.")       val input:  Seq[PathToFastq],
+    @arg(flag="o", doc="A matching number of outpuf files.")   val output: Seq[PathToFastq],
+    @arg(flag="l", doc="Length to trim reads to.")             val length: Int,
+    @arg(flag="x", doc="Exclude reads below the trim length.") val exclude: Boolean = false
+) extends FgBioTool with LazyLogging {
+
+  override def execute(): Unit = {
+    assert(input.size == output.size, "Number of input and output files must match.")
```
nh13: I am curious why this is in the `execute` method and not the constructor?

tfenne: Habit.  I changed it.

PR #14, opened 2016-03-25
Add a tool to annotate reads with their molecular barcode(s).

nh13:


nh13:
@tfenne so you are aware, I have a local branch of the `DemuxFastqs` PR that uses the `ReadStructure` code introduced here, and that is why there is more implemented in `ReadStructure` than what is needed by `AnnotateMolecularBarcodes`.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+import dagr.commons.io.Io
+import dagr.commons.util.LazyLogging
+import dagr.sopt._
+import dagr.sopt.cmdline.ValidationException
+import htsjdk.samtools._
+import htsjdk.samtools.util._
+
+import scala.collection.JavaConversions._
+
+@clp(description =
```
tfenne: I think we need to do two things:
1) Give this a better name
2) Make it clear how this is different from `AnnotateBamWithUmis`

On the naming front I would suggest:
a) We use `UMI` consistently as our term, instead of Molecular Barcode, Molecular Index, etc. so that it's clear we're talking about the same thing
b) Using a word like `Extract` or a synonym to indicate that the UMIs are being extracted from the reads in the BAM, not an external source

Perhaps calling it `ExtractUmisFromReads` or similar would be better?

On 2, I think it would be good to make the distinction in the first line of the documentation in both tools.  For this one perhaps we write `Extracts unique molecular indices from reads in a BAM file into tags.`, and edit `AnnotateBamWithUmis` to read `Annotates existing BAM files with UMIs (Unique Molecular Indices, aka Molecular IDs, Molecular barcodes) from a separate FASTQ file.`

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+    |structure.  If only one molecular barcode tag is given, then the molecular barcodes will be concatenated and stored
+    |in that tag.  Otherwise the number of molecular barcodes should match the number of tags.  Mapping information will
+    |be unchanged, as such, this tool should not be used on reads that have been mapped since it would lead to an BAM
+    |with inconsistent records.
+  """,
+  group = ClpGroups.SamOrBam)
+class AnnotateMolecularBarcodes
+( @arg(flag = "i",   doc = "Input BAM file.")                                     val input: PathToBam,
+  @arg(flag = "o",   doc = "Output BAM file.")                                    val output: PathToBam,
+  @arg(flag = "rs1", doc = "The read structure for read 1.")                       val readStructure1: String,
```
tfenne: Does this work?  I thought sopt only supported single letter flags?

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+  @arg(flag = "rs2", doc = "The read structure for read 2.")                       val readStructure2: String,
+  @arg(flag = "b1",  doc = "SAM tags in which to store the molecular barcodes (one-per segment) for read.")
```
tfenne: I personally think it would be nicer to supply just a single list of tags, with the understanding that UMIs are extracted from R1 first then R2 and matched to tags in that order.  Then you could also just call this `tags`.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+                                                                                  val molecularBarcodeTags1: Seq[String] = Seq.empty,
+  @arg(flag = "b2",  doc = "SAM tags in which to store the molecular barcodes (one-per segment).")
+                                                                                  val molecularBarcodeTags2: Seq[String] = Seq.empty,
+  @arg(flag = "a",   doc = "Annotate the read names with the molecular barcodes.") val annotateReadNames: Boolean = true,
```
tfenne: I think this needs more explanation in the doc.  How are the read names annotated?  Are multiple UMI's concatenated directly, with a separator.  Is anything else done to the read names?

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+  @arg(flag = "t",   doc = "Annotate both reads from a read pair with each others tags.")
+                                                                                  val tagBothEnds: Boolean = true
+) extends FgBioTool with LazyLogging {
+
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+  molecularBarcodeTags1.foreach(tag => if (tag.length != 2) throw new ValidationException("SAM tags must be of length two: " + tag))
+  molecularBarcodeTags2.foreach(tag => if (tag.length != 2) throw new ValidationException("SAM tags must be of length two: " + tag))
+  if (readStructure1.isEmpty) throw new ValidationException("readStructure1 may not be empty")
+  if (readStructure2.isEmpty) throw new ValidationException("readStructure2 may not be empty")
```
tfenne: What if it's an SE bam?

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+  if (tagBothEnds) {
+    molecularBarcodeTags1.foreach { tag => assert(!molecularBarcodeTags2.contains(tag)) }
+    molecularBarcodeTags2.foreach { tag => assert(!molecularBarcodeTags1.contains(tag)) }
+  }
+
+  override def execute(): Unit = {
+    val rs1 = ReadStructure(readStructure1)
+    val rs2 = ReadStructure(readStructure2)
+    assert(molecularBarcodeTags1.length == rs1.count {
```
tfenne: Does this actually work without the `match` keyword?

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+      case _ => false
+    })
+    assert(molecularBarcodeTags2.length == rs2.count {
+      case m: MolecularBarcode => true
+      case _ => false
+    })
+    val in: SamReader = SamReaderFactory.make.open(input.toFile)
+    val iterator: SAMRecordIterator = in.iterator()
+    val out: SAMFileWriter = new SAMFileWriterFactory().makeSAMOrBAMWriter(in.getFileHeader, true, output.toFile)
+    iterator.sliding(2,2).foreach {
```
tfenne: Ok, so really not supporting non-paired reads.  That should be made clear in the documentation.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+      case Seq(r1: SAMRecord, r2: SAMRecord) =>
+        assert(r1.getReadName.compareTo(r2.getReadName) == 0)
```
tfenne: Please add a message to the assert call so user's will know what the problem is without looking at the code.

tfenne: Probably also should assert that `r1.getFirstOfPairFlag()` and `r2.getSecondOfPairFlag()`

tfenne: I think also it should assert that all the reads coming through are unmapped, since this will mangle things horribly if they are mapped and the tool states that mapped reads are not supported.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+        val bases1 = annotateRecord(record=r1, readStructure=rs1, molecularBarcodeTags=molecularBarcodeTags1)
+        val bases2 = annotateRecord(record=r2, readStructure=rs2, molecularBarcodeTags=molecularBarcodeTags2)
+        if (annotateReadNames) {
+          r1.setReadName(r1.getReadName + "|" + bases1 + bases2)
+          r2.setReadName(r2.getReadName + "|" + bases1 + bases2)
+        }
+        if (tagBothEnds) {
```
tfenne: My 2c is that I can't think of a time when I would want to tag the two ends independently (i.e. `tagBothEnds == false`), and I think this option makes it all more complicated.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+  /**
+    * Extracts bases associated with molecular barcodes and adds them as SAM tags, and optionally concatenates
+    * them to the read name.  The read's bases will only contain template bases as defined in the given read structure.
+    */
+  private def annotateRecord(record: SAMRecord,
+                             readStructure: ReadStructure,
+                             molecularBarcodeTags: Seq[String]): String = {
+    val bases = record.getReadString
+    val qualities = record.getBaseQualityString
+    assert(bases.length == readStructure.totalBases)
```
tfenne: Won't this fail on any run that's had adapter trimming performed and thus has reads of variable length?

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+    // get the bases associated with each segment
+    val readStructureBases = ReadStructure.extractWithQualities(bases, qualities, readStructure)
+    // get the molecular barcode segments
+    val molecularBarcodeBases = readStructureBases.flatMap {
```
tfenne: I feel like you ought to be able to write `readStructureBases.filter(_.kind == MolecularBarcode).map(_.bases)` here.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
```diff
@@ -0,0 +1,148 @@
...
+    }
+    out.close()
+    CloserUtil.close(iterator)
+  }
+
+  /**
+    * Extracts bases associated with molecular barcodes and adds them as SAM tags, and optionally concatenates
+    * them to the read name.  The read's bases will only contain template bases as defined in the given read structure.
+    */
+  private def annotateRecord(record: SAMRecord,
```
tfenne: I find it kind of icky that this method is very side-effecty on the `record` (even to the point of using instance state to determine side effects) and _also_ returns the concatenated molecular barcodes.  Not sure what to suggest instead though.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+ */
+
+package com.fulcrumgenomics.util
+
+import com.fulcrumgenomics.util.BaseSymbol._
+import com.fulcrumgenomics.util.ReadSegment._
+
+object ReadStructure {
+  private val segmentPattern = s"\\d+[${SymbolChars.mkString}]".r
+  private val readStructurePattern = s"(\\d+[${SymbolChars.mkString}])+".r
```
tfenne: Two thoughts:
1) You could just add `^` and `$` to the start and end of your read structure pattern to make it exhaustive on the input string, which I think would be clearer.
2) Using regexes here means that the error messaging is very coarse (i.e. "invalid read structure") without being able to say what was invalid (lower case letter?  wrong letter?  something else?).

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+        val length = lengths(idx)
+        if (length == 0) throw new IllegalArgumentException("Read structure had a segment with zero length: " + seg)
+        seg.substring(seg.length-1).head match {
+          case TemplateSymbol.symbol         => new Template(offset=offset, length=length)
+          case SampleBarcodeSymbol.symbol    => new SampleBarcode(offset=offset, length=length)
+          case MolecularBarcodeSymbol.symbol => new MolecularBarcode(offset=offset, length=length)
+          case SkipSymbol.symbol             => new Skip(offset=offset, length=length)
+          case s                             => throw new IllegalArgumentException("Read segment type unknown: " + s)
+        }
+    }.toTraversable)
```
tfenne: Traversable does not require ordering, and I believe this should.  `Seq`?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+  private val readStructurePattern = s"(\\d+[${SymbolChars.mkString}])+".r
+
+  /** Creates a read structure from a string */
+  def apply(readStructure: String): ReadStructure = {
+    if (readStructurePattern.findAllIn(readStructure).mkString.length != readStructure.length) {
+      throw new IllegalArgumentException("Read structure was not valid: " + readStructure)
+    }
+    val segments = segmentPattern.findAllIn(readStructure).toSeq
+    val lengths = segments.map(seg => Integer.parseInt(seg.substring(0, seg.length-1)))
+    new ReadStructure(segs = segments.zipWithIndex.map {
```
tfenne: I think this whole section could be made more readable by doing the splitting from a string once, and then doing a single map.  Something like:
```scala
segmentPattern.findAllIn(readStructure)
  .map (str => str.splitAt(str.length-1))
  .zipWithIndex
  .map { case ((len, kind), idx) => (len.toInt, kind.charAt(0), idx) }
  .map { 
    case (0, _, _) => throw new IllegalArgumentException("Read structure had a segment with zero length: " + seg)
    case (n, TemplateSymbol.symbol, idx) => ....
    ....
  }
```

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+      case (seg, idx) =>
+        val offset = lengths.slice(0, idx).sum
+        val length = lengths(idx)
+        if (length == 0) throw new IllegalArgumentException("Read structure had a segment with zero length: " + seg)
+        seg.substring(seg.length-1).head match {
+          case TemplateSymbol.symbol         => new Template(offset=offset, length=length)
+          case SampleBarcodeSymbol.symbol    => new SampleBarcode(offset=offset, length=length)
+          case MolecularBarcodeSymbol.symbol => new MolecularBarcode(offset=offset, length=length)
+          case SkipSymbol.symbol             => new Skip(offset=offset, length=length)
+          case s                             => throw new IllegalArgumentException("Read segment type unknown: " + s)
```
tfenne: `unreachable`?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }.toTraversable)
+  }
+
+  /** Splits the given bases into tuples with its associated read segment. */
+  def extract(bases: String, readStructure: ReadStructure, symbol: Option[BaseSymbol] = None, strict: Boolean = true): Seq[(String, ReadSegment)] = {
```
tfenne: Why isn't this an instance method on the class?  I think that object methods that take in a single instance of the companion class are largely an anti-pattern.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+  def molecularBarcode(rs: ReadStructure): Seq[MolecularBarcode] = rs.filter(_.symbol == MolecularBarcodeSymbol).map(_.asInstanceOf[MolecularBarcode])
+  def skip(rs: ReadStructure): Seq[Skip] = rs.filter(_.symbol == SkipSymbol).map(_.asInstanceOf[Skip])
+}
+
+/**
+  * Describes the structure of a give read.  A read contains one or more read segments. A read segment describes
+  * a contiguous stretch of bases of the same type (ex. template bases) of some length and some offset from the start
+  * of the read.
+  */
+class ReadStructure private(segs: Traversable[ReadSegment]) extends scala.collection.immutable.Seq[ReadSegment] {
```
tfenne: Why not just extend `immutable.Seq`?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.util
+
+import com.fulcrumgenomics.util.BaseSymbol._
+import com.fulcrumgenomics.util.ReadSegment._
+
+object ReadStructure {
```
tfenne: I think we really need to support having a final segment of indeterminate length.  Probably it would be good to have support for `*` and `+` to indicate zero or more and one or more bases.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
tfenne: I'm not sure we should ever allow unstrict in the sense you have it here - I think all reads should observe the _start_ of all segments.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+        if (bases.length < end) {
+          if (strict) throw new IllegalStateException("Not enough bases for read structure: " + readStructure)
+          else (bases.substring(segment.offset), ReadSegment(segment, bases.length-segment.offset))
+        }
+        else (bases.substring(segment.offset, end), segment)
+    }
+  }
+
+  /** Splits the given bases and qualities into triples with its associated read segment. */
+  def extractWithQualities(bases: String, qualities: String, readStructure: ReadStructure, symbol: Option[BaseSymbol] = None, strict: Boolean = true): Seq[(String, String, ReadSegment)] = {
```
tfenne: Ditto on moving this into the class, and removing the `strict` option.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+  private val segments: Seq[ReadSegment] = segs.toSeq
+  def length: Int = segments.length
+  def apply(idx: Int): ReadSegment = segments(idx)
+  def iterator: Iterator[ReadSegment] = segments.iterator
+  override def toString: String = segments.map(_.toString).mkString
+  def totalBases: Int = segments.map(_.length).sum
+}
+
+object BaseSymbol {
+  sealed trait BaseSymbol {
```
tfenne: I think it's weird to have a trait nested inside an object of the same name.  E.g. I would call the object `BaseSymbols` in this case.

But also, I think the name is kind of weird.  These are really describing the types of kinds of read segments.  They _contain_ the symbol used for that type of segment.  And other than being single characters don't, I think, resemble bases.  How about `SegmentType` instead?  And then I would suggest the case objects could be just `Template`, `SampleBarcode`, `MolecularBarcode`, `Skip` and reference them elsewhere as `SegmentType.Template` etc.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+    def symbol: Char
+    override def toString: String = s"$symbol"
+  }
+  case object TemplateSymbol         extends BaseSymbol { override val symbol: Char = 'T' }
+  case object SampleBarcodeSymbol    extends BaseSymbol { override val symbol: Char = 'B' }
+  case object MolecularBarcodeSymbol extends BaseSymbol { override val symbol: Char = 'M' }
+  case object SkipSymbol             extends BaseSymbol { override val symbol: Char = 'S' }
+  val Symbols = Seq(TemplateSymbol, SampleBarcodeSymbol, MolecularBarcodeSymbol, SkipSymbol)
+  val SymbolChars = Seq(TemplateSymbol, SampleBarcodeSymbol, MolecularBarcodeSymbol, SkipSymbol).map(_.symbol)
+}
```
tfenne: Might be nice to provide `def apply(s: String): SegmentType` and `def apply(ch: Char): SegmentType` method to resolve the segment types from symbols without having to write `match` statements in client code.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,143 @@
...
+
+object ReadSegment {
+
+  /** Base for all types of segments of a read */
+  sealed abstract class ReadSegment(val offset: Int, val length: Int, val symbol: BaseSymbol) {
+    override def toString: String = s"$length$symbol"
+  }
+
+  /** Bases from the template */
+  case class Template(override val offset: Int, override val length: Int) extends ReadSegment(offset = offset, length = length, symbol = TemplateSymbol)
```
tfenne: Now I'm confused.  I'm struggling to see the value in having parallel hierarchies (`BaseSymbol` and `ReadSegment`) that must be synchronized.  Why not just do one or the other?

tfenne:
Lots of comments on the read structure stuff.  I think it's a good start, but I think we should spend the time to make it really nice before merging since I feel like we'll use it a lot quite quickly and make refactoring of it later hard.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
nh13: Huh, I think it does "support" flags being multi-characters.  Lets discuss in person if this is what we want (the README in sopt says no).

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
nh13: Currently only PE bams are supported.  I would like to defer SE support until it is needed.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
nh13: It does, actually, since `count` is a method not a value.  `count` takes a function from `ReadSegment` to `boolean` here.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
nh13: Will do.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
nh13: Someone has to modify the sam record.  What instance state determines side effects here?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: 2) can always be improved since the parsing is not exposed beyond the coarse error message.  I'll think about it.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: Close, I see what you are saying though.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: 1. Unstrict would allow you to pass in 50bp read and a read structure of `75T50T`.  The second segment is filtered out in this case.  Strict would throw an exception.
2. Unstrict would allow you to pass in 50bp read and a read structure of `75T`.  Strict would throw an exception.
I think you object to (1) but not (2).  

If you think about each read from a pair independently I agree with you, since I think in all cases the last segment would be the template bases.  For index reads I agree too.  But for a read structure that describes the pair jointly (see `IlluminaBasecallsToSam`) then this may not be true.  Consider if the second read of the pair fails completely or is completely trimmed.  We may never use `ReadStructure` across multiple reads (ex. for a whole template) in `fgbio`, and if so, I agree with you and we should document that clearly.  What's your thought?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: so just put `import scala.collection.immutable` up top and `immutable.Seq[ReadSegment]` here?  What does that buy you but an extra line?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: Besides special values for `length` in `ReadSegment`, how would you suggest storing this?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: `ReadSegment` stores the number of bases in the segment (`length`),  the offset from the start of the read structure (`offset`), and the symbol use to represent it. `BaseSymbol` (now `SegmentType`) stores the character that represents a `ReadSegment`.  The reasons for their separation are:
1. I can access `SegmentType.symbol` from a static context.  This is useful from converting a string to a read structure.  I need a way to build a specific segment type from a character.  I suppose I can a generic apply method that takes in a character.
2. If I have a `Seq` of `ReadSegment`s, I can filter them by `==` on their symbol (which is a case class of `SegmentType`). I do not want client code to have `'T'` hardcoded everywhere, but have the ability to use `==` and match statements too. I suppose I could use `collect { case t: Template => t }` instead.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: I have a simpler method that can provide better error messages. I'll push here once we get the rest of the questions squared away.

nh13:
There are a few remaining questions in the previous review.  The outstanding ones added them as TODOs in the current commit.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
tfenne: Ah, cool.

src/main/scala/com/fulcrumgenomics/bam/AnnotateMolecularBarcodes.scala
tfenne: Agreed someone has to modify - it's the modify _and_ return something else that I'm objecting to.  I might be wrong on instance state since I can no longer see what I was talking about either ;)

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
tfenne: Sorry - I misspoke/typed.  What I really meant was why not have `ReadStructure` extend something that provides a full implementation of `Seq` (e.g. `List`) instead of having an internal `Seq[ReadSegment]`.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
tfenne: I'm not sure tbh.  Maybe length becomes a case class that is either `Exactly(Int)`, `ZeroOrMore` or `OneOrMore`?

src/main/scala/com/fulcrumgenomics/bam/ExtractUmisFromBam.scala
```diff
@@ -0,0 +1,154 @@
...
+@clp(description =
+  """
+    |Extracts unique molecular indices from reads in a BAM file into tags.
+    |
+    |Currently only paired end read data are supported.
+    |
+    |Only template bases will be retained as read bases (stored in the SEQ field) as specified by the read structure.
+    |
+    |For a given read, if only one molecular barcode tag is given, then the molecular barcodes will be concatenated and
+    |stored in that tag.  Otherwise the number of molecular barcodes should match the number of tags.  Each  end of a
```
tfenne: Extra space between "Each" and "end".

src/main/scala/com/fulcrumgenomics/bam/ExtractUmisFromBam.scala
```diff
@@ -0,0 +1,154 @@
...
+    |pair will store the molecular barcode(s) for both pairs.  In this case, the tags for the molecular barcodes should
```
tfenne: `pairs` -> `ends`?

src/main/scala/com/fulcrumgenomics/bam/ExtractUmisFromBam.scala
```diff
@@ -0,0 +1,154 @@
...
+  """,
+  group = ClpGroups.SamOrBam)
+class ExtractUmisFromBam
+( @arg(flag = "i", doc = "Input BAM file.")                                      val input: PathToBam,
+  @arg(flag = "o", doc = "Output BAM file.")                                     val output: PathToBam,
+  @arg(flag = "r", doc = "The read structure for read 1.")                       val readStructure1: String,
+  @arg(flag = "R", doc = "The read structure for read 2.")                       val readStructure2: String,
+  @arg(flag = "b", doc = "SAM tags in which to store the molecular barcodes (one-per segment) for read 1.")
+                                                                                 val molecularBarcodeTags1: Seq[String] = Seq.empty,
+  @arg(flag = "B", doc = "SAM tags in which to store the molecular barcodes (one-per segment) for read 2.")
```
tfenne: I still don't love having two arguments to specify the set of tags to use.  I (personally) think:
`ExtractUmisFromBam ... -r 8M75T -R 8M75T -b ZA ZB`
is more natural.

src/main/scala/com/fulcrumgenomics/bam/ExtractUmisFromBam.scala
```diff
@@ -0,0 +1,154 @@
...
+    val iterator: SAMRecordIterator = in.iterator()
+    val out: SAMFileWriter = new SAMFileWriterFactory().makeSAMOrBAMWriter(in.getFileHeader, true, output.toFile)
+    iterator.sliding(2,2).foreach {
+      case Seq(r1: SAMRecord, r2: SAMRecord) =>
+        // verify everything is in order
+        Seq(r1, r2).foreach(r => {
+          if (!r.getReadPairedFlag) throw new IllegalStateException(s"Read ${r.getReadName} was not paired")
+          if (!r1.getReadUnmappedFlag) throw new IllegalStateException(s"Read ${r.getReadName} was mapped")
+        })
+        if (!r1.getFirstOfPairFlag) throw new IllegalStateException(s"Read ${r1.getReadName} was the first end of a pair")
```
tfenne: I think your error messages here and on the next line need "was" -> "was not", right?

src/main/scala/com/fulcrumgenomics/bam/ExtractUmisFromBam.scala
nh13: If it's not symmetric, which we have seen, then I don't like `ExtractUmisFromBam ... -r 8M8M75T -R 8M75T -b ZA ZB ZC` since the read structure was in two args but the tags are not.  I (personally) like: 
`ExtractUmisFromBam ... -r 8M8M75T -R 8M75T -b ZA ZB -B ZC`.

nh13:
@tfenne only thing left to resolve is how the SAM tags are specified.  I would argue that since the read structures are specified per end that the SAM tags should also be specified per end.  I think this is more intuitive.  I am interested in hearing the reason for your preference.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,167 @@
...
+package com.fulcrumgenomics.util
+
+import scala.collection.immutable
+import scala.reflect.ClassTag
+
+
+// TODO TF: I think we really need to support having a final segment of indeterminate length. Probably it would be good to have support for * and + to indicate zero or more and one or more bases
+object ReadStructure {
+  /** Creates a read structure from a string */
+  def apply(readStructure: String): ReadStructure = {
```
tfenne: FWIW I find this method very hard to follow.  I'm going to assume it just works given the tests, but a handful of comments would probably help my (and perhaps future-your) understanding.  In addition more verbose variable names might help (e.g. length => segmentLength, offset => ???).

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,167 @@
...
+  * a contiguous stretch of bases of the same type (ex. template bases) of some length and some offset from the start
+  * of the read.
+  */
+class ReadStructure(segs: Seq[ReadSegment]) extends immutable.Seq[ReadSegment] {
+  private val segments: Seq[ReadSegment] = segs
+  def length: Int = segments.length
+  def apply(idx: Int): ReadSegment = segments(idx)
+  def iterator: Iterator[ReadSegment] = segments.iterator
+  override def toString: String = segments.map(_.toString).mkString
+  def totalBases: Int = segments.map(_.length).sum
```
tfenne: lazy val?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,167 @@
...
+
+  /** Splits the given bases into tuples with its associated read segment. */
+  def extract(bases: String, strict: Boolean = true): Seq[(String, ReadSegment)] = {
```
tfenne: Coming at this fresh it took me a minute to realize this is breaking apart an actual read.  Perhaps calling it something like `structureRead` would be clearer?  A small example in the scaladoc would also be super helpful.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -0,0 +1,167 @@
...
+  def apply(segment: ReadSegment, length: Int): ReadSegment = segment match {
+    case seg: Template         => Template(seg.offset, length)
+    case seg: SampleBarcode    => SampleBarcode(seg.offset, length)
+    case seg: MolecularBarcode => MolecularBarcode(seg.offset, length)
+    case seg: Skip             => Skip(seg.offset, length)
+  }
+}
+
+/** Base for all types of segments of a read */
+sealed abstract class ReadSegment(val offset: Int, val length: Int, val symbol: Char) {
```
tfenne: Having this as an abstract class, and then having the subclasses be case classes that override the attributes feels strange - particularly if you override an attribute it's odd to then also have to pass the value to the super-class constructor.

How about:

```scala
sealed trait ReadSegment {
    def offset: Int
    def length: Int
    def symbol: Char
}
```

Then the `val`s in the case classes will fulfill the `def`s without the possible duplications?

src/test/scala/com/fulcrumgenomics/util/ReadStructureTest.scala
```diff
@@ -0,0 +1,75 @@
...
+class ReadStructureTest extends UnitSpec {
+
+  "ReadStructure" should "be built from a string" in {
+    ReadStructure("1T").toString shouldBe "1T"
+    ReadStructure("1B").toString shouldBe "1B"
+    ReadStructure("1M").toString shouldBe "1M"
+    ReadStructure("1S").toString shouldBe "1S"
+    ReadStructure("101T").toString shouldBe "101T"
+    ReadStructure("123456789T").toString shouldBe "123456789T"
+    ReadStructure("10T10B10B10S10M").toString shouldBe "10T10B10B10S10M"
```
tfenne: Would be nice on this last one to validate that each of the segments created is also the right sub-type of `ReadSegment`.

tfenne:
Back to you @nh13.  I think the one vs. two inputs for the tag names is just a matter of preference.  My tendency is to this of it in the output space - since all the tags go on to all reads from a template, I tend to think of the tags as being specified once per template.  In addition, if you want to do concatenation of multiple UMIs into one tag, I think that:
`ExtractUmisFromBam ... -r 8M75T -R 8M75T -b ZA ZA`
is clearer than:
`ExtractUmisFromBam ... -r 8M75T -R 8M75T -b ZA -B ZA`
because the second one makes me wonder if it will overwrite ZA instead of appending to it.

At the end of the day either is fine though, so do as you like.

PR #16, opened 2016-04-09
Adding a tool to update read groups.

nh13:
This differs from Picard's AddOrReplaceReadGroups since we do not add
any read groups but merely update them.  We also can also update more
than one read group.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
@@ -1,0 +1,149 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.bam
+
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.util.ProgressLogger
+import dagr.commons.CommonsDef._
+import dagr.commons.io.Io
+import dagr.commons.util.LazyLogging
+import dagr.sopt._
+import htsjdk.samtools._
+
+import scala.collection.JavaConversions._
+
+/**
+  * Updates one or more read groups.
+  *
+  * @author Nils Homer
+  */
+@clp(description =
+  """
+      |Updates one or more read groups and their identifiers.
+      |
+      |This tool will replace each read group with a new read group, including a new read group identifier.  If the read
+      |group identifier is not to be changed, it is recommended to use `samtools reheader` or Picard's
+      |`ReplaceSamHeader` instead as in this case only the header needs modification. If all read groups are to be
+      |assigned to one read group, it is recommended to use Picard's `AddOrReplaceReadGroups`.  Nonetheless, if the read
+      |group identifier also needs to be changed, use this tool.
+      |
+      |Each read group in the input file will be mapped to one and only one new read group identifier, unless
+      |`ignoreMissingReadGroups` is set.  A SAM header file should be given with the new read groups and the ID field
+      |foreach read group containing the new read group identifier.  An additional attribute ("FR") should be provided
+      |that gives the original read group identifier ("ID") to which this new read group corresponds.
+      |
+      |If `keepReadGroupAttributes` is true, then any read group attribute not replaced will be kept in the new read
+      |group.  Otherwise, only the attributes in the provided SAM header file will be used.
+    """,
+  group = ClpGroups.SamOrBam)
+class UpdateReadGroups
+(@arg(flag = "i", doc = "Input BAM file.") val input: PathToBam,
+ @arg(flag = "o", doc = "Output BAM file.") val output: PathToBam,
+ @arg(flag = "r", doc = "A SAM header file with the replacement read groups (see detailed usage).") val readGroupsFile: PathToBam,
+ @arg(flag = "k", doc = "Keep all read group attributes that are not replaced.") val keepReadGroupAttributes: Boolean = false,
+ @arg(flag = "g", doc = "Keep all read groups not found in the replacement header, otherwise throw an error.") val ignoreMissingReadGroups: Boolean = false
+) extends FgBioTool with LazyLogging {
+
+  private val fromReadGroupKey = "FR"
+
+  Io.assertReadable(input)
+  Io.assertReadable(readGroupsFile)
+  Io.assertCanWriteFile(output)
+
+  override def execute(): Unit = {
+    val progress: ProgressLogger = new ProgressLogger(logger)
+
+    val in: SamReader = SamReaderFactory.make.open(input.toFile)
+    val toReadGroupsMap = getReadGroupMap(readGroupsFile, in.getFileHeader)
+    val outHeader = in.getFileHeader.clone()
+    outHeader.setReadGroups(toReadGroupsMap.values.toList.sortBy(_.getId))
+    val out: SAMFileWriter = new SAMFileWriterFactory().makeSAMOrBAMWriter(outHeader, true, output.toFile)
+
+    // main loop
+    in.foreach { record =>
+      updateRecord(record, outHeader, toReadGroupsMap)
+      out.addAlignment(record)
+      progress.record(record)
+    }
+    logger.info(s"Processed ${progress.getCount} in ${progress.getElapsedSeconds} seconds")
+
+    in.safelyClose()
+    out.close()
+  }
+
+  private def updateRecord(record: SAMRecord,
+                           header: SAMFileHeader,
+                           readGroupMap: Map[String, SAMReadGroupRecord]): SAMRecord = {
+    val fromReadGroup = record.getReadGroup match {
+      case null => fail(s"Record '${record.getReadName}' has no read group")
+      case rg => rg
+    }
+    readGroupMap.get(fromReadGroup.getId) match {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,201 @@
...
+  **
+  * @author Nils Homer
+  */
+@clp(description =
+  """
+      |Updates one or more read groups.
+      |
+      |The mapping CSV file should contain a header specifying each read group datum as described below.  Each subsequent
+      |row should contain at least the old and new read group ID.  Any other datum can be omitted, and the columns can be
+      |in any order.
```
tfenne: What's the behaviour if they are omitted?  Preserve the old value, or have no value in the output?

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,201 @@
...
+      |DT - Read Group run date
+      |PI - Read Group predicted insert size
+      |PG - Read Group program group
+      |PM - Read Group platform model
+    """,
+  group = ClpGroups.SamOrBam)
+class UpdateReadGroups
+(@arg(flag = "i", doc = "Input BAM file.") val input: PathToBam,
+ @arg(flag = "o", doc = "Output BAM file.") val output: PathToBam,
+ @arg(flag = "m", doc = "CSV file with the read group mapping (see detailed usage).") val readGroupsFile: FilePath,
```
tfenne: CSV, boo :frowning:

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,201 @@
...
+  private def getReadGroupMap(readGroupsFile: FilePath, samFileHeader: SAMFileHeader): Map[String, SAMReadGroupRecord] = {
+    // create the new read groups
+    val fromReadGroups = samFileHeader.getReadGroups.map { rg => rg.getId -> rg }.toMap
+
+    // read in the lines
+    val lines = Io.readLines(readGroupsFile).map(_.trim).toSeq
+    if (lines.size <= 1) mappingFileError("No read groups to replace were given in the mapping file.", 1)
+
+    // parse in the header
+    val header = lines.head.split(",")
```
tfenne: `header` -> `columnNames`

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,201 @@
...
+    // make sure we recognize all values in the header
+    header.foreach { key =>
+      if (key != fromReadGroupKey && !SAMReadGroupRecord.STANDARD_TAGS.contains(key)) mappingFileError(s"Read group key '$key' not recognized.", 1)
+    }
+    // make sure there are no duplicates
+    val headerKeySet = header.toSet
+    header.find(!headerKeySet.contains(_)) match {
```
tfenne: I've been staring at this and can't figure out how this detects duplicate keys.  Won't the names from the header always be in the set?

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,201 @@
...
+      case None =>
+      case Some(key) => mappingFileError(s"Duplicate read group key '$key'", 1)
+    }
+    // make sure there is at least the id
+    if (!header.contains(SAMReadGroupRecord.READ_GROUP_ID_TAG)) mappingFileError(s"Missing header key '${SAMReadGroupRecord.READ_GROUP_ID_TAG}'", 1)
+
+    // parse in the rest of the lines
+    val readGroupMaps = lines.tail.zipWithIndex.map { case (line: String, idx: Int) =>
+      line.split(",").toList match {
```
tfenne: I think technically `split(String)` is by regex, and `split(Char)` is by character, so you might want to use `split(',')`

tfenne:
I'm curious about the use case of this, vs. just doing something like:
```bash
samtools view -H foo.bam > header.txt
// munge header
samtools reheader header.txt foo.bam > bar.cam
```

It seems very complex, and there are no tests :disappointed:

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
nh13: I'll make an option for the delimiter.

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
nh13: Will add to the doc, but `keepReadGroupAttributes` controls this.

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
nh13: Replaced with
```
    val duplicateColumnNames = columnNames.toList.groupBy(identity).collect {  case (x, List(_, _, _*)) => x }
    if (duplicateColumnNames.nonEmpty) {
      mappingFileError(s"Duplicate read group keys: " + duplicateColumnNames.mkString(", "), 1)
    }
```

nh13:
@tfenne I'll add tests.

If I understand correctly, `samtools reheader` (and Picard's `ReplaceSamHeader`) don't update `RG` tags in the sam records.  This does.  

Nonetheless, perhaps this tool should instead take in a SAM header instead of comma/tab-delimited file.  I'll do that too, and we'll see if the code gets cleaner.

nh13:
@tfenne take a look as I have used a SAM header as input instead.  It simplifies the code.  I think other tools don't modify the read group ID where as I want to modify the read group IDs.

nh13:
The two uses cases I have come across are:
1. Combining two BAM files where the read group ID and meta information are different in each file, but I want one read group in the merged BAM.
2. Combining multiple BAMs where the read group IDs are the same but the meta data are different, and I want one read group in the merged BAM.

I suppose I could use Picard's `AddOrReplaceReadGroups` if the read group IDs are the same in the second case, but in the first they aren't.  This tool was super handy since it allowed me to modify not only the header but the read group ID stored in each record.

nh13:
@tfenne bump.

tfenne:
I'm still confused by your need for this.  `AddOrReplaceReadGroups` will handle both of your use cases as described - it replaces all the read groups in the BAM with a single read group, and updates all the reads to point to the new `@RG`.

nh13:
Suppose I have the following read groups in file one:
```
@RG ID:A.1
@RG ID:A.2
```
and the following read groups in file two:
```
@RG ID:B.1
@RG ID:B.2
```
and I want both `A.1` and `B.1` to be mapped to `C.1`, and similarly, both `A.2` and `B.2` to be mapped to `C.2`, in file three:
```
@RG ID:C.1
@RG ID:C.2
```
I think this is a clearer example of what this tool does that `AddOrReplaceReadGroups` doesn't.

Also, I like being able to give a SAM header as input just like `samtools reheader` while having the IDs in the SAM records updated.

nh13:
@tfenne I am going to add some tests, as a few colleagues need this tool, and I want to make sure it does what it says it does via tests and then I would like a review.

nh13:
@tfenne added tests, ready for review.

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,152 @@
...
+      |If `keepReadGroupAttributes` is true, then any read group attribute not replaced will be kept in the new read
+      |group.  Otherwise, only the attributes in the provided SAM header file will be used.
+    """,
+  group = ClpGroups.SamOrBam)
+class UpdateReadGroups
+(@arg(flag = "i", doc = "Input BAM file.") val input: PathToBam,
+ @arg(flag = "o", doc = "Output BAM file.") val output: PathToBam,
+ @arg(flag = "r", doc = "A SAM header file with the replacement read groups (see detailed usage).") val readGroupsFile: PathToBam,
+ @arg(flag = "k", doc = "Keep all read group attributes that are not replaced.") val keepReadGroupAttributes: Boolean = false,
+ @arg(flag = "g", doc = "If true ignore read groups not found in the replacement header, otherwise throw an error.") val ignoreMissingReadGroups: Boolean = false
```
tfenne: Does ignore mean a) don't futz with them in the header, or b) don't output reads from other read groups?

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,152 @@
...
+
+    CloserUtil.close(in)
+    out.close()
+  }
+
+  private def updateRecord(record: SAMRecord,
+                           header: SAMFileHeader,
+                           readGroupMap: Map[String, SAMReadGroupRecord]): SAMRecord = {
+    val fromReadGroup = record.getReadGroup match {
+      case null => throw new IllegalStateException(s"Record '${record.getReadName}' has no read group")
```
tfenne: Since this is in a FgBioTool you can use `fail()` here instead which will render the message without printing out the stack trace.

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,152 @@
...
+        record.setHeader(header)
+    }
+    record
+  }
+
+  private def getReadGroupMap(readGroupsFile: FilePath, samFileHeader: SAMFileHeader): Map[String, SAMReadGroupRecord] = {
+    val fromReadGroups = samFileHeader.getReadGroups.map { rg => rg.getId -> rg }.toMap
+
+    // get the header
+    val reader = SamReaderFactory.make.open(readGroupsFile.toFile)
```
tfenne: I think the latest version actually has `open(Path)` finally

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,152 @@
...
+    val h = reader.getFileHeader
+    CloserUtil.close(reader)
+
+    // map from the "FR" attribute value to the new read group
+    val readGroupMap = h.getReadGroups.map { rg =>
+      rg.getAttribute(fromReadGroupKey) match {
+        case null => throw new IllegalArgumentException(s"'$fromReadGroupKey' attribute not found in read group: " + rg.toString)
```
tfenne: `fail()`

tfenne:
Looks good to me @nh13.  A few minor comments for you to review then merge away.

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
nh13: Replaced with `Keep all read groups not found in the replacement header, otherwise throw an error`

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
```diff
@@ -0,0 +1,152 @@
...
+
+    // main loop
+    in.foreach { record =>
+      updateRecord(record, outHeader, toReadGroupsMap)
+      out.addAlignment(record)
+      progress.record(record)
+    }
+    logger.info(s"Processed ${progress.getCount} in ${progress.getElapsedSeconds} seconds")
+
+    CloserUtil.close(in)
```
nh13: At some point I wont have outdated close methods.

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
nh13: Done, here and elsewhere.

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/bam/UpdateReadGroups.scala
nh13: Done

PR #17, opened 2016-05-30
Adding a tool to create intervals for WGS variant calling.

nh13:


src/main/scala/com/fulcrumgenomics/wgs/CreateWgsCallableIntervalList.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  private var counter: Int = 0
+  private var refName: String = ""
+  private var refIndex: Int = -1
+  private var start: Int = -1
+  private var end: Int = -1
+  private val intervals = ListBuffer[Interval]()
+
+  def add(locus: LocusInfo): Boolean = {
+    val coverage = locus.getRecordAndPositions.size()
```
nh13: Could be improved with https://github.com/samtools/htsjdk/pull/616/ since then we wouldn't have to create a new list each time.

src/main/scala/com/fulcrumgenomics/wgs/CreateWgsCallableIntervalList.scala
```diff
@@ -0,0 +1,126 @@
...
+import scala.collection.mutable.ListBuffer
+
+/**
+  * Creates a list of intervals for WGS variant calling.
+  */
+@clp(description =
+  """
+    |Creates a list of intervals for WGS variant calling.
+    |
+    |The output intervals will only contain sites that have both a minimum and maximum coverage as set by the user.
```
tfenne: have both a -> between the

tfenne:
A few quick thoughts:

1. By default SamLocusIterator filters out both secondary and supplementary alignments.  You might want to change that to include supplemental.
2. Do you want to do any filtering on mapq?  Maybe exclude mapq 0 or make that a parameter?
3. Would it be beneficial to also take in the reference and exclude bases that are non-ACGT in the reference?

src/main/scala/com/fulcrumgenomics/wgs/CreateWgsCallableIntervalList.scala
```diff
@@ -0,0 +1,126 @@
...
+) extends FgBioTool with LazyLogging {
+
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+  if (maxToAccumulatePerLocus < 1.0f) throw new ValidationException("The scaling factor for maximum number of reads to accumulate per locus must be => 1.")
+
+  override def execute(): Unit = {
+    val progress: ProgressLogger = new ProgressLogger(logger, noun="positions", unit=10*1000*1000)
+    val in: SamReader = SamReaderFactory.make.open(input.toFile)
+    val locusIterator = new SamLocusIterator(in)
```
tfenne: I think you probably want to `setIncludeIndels()` here too, otherwise you'll exclude any sites that are homozygously deleted.

src/main/scala/com/fulcrumgenomics/wgs/CreateWgsCallableIntervalList.scala
tfenne: True, although it's only a wrapping of an already-created list in an unmodifiable.

src/main/scala/com/fulcrumgenomics/wgs/CreateWgsCallableIntervalList.scala
```diff
@@ -0,0 +1,126 @@
...
+    converter.intervalList.write(output.toFile)
+
+    CloserUtil.close(locusIterator)
+  }
+}
+
+private class CoveredLociToIntervals(val header: SAMFileHeader, val minCoverage: Int, val maxCoverage: Int) {
+
+  private var counter: Int = 0
+  private var refName: String = ""
```
tfenne: Would be nicer to use `Option[]` here instead of `String`/`Int` with magic "no value" values.

src/main/scala/com/fulcrumgenomics/wgs/CreateWgsCallableIntervalList.scala
```diff
@@ -0,0 +1,126 @@
...
+  private var refIndex: Int = -1
+  private var start: Int = -1
+  private var end: Int = -1
+  private val intervals = ListBuffer[Interval]()
+
+  def add(locus: LocusInfo): Boolean = {
```
tfenne: I find this whole method a little hard to follow, and I'm not convinced that it won't lose the last covered interval in the genome.  I would suggest making a four-way case statement along the lines of:

```scala
val acceptableCoverage = coverage < minCoverage || maxCoverage < coverage
val inCoverage = this.refIndex != -1

(inCoverage, acceptableCoverage) match {
    ...
}
```

Alternatively I'd suggest having test cases that actually test all the possible state transitions.

tfenne:
@nh13 My only real concern here is handling all the state transitions in the `CoveredLociToIntervals` class.  It seems like it's handling a lot (good->not good, not good-> good, "skipped over" bases, switch to next reference, etc) implicitly.  If it were me I'd either want to handle more of those explicitly or have test cases for all of them.

src/main/scala/com/fulcrumgenomics/wgs/CreateWgsCallableIntervalList.scala
nh13: I'll need update the picard/htsjdk versions.  Will do.

nh13: Waiting for https://github.com/samtools/htsjdk/pull/408 to be in master.

src/main/scala/com/fulcrumgenomics/wgs/CreateWgsCallableIntervalList.scala
nh13: The last interval is checked in the `intervalList` method.  If we are the next base, we extend the interval, otherwise we start a new one.

nh13: @tfenne do we ever get a locus with unmapped reads?  If not, then I don't think `inCoverage` above is necessary.  I am bit lost about what is confusing here.

PR #19, opened 2016-06-11
Simple class to add slurping/piping of process stdout and/or stderr t…

tfenne:
…o another location.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/AyncStreamSink.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/AyncStreamSink.scala
@@ -1,0 +1,60 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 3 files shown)


tfenne:
@nh13 Small PR here for a class that can be used to suck from stdout and/or stderr when running a `Process` since if you are not merging the streams you need to have another thread pulling from one of them to ensure you don't block/deadlock.

src/main/scala/com/fulcrumgenomics/util/AyncStreamSink.scala
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/**
+  * Class that is capable of reading output from a sub-process (or arbitrary other InputStream)
+  * and processing it through any method that accepts a String, while managing the extra thread
+  * necessary to ensure nothing blocks.
+  */
+class AsyncStreamSink(in: InputStream, private val sink: String => Unit) extends Closeable {
+  private val source = Source.fromInputStream(in).withClose(() => in.close())
+  private val thread = new Thread( new Runnable { def run() : Unit = { source.getLines().foreach(sink) } } )
```
nh13: is `source.getLines` lazy?  I'd like it to return lines when they become available, rather than when EOF has been reached.

nh13:
👍  just one question.

src/main/scala/com/fulcrumgenomics/util/AyncStreamSink.scala
tfenne: Yes, it returns an iterator that will keep returning lines when invoked - block if necessary - and then self-close when it hits EOF.

tfenne:
I just realized that I think this would be better off in dagr-commons than here.  Unless you object I'm going to close this PR and move it there instead.

nh13:
👍

PR #21, opened 2016-06-16
WIP: Adding a simple consensus caller.

nh13:
Things to do before merging:
- [ ] use the `Umi` `ClpGroup` from https://github.com/fulcrumgenomics/fgbio/pull/22


src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
```diff
@@ -0,0 +1,450 @@
...
+          val qualsAtIdx         = quals.filter { q => baseIdx < q.length}.map { q => q(baseIdx) }
+          val basesAndQualsAtIdx = basesAtIdx.zip(qualsAtIdx)
+
+          // TODO: handle Ns in the reads
+          // get the phred-scaled probability of each potential DNA base
+          var phreds = SequenceUtil.VALID_BASES_UPPER.map(_.toChar).map {
+            candidateBase =>
+              basesAndQualsAtIdx.map {
+                case (base, qual) =>
+                  if (base == candidateBase) qual.inv() // probability of not an error
```
nh13: what if the quality is zero, this could make the probability zero.  There should be some minimum, say `PhredValue.Precision`?

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
```diff
@@ -0,0 +1,450 @@
...
+  group = ClpGroups.SamOrBam)
+class CallConsensusFromUmis
+(  @arg(flag="i", doc="The input SAM or BAM file.") val input: PathToBam,
+   @arg(flag="o", doc="Output SAM or BAM file to write consensus reads.") val output: PathToBam,
+   @arg(flag="r", doc="Output SAM or BAM file to write reads not used.") val rejects: PathToBam,
+   @arg(flag="t", doc="The SAM attribute with the unique UMI identifier.") val attribute: String = DefaultAttribute,
+   @arg(flag="p", doc="The Prefix all consensus read names") val readNamePrefix: Option[String] = None,
+   @arg(flag="R", doc="The Read Group for all the consensus reads.") val readGroup: String = "A",
+   @arg(flag="1", doc="The Phred-scaled error rate for an error prior to the UMIs being integrated.") val errorRatePreUmi: PhredValue = DefaultErrorRatePreUmi,
+   @arg(flag="2", doc="The Phred-scaled error rate for an error prior to the UMIs being integrated.") val errorRatePostUmi: PhredValue = DefaultErrorRatePostUmi,
```
nh13: prior -> post

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
```diff
@@ -0,0 +1,118 @@
...
+
+
+@clp(description =
+  """
+    |Calls consensus sequences from reads with the same UMI identifier.
+    |
+    |This tool assumes that reads with the same identifier are grouped together (consecutive in the file).
+  """,
+  group = ClpGroups.SamOrBam)
+class CallConsensusFromUmis
```
tfenne: Consenses?  Since it makes multiple, and they use but are not from the UMIs, perhaps 'CallMolecularConsensusReads`?

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
```diff
@@ -0,0 +1,118 @@
...
+(  @arg(flag="i", doc="The input SAM or BAM file.") val input: PathToBam,
+   @arg(flag="o", doc="Output SAM or BAM file to write consensus reads.") val output: PathToBam,
+   @arg(flag="r", doc="Output SAM or BAM file to write reads not used.") val rejects: PathToBam,
+   @arg(flag="t", doc="The SAM attribute with the unique UMI identifier.") val attribute: String = DefaultAttribute,
```
tfenne: `attribute` -> `t` is slightly odd, maybe `tag` & `t`, or `attribute` and `a`?  What do we do elsewhere?

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
```diff
@@ -0,0 +1,118 @@
...
+   @arg(flag="p", doc="The Prefix all consensus read names") val readNamePrefix: Option[String] = None,
+   @arg(flag="R", doc="The Read Group for all the consensus reads.") val readGroup: String = "A",
```
tfenne: Is this a new RG ID?  Or the ID of an RG in the header to re-use?

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
```diff
@@ -0,0 +1,118 @@
...
+   @arg(flag="1", doc="The Phred-scaled error rate for an error prior to the UMIs being integrated.") val errorRatePreUmi: PhredValue = DefaultErrorRatePreUmi,
+   @arg(flag="2", doc="The Phred-scaled error rate for an error post the UMIs have been integrated.") val errorRatePostUmi: PhredValue = DefaultErrorRatePostUmi,
+   @arg(flag="q", doc="Cap the maximum base quality in the input.") val maxBaseQuality: PhredValue = DefaultMaxBaseQuality,
```
tfenne: Would be good to document how this interacts with `s` - is the shift applied then the cap, or vice-versa?

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
```diff
@@ -0,0 +1,118 @@
...
+   @arg(flag="s", doc="Subtract this base quality from the input base qualities.") val baseQualityShift: PhredValue = DefaultBaseQualityShift,
+   @arg(flag="N", doc="Mask (make 'N') consensus bases with quality less than this threshold.") val minConsensusBaseQuality: PhredValue = DefaultMinConsensusBaseQuality,
+   @arg(flag="M", doc="The minimum number of reads to produce a consensus base.") val minReads: Int = DefaultMinReads,
+   @arg(flag="Q", doc="The minimum mean base quality across a consensus base to output.") val minMeanConsensusBaseQuality: PhredValue = DefaultMinMeanConsensusBaseQuality
+  ) extends FgBioTool with LazyLogging {
+
+  Io.assertReadable(input)
+  Seq(output, rejects).foreach(Io.assertCanWriteFile(_))
+  if (attribute.length != 2) throw new ValidationException("attribute must be of length 2")
+  if (errorRatePreUmi < 0)   throw new ValidationException("Error rate pre UMI must be >= 0")
```
tfenne: Floor should probably be Q2 or Q3 since lower values mean it's more likely wrong than right!

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
```diff
@@ -0,0 +1,118 @@
...
+      input          = in.iterator().asScala,
+      header         = in.getFileHeader,
+      readNamePrefix = readNamePrefix,
+      readGroup      = readGroup,
+      options        = options,
+      rejects        = Some(rej),
+      progress       = Some(progress)
+    )
+
+    consensusCaller.foreach { rec => out.addAlignment(rec) }
```
tfenne: Why bother passing the progress logger into the caller - you could just do it here?

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+                                  maxBaseQuality: PhredValue              = DefaultMaxBaseQuality,
+                                  baseQualityShift: PhredValue            = DefaultBaseQualityShift,
+                                  minConsensusBaseQuality: PhredValue     = DefaultMinConsensusBaseQuality,
+                                  minReads: Int                           = DefaultMinReads,
+                                  minMeanConsensusBaseQuality: PhredValue = DefaultMinMeanConsensusBaseQuality
+                                 )
+
+object ConsensusCaller {
+  /** Creates a consensus read from the given read and quality string tuples.  If no consensus read was created, None is returned. */
+  def fromBasesAndQualities(basesAndQualities: Seq[(String, String)],
```
tfenne: This sounds to me like you're create a `ConsensusCaller` from bases and qualities, not a `consensus`.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+                            options: ConsensusCallerOptions = new ConsensusCallerOptions()
+                           ): Option[(String, String)] = {
```
tfenne: Maybe we should use a case class for `ConsensusRead` instead of a tuple?  Works the same, but will give us place to hang extra stats or anything else we want.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+    if (basesAndQualities.exists { case (b, q) => b.length != q.length }) {
+      throw new IllegalArgumentException("Found a read with the bases and qualities having different length")
+    }
+    // check to see if we have enough reads.
+    if (basesAndQualities.length < options.minReads) return None
+
+    // extract the bases and qualities, and adjust the qualities based on the given options.
+    val bases: Seq[String] = basesAndQualities.map { _._1.toUpperCase }
```
tfenne: While true, it confused me for a minute that `bases` is the bases from all the (likely multiple) reads, not just the bases from a single read.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+    // check that the mean base quality is high enough
+    if (PhredValue.mean(consensusQualities:_*) < options.minMeanConsensusBaseQuality) None
+    else Some((consensusBases, SAMUtils.phredToFastq(consensusQualities.map(_.toByte).toArray[Byte])))
+  }
+
+  /** Creates a consensus read from the given records.  If no consensus read was created, None is returned. */
+  def fromSamRecords(records: Seq[SAMRecord],
+                     options: ConsensusCallerOptions = new ConsensusCallerOptions()
+                    ): Option[(String, String)] = {
+    // TODO: if mapped, this gets the bases in genomic order and not sequencing order.  Does it matter?
```
tfenne: Yes, I think it does.  Either here, or at the callsite, we'll have to RC if the consensus is from reads mapped to the negative strand.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+    val bases = records.map { rec => rec.getReadString.toUpperCase }
+    val quals = records.map { rec => rec.getBaseQualityString }
+
+    fromBasesAndQualities( basesAndQualities=bases.zip(quals), options = options)
```
tfenne: This will be a little inefficient when processing BAM as I believe it will force base qualities to go:
`BAM-6bit -> Array[Byte] numeric Phred -> String -> Array[Byte] numeric Phred -> PhredValue`
for each record.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  /** Get the most likely consensus bases and qualities. */
+  private[bam] def consensusCalls(bases: Seq[String],
+                                  quals: Seq[Seq[PhredValue]],
+                                  errorRatePreUmi: PhredValue         = DefaultErrorRatePreUmi,
+                                  minConsensusBaseQuality: PhredValue = DefaultMinConsensusBaseQuality,
+                                  minReads: Int                       = DefaultMinReads): (String, Seq[PhredValue]) = {
+    val maxBaseLength = bases.map(_.length).max
```
tfenne: maxReadLength?

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+    val (consensusBases, consensusQualities) = Seq.range(0, maxBaseLength, 1).map { baseIdx =>
+      val basesAtIdx = bases.filter { str => baseIdx < str.length}.map { str => str.charAt(baseIdx) }
+
+      // check to see if we heave a enough reads to produce a consensus base.
+      if (basesAtIdx.length < minReads) (SequenceUtil.N.toChar, ZeroPhred)
+      else {
+        val qualsAtIdx         = quals.filter { q => baseIdx < q.length}.map { q => q(baseIdx) }
+        val basesAndQualsAtIdx = basesAtIdx.zip(qualsAtIdx)
+
+        // TODO: handle Ns in the reads
```
tfenne: I think it would be fine to handle this by adding a ` && str.charAt(baseIdx) != 'N'` in the `filter` on line 118.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+        // get the phred-scaled probability of each potential DNA base
+        var phreds = SequenceUtil.VALID_BASES_UPPER.map(_.toChar).map {
```
tfenne: I would just create your own constant for Array('A', 'C', 'G', 'T') since I wouldn't trust someone not to add 'N' or other things to that constant in HTSJDK.

tfenne: Then they could also just be `Char`s to begin with instead of having to `map` them each time through the loop.

src/main/scala/com/fulcrumgenomics/util/PhredValue.scala
```diff
@@ -0,0 +1,132 @@
...
+  /** Phred-scaled zero probability. */
+  val ZeroProbability = PhredValue(Double.PositiveInfinity)
+  /** Phred-scaled probability of one. */
+  val OneProbability  = PhredValue(0.0)
+  /** Phred value of 0. */
+  val ZeroPhred       = PhredValue(0.0)
+  /** The precision to use when converting to an Integer. */
+  val Precision       = 0.0000000001 // Q100
+  /** converts a probability to a phred score, with no validation. */
+  def toPhred(pr: Double) = -10.0*Math.log10(pr)
```
tfenne: `toPhredScale` and annotate the return type as `Double` since this doesn't actually return a PhredValue.

src/main/scala/com/fulcrumgenomics/util/PhredValue.scala
```diff
@@ -0,0 +1,132 @@
...
+  /** converts to a probability from a phred score, with no validation. */
+  def toPr(phred: Double) = Math.pow(10.0, phred / -10.0)
```
tfenne: `toProbability` please

src/main/scala/com/fulcrumgenomics/util/PhredValue.scala
```diff
@@ -0,0 +1,132 @@
...
+  /** Gets the mean of the given phred values. */
+  def mean(phred: PhredValue*): PhredValue = phred.foldLeft(ZeroProbability)((sum, qual) => sum + qual) / phred.length.toDouble
+  /*
+  implicit def phredValueWrapper[T](v: T)(implicit numeric: Numeric[T]): PhredValue = new PhredValue(value=numeric.toDouble(v))
+  implicit class NumericPhredValue[T](v: T)(implicit numeric: Numeric[T]) {
+    def toPhredValue: PhredValue = new PhredValue(value=numeric.toDouble(v))
+  }
+  */
+  implicit def intToPhredWrapper(v: Int): PhredValue = new PhredValue(value=v)
```
tfenne: I really don't like these implicit conversions for a number of reasons.  Do you really need them?

src/main/scala/com/fulcrumgenomics/util/PhredValue.scala
```diff
@@ -0,0 +1,132 @@
...
+  implicit class DoublePhredValue(v: Double) {
+    def toPhredValue: PhredValue = {
+      //if (v < 0 || v > 1) throw new IllegalArgumentException(s"Double value is out of range to convert to phred: $v")
+      new PhredValue(value=toPhred(v.toDouble))
+    }
+  }
+}
+
+/** Represents Phred-scaled probabilities, which are `-10 * log10(Pr)`. */
+case class PhredValue(var value: Double) extends Ordered[PhredValue] {
```
tfenne: Yikes - I was assuming that PhredValue was immutable.  I think it should probably be immutable and also declared as a value class (see: http://docs.scala-lang.org/overviews/core/value-classes.html).  This will have the same performance semantics, but make it much easier to reason about PhredValues as they'll behave like other value types (Ints, Doubles).

src/main/scala/com/fulcrumgenomics/util/PhredValue.scala
```diff
@@ -0,0 +1,132 @@
...
+  import PhredValue._
+  def this(v: String) = this(if (v.contains(".")) v.toDouble else v.toInt)
+  //if (value < 0) throw new IllegalArgumentException("PhredValue was less than zero")
+  def /(that: PhredValue): PhredValue = {
+    if (that.value == Double.PositiveInfinity) throw new IllegalStateException("Trying to divide by 0")
```
tfenne: `java.lang.ArithmeticException` is what you get from the native numeric types.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+          candidateBase =>
+            basesAndQualsAtIdx.map { case (base, qual) =>
+              if (base == candidateBase) qual.inv() // probability of not an error
```
tfenne: Something doesn't sit right with me here.  You use `qual.inv()`, which I would interpret as `1 / qual`.  But other operators on `PhredValue` are named as if dealing with non-log values (e.g. `phred * phred` is multiplication in linear space, and implemented as summing in log space).  Should `inv()` therefore be `oneMinus()`?

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+              else qual / 3.0 // probability of an error for this specific base
+            }.foldLeft(OneProbability)((phred, qual) => phred * qual)
+        }
+        // normalize
+        val phredsSum = phreds.foldLeft(ZeroProbability)((a, b) => a + b)
+        phreds = phreds.map { phred => phred / phredsSum }
+
+        // Find the base with the maximum probability, which means the smallest phred
```
tfenne: So confusing!  I have a problem with the idea of `PhredValue` since although you could argue it's just a scaling, it is really always used as `-10 * log10(p[error])` - i.e. as the scaling of the probability of error.  So to use it to store `p(true)` as opposed to `p(error)` is super-confusing.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+        // get the phred-scaled probability of each potential DNA base
+        var phreds = SequenceUtil.VALID_BASES_UPPER.map(_.toChar).map {
+          candidateBase =>
+            basesAndQualsAtIdx.map { case (base, qual) =>
+              if (base == candidateBase) qual.inv() // probability of not an error
+              else qual / 3.0 // probability of an error for this specific base
+            }.foldLeft(OneProbability)((phred, qual) => phred * qual)
+        }
+        // normalize
+        val phredsSum = phreds.foldLeft(ZeroProbability)((a, b) => a + b)
```
tfenne: This causes a lot of round-tripping to non-log space and back, which might result in a precision loss vs. doing a `.map(toProbability).sum`

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+  private object ReadType extends Enumeration {
+    val Fragment, FirstOfPair, SecondOfPair = Value
+  }
+
+  /** Gets the longest common prefix of the given strings, None if there is only one string or if there is an empty string. */
+  @annotation.tailrec
+  private def longestCommonPrefix(strs: Iterable[String], accu: Option[String] = None): Option[String] = {
+    if (strs.exists(_.isEmpty) || strs.size <= 1) accu
+    else {
+      val firsts = strs.map(_.head)
```
tfenne: This could be more efficient by stopping at the first non-matching character.  E.g. 
```scala
val first = strs.head.head
if (strs.tail.forall(_.head == first)) recurse
else accu
```

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+
+  /** True if there are more consensus reads, false otherwise. */
+  def hasNext(): Boolean = {
+    if (this.nextConsensusRecords.nonEmpty) true
+    else this.advance()
+  }
+
+  /** Returns the next consensus read. */
+  def next(): SAMRecord = {
+    if (!this.hasNext()) throw new IllegalStateException("Calling next() when hasNext() is false.")
```
tfenne: should be a `NoSuchElementException`

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+  import ConsensusCaller.ReadType._
+  import ConsensusCaller._
+
+  private val iter = input.buffered
+  private val nextConsensusRecords: mutable.Queue[SAMRecord] = mutable.Queue[SAMRecord]() // one per UMI group
+  private var readIdx = 1
+
+  /** True if there are more consensus reads, false otherwise. */
+  def hasNext(): Boolean = {
+    if (this.nextConsensusRecords.nonEmpty) true
```
tfenne:
```scala
def hasNext(): Boolean = this.nextConsensusRecords.nonEmpty || advance()
```

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+    else this.advance()
+  }
+
+  /** Returns the next consensus read. */
+  def next(): SAMRecord = {
+    if (!this.hasNext()) throw new IllegalStateException("Calling next() when hasNext() is false.")
+    this.nextConsensusRecords match {
```
tfenne: Why not just directly `dequeue()` here - I don't think the matcha adds anything.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+      case mutable.Queue(x, _*) => this.nextConsensusRecords.dequeue()
+      case mutable.Queue() => unreachable("hasNext() was true but nextConsensusRecords is empty")
+    }
+  }
+
+  /** Consumes records until a consensus read can be created, or no more input records are available. Returns
+    * true if a consensus read was created, false otherwise. */
+  private def advance(): Boolean = {
+    if (this.iter.isEmpty) return false
```
tfenne: Eewwwww, `return`

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+    // build a consensus for each sub-group that has a read
+    val fragmentName = nextReadName(fragments)
+    val pairName = nextReadName(firstOfPair ++ secondOfPair) // must have the same read name
+    var success = false
+    success |= this.callConsensus(fragments,    readName=fragmentName, readType=Fragment)
+    success |= this.callConsensus(firstOfPair,  readName=pairName,     readType=FirstOfPair)
+    success |= this.callConsensus(secondOfPair, readName=pairName,     readType=SecondOfPair)
+
+    if (success) true
+    else if (this.iter.nonEmpty) this.advance() // try again
```
tfenne: Could be simpified to `else this.advance()` which would make this `@tailrec` and guard against stack overlfow in the case of crappy data that can't form consensuses.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+  private def getNextGroupOfRecords(buffer: ListBuffer[SAMRecord] = new ListBuffer[SAMRecord], attr: String): ListBuffer[SAMRecord] = {
+    if (this.iter.isEmpty || this.iter.head.toAttr != attr) return buffer
+    val rec = this.iter.next()
+    progress.map(_.record(rec))
+    buffer.append(rec)
+    getNextGroupOfRecords(buffer=buffer, attr=attr)
+  }
+
+  /** Adds a method to the SAMRecord class to retrieve the attribute for the UMI identifier. */
+  implicit private class SAMRecordToString(r: SAMRecord) {
```
tfenne: Is this really necessary?  I see three usages that could easily be replaces with the inlined code without causing all the extra overhead.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+    }
+  }
+
+  /** Consumes records until a consensus read can be created, or no more input records are available. Returns
+    * true if a consensus read was created, false otherwise. */
+  private def advance(): Boolean = {
+    if (this.iter.isEmpty) return false
+
+    // get the records to create the consensus read
+    val buffer = getNextGroupOfRecords(attr=this.iter.head.toAttr)
```
tfenne: Seems odd to me to pass in the attribute value, which is then compared to.  Why not have this handled all in one place?

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+    else false // failed to build a consensus, and no more records in the input iterator
+  }
+
+  /** Adds a SAM record from the underlying iterator to the buffer if either the buffer is empty or the SAM attribute is
+    * the same for the records in the buffer as the next record in the input iterator.  Returns true if a record was
+    * added, false otherwise.
+    */
+  @annotation.tailrec
+  private def getNextGroupOfRecords(buffer: ListBuffer[SAMRecord] = new ListBuffer[SAMRecord], attr: String): ListBuffer[SAMRecord] = {
+    if (this.iter.isEmpty || this.iter.head.toAttr != attr) return buffer
```
tfenne:
```scala
if (this.iter.isEmpty) Nil
else {
    val attr = this.iter.head.toAttr
    val buffer = ListBuffer[SAMRecord]()
    while (this.iter.head.toAttr = attr) buffer += this.iter.next
    buffer.toList 
}
```

Same amount of code, but no recursion and no odd passing in of parameters from the callsite.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+  private def nextReadName(records: Seq[SAMRecord]): String = {
+    if (records.isEmpty) return ""
+    val curIdx = yieldAndThen(readIdx)(readIdx += 1)
+    val prefix = readNamePrefix.getOrElse(longestCommonPrefix(records.map(_.getReadName)).getOrElse("CONSENSUS"))
+    s"$prefix:$curIdx"
+  }
+
+  /** Attempts to create a consensus read from the given records adds it to `nextConsensusRecords`. Returns true if a
+    * consensus read was created, false otherwise. */
+  private def callConsensus(records: Seq[SAMRecord],
```
tfenne: `callConsensus` doesn't really imply side-effecting like adding it to the output 😬

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
```diff
@@ -0,0 +1,369 @@
...
+
+    // partition the records to which end of a pair it belongs, or if it is a fragment read.
+    val (fragments, firstOfPair, secondOfPair) = subGroupRecords(records = buffer)
+
+    // build a consensus for each sub-group that has a read
+    val fragmentName = nextReadName(fragments)
+    val pairName = nextReadName(firstOfPair ++ secondOfPair) // must have the same read name
+    var success = false
+    success |= this.callConsensus(fragments,    readName=fragmentName, readType=Fragment)
+    success |= this.callConsensus(firstOfPair,  readName=pairName,     readType=FirstOfPair)
```
tfenne: This raises an interesting question - what _should_ happen if `firstOfPair` creates a consensus and `secondOfPair` does not?  I would argue that we should either emit both of those reads or neither of those reads (at least by default).  To do that I think you'd need to modify `callConsensus` to return the consensus read, and then only write them if you got both, and only report success if you got both (or a fragment).

tfenne:
Ok - done for now.  I didn't review the tests, but will do so once you've had a chance to respond to the rest of the comments.

src/main/scala/com/fulcrumgenomics/util/PhredValue.scala
```diff
@@ -0,0 +1,132 @@
...
+package com.fulcrumgenomics.util
+
+import dagr.commons.CommonsDef._
+
+/** Methods and values to help convert to and from Phred-scaled probabilities.
+  *
+  * A number of implicit conversions are defined in this object.  Integers are implicitly are interpreted as
+  * literal phred values, while Floats and Doubles converted to Phred values using `toPhred`.
+  */
+object PhredValue {
```
nh13: change to `PrLog10` or something

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
nh13: using 't' and 'tag'

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
nh13: A new read group.

nh13: At some point, the specification of this should include the other values.

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
nh13: If you want to not use `errorRatePreUmi` or `errorRatePostUmi` then you should set it to `0`, so I respectively disagree.

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
nh13: I am logging the progress of the short reads, not the consensus reads.  In my opinion, the user will more likely know how many short reads there are, so logging that count makes more sense.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: Changing to `baseStrings` and `qualStrings`.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: done.

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
tfenne: _respectfully_ :trollface:

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
tfenne: Ah, ok.

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
nh13: Yes, `0` will be converted to `PositiveInfinity` as it should be, thus zero'ing out the cases where there is an error pre or post UMI (not in sequencing).  The conversion is implicit for now, as I am likely going to remove those implicits since they are bug prone.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: It should be `oneMinus()`.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: This is done _all_ in log space, the addition and the division.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: Agreed, I will change the `PhredValue` class to `PrLog10` or something like it.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: I can get rid of it.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: This is the method recursively accumulates records that have the same tag.  We pass in the value of the tag from the first record, and all subsequent records match this tag (so multiple comparisons).  I don't see how to handle it in once place, since it is already being handled in one place, namely `getNextGroupOfRecords`

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
tfenne: I'm _so_ confused.  I thought these arguments were specified as Phred scaled values.  `-10*log10(0)` is infinity, but if 0 is interpreted as a Phred scale value then it would convert to Q1.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: I see, with your `getNextGroupOfRecords` re-implementation below, the point is made.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: I got implicit happy.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
tfenne: But it's not.  `+` vectors to `add` which contains this line:
`+        -10.0 * (bLog10Pr + Math.log10(1.0 + Math.pow(10.0, aLog10Pr - bLog10Pr)))`

nh13:
@tfenne I did a major refactor of `PhredValue` and turned it into two classes: `PhredScore` and `LogDouble`.  I also updated based on your comments.

src/main/scala/com/fulcrumgenomics/bam/CallConsensusFromUmis.scala
nh13: These args in the first commit could be phred values if specified as integers, or probabilities if specified as doubles.  That's a bit confusing.  See the next commit.

src/main/scala/com/fulcrumgenomics/bam/ConsensusCaller.scala
nh13: I think this needs to be proved either way.

src/main/scala/com/fulcrumgenomics/util/LogDouble.scala
```diff
@@ -0,0 +1,172 @@
...
+  def apply(logValue: Double): LogDouble = new LogDouble(logValue=logValue)
+}
+
+/**
+  * Represents a Double in log-space, to prevent underflow and overflow.
+  * @param logValue the Double in log-space
+  */
+class LogDouble(val logValue: Double) extends AnyVal with Ordered[LogDouble] {
+  /** Returns the value as a Double. */
+  def value: Double = exp(logValue)
```
tfenne: Should we call this `linearValue` or something to make it super clear?

nh13:
@tfenne not ready for merge since I didn't squash the commits just in case you wanted to see the different ones.  Once I get a 👍  I will squash and merge.

tfenne:
Closing, so we can re-review with changes in a new PR.

PR #22, opened 2016-06-17
WIP: Umi Grouping/Assignment Tool

tfenne:
Very rough work in progress.  Due to a weird bug that I can't track down, also requires a custom HTSJDK build that is tag `2.4.1` plus [this commit](https://github.com/samtools/htsjdk/pull/642/commits/34df1a77d99b4a77c7c9231f88eb8ae1ed5604cc)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -94,6 +94,7 @@ lazy val commonSettings = Seq(
   //testOptions in Test  += Tests.Argument("-oD"),
   fork in Test         := true,
   resolvers            += Resolver.sonatypeRepo("public"),
+  resolvers            += Resolver.mavenLocal,
   shellPrompt          := { state => "%s| %s> ".format(GitCommand.prompt.apply(state), version.value) },
   updateOptions        := updateOptions.value.withCachedResolution(true)
 ) ++ Defaults.coreDefaultSettings
@@ -122,7 +123,7 @@ lazy val root = Project(id="fgbio", base=file("."))
       "org.scala-lang"            %  "scala-reflect" %  scalaVersion.value,
 	    "com.fulcrumgenomics"       %% "dagr-commons"  % "0.1.1-SNAPSHOT",
 	    "com.fulcrumgenomics"       %% "dagr-sopt"     % "0.1.1-SNAPSHOT",
-      "com.github.samtools"       %  "htsjdk"        % "2.1.0" excludeAll(htsjdkExcludes: _*),
+      "com.github.samtools"       %  "htsjdk"        % "2.5.0" excludeAll(htsjdkExcludes: _*),
       //---------- Test libraries -------------------//
       "org.scalatest"             %% "scalatest"     % "2.2.4"  % "test->*" excludeAll ExclusionRule(organization="org.junit", name="junit"),
       "org.apache.commons"        %  "commons-math3" % "3.6.1"  % "test"
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -1,0 +1,390 @@
+/**
+ * Copyright (c) 2016, Fulcrum Genomics LLC
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.fulcrumgenomics.umi
+
+import java.nio.file.Paths
+import java.util.concurrent.atomic.AtomicLong
+
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.util.{Io, ProgressLogger, Sequences, SimpleCounter}
+import com.fulcrumgenomics.util.Sequences.countMismatches
+import dagr.commons.CommonsDef.{DirPath, PathToBam}
+import dagr.commons.util.LazyLogging
+import dagr.sopt.cmdline.ValidationException
+import dagr.sopt.{arg, clp}
+import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
+import htsjdk.samtools._
+import htsjdk.samtools.util.SortingCollection
+
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+
+object GroupReadsByUmi {
+  private val SortPositionAttribute        = "__GRs"
+  private val LibraryAttribute             = "__GRlb"
+  private val InsertEndsAndStrandAttribute = "__Insert"
+
+  /** Trait that can be implemented to provide a UMI assignment strategy. */
+  private[umi] sealed trait UmiAssigner {
+    def assign(rawUmis: Seq[String]) : Map[String, String]
+  }
+
+  /** Assigns UMIs based solely on sequence identity. */
+  private[umi] class IdentityUmiAssigner extends UmiAssigner {
+    override def assign(rawUmis: Seq[String]): Map[String, String] = rawUmis.map(umi => umi -> umi.toUpperCase).toMap
+  }
+
+  /** Assigns UMIs based solely on sequence identity. */
+  private[umi] class SimpleErrorUmiAssigner(val maxMismatches: Int) extends UmiAssigner {
+    override def assign(rawUmis: Seq[String]): Map[String, String] = {
+      type UmiSet = mutable.SortedSet[String]
+      val umiSets = ListBuffer[UmiSet]()
+      for (umi <- rawUmis) {
+        val matchedSets = mutable.Set[UmiSet]()
+        for (set <- umiSets) {
+          if (set.exists(other => countMismatches(other, umi) <= maxMismatches)) {
+            set.add(umi)
+            matchedSets.add(set)
+          }
+        }
+
+        matchedSets.size match {
+          case 0 => umiSets += mutable.SortedSet(umi)
+          case 1 => matchedSets.head.add(umi)
+          case _ =>
+            // Merge the sets and remove all but one from umiSets
+            val pick = matchedSets.head
+            matchedSets.tail.foreach(set => {
+              umiSets -= set
+              pick ++= set
+            })
+        }
+      }
+
+      umiSets.flatMap(set => set.map(umi => umi -> set.head)).toMap
+    }
+  }
+
+  /**
+    * Class that implements the directed adjacency graph method from umi_tools.
+    * See: https://github.com/CGATOxford/UMI-tools
+    */
+  private[umi] class AdjacencyUmiAssigner(val maxMismatches: Int) extends UmiAssigner {
+    /** Represents a node in the adjacency graph; equality is just by UMI sequence. */
+    private class Node(val umi: String, val count: Long, val children: mutable.Buffer[Node] = mutable.Buffer()) {
+      /** Gets the full set of descendants from this node. */
+      def descendants: List[Node] = {
+        val buffer = ListBuffer[Node]()
+        addDescendants(buffer)
+        buffer.toList
+      }
+
+      private def addDescendants(buffer: mutable.Buffer[Node]): Unit = children.foreach(child => {
+        buffer += child
+        child.addDescendants(buffer)
+      })
+
+      override def equals(other: scala.Any): Boolean = other.isInstanceOf[Node] && this.umi == other.asInstanceOf[Node].umi
+    }
+
+    override def assign(rawUmis: Seq[String]): Map[String, String] = {
+      // Make a list of counts of all UMIs in order from most to least abundant
+      val nodes = SimpleCounter(rawUmis).map{case(umi,count) => new Node(umi, count)}.toBuffer.sortBy((n:Node) => -n.count)
+      val roots = mutable.Buffer[Node]()
+
+      /** Function that takes a root or starting node and finds all children in the "others"
+        * that are within a fixed edit mismatch distance (specified by this.maxMismatches) and
+        * where the child has a count that is more plausibly explained by being an error from
+        * the root than by being a different UMI. The relationship used in UMI tools is:
+        *   count(root) >= 2 * count(child) - 1
+        *
+        * This allows a root with a single observation to have a child with a single observation,
+        * but as counts of the root go up they must be >= approximately twice the child count.
+        */
+      def addChildren(root: Node, others: mutable.Buffer[Node]) : Unit = {
+        val xs = others.filter(other => countMismatches(root.umi, other.umi) <= maxMismatches && root.count >= 2 * other.count - 1)
+        root.children ++= xs
+        others --= xs
+        root.children.foreach(r => addChildren(r, others))
+      }
+
+      // Now build one or more graphs starting with the most abundant remaining umi
+      while (nodes.nonEmpty) {
+        val root = nodes.remove(0)
+        addChildren(root, nodes)
+        roots += root
+      }
+
+      // And finally make the output mapping
+      val mappings = mutable.Buffer[(String,String)]()
+      roots.foreach(root => {
+        mappings.append((root.umi, root.umi))
+        root.descendants.foreach(child => mappings.append((child.umi, root.umi)))
+      })
+
+      mappings.toMap
+    }
+  }
+
+  /** Looks in all the places the library name can be hiding. Returns the library name
+    * if one is found, otherwise returns "unknown".
+    */
+  final def library(rec: SAMRecord): String = {
+    var lib: String = rec.getTransientAttribute(LibraryAttribute).asInstanceOf[String]
+    if (lib == null) {
+      val rg = rec.getReadGroup
+      if (rg != null && rg.getLibrary != null) lib = rg.getLibrary
+      else lib = rec.getStringAttribute("LB")
+
+      if (lib == null) lib = "unknown"
+      rec.setTransientAttribute(LibraryAttribute, lib)
+    }
+
+    lib
+  }
+
+  /** Comparator to sort the reads by the earlier of the unclipped 5' position of the
+    * first or second read.  Groups reads in a convenient way for duplicate-marking and
+    * UMI assignment, and ensures that both ends of a read come together.
+    *
+    * Private because it has some serious restrictions!  Only allows:
+    *   - Paired reads only
+    *   - Mapped reads with mapped mates only
+    *   - No secondary or supplementary reads
+    *   - Read and mate _must_ be mapped to the same chromosome
+    * If any of these conditions are violated it will go badly!
+    */
+  private[umi] class EarlierReadComparator extends SAMRecordComparator {
+    override def fileOrderCompare(lhs: SAMRecord, rhs: SAMRecord): Int = compare(lhs, rhs)
+
+    /** A set of functions that extract a value from a read, used in order to compare reads. */
+    private val comparisons: Seq[(SAMRecord,SAMRecord) => Int] = Seq(
+      (lhs, rhs) => lhs.getReferenceIndex - rhs.getReferenceIndex,
+      (lhs, rhs) => sortPosition(lhs) - sortPosition(rhs),
+      (lhs, rhs) => library(lhs).compareTo(library(rhs)),
+      (lhs, rhs) => lhs.getReadName.compareTo(rhs.getReadName),
+      (lhs, rhs) => if (lhs.getFirstOfPairFlag) -1 else 1
+    )
+
+    /** Compares two reads for sort order. */
+    override def compare(lhs: SAMRecord, rhs: SAMRecord): Int = {
+      // Do some asserting!
+      assertValidRead(lhs)
+      assertValidRead(rhs)
+      comparisons.iterator.map(f => f(lhs, rhs)).find(_ != 0).getOrElse(0)
+    }
+
+    /** Asserts that we didn't get reads we are unable to sort. */
+    final def assertValidRead(rec: SAMRecord) = {
+      assert(rec.getReadPairedFlag,    "Unpaired read: " + rec.getReadName)
+      assert(!rec.getReadUnmappedFlag, "Unmapped read: " + rec.getReadName)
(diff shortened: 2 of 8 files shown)


tfenne:
@nh13 I think this is ready for a first pass review.  There are some tests now, and the structure is there to add more easily, which I will do later when I'm less brain-dead.  The `adjacency` method is yet to be implemented, but that shouldn't be too hard either.

`SamRecordSetBuilder` is a thin wrapper around `SAMRecordSetBuilder` and I would suggest that we slowly add things to it as we need them, trying to keep the API fairly minimal, so as not to end up just replicating the thing we're trying to make nicer.

build.sbt
```diff
@@ -93,6 +93,7 @@ lazy val commonSettings = Seq(
   //testOptions in Test  += Tests.Argument("-oD"),
   fork in Test         := true,
   resolvers            += Resolver.sonatypeRepo("public"),
+  resolvers            += Resolver.mavenLocal,
```
nh13: TODO: remove

build.sbt
```diff
@@ -121,7 +122,7 @@ lazy val root = Project(id="fgbio", base=file("."))
       "org.scala-lang"            %  "scala-reflect" %  scalaVersion.value,
 	    "com.fulcrumgenomics"       %% "dagr-commons"  % "0.1.1-SNAPSHOT",
 	    "com.fulcrumgenomics"       %% "dagr-sopt"     % "0.1.1-SNAPSHOT",
-      "com.github.samtools"       %  "htsjdk"        % "2.1.0" excludeAll(htsjdkAndPicardExcludes: _*),
+      "com.github.samtools"       %  "htsjdk"        % "2.4.1-SNAPSHOT" excludeAll(htsjdkAndPicardExcludes: _*),
```
nh13: TODO: remove

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
+package com.fulcrumgenomics.umi
```
nh13: license

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+
+
+object GroupReadsByUmi {
+  private val SortPositionAttribute        = "__GRs"
+  private val LibraryAttribute             = "__GRlb"
+  private val InsertEndsAndStrandAttribute = "__Insert"
+
+  /** Trait that can be implemented to provide a UMI assignment strategy. */
+  sealed trait UmiAssigner {
+    def assign(rawUmis: Seq[String]) : Map[String, String]
```
nh13: thinking out loud, is there every a time we would want to not return an assignment for a particular `rawUmi`?   If so, then we should state that there is no guarantee that the resulting map contains all the `rawUmi`s.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+    override def assign(rawUmis: Seq[String]): Map[String, String] = {
+      type UmiSet = mutable.SortedSet[String]
+      val umiSets = ListBuffer[UmiSet]()
+      for (umi <- rawUmis) {
+        val matchedSets = mutable.Set[UmiSet]()
+        for (set <- umiSets) {
+          if (set.exists(other => countMismatches(other, umi) <= maxMismatches)) {
+            set.add(umi)
+            matchedSets.add(set)
+          }
```
nh13: How wedded are we to wanting to use `map/filter/foreach` versus a nice loop like you have, that to me reads better?
```umiSets.filter {set => set.exists(other => countMismatches(other, umi) <= maxMismatches }.foreach { set => { set.add(umi); matchSets.add(set) }}```?

File: src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala [hunk 6]
```diff
@@ -0,0 +1,350 @@
...
+              pick ++= set
+            })
+        }
+      }
+
+      umiSets.flatMap(set => set.map(umi => umi -> set.head)).toMap
+    }
+  }
+
+  /** Class that implements the directed adjacency graph method from umi_tools. */
```
nh13: Probably add an http link to the paper or repo for `umi_tools`.

build.sbt
tfenne: I'll bump this to 2.5 when available later today.  I forget why now, but I did need something not in 2.4.1.

build.sbt
tfenne: I think we either need to figure out how to have gradle publish locally to .ivy in htsjdk, or leave this in if we want to be able to use locally published htsjdk artifacts.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+      /** Gets the full set of descendants from this node. */
+      def descendants: List[Node] = {
+        val buffer = ListBuffer[Node]()
+        addDescendants(buffer)
+        buffer.toList
+      }
+
+      private def addDescendants(buffer: mutable.Buffer[Node]): Unit = children.foreach(child => {
+        buffer += child
+        child.addDescendants(buffer)
```
nh13: Can we have a case where a `child` is added twice?  Consider `A->B->C` and `A->D->C`, where `C` is a child of both `B` and `D`.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+      })
+
+      override def equals(other: scala.Any): Boolean = other.isInstanceOf[Node] && this.umi == other.asInstanceOf[Node].umi
+    }
+
+    override def assign(rawUmis: Seq[String]): Map[String, String] = {
+      // Make a list of counts of all UMIs in order from most to least abundant
+      val nodes = SimpleCounter(rawUmis).map(x => new Node(x._1, x._2)).toBuffer.sortBy((n:Node) => -n.count)
```
nh13: ```val nodes = SimpleCounter(rawUmis).map { case (umi, count) => new Node(umi, count) }.toBuffer.sortBy((n: Node) => -n.count)```

nh13: Also `sortBy((n: Node) => -n.count)` could be `sortBy((n: Node) => Ordering.by(n.count).reverse)` if you want verbosity.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+      val roots = mutable.Buffer[Node]()
+
+      // Now build one or more graphs starting with the most abundant remaining umi
+      def addChildren(root: Node, others: mutable.Buffer[Node]) : Unit = {
+        val xs = others.filter(other => countMismatches(root.umi, other.umi) <= maxMismatches && root.count >= 2 * other.count - 1)
```
nh13: some explanation would be nice here about the intent of what it is doing, especially `root.count >= 2 * other.count - 1`

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+      }
+
+      while (nodes.nonEmpty) {
+        val root = nodes.remove(0)
+        addChildren(root, nodes)
+        roots += root
+      }
+
+      // And finally make the output mapping
+      val mappings = mutable.Buffer[(String,String)]()
```
nh13: I thought `mutable.Buffer` was a trait. What's going on here?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+      })
+
+      mappings.toMap
+    }
+  }
+
+  /** Looks in all the places the library name can be hiding. Returns the library name
+    * if one is found, otherwise returns "unknown".
+    */
+  final def library(rec: SAMRecord): String = {
```
nh13: this is so sad.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I'm not sure - I think for now I'd rather just make this basically private to GroupReadsByUmi.  I certainly don't want to update all the code here to handle possible non-assignments when it currently always assigns.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+    override def fileOrderCompare(lhs: SAMRecord, rhs: SAMRecord): Int = compare(lhs, rhs)
+
+    /** A set of functions that extract a value from a read, used in order to compare reads. */
+    private val comparisons: Seq[(SAMRecord,SAMRecord) => Int] = Seq(
+      (lhs, rhs) => lhs.getReferenceIndex - rhs.getReferenceIndex,
+      (lhs, rhs) => sortPosition(lhs) - sortPosition(rhs),
+      (lhs, rhs) => library(lhs).compareTo(library(rhs)),
+      (lhs, rhs) => lhs.getReadName.compareTo(rhs.getReadName),
+      (lhs, rhs) => if (lhs.getFirstOfPairFlag) -1 else 1,
+      (lhs, rhs) => lhs.getFlags - rhs.getFlags
```
nh13: what's the purpose of comparing their flags, which obviously must be different due to one being the first or second end of the pair?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I tend to prefer for comprehensions when thing start to get complicated and I'm only looping (i.e. not mapping or folding etc.).  For a string of simple operations, or one big operation, I like using `foreach` and friends.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+
+    /** Asserts that we didn't get reads we are unable to sort. */
+    final def assertValidRead(rec: SAMRecord) = {
+      assert(rec.getReadPairedFlag,    "Unpaired read: " + rec.getReadName)
+      assert(!rec.getReadUnmappedFlag, "Unmapped read: " + rec.getReadName)
+      assert(!rec.getMateUnmappedFlag, "Read w/unmapped mate: " + rec.getReadName)
+      assert(rec.getReferenceIndex == rec.getMateReferenceIndex, "Read w/mate on different chr: " + rec.getReadName)
+      assert(SAMUtils.getMateCigarString(rec) != null, "Read w/o Mate Cigar tag: " + rec.getReadName)
+      assert(rec.getAttribute("MQ") != null, "Read w/o Mate MQ tag: " + rec.getReadName)
+    }
```
nh13: should also include secondary or supplementary.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+        rec.setTransientAttribute(SortPositionAttribute, pick)
+        pick
+      }
+    }
+  }
+}
+
+@clp(group=ClpGroups.Umi, description =
+  """
+    |Groups reads together that appear to have come from the same original molecule. Reads
```
nh13: Not sorted by the outer ends, but by the first based sequenced, which could be inner for non-standard pairings.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+    |are sorted by the alignment positions of the outer-ends of the template, and then
+    |reads that have the same end positions are sub-grouped by UMI sequence.
+    |
+    |Accepts reads in any order (including unsorted) and outputs reads sorted by:
+    |   1. The lower genome coordinate of the two outer ends of the templates
+    |   2. The assigned UMI tag
+    |   3. Read Name
```
nh13: library?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: No - each child is only allowed a single parent in the way I've implemented it.  So I guess really it's an adjacency tree ;)

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: You also sort by 1st of pair, but I don't think that should be required.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+    |
+    |Both 'edit' and 'adjacency' make use of the 'edits' parameter to control the matching of
+    |non-identical UMIs.
+  """
+)
+class GroupReadsByUmi
+( @arg(flag="i", doc="The input BAM file.")              input: PathToBam  = Io.StdIn,
+  @arg(flag="o", doc="The output BAM file.")             output: PathToBam = Io.StdOut,
+  @arg(flag="t", doc="The tag containing the raw UMI.")  rawTag: String    = "RX",
+  @arg(flag="T", doc="The output tag for UMI grouping.") assignTag: String = "MI",
```
nh13: The tag isn't in the spec.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+  @arg(flag="e", doc="The allowable number of edits between UMIs.") edits: Int = 1,
+  @arg(          doc="Temporary directory for sorting.") tmpDir: DirPath = Paths.get(System.getProperty("java.io.tmpdir"))
+)extends FgBioTool with LazyLogging {
+  import GroupReadsByUmi._
+
+  private val assigner = strategy match {
+    case "identity"  => new IdentityUmiAssigner
+    case "edit"      => new SimpleErrorUmiAssigner(edits)
+    case "adjacency" => new AdjacencyUmiAssigner(edits)
+    case other       => throw new ValidationException("Unknown strategy: $other")
```
nh13: `"Unknown` => `s"Unknown`

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+    val sortProgress = new ProgressLogger(logger, verb="Sorted")
+
+    // Filter and sort the input BAM file
+    logger.info("Filtering and sorting input.")
+    in.iterator
+      .filter(r => !r.isSecondaryOrSupplementary)
+      .filter(r => r.getReadPairedFlag)
+      .filter(r => !r.getReadUnmappedFlag && !r.getMateUnmappedFlag)
+      .filter(r => r.getReferenceIndex == r.getMateReferenceIndex)
+      .filter(r => r.getMappingQuality >= this.minMapQ && r.getIntegerAttribute(SAMTag.MQ.name()) >= this.minMapQ)
```
nh13: may cause a problem if `MQ` isn't present, perhaps guard against it anyhow?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+    while (iterator.hasNext) {
+      // Take the next set of pairs by position and assign UMIs
+      val pairs = takeNextGroup(iterator)
+      pairs.foreach { case(r1, r2) => assert(r1.getReadName == r2.getReadName, "Reads out of order @ " + r1.getReadName) }
+      assignUmiGroups(pairs)
+
+      // Then output the records in the right order (assigned tag, read name, r1, r2)
+      val pairsByAssignedTag = pairs.groupBy(pair => pair._1.getStringAttribute(this.assignTag))
+      pairsByAssignedTag.keys.toSeq.sorted.foreach(tag => {
+        pairsByAssignedTag(tag).sortBy(pair => pair._1.getReadName).flatMap(pair => Seq(pair._1, pair._2)).foreach(rec => {
```
nh13: gorgeous!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Like a lot of the traits, there's an object with an `apply` that provides a default implementation.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Agreed.  Ideally this code would be inside SAMRecord.  I blame the spec for having multiple places to store `LB`.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+    */
+  def assignUmiGroups(pairs: Seq[ReadPair]): Unit = {
+    val rawUmis       = pairs.map(_._1).map(_.getStringAttribute(this.rawTag).toUpperCase)
+    val rawToAssigned = this.assigner.assign(rawUmis)
+    val assignedToId  = rawToAssigned.values.toSet[String].map(_ -> counter.getAndIncrement().toString).toMap
+
+    pairs.foreach(pair => Seq(pair._1, pair._2).foreach(rec => {
+      val raw = rec.getStringAttribute(this.rawTag).toUpperCase
+      val assigned = rawToAssigned(raw)
+      val id = assignedToId(assigned)
```
nh13: this is purely for counting purposes right?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Admittedly it should never be reached, but for some reason I thought there should be a tie breaker in case duplicate reads got fed through.  I'll kill it until proved necessary.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -0,0 +1,350 @@
...
+    val tmp = rec.getTransientAttribute(GroupReadsByUmi.InsertEndsAndStrandAttribute)
+    if (tmp != null) {
+      tmp.asInstanceOf[(Int,Int,Int,Boolean,Boolean)]
+    }
+    else {
+      val chrom = rec.getReferenceIndex.toInt
+      val recNeg = rec.getReadNegativeStrandFlag
+      val recPos = if (recNeg) rec.getUnclippedEnd else rec.getUnclippedStart
+      val mateNeg = rec.getMateNegativeStrandFlag
+      val matePos = if (mateNeg) SAMUtils.getMateUnclippedEnd(rec) else SAMUtils.getMateUnclippedStart(rec)
```
nh13: align the `=`

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -0,0 +1,14 @@
+package com.fulcrumgenomics.util
```
nh13: license

src/test/scala/com/fulcrumgenomics/testing/SamRecordSetBuilder.scala
```diff
@@ -0,0 +1,94 @@
+package com.fulcrumgenomics.testing
```
nh13: license

src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
```diff
@@ -0,0 +1,128 @@
+package com.fulcrumgenomics.umi
```
nh13: license

nh13:
Review done, back to you.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: A tree is a graph! :shipit:

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Agreed - I wanted to discuss this, and then will make a PR once we're happy.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Good catch.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I'm not sure I understand.  The way it's implemented it does a double-mapping: `raw-umi` -> `assigned-umi` -> `serial-id`.  I did this so that the assignment strategies can just map input to output UMIs, and then one piece of code can "name" each group.

One question is whether we would prefer to use more meaningful IDs, e.g. instead of `1`, maybe something like `chr1:1000-1400:+-:ACTGCATG`.  The latter is clearly more human meaningful, but to guaranteed to be unique we'd have to include a bunch of fields.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: Good point, neither `RX`!

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -0,0 +1,40 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.fulcrumgenomics.util
+
+/**
+  * Utility methods for working with DNA or RNA sequences
+  */
+object Sequences {
```
nh13: `Sequences` seems too generic, perhaps, `GenomicSequences`?

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
```diff
@@ -0,0 +1,54 @@
...
+class SimpleCounter[T] extends Iterable[(T, Long)] {
+  private val counts = mutable.Map[T,Long]()
+
+  /** Increment the count of object T by 1. */
+  def count(t: T): Unit = count(t, 1)
+
+  /** Increment the count of object T by the specified value. */
+  def count(t: T, n: Long): Unit = counts(t) = counts.getOrElse(t, 0L) + n
+
+  /** Gets the count of the provided object. */
```
nh13: ... or zero if the object hasn't been seen.

src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  {
+    "SimpleErrorUmiAssigner" should "group UMIs together by mismatches" in {
+      val assigner = new GroupReadsByUmi.SimpleErrorUmiAssigner(1)
+      val umis = Seq("AAAAAA", "AAAATT", "AAAATA",
+                     "TGCACC", "TGCACG",
+                     "GGCGGC", "GGCGGC", "GGCGGC")
+
```
nh13: nls

tfenne: The newlines are on purpose - each row represents a set of UMIs that are expected to be grouped together.

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
tfenne: Is that not then the count of times that object has been observed?

src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
nh13: There are two newlines (line 66 and 67), perhaps just a preference thing.  Feel free to ignore.

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
nh13: I think there is a difference between knowing that you have seen it zero times and never knowing about it at all.  For example, one could have a method:
```
def register(t: T): Unit = counts(t) = 0
```
and now I am definitively saying I have seen this zero times.

PR #23, opened 2016-06-23
New tool that is similar to, but less complicated than, Picard's FixMateInformation.

tfenne:
@nh13 This is a pretty simple tool, but since we are starting to have tools that rely on correctly set mate information I thought it would be nice to be able to point people to a tool in `fgbio` in the usage of other tools that require proper mate information.


src/main/scala/com/fulcrumgenomics/bam/SetMateInformation.scala
```diff
@@ -0,0 +1,57 @@
+package com.fulcrumgenomics.bam
```
nh13: license

src/main/scala/com/fulcrumgenomics/bam/SetMateInformation.scala
```diff
@@ -0,0 +1,57 @@
...
+import dagr.commons.util.LazyLogging
+import dagr.sopt.cmdline.ValidationException
+import dagr.sopt.{arg, clp}
+import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
+import htsjdk.samtools.SamPairUtil.SetMateInfoIterator
+import htsjdk.samtools.{SAMFileWriterFactory, SamReaderFactory}
+
+import scala.collection.JavaConversions.asScalaIterator
+
+
```
nh13: nl

src/main/scala/com/fulcrumgenomics/bam/SetMateInformation.scala
```diff
@@ -0,0 +1,57 @@
...
+  @arg(flag="r", doc="Reference fasta, only needed if writing CRAM.") val ref: Option[PathToFasta] = None,
+  @arg(flag="x", doc="If specified, do not fail when reads marked as paired are missing their mate pairs.")
+                                                                      val allowMissingMates: Boolean = false
+) extends FgBioTool with LazyLogging {
+
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+  private val in = SamReaderFactory.make().open(input.toFile)
+  if (in.getFileHeader.getSortOrder != SortOrder.queryname && in.getFileHeader.getGroupOrder != GroupOrder.query) {
+    in.close()
```
nh13: `CloserUtil.close`?

src/main/scala/com/fulcrumgenomics/bam/SetMateInformation.scala
```diff
@@ -0,0 +1,57 @@
...
+    val out = new SAMFileWriterFactory().makeWriter(in.getFileHeader, true, output.toFile, ref.map(_.toFile).orNull)
+    val progress = new ProgressLogger(logger)
+    val iterator = new SetMateInfoIterator(in.iterator(), true, allowMissingMates).toIterator
+
+    for (rec <- iterator) {
+      out.addAlignment(rec)
+      progress.record(rec)
+    }
+
+    out.close()
```
nh13: add `CloserUtil.close(in)`.

nh13: Perhaps we need our own `CloserUtil`?

src/test/scala/com/fulcrumgenomics/bam/SetMateInformationTest.scala
```diff
@@ -0,0 +1,49 @@
+package com.fulcrumgenomics.bam
```
nh13: license

nh13:
✔️

src/main/scala/com/fulcrumgenomics/bam/SetMateInformation.scala
tfenne: Perhaps an implicit that adds a `safelyClose()` method to `Any` that we stick in some common def?  Though we should only use that kind of `close()` on readers, since we want to know if a writer has a problem.

tfenne:
@nh13 Updated (and fixed my project prefs to auto-populate licenses for fgbio).  Build will probably fail until https://github.com/fulcrumgenomics/dagr/pull/173 is merged though.

PR #24, opened 2016-06-23
Removed usage of PicardException, replace with fail()

tfenne:
@nh13 I was thinking that since we have minimal dependencies on Picard it would be nice to be able to drop that dependency, and just depend on htsjdk.  What do you think?

The only remaining usages of Picard are:

1. `PickIlluminaIndices` references `IlluminaUtil.IlluminaAdapterPair`
1. `SplitTag` references `IlluminaUtil.BARCODE_DELIMITER` aka `-`.

The options as I see it are: 1) make a PR to push IlluminaUtil down into HTSJDK (it is standalone as far as I can tell) or 2) copy just the adapters and constants we need into a class in fgbio.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,7 +100,7 @@ lazy val commonSettings = Seq(
 ////////////////////////////////////////////////////////////////////////////////////////////////
 // root project
 ////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


nh13:
Lets copy things over that we need.  I don't think Illumina adapter sequences belong in htsjdk.

tfenne:
@nh13 Done!

PR #27, opened 2016-06-24
Some basic tests for the cmdline package.

tfenne:


tfenne:
@nh13 Nothing fancy, just makes sure we didn't accidentally break the CLP system.

PR #32, opened 2016-06-26
Adding a tool to compress and decompress bytes in blocks.

nh13:


build.sbt
```diff
@@ -94,6 +94,7 @@ lazy val commonSettings = Seq(
   //testOptions in Test  += Tests.Argument("-oD"),
   fork in Test         := true,
   resolvers            += Resolver.sonatypeRepo("public"),
+  resolvers            += Resolver.mavenLocal,
```
nh13: :/

build.sbt
```diff
@@ -119,10 +120,11 @@ lazy val root = Project(id="fgbio", base=file("."))
...
     libraryDependencies ++= Seq(
-      "org.scala-lang"            %  "scala-reflect" %  scalaVersion.value,
-	    "com.fulcrumgenomics"       %% "dagr-commons"  % "0.1.1-SNAPSHOT",
-	    "com.fulcrumgenomics"       %% "dagr-sopt"     % "0.1.1-SNAPSHOT",
-      "com.github.samtools"       %  "htsjdk"        % "2.1.0" excludeAll(htsjdkExcludes: _*),
+      "org.scala-lang"            %  "scala-reflect" % scalaVersion.value,
+      "com.github.dblock"         %   "oshi-core"    % "2.1",
+      "com.fulcrumgenomics"       %% "dagr-commons"  % "0.1.1-SNAPSHOT",
+      "com.fulcrumgenomics"       %% "dagr-sopt"     % "0.1.1-SNAPSHOT",
+      "com.github.samtools"       %  "htsjdk"        % "2.4.1-44-gec44a54-SNAPSHOT" excludeAll(htsjdkExcludes: _*),
```
nh13: this will require the next release of htsjdk

nh13:
@tfenne this will fail tests until htsjdk is released.  In addition to a review, would you care to take it for a spin?  I have some large test data on which it works.  I have found that the time it takes to get to the actual processing of data (i.e. the time it takes to do all the CLP work) is 3-5 seconds, so when comparing it to `pbgzip` I see about that shift in run time.

src/main/scala/com/fulcrumgenomics/util/CloseableBlockingQueue.scala
```diff
@@ -0,0 +1,200 @@
...
+import java.util.concurrent.locks.ReentrantLock
+import java.util.concurrent.{BlockingQueue, LinkedBlockingQueue}
+
+import scala.collection.JavaConversions._
+import scala.collection.mutable
+
+/**
+  * A blocking queue that has a close method.
+  *
+  * The following methods are blocking: [[put]], [[putAll]], [[take]], [[takeToMax]], and [[close]].
```
nh13: add `[[putSome]]` and `[[takeAtLeast]]` in this list

src/main/scala/com/fulcrumgenomics/util/zip/ZipBlocks.scala
```diff
@@ -0,0 +1,95 @@
...
+import dagr.commons.io.Io
+import dagr.commons.util.LazyLogging
+import dagr.sopt._
+import dagr.sopt.cmdline.ValidationException
+
+@clp(description =
+  """
+    |Performs block compression or decompression.
+    |
+    |This tool compressed data into BGZF compression format (see https://samtools.github.io/hts-specs/SAMv1.pdf).
```
nh13: this sentence needs work.

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016Fulcrum Genomics
```
nh13: needs a space

nh13:
Note: htsjdk still needs a release (2.6.0+) for this to compile.

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+    // Create the threads
+    val threads = (Seq(reader, writer) ++ consumers).map { runnable => new Thread(runnable) }
+    threads.foreach { thread => thread.setUncaughtExceptionHandler(threadExceptionHandler) }
+
+    // start your engines
+    threads.foreach { thread => thread.start() }
+
+    while (!isDone) {
+      // let others do stuff for a bit
+      Thread.sleep(Zipper.TimeoutInMilliseconds)
```
tfenne: I figured you'd just run the reader or writer on this thread.  Since you're not doing that I would suggest a longer sleep than 1ms - 10-50ms wouldn't affect response time for a user.

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+
+      // if the reader reaches an EOF, close the toConsumeQueue only if it is empty
+      if (!toConsumeQueue.isClosed && reader.isStopped && toConsumeQueue.isEmpty) toConsumeQueue.close()
+
+      // if all the consumers are done, close the toWriteQueue
+      if (!toWriteQueue.isClosed && !consumers.exists(c => !c.isStopped)) toWriteQueue.close()
+
+      /*
```
tfenne: Uncomment or delete.

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+        ", R=" + reader.blocksBuffered + ")"
+      )
+      */
+    }
+
+    // throw the exception if one was raised in a thread
+    threadException.map { thr => throw thr }
+
+    // stop all runnables
+    require(reader.isStopped) // should be done, so need to stop it
```
tfenne: provide a message in case it's not though?

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+    consumers.foreach { consumer => consumer.stopIt() }
+    writer.stopIt()
+
+    // close the queues
+    unusedQueue.close()
+    toConsumeQueue.close()
+    toWriteQueue.close()
+
+    // join
+    threads.foreach { thread => thread.join() } // TODO: timeout?
```
tfenne: Yes, definitely a timeout!

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+
+    // forces the writer to perform any last steps (ex. write the last empty GZIP block).
+    writer.finish()
+
+    // stop itself
+    stopIt()
+  }
+
+  /** Closes the input and output channels, and any other streams opened using them. */
+  def close(): Unit = {
```
tfenne: What happens if an external user calls this without first calling `stopIt`?

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+    reader.close()
+    writer.close()
+  }
+}
+
+/** Companion methods for input and output blocks of data. */
+private[zip] object Blocks {
+  /** Returns the approximate size in bytes for Blocks */
+  val sizeInBytes: Long = {
```
tfenne: No need for braces or multiple lines.

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+    SequentialBlocks.sizeInBytes(true) + SequentialBlocks.sizeInBytes(false)
```
tfenne: I find it really confusing that things are parameterized (e.g. `def sizeInBytes(compress: Boolean, capacity: Int = DefaultCapacity)`) but then everywhere just assumes the default is in use.  If that's the case, why not just turn them all into constant values and be done?

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+  }
+
+  /** The default block length in bytes of uncompressed data. */
+  val DefaultUncompressedFixedBlockLength = DEFAULT_UNCOMPRESSED_BLOCK_SIZE
```
tfenne: I spent quite a bit of time trying to locate `DEFAULT_UNCOMPRESSED_BLOCK_SIZE` before finding the import.  Can you please prefix these values coming in from a separate library (htsjdk) with their class name?

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+  /**
+    * Stores multiple blocks to compress or decompress.  This facilitates more efficient reading and writing
+    * of the data, since more data can be read or written at any one time.  It also allows for multiple blocks to
+    * be compressed or decompressed by a single consumer.
+    *
+    * @param capacity the capacity of fixed-size blocks.
+    * @param compress true if the data will be compressed, false if it is to be decompressed.
+    */
+  private class SequentialBlocks private(capacity: Int, compress: Boolean) {
+    def this(compress: Boolean) = {
```
tfenne: No need for braces or multiple lines.

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+      this(capacity=SequentialBlocks.DefaultCapacity, compress=compress)
```
tfenne: Why use a secondary constructor to do this instead of just making capacity defaulted in the primary constructor?

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+    }
+    byteBuffer.order(ByteOrder.LITTLE_ENDIAN)
+
+    /** The number of blocks stored. */
+    var numStoredBlocks: Int = 0
+
+    /** The size of each stored block. */
+    var blockSizes = new ListBuffer[Int]() // NB: for uncompressed data, only the last size can be different
+
+    /** Gets the underlyng array of bytes. */
```
tfenne: underlyng -> underlying

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+    }
+  }
+}
+
+/** Stores the input data to be processed, and the output data after processing.
+  *
+  * @param blockId the block identifier, typically the index of the block as it was read.
+  * @param compress true if the input data will be compressed, false otherwise.
+  */
+private class Blocks(var blockId: Long, compress: Boolean) extends Ordered[Blocks] {
```
tfenne: The names of these classes make me think that `SequentialBlocks` will contain many `Block[s]`, not the other way around.

src/main/scala/com/fulcrumgenomics/util/zip/Zipper.scala
```diff
@@ -0,0 +1,840 @@
...
+
+  /** Clears the underling input and output data. */
+  def clear(): Unit = {
+    this.inData.clear()
+    this.outData.clear()
+  }
+
+  /** Compares two blocks by their id. */
+  def compare(that: Blocks): Int = this.blockId.compare(that.blockId)
+}
```
tfenne: Both this class and `SequentialBlocks` have a very struct-like feel to them, in that they seem to contain a bunch of state but don't have methods to mutate that state and instead (I'm assuming) expose their state for others to manipulate.  Makes it very hard to understand what's going on and whether things are maintained consistently (e.g. keeping the buffers, block sizes etc. in sync).

PR #33, opened 2016-06-27
WIP: Adding a simple consensus caller (take 2)

tfenne:


src/main/scala/com/fulcrumgenomics/util/LogDouble.scala
```diff
@@ -0,0 +1,172 @@
...
+
+  /** Convenience constructor. */
+  def apply(logValue: Double): LogDouble = new LogDouble(logValue=logValue)
+}
+
+/**
+  * Represents a Double in log-space, to prevent underflow and overflow.
+  * @param logValue the Double in log-space
+  */
+class LogDouble(val logValue: Double) extends AnyVal with Ordered[LogDouble] {
```
tfenne: I found it odd that this uses natural log instead of log10, despite that in bfx log10 is far more common.  Any reason not to just use log10?  Or call the class `LnDouble`?

tfenne: Also, it makes me wonder if the `logValue` shouldn't be private and accessible only through methods that make it clear what the base is?

src/main/scala/com/fulcrumgenomics/util/LogDouble.scala
```diff
@@ -0,0 +1,172 @@
...
+  def linearValue: Double = exp(logValue)
+
+  /** Division. */
+  def /(that: LogDouble): LogDouble = {
+    if (that.logValue == LogDouble.Zero.logValue) throw new IllegalStateException("Trying to divide by 0")
+    new LogDouble(logValue=this.logValue - that.logValue)
+  }
+
+  /** Multiplication. */
+  def * (that: LogDouble): LogDouble = new LogDouble(logValue=this.logValue + that.logValue)
```
tfenne: I still struggle with these operators.  I'm never going to be not confused that, e.g.: `LogDouble(5)` + `LogDouble(5)` is doing non-log space addition, and doesn't yield `LogDouble(10)`.

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -0,0 +1,122 @@
...
+  """,
+  group = ClpGroups.Umi)
+class CallMolecularConsensusReads
+( @arg(flag="i", doc="The input SAM or BAM file.") val input: PathToBam,
+  @arg(flag="o", doc="Output SAM or BAM file to write consensus reads.") val output: PathToBam,
+  @arg(flag="r", doc="Output SAM or BAM file to write reads not used.") val rejects: PathToBam,
+  @arg(flag="t", doc="The SAM attribute with the unique UMI identifier.") val tag: String = DefaultTag,
+  @arg(flag="p", doc="The Prefix all consensus read names") val readNamePrefix: Option[String] = None,
+  @arg(flag="R", doc="The new read group ID for all the consensus reads.") val readGroupId: String = "A",
+  @arg(flag="1", doc="The Phred-scaled error rate for an error prior to the UMIs being integrated.") val errorRatePreUmi: LogDouble = DefaultErrorRatePreUmi,
```
tfenne: What does it mean for these to be `LogDouble`?  I don't think we're asking users to specify these values in `ln(error probability)`, but that would seem logical based on asking for it as a `LogDouble`.  Why not have `PhredScore` as a type that can be used as arguments?

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -0,0 +1,122 @@
...
+    |Calls consensus sequences from reads with the same UMI identifier.
+    |
+    |This tool assumes that reads with the same identifier are grouped together (consecutive in the file).
+  """,
+  group = ClpGroups.Umi)
+class CallMolecularConsensusReads
+( @arg(flag="i", doc="The input SAM or BAM file.") val input: PathToBam,
+  @arg(flag="o", doc="Output SAM or BAM file to write consensus reads.") val output: PathToBam,
+  @arg(flag="r", doc="Output SAM or BAM file to write reads not used.") val rejects: PathToBam,
+  @arg(flag="t", doc="The SAM attribute with the unique UMI identifier.") val tag: String = DefaultTag,
```
tfenne: I don't think this is strictly a UMI tag - it's a `UM` tag ;)  But seriously, in `GroupReadsByUmi` I'm emitting a tag that is a unique identifier for an original molecule, file-wide, i.e. it's broader scoped than a corrected UMI.

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  Io.assertReadable(input)
+  Seq(output, rejects).foreach(Io.assertCanWriteFile(_))
+  if (tag.length != 2)         throw new ValidationException("attribute must be of length 2")
+  if (errorRatePreUmi < Zero)  throw new ValidationException("Error rate pre UMI must be >= 0")
+  if (errorRatePostUmi < Zero) throw new ValidationException("Error rate post UMI must be >= 0")
+
+  /** TODO
+    * - what about when a read has an indel causing the rest of the bases to mismatch?
+    * - what about if a read is soft-clipped and the rest of the bases mismatch (ex. 30M120S) [second pass]?
```
tfenne: I thought we would just ignore soft-clipping and process the entire read vs. the other reads?

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -0,0 +1,122 @@
...
+
+    CloserUtil.close(in)
+    out.close()
+    rej.close()
+    logger.info(s"Processed ${progress.getCount} records.")
+  }
+}
+
+// TODO
+class ConsensusCallingMetricsCollector {
```
tfenne: Can we kill this class until we're ready please?

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -0,0 +1,419 @@
...
+import dagr.commons.CommonsDef._
+import htsjdk.samtools._
+import htsjdk.samtools.util.SequenceUtil
+
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+object ConsensusCallerOptions {
+  /** Various default values for the consensus caller. */
+  val DefaultTag: String                             = "RX"
```
tfenne: `MI` to match `GroupReadsByUmi` unless we want to standardize on something else.

tfenne:
@nh13 This is the new PR.  Sorry, since I opened it you won't have gotten emails for prior comments.

src/main/scala/com/fulcrumgenomics/util/LogDouble.scala
```diff
@@ -169,4 +175,6 @@ class LogDouble(val logValue: Double) extends AnyVal with Ordered[LogDouble] {
     else if (a < b) throw new IllegalArgumentException("Subtraction will be less than zero.")
     else a + log1p(0.0 - exp(b - a)) // using log1p, otherwise it would be `a + log(1.0 - exp(b - a))`
   }
+
+  override def toString: String = s"$linearValue.to ($logValue)"
```
nh13: @tfenne I have no clue what we want here for a string, but this is clearly wrong.

src/main/scala/com/fulcrumgenomics/util/LogDouble.scala
nh13: `LogDouble` is really the standard name, I swear it's not just me.  I am open to calling in `PreciseDouble` of just `Probability`, and then just hide the fact it is in log-scale?

 *http://colorfulengineering.org/logmath-notes.pdf
* https://github.com/scalanlp/breeze/blob/master/math/src/main/scala/breeze/math/LogDouble.scala
* https://en.wikipedia.org/wiki/Log_probability

nh13:
@tfenne Added a large explanation of the likelihood model in the description of `CallMolecularConsensusReads`.  As a reviewer once commented on one of my papers, I should probably have a native English speaker check it.  Can you do a quick read?

nh13:
@tfenne I think we can speed up this further.  We've discussed a few offline, and I have a few example branches which give a 50% speedup when re-writing some of the code in Java where I use only arrays of bytes. 

My suggestion would be a 👍 you if you think the review is done, and I will create an issue to speed it up, and list ideas there, and we then aren't holding up this PR.

nh13:
Closing as the consensus caller was merged in #48

PR #35, opened 2016-06-28
Adding tests for ExtractUmisFromBam.

nh13:
See #30


nh13:
@tfenne code coverage is 100% for `ExtractUmisFromBam`.  I want feedback on the first commit, where I moved two tools to the new `umi` package.

src/main/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmis.scala
```diff
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2016 Fulcrum Genomics LLC
+ * Copyright (c) $year Fulcrum Genomics
```
tfenne: Is this an accident?

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2016 Fulcrum Genomics LLC
+ * Copyright (c) $year Fulcrum Genomics
```
tfenne: Ditto

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -113,14 +114,14 @@ class ExtractUmisFromBam
         // verify everything is in order
         Seq(r1, r2).foreach(r => {
           if (!r.getReadPairedFlag) throw new IllegalStateException(s"Read ${r.getReadName} was not paired")
-          if (!r1.getReadUnmappedFlag) throw new IllegalStateException(s"Read ${r.getReadName} was not mapped")
+          if (!r1.getReadUnmappedFlag) throw new IllegalStateException(s"Read ${r.getReadName} was not unmapped")
```
tfenne: I think this should be `!r.getReadUnmappedFlag` not `!r1...`

tfenne: Also, if you care, you could use `fail(msg)` instead of throwing here, and it will result in a message on the console and exiting, without a big stack trace getting printed.

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
```diff
@@ -0,0 +1,177 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics
```
tfenne: I believe we should still have `LLC` in the license.

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
```diff
@@ -0,0 +1,177 @@
...
+import htsjdk.samtools.{SAMFileWriterFactory, SAMRecord}
+
+/**
+  * Tests for ExtractUmisFromBam.
+  */
+class ExtractUmisFromBamTest extends UnitSpec {
+
+  def newBam = makeTempFile("extract_umis_from_bam_test.", ".bam")
+
+  def annotateRecordFragment: SAMRecord = {
```
tfenne: This could be simplified to:
```scala
new SamRecordSetBuilder(readLength = 100).addFrag(name = "Frag", start = 1).map(_.setReadString("A" * 100)).head
```

tfenne:
Some minor weirdness with the licenses, otherwise looks good.

nh13:
Yeah, I made changes in the ReadStructure tests to fix license issues. I'll make sure to rebase and check the licenses again.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
nh13: I think IntelliJ did this when I refactored, weird that it did this, and weird it isn't putting in the year based on my template.

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
nh13: `addFrag` doesn't return a list, and `setReadString` returns void so I had to do this:
    `Seq(new SamRecordSetBuilder(readLength = 100).addFrag(name = "Frag", start = 1)).map {rec => rec.setReadString("A" * 100); rec }.head`

PR #38, opened 2016-06-28
Added tests for PickIlluminaIndices.

nh13:
I started adding some tests, and had to add an option for ignoring predicting structure.  I also found a bug (factorial(0) is one).


tfenne:
Looks good.  The only thing I would be tempted to add is reading back in the indices and asserting the edit distance between each pair that was picked.  Not essential though.

PR #41, opened 2016-06-30
Tool for identifying and extracting reads that contain technical sequences

tfenne:
Should be a quick review @nh13

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
@@ -1,0 +1,113 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
```diff
@@ -0,0 +1,113 @@
...
+import dagr.sopt._
+import htsjdk.samtools.util.SequenceUtil
+import htsjdk.samtools.{SAMFileWriterFactory, SamReaderFactory}
+
+import scala.annotation.tailrec
+import scala.collection.JavaConversions.asScalaIterator
+
+@clp(group = ClpGroups.SamOrBam, description =
+  """
+     |Find reads that are from technical or synthetics sequences in a BAM file. Takes in
```
nh13: synthetics -> synthetic

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
```diff
@@ -0,0 +1,113 @@
...
+    while (iter.hasNext) {
+      val r1 = iter.next()
+      val r2 = if (iter.hasNext && iter.head.getReadName == r1.getReadName) Some(iter.next()) else None
+      val recs = Seq(Some(r1), r2).flatten
+      if (recs.exists(r => matcher.matches(r.getReadBases))) {
+        recs.foreach(out.addAlignment)
+        n += recs.size
+      }
+
+      recs.foreach(progress.record)
```
nh13: does progress have a `counts` method that returns the # of records it has encountered?  If so, no need for `n`.

nh13:
One minor typo, then merge.

PR #43, opened 2016-06-30
Adding start and end messages, v2

tfenne:


tfenne:
An alternative to #40 - tidies up the code a bit, and condenses things down so only a single line is output at startup and shutdown.

PR #44, opened 2016-07-01
Adding a tool HapCutToVcf.

nh13:
This tool will convert the output of HapCut to a valid VCF, in a similar
fashion to how Picard's MergeBamAlignment post-processes the output of
an aligner.


src/main/scala/com/fulcrumgenomics/cmdline/ClpGroups.scala
```diff
@@ -58,11 +58,17 @@ object ClpGroups {
...
+    override def description: String = "Tools for working and producing with phased variants"
+  }
+
   final val Personal  = classOf[_PersonalGroup]
   final val SamOrBam  = classOf[_SamOrBamGroup]
   final val Fasta     = classOf[_FastaGroup]
   final val Fastq     = classOf[_FastqGroup]
   final val Umi       = classOf[_Umi]
   final val Utilities = classOf[_UtilitiesGroup]
+  final val Phasing   = classOf[_PhasingGroup]
```
tfenne: Are we likely to have > 1 tool for phasing over time?  If not, perhaps just a variants group?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+import htsjdk.variant.variantcontext.writer.{Options, VariantContextWriter, VariantContextWriterBuilder}
+import htsjdk.variant.vcf._
+
+import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
+
+@clp(
+  description =
+    """
+      |Converts the output of HapCut to a VCF.
```
tfenne: Maybe add a reference to HapCut here or at the end of the documentation?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      |
+      |The output of HAPCUT does not include all input variants, but simply those variants that are in phased blocks.
+      |This tool takes the original VCF and the output of HapCut, and produces two VCFs: (1) a VCF containing variants
+      |that were phased, and (2) a VCF containing variants that were not phased, such that all variants in the original
```
tfenne: I'm guessing here, but is this because it's not possible to output unphased genotypes between phased genotypes in VCF?  I.e. if you have variants A-B-c-D with A, B and D phased, you can't stick `c` in the middle?  Otherwise it seems like it would be much nicer to have a single VCF with all variants, with phasing where possible.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      |file are in one of two output VCFs.
+      |
+      |The VCF with phased variants will have the first variant of a haplotype block with a genotype "M\F", while the
```
tfenne: This doesn't mean much to me - so perhaps a little more elaboration?  I'm guessing Mother and Father?  But we don't know parent, we just know what goes together.  Also is the `\` vs `|` VCF spec stuff?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      |
+    """,
+  group=ClpGroups.Phasing
+)
+class HapCutToVcf
+( @arg(flag="v", doc="The original VCF provided to HAPCUT.") val vcf: PathToVcf,
+  @arg(flag="i", doc="The output file from HAPCUT.") val input: FilePath,
+  @arg(flag="o", doc="The output VCF with variants that were phased by HAPCUT.") val output: PathToVcf,
+  @arg(flag="u", doc="The output VCF with variants that were not phased by HAPCUT.") val unphasedOutput: PathToVcf,
+  @arg(flag="m", doc="Allow missing fields in the VCF header.") val allowMissingFieldsInVcfHeader: Boolean = true,
```
tfenne: I would vote for making this hardcoded and not exposing it as a parameter.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      |
+      |The VCF with phased variants will have the first variant of a haplotype block with a genotype "M\F", while the
+      |subsequent variants in that block having "M|F" as their genotype.  This first variant in a block can optionally
+      |have "M|F".
+      |
+    """,
+  group=ClpGroups.Phasing
+)
+class HapCutToVcf
+( @arg(flag="v", doc="The original VCF provided to HAPCUT.") val vcf: PathToVcf,
```
tfenne: Perhaps this is already covered by "provided to HAPCUT", but are there any limitations?  Must this be a single-sample file for example?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+  Io.assertReadable(vcf)
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+
+  override def execute(): Unit = {
+    val vcfReader      = new VCFFileReader(vcf.toFile, false)
+    val phasedWriter   = makeWriter(output, vcfReader, addPhasedHeaderLines=true)
+    val unphasedWriter = makeWriter(unphasedOutput, vcfReader, addPhasedHeaderLines=false)
+
+    val iterator = new MergingIterator(input, vcfReader, readBackedPhasingFormat)
```
tfenne: I know it's a private class but `MergingIterator` is a horrible name for this class!  How about `HapCutAndVcfMergingIterator`?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+    iterator.foreach {
+      case Left(ctx)  => phasedWriter.add(ctx)
+      case Right(ctx) => unphasedWriter.add(ctx)
+    }
+
+    CloserUtil.close(vcfReader)
```
tfenne: `vcfReader.safelyClose()`

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      .setReferenceDictionary(inputHeader.getSequenceDictionary)
+      .setOption(Options.INDEX_ON_THE_FLY)
+      .modifyOption(Options.ALLOW_MISSING_FIELDS_IN_HEADER, allowMissingFieldsInVcfHeader)
+    val writer: VariantContextWriter = builder.build
+
+    val outHeader = if (addPhasedHeaderLines) {
+      // get the header lines to skip
+      val headerLinesToSkip = HapCutCallGenotypeInfo.formatHeaderKeys
+        .filter { key => inputHeader.hasFormatLine(key) }
+        .map { key => inputHeader.getFormatHeaderLine(key) }
```
tfenne: if `inputHeader.getFormatHeaderLine(key)` returns null on a missing key, you could reduce this to:
```scala
HapCutCallGenotypeInfo.formatHeaderKeys.flatMap(key => Option(inputHeader.getFormatHeaderLine(key))
```

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+
+    val builder = new VariantContextWriterBuilder()
+      .setOutputFile(path.toFile)
+      .setReferenceDictionary(inputHeader.getSequenceDictionary)
+      .setOption(Options.INDEX_ON_THE_FLY)
+      .modifyOption(Options.ALLOW_MISSING_FIELDS_IN_HEADER, allowMissingFieldsInVcfHeader)
+    val writer: VariantContextWriter = builder.build
+
+    val outHeader = if (addPhasedHeaderLines) {
+      // get the header lines to skip
```
tfenne: `// get the header lines in the input header that we wish to skip/replace with our own definitions`

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      val headerLinesToSkip = HapCutCallGenotypeInfo.formatHeaderKeys
+        .filter { key => inputHeader.hasFormatLine(key) }
+        .map { key => inputHeader.getFormatHeaderLine(key) }
+      // get all the header lines from the input
+      val headerLines: util.Set[VCFHeaderLine] = new util.HashSet[VCFHeaderLine](
+        inputHeader.getMetaDataInSortedOrder.filter { line =>
+          if (headerLinesToSkip.contains(line)) {
+            logger.debug(s"Will overwrite header line: '${line.toString}")
```
tfenne: Is this logging still useful after debugging whatever was going on?  If not, maybe kill it and simplify the whole thing down?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+          }
+          else true
+        }.toSet
+      )
+      // add standard header lines
+      VCFStandardHeaderLines.addStandardFormatLines(headerLines, false, Genotype.PRIMARY_KEYS)
+      // create the header
+      val header: VCFHeader = new VCFHeader(headerLines, inputHeader.getSampleNamesInOrder)
+      // add the new header lines
+      HapCutCallGenotypeInfo.formatHeaderLines.foreach(header.addMetaDataLine)
```
tfenne: Why add the standard ones before building the header and these ones after?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+    }
+    else {
+      inputHeader
+    }
+    writer.writeHeader(outHeader)
+    writer
+  }
+}
+
+/** An iterator over all variants in the provided VCFFileReader.  The `input` is the output of HapCut.
```
tfenne: This line is confusing on first (and second) read.  I would suggest killing `The input is the output of HapCut.` from here, and adding it back as a `@param` doc line to the class.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
tfenne: Ok, having read a bunch more, I can't find anything in the spec about using `\` as a separator.  It says genotypes should always be specified with `|` or `/`.  What gives?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+  *
+  * If the variant in the VCFFileReader was not phased by HapCut, then we will return Right(VariantContext).  If the
+  * variant was phased by HapCut, we will return Left(VariantContext).
+  */
+private class MergingIterator(input: FilePath, vcfReader: VCFFileReader, readBackedPhasingFormat: Boolean) extends Iterator[Either[VariantContext,VariantContext]] {
+
+  private val sourceIterator = vcfReader.toStream.map { ctx =>
+    // NB: unsets the phase in case it was already set
+    new VariantContextBuilder(ctx).genotypes(new GenotypeBuilder(ctx.getGenotype(0)).phased(false).make()).make()
```
tfenne: Given that probably 99% of the time you would be using unphased input, do you want to make this conditional on the input genotypes being phased?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      call.pos.toLong,
+      call.pos.toLong,
+      allelesCollection
+    )
+      .computeEndFromAlleles(seqAsJavaList(alleles.toSeq), call.pos)
+      .genotypes(genotype).make()
+  }
+}
+
+/** Reads phased blocks and variants from a HapCut output file, returning a HapCut call for each input variant call to
```
tfenne: Something here (and in the usage of HapCutReader) doesn't make sense to me.  The comment says it returns an HapCut call for each input variant, then turns around and says that if HapCut didn't phase a variant `None` is returned because it's not present in the HapCut output.  But this reader class doesn't take in anything _but_ the HapCut output, so how can it know about variants that were not present in the output?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+
+  private var block: Option[HapCutBlockInfo] = None
+  private var calls: Iterator[Option[HapCutCall]] = Iterator.empty
+
+  // initialize
+  this.advance()
+
+  def hasNext(): Boolean = {
+    this.block match {
+      case None => this.advance()
```
tfenne: This can be reduced to 2 cases - drop the `None` and make the last case `_`.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+  }
+
+  def close(): Unit = {
+    this.reader.close()
+  }
+
+  /** Reads in the next block of data */
+  private def advance(): Boolean = {
+    val blockLine = reader.readLine() match {
+      case null => return false
```
tfenne: Using `return` here is not ideal.  You could move the rest of the method into the `case line` block, and just make the whole method a match statement that yields `true` or `false`.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      case line => line.trim
+    }
+    if (blockLine == "********") return this.advance()
```
tfenne: I worry a little bit about hardcoding exactly eight stars - seems like an odd choice from HapCut.  How would you feel about just testing for `line.startsWith('*')`?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+    val lastOffset = this.block match {
+      case None => 0
+      case Some(b) => b.offset + b.len - 1
```
tfenne: Could this be a method on the block?  I.e. b.offsetOfLastGenotype?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+    }
+    val currentBlock = HapCutBlockInfo(blockLine, -1)
```
tfenne: FWIW putting the parsing into the case class objects makes this much harder to follow.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+          val call = HapCutCall(line, currentBlock.offset)
+          if (currentBlock.phaseSet < 0) currentBlock.phaseSet = call.pos
+          (call.offset, call)
+      }
+      call
+    }).toMap
+    // add in the empty calls
+    this.calls = Stream.range(lastOffset+1, currentBlock.len+currentBlock.offset).map { i =>
+      callMap.get(i) match {
+        case None => None
```
tfenne: Why do you need this match?  You're always returning the thing that was found.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+    val (sourceContext, sourceOffset) = sourceIterator.next()
+    if (!hapCutIterator.hasNext()) {
+      Right(sourceContext)
+    }
+    else {
+      hapCutIterator.next() match {
+        case None => Right(sourceContext)
+        case Some(hapCutCall) =>
+          if (hapCutCall.offset != sourceOffset+1) throw new IllegalStateException("BUG: calls are out of order")
+          val hapCutContext = HapCutCall.toVariantContext(hapCutCall, sampleName)
```
tfenne: Why isn't this just a method on the `HapCutCall` class?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+  *
+  * @param phaseSet the phase set identifier for this block (typically the position of the first variant).
+  * @param offset the 1-based variant offset from the first variant given to HapCut.
+  * @param len the total number of variants that this block spans, including unphased variants.
+  * @param phased the total number of variants phased.
+  * @param span the number of bases this block spans.
+  * @param mecScore the minimum error correction (MEC) score.
+  * @param fragments the number of fragments contributing to variants in this block.
+  */
+private class HapCutBlockInfo private(var phaseSet: Int, val offset: Int, val len: Int, val phased: Int, val span: Int, val mecScore: Float, val fragments: Int)
```
tfenne: Is `phaseSet` a `var` on purpose?

tfenne: Looking forward, I see why, but I still don't like it.  Options I think would be nicer:
1) Parse the info in the block line outside the case class, and then only create the block when you have all the information
2) Make it a val and use `copy()` to set it once you have the value (and perhaps use `Option[Int]` instead of `Int` with a special value of `-1`.
3) Screw HapCut's block IDs's and just use the offset as the block ids since those are unique and show up in the block line.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Good point, the VCF spec says that all phased variants should have `|`, and recommends to use the `PS` format field to with the coordinate first variant in the phasing set for all variants phased in one block.  

GATK, funnily enough, doesn't do this, instead it does what I describe above, which is the first (genomic coordinate) variant in a phase set is set to `\`, then the rest `|` (see "Note that the first heterozygous genotype ..." http://gatkforums.broadinstitute.org/gatk/discussion/45/purpose-and-operation-of-read-backed-phasing). 

What if I make it conform to the spec (all phased variants are `\` and `PS` is set correctly), but have an option to output "GATK-style" phased blocks?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+
+private object HapCutBlockInfo {
+  val BlockLinePattern = """^BLOCK: offset: (\d+) len: (\d+) phased: (\d+) SPAN: (\d+) MECscore (\d+.\d+) fragments (\d+)$""".r
+
+  def apply(blockLine: String, phaseSet: Int): HapCutBlockInfo = {
+    BlockLinePattern findFirstMatchIn blockLine match {
```
tfenne: This can be written more idiomatically as the following (I think):
```scala
blockLine match {
  case BlockLinePattern(offset, len, phased, span, mecScore, fragments) =>
    new HapCutBlockInfo(phaseSet, offset.toInt, len.toInt, phased.toInt, span.toInt, mecScore.toFloat, fragments.toInt)
  case _ => throw new IllegalStateException("Could not parse block line: " + blockLine)
}
```

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Suppose you are trying to re-phase the truth data, rather than rely on your own calls, then this makes a lot of sense to unphase, so it is likely "at some point in the future it is 99% of the time" but your point is well taken.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      .attribute(LikelihoodChangeFormatTag, info.likelihoods.asJava)
+      .attribute(MaxLikelihoodChangeFormatTag, info.delta)
+      .attribute(MecReductionFormatTag, info.rMEC)
+  }
+}
+
+/** Information about a variant phased by HapCut.
+  *
+  * @param offset the 1-based offset of the variant relative to the variants given to HapCut.
+  * @param haplotype1 the first allele.
```
tfenne: hap1Allele?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
tfenne: No, I get it - that's why I suggested adding an `if (alreadyPhased)` check here and not just removing it.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+                                 val ref: String, val alts: String,
+                                 val genotype: String, val info: HapCutCallGenotypeInfo,
+                                 val firstInBlock: Boolean
+                                )
+
+private object HapCutCall {
+
+  /** Parse a variant line. */
+  def apply(callLine: String, blockOffset: Int): HapCutCall = {
+    val tokens = callLine.split("\t")
```
tfenne: This may end up being rather slow as it's a regex split on a single character.  You might want to try `htsjdk`'s `StringUtil.split()` instead.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
```diff
@@ -0,0 +1,413 @@
...
+      ref          = tokens(5),
+      alts         = tokens(6),
+      genotype     = tokens(7),
+      info         = HapCutCallGenotypeInfo(tokens(8)),
+      firstInBlock = blockOffset == offset
+    )
+  }
+
+  /** Converts the HapCut variant representation to a VariantContext. */
+  def toVariantContext(call: HapCutCall, sampleName: String): VariantContext = {
```
tfenne: I'm going to suggest that the following is an anti-pattern: methods in a companion object that take in, as a parameter, a single instance of the companion class and operate on it.

tfenne:
@nh13 A bunch of fairly minor comments, and then one or two more major things I'd like to discuss.

Firstly I think the phasing stuff in VCF is under-specified, and that the GATK docs recommend things that are not ideal.  E.g. I can't find anything in the specification about whether or not the first genotype in a phase set should use `|` or `/` as it's separator.  The example in section `1.1` of the VCF spec shows using `|` in the first genotype in the file(!), while the GATK page you pointed to suggests it should be `/` to differentiate from the prior phasing block.  I would suggest that phasing information in VCF is largely useless without annotated phase sets using the `PS` tag, and so all phased sets should have `PS` set as you are doing, in which case you always know which set/block a genotype belongs to.  Which leads me to saying that all genotypes in a phase set should just use `|`.

Secondly I think the output of a the tool would be better as a single VCF that has genotypes phased where they can be, and unphased where they cannot be.  In addition I would _"phase"_ homozygous genotypes internal to a block.  So if you had say, six genotypes, three of which are phased, two of which are homozygous and one is unphased, you might get the following genotypes in a single file:
```
0|1:PhaseSetA
1|1:PhaseSetA
0|1:PhaseSetA
0/1
0|0:PhaseSetA
1|0:PhaseSetA
```

I could see going either way on the homozygous genotypes, it's just easier to see things if those are also labeled as phased, even though their phase is not really a thing.  But having it all in once file is, I think, more generally useful, and you can always filter to just phased genotypes at runtime by only looking at genotypes with `PS` set.  What do you think?

Lastly, one thing I realized from reading through this and the HapCut doc a little bit is that HapCat is presumably not capable of outputting two parallel "zippered" phase sets.  E.g. imagine if you had intermittent or gappy long reads like the old PacBio strobed reads, that happened to give you phasing of every other het over a region, and another set of long reads that gave you the opposite every-other set.  You have enough information to generate two interleaved phase sets, but I don't think HapCut can do that given the restrictions of it's output format.

src/main/scala/com/fulcrumgenomics/cmdline/ClpGroups.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: I think the original intent was to have a VCF that was equivalent to what GATK's read-backed phasing produced, which did not produce A-B-c-D like you described, though it might well produce A-B-c-D where c was marked as not PASS, though I haven't checked.  Nonetheless, I think you are right to question the original intent and ask to just output a clean VCF with phasing information as the spec describes. Will do.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Awesome! It's in.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: The first sentence is wrong, thanks for catching that, done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: I think I will change this all around and use `Files.lines` to create an iterator and then it should be trivial to find the next line that starts with "BLOCK".

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Huh, I don't understand why it is harder to follow, since having it in the case class allows me not to be distracted by all the parsing code in this simple method.  What in the case class would you have liked seen here in regards to the parsing code?

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: This is moot, since I have moved it out anyhow, due to the comment about "phaseSet a var" later.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done, implemented 2).

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
nh13: Yup, it is :/

nh13: I will fix this here and elsewhere, as I am guilty in multiple places.

nh13:
@tfenne I really appreciate the thorough review, and thinking about the context of the tool and what it tries to do!

1. I have updated the output format to conform to the spec: 
  1. including all phased variants have their alleles separated by `|`, unlike GATK's output.
  2. annotating all phased variants with a PS tag.
2. The output is one file instead of two.
3. I incorporated almost all your suggestions, except one, where I learned about and then used the `java.nio.Files.lines` method, which is quite handy.
4. I agree that HAPCUT doesn't handle the case where you have zipped or overlapping phase sets, which can still be useful in many contexts.  I think this is something to remember after the low hanging fruit has been plucked (phased).  For now, most folks care about the length of the phased blocks, and certain types of switch errors (short, long, isolated), and almost all of these metrics assume no overlapping phased blocks.
5. I tried making the description of the tool more verbose, especially as it relates to the output format.
6. I think the main `execute` method is one of the shortest in any tools we have 👍

src/main/scala/com/fulcrumgenomics/phasing/HapCutToVcf.scala
tfenne: I think mostly it’s that I didn’t know the details of the hapcut output format, so it wasn’t clear to me (without detouring) what the values were that were parsed out of the block line, and how those values then influenced the parsing of the subsequent lines.  Now that I have read up on the HapCut format it’s probably a non-issue.

src/main/scala/com/fulcrumgenomics/vcf/HapCutToVcf.scala
```diff
@@ -0,0 +1,462 @@
...
+import scala.collection.JavaConverters._
+
+@clp(
+  description =
+    """
+      |Converts the output of HapCut to a VCF.
+      |
+      |The output of HAPCUT does not include all input variants, but simply those variants that are in phased blocks.
+      |This tool takes the original VCF and the output of HapCut, and produces a VCF containing both the variants that
+      |were phased the variants that were not phased, such that all variants in the original file are in the output.
```
tfenne: `|were phased _and_ the`

src/main/scala/com/fulcrumgenomics/vcf/HapCutToVcf.scala
```diff
@@ -0,0 +1,462 @@
...
+) extends FgBioTool with LazyLogging {
+  Io.assertReadable(vcf)
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+
+  override def execute(): Unit = {
+    val vcfReader      = new VCFFileReader(vcf.toFile, false)
+    val vcfWriter      = makeWriter(output, vcfReader)
+
+    new HapCutAndVcfMergingIterator(input, vcfReader, gatkPhasingFormat).foreach(vcfWriter.add)
```
tfenne: Very nice!

src/main/scala/com/fulcrumgenomics/vcf/HapCutToVcf.scala
```diff
@@ -0,0 +1,462 @@
...
+    }
+  }
+
+  /** Returns a new variant context with the phase unset (if it was set) and appropriately formatted (filtered as
+    * not phased if `gatkPhasingFormat` is true).
+    */
+  private def formatSourceContext(sourceContext: VariantContext): VariantContext = {
+    val builder = new VariantContextBuilder(sourceContext)
+    if (sourceContext.getGenotypes.exists(_.isPhased)) { // unset the phase if the input has phase set
+      builder.genotypes(sourceContext.getGenotypes().map(g => new GenotypeBuilder(g).phased(false).make()))
```
tfenne: Isn't this already handled on line 182-184?

tfenne:
Looks great @nh13.  I'm curious about the decision to set filters on any unphased variants and set `PASS` on any phased variants.  But I think it's fine as is and we could always add an option later to set vs. not set the filters based on phasing status.

tfenne:
Oh - one more thought: it might be nice to either open an issue at HapCut, or drop the author a line and let them know about the tool.  I'm sure other HapCut users would like to know about it.

src/main/scala/com/fulcrumgenomics/vcf/HapCutToVcf.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/HapCutToVcf.scala
nh13: Will move this out of here, so I only need to make at most one builder.

nh13:
@tfenne the details of how GATK represents phase can be seen here: http://gatkforums.broadinstitute.org/gatk/discussion/45/purpose-and-operation-of-read-backed-phasing

In particular, see "More detailed aspects of semantics of phasing in the VCF format".  I am deviating from GATK's output format by not "appending" homozygous variants to the previous phase set.  Note, this output format is only used when the gatk format option is set to true, which is false by default.

PR #49, opened 2016-07-06
Exposing some methods for other tools to use Consensus Calling.

nh13:


nh13:
@tfenne I moved some methods from inside the consensus caller to a trait, so that other tools that call consensus reads can benefit from some of these common computations, especially some of the pre-computed phred to log probabilities and adjustment of base qualities.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -87,6 +87,89 @@ case class ConsensusRead(bases: Array[Byte], quals: Array[Byte]) {
...
+      //         = X + Y + XY*(2/3 - 6/3)
+      //         = X + Y - 4/3*XY
+      val term1 = LogProbability.or(prErrorTrialOne, prErrorTrialTwo) // X + Y
+      val term2 = LogFourThirds + prErrorTrialOne + prErrorTrialTwo // 4/3*XY
+      LogProbability.aOrNotB(term1, term2)
+    }
+  }
+
+  /** Pre-computes the the log-scale probabilities of an error for each a phred-scaled base quality from 0-100. */
+  protected lazy val phredToLogProbError: Array[Double] = Range(0, 100).toArray.map(p => {
```
tfenne: Turning these into lazy vals is going to add a lot of overhead - previously use of these was a direct access into an array, now each access into the array will be via a method call that does an initialization check (with, I think, synchronization overhead).

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -87,6 +87,89 @@ case class ConsensusRead(bases: Array[Byte], quals: Array[Byte]) {
...
+    val e1 = LogProbability.fromPhredScore(errorRatePostUmi)
+    val e2 = LogProbability.fromPhredScore(p.toByte)
+    probabilityOfErrorTwoTrials(e1, e2)
+  })
+
+  /** Pre-computes the the log-scale probabilities of an not an error for each a phred-scaled base quality from 0-100. */
+  protected lazy val phredToLogProbCorrect: Array[Double] = phredToLogProbError.map(LogProbability.not)
```
tfenne: Ditto here.

tfenne:
None of this code is really new, so I don't have many line comments to make.  But I do have a few general points:

1. Having this as a trait sits wrong with me.  I'm not keen on using traits as a way to "inject" a bunch of shared utility methods into classes.  I'd be happier to see this as a simple class that users instantiate and call methods on instead of mixing in.

2. It also seems like a fairly random grab-bag of methods that went into the trait - I'm sure based on what you needed elsewhere, but it doesn't feel cohesive.

3. None of the tests got updated, which means the trait is only tested through it's usage in `ConsensusCaller`, which is not ideal.

4. Having `errorRatePostUmi` defined in the trait makes it feel like this is too tied to the single-umi case, limiting re-usability.

I think it would be better to spend a little extra time and decide what would make good sense here given the various consensus callers we want to write, and then to extract a small class that is self-contained and performs the various operations we want.

tfenne:
Closing in favor of #56

PR #50, opened 2016-07-07
Adding more precision to the "or" and "aOrNotB" methods.

nh13:
This implements some heuristics for computing `log(1 - exp(x))` and `log(1+exp(x))` accurately.  Both of these methods are used in adding and subtracting probabilities while in log space.  The implementation comes from `R` (pronounced Aargh in a pirate's voice): https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/NumericTypes.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/NumericTypes.scala
@@ -111,12 +111,33 @@ object NumericTypes {
       else fromPhredScore(s.toByte)
     }
 
+    /**
+      * Precise computation of log(1+exp(x)) for use in the `or` method.
+      * See Equation (10) in https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf
+      */
+    private def log1pexp(value: Double): Double = {
+      if (value <= -37) exp(value)
+      else if (value <= 18) log1p(exp(value))
+      else if (value <= 33.3) value + exp(-value)
+      else value
+    }
+
+    /**
+      * Precise computation of log(1-exp(x)) for use in the `aOrNotB` method.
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/util/NumericTypes.scala
```diff
@@ -111,26 +111,47 @@ object NumericTypes {
...
     def or(values: Array[LogProbability]): LogProbability = {
       if (values.forall(_.isNegInfinity)) Double.NegativeInfinity
       else {
         val (minValue, minValueIndex) = MathUtil.minWithIndex(values)
-        var sum: LogProbability = 0.0
+        var sum: LogProbability = minValue
         var i = 0
         while (i < values.length) {
-          if (minValueIndex != i) sum += exp(values(i) - minValue)
+          if (minValueIndex != i) sum = or(sum, values(i))
```
nh13: I am not going to pretend that I can do the math anymore, so I am just going to move this to a rolling sum.

src/test/scala/com/fulcrumgenomics/util/NumericTypesTest.scala
```diff
@@ -104,5 +105,6 @@ class NumericTypesTest extends UnitSpec {
 
   it should "or multiple values in log space" in {
     LogProbability.or(Array(log(10), log(10), log(10))) shouldBe log(30) +- 0.00001
+    LogProbability.or(Array(-718.3947756282423, -8.404216861178751, -710.0756239287693, -718.3947756282423)) shouldBe -8.404216861178751 +- 0.00001
```
nh13: We could use a test for precision with `aOrNotB` but I don't have one.

tfenne:
Looks reasonable.  Merge once the tests pass.

PR #55, opened 2016-07-13
Added options to FindTechnicalReads to output all reads and tag match…

tfenne:
…es with an attribute.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
@@ -30,10 +30,11 @@ import com.fulcrumgenomics.util.{IlluminaAdapters, Io, ProgressLogger}
 import dagr.commons.util.LazyLogging
 import dagr.sopt._
 import htsjdk.samtools.util.SequenceUtil
-import htsjdk.samtools.{SAMFileWriterFactory, SamReaderFactory}
+import htsjdk.samtools.{SAMFileWriterFactory, SAMRecord, SamReaderFactory}
 
 import scala.annotation.tailrec
 import scala.collection.JavaConversions.asScalaIterator
+import scala.collection.mutable
 
 @clp(group = ClpGroups.SamOrBam, description =
   """
@@ -47,8 +48,12 @@ import scala.collection.JavaConversions.asScalaIterator
      |technical sequences.  Sub-sequences are generated from both the sequences and the
      |reverse complement of the sequences, ignoring any sub-sequences that include 'N's.
      |
-     |The output BAM file will contain all reads that matched to a sub-sequence of the
-     |technical sequences and, if the read is paired, the read's mate pair.
+     |By default the output BAM file will contain all reads that matched to a sub-sequence of the
+     |technical sequences and, if the read is paired, the read's mate pair.  An option is
+     |available to apply a tag to matched reads (--tag/-t), and if specified each matching
+     |read will be tagged with the 0-based index of the sequence to which it matched. In
+     |combination with tagging it is possible to output all reads (-a/--all-reads) which will
+     |re-create the input BAM with the addition of tags on matching reads.
      |
      |The default set of sequences include a range of different Illumina adapter sequences
      |with the sample index/barcode region masked to Ns.
@@ -59,15 +64,26 @@ class FindTechnicalReads
   @arg(flag="o", doc="Output SAM or BAM file") val output: PathToBam,
   @arg(flag="m", doc="The number of bases at the start of the read to match against.") val matchLength: Int = 15,
   @arg(flag="e", doc="The maximum number of errors in the matched region.") val maxErrors: Int = 1,
-  @arg(flag="s", doc="The set of technical sequences to look for.") val sequences: Seq[String] = IlluminaAdapters.all.flatMap(_.both)
+  @arg(flag="s", doc="The set of technical sequences to look for.") val sequences: Seq[String] = IlluminaAdapters.all.flatMap(_.both),
+  @arg(flag="a", doc="Output all reads.") val allReads: Boolean = false,
+  @arg(flag="t", doc="Tag to set to indicate a read is a technical sequence.") val tag: Option[String] = None
 ) extends FgBioTool with LazyLogging {
+
   Io.assertReadable(input)
   Io.assertCanWriteFile(output)
+  if (allReads && tag.isEmpty) fail("Tag option must be used when outputting all reads.")
+
+  private val sequenceToIndex = sequences.zipWithIndex.toMap
 
   override def execute(): Unit = {
-    val in       = SamReaderFactory.make().open(input.toFile)
-    val iter     = if (in.hasIndex) in.queryUnmapped().buffered
-                   else in.iterator().filter(r => r.getReadUnmappedFlag && (!r.getReadPairedFlag || r.getMateUnmappedFlag)).buffered
+    // Built the appropriate input iterator
+    val in = SamReaderFactory.make().open(input.toFile)
+    val iter = (allReads, in.hasIndex) match {
+      case (true, _)      => in.iterator().buffered
+      case (false, true)  => in.queryUnmapped().buffered
+      case (false, false) => in.iterator().filter(isUnmappedTemplate).buffered
+    }
+
     val out      = new SAMFileWriterFactory().makeWriter(in.getFileHeader, true, output.toFile, null)
     val matcher  = new Matcher(sequences=sequences, matchLength=matchLength, maxErrors=maxErrors)
     val progress = new ProgressLogger(logger, unit=250000)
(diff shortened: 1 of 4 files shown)


tfenne:
@nh13 Any chance you could do a quick review this morning?

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
```diff
@@ -59,15 +64,26 @@ class FindTechnicalReads
...
   @arg(flag="e", doc="The maximum number of errors in the matched region.") val maxErrors: Int = 1,
-  @arg(flag="s", doc="The set of technical sequences to look for.") val sequences: Seq[String] = IlluminaAdapters.all.flatMap(_.both)
+  @arg(flag="s", doc="The set of technical sequences to look for.") val sequences: Seq[String] = IlluminaAdapters.all.flatMap(_.both),
+  @arg(flag="a", doc="Output all reads.") val allReads: Boolean = false,
+  @arg(flag="t", doc="Tag to set to indicate a read is a technical sequence.") val tag: Option[String] = None
 ) extends FgBioTool with LazyLogging {
+
   Io.assertReadable(input)
   Io.assertCanWriteFile(output)
+  if (allReads && tag.isEmpty) fail("Tag must be specified in all reads mode.")
```
nh13: How about "Tag option must be used when outputting all reads"?

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
```diff
@@ -47,8 +48,12 @@ import scala.collection.JavaConversions.asScalaIterator
      |technical sequences.  Sub-sequences are generated from both the sequences and the
      |reverse complement of the sequences, ignoring any sub-sequences that include 'N's.
      |
-     |The output BAM file will contain all reads that matched to a sub-sequence of the
-     |technical sequences and, if the read is paired, the read's mate pair.
+     |By default the output BAM file will contain all reads that matched to a sub-sequence of the
+     |technical sequences and, if the read is paired, the read's mate pair.  An option is
+     |available to apply a tag to matched reads (--tag/-t), and if specified each matching
+     |read will be tagged with the 0-based index of the sequence to which it matched. In
```
nh13: What about matches to multiple technical sequences, not just one or the first?

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
```diff
@@ -77,33 +93,66 @@ class FindTechnicalReads
...
-        recs.foreach(out.addAlignment)
-        n += recs.size
+
+      val isTechnical = if (isUnmappedTemplate(r1)) {
+        val hit      = recs.toStream.flatMap(r => matcher.findMatch(r.getReadBases)).headOption
+        val hitIndex = hit.map(h => sequenceToIndex(h))
+
+        for (t <- tag; hi <- hitIndex; r <- recs) {
+          r.setAttribute(t, hi)
+          n += 1
```
nh13: Is n only updated of a tag is given?

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
```diff
@@ -77,33 +93,66 @@ class FindTechnicalReads
       val r1 = iter.next()
       val r2 = if (iter.hasNext && iter.head.getReadName == r1.getReadName) Some(iter.next()) else None
       val recs = Seq(Some(r1), r2).flatten
-      if (recs.exists(r => matcher.matches(r.getReadBases))) {
-        recs.foreach(out.addAlignment)
-        n += recs.size
+
+      val isTechnical = if (isUnmappedTemplate(r1)) {
+        val hit      = recs.toStream.flatMap(r => matcher.findMatch(r.getReadBases)).headOption
```
nh13: Why not match on the headOption which avoids the map on hit below?

nh13: Also, you could zipWithIndex on the recs then return the first index that has a match:
'recs.zipWithIndex...'

nh13: Nevermind, reading this on my phone, looks fine, I just prefer matching:

```
recs.toStream.flatMap(r => matcher.findMatch(r.getReadBases)).headOption match {
  case None => false
  case Some(hit) => 
    val hitIndex = sequentToIndex(hit)
    ....
}
```

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
```diff
@@ -77,33 +93,66 @@ class FindTechnicalReads
...
+      else {
+        false
       }
 
+      if (isTechnical || allReads) recs.foreach(out.addAlignment)
       recs.foreach(progress.record)
     }
 
-    logger.info(s"Wrote out ${n} reads that matched to technical sequences.")
+    logger.info(s"Found ${n} reads that matched to technical sequences.")
```
nh13: `${n}` -> `$n`?

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
```diff
@@ -77,33 +93,66 @@ class FindTechnicalReads
...
     in.safelyClose()
     out.close()
   }
+
+  /** Returns true if all reads from the template are unmapped, otherwise false. */
+  private[bam] def isUnmappedTemplate(rec: SAMRecord) : Boolean = rec.getReadUnmappedFlag && (!rec.getReadPairedFlag || rec.getMateUnmappedFlag)
```
nh13: why is this package private and not just private?

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
nh13: Also, since you care only about the index of the hit, why not just have a method on matcher `findMatchIndex`?

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
```diff
@@ -77,33 +93,66 @@ class FindTechnicalReads
...
     read.length >= matchLength && kmers.exists(kmer => matches(read=read, kmer=kmer, len=matchLength))
   }
 
+  /** Identifies the first sequence with a matching kmer. */
+  def findMatch(read: Array[Byte]): Option[String] = {
+    if (read.length < matchLength) {
+      None
+    }
+    else {
+      kmers.find(kmer => matches(read=read, kmer=kmer, len=matchLength)) match {
```
nh13: for `findMatchIndex` could use `kmers.zipWithIndex.find { case (kmer, i) => matches(...) } .map (_._2) ...

src/test/scala/com/fulcrumgenomics/testing/SamRecordSetBuilder.scala
```diff
@@ -102,23 +102,33 @@ class SamRecordSetBuilder(val readLength: Int=100,
...
-              start: Int,
+              contig: Int = 0,
+              start: Int = SAMRecord.NO_ALIGNMENT_START,
               unmapped: Boolean = false,
               cigar: String = readLength + "M",
               mapq: Int = 60,
               strand: Strand = Plus,
               baseQuality: Int = this.baseQuality,
-              attrs: Map[String,AnyRef] = Map.empty) : SAMRecord = {
+              attrs: Map[String,AnyRef] = Map.empty) : Option[SAMRecord] = {
```
nh13: :/ I actually rely on this elsewhere, so changing the return type is a good idea to make it similar to `addPair`, we should be careful about changing things like this.

nh13:
@tfenne I think there is an issue with updating the count of matching technical sequences for logging, a question about whether we want to output all indices for technical sequences that match, and a few optimizations that may avoid some lookups to find the index of a match.  Your call on the latter, but take a look at the first two.

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
tfenne: That's going to make everything much slower as it would have to do an exhaustive search instead of aborting at the first kmer that matched.

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
nh13: But is it what the user wants?  I may want to know which reads matched a specific technical sequence which would require me to run the tool once per technical sequence.

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
tfenne: I have a strong preference for using `${}` over bar `$` in these strings - I think it makes it both much easier to spot where the interpolation is happening and also to adjust if necessary, e.g. `${n}` -> `${n-1}`.

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
tfenne: The match is tidier, you're right.  I wanted to keep the `Matcher` working in just sequence as I figure it might be more generally useful at some point, and it's only in the tagging that we currently care about the index.

src/main/scala/com/fulcrumgenomics/bam/FindTechnicalReads.scala
tfenne: Oops, good catch, thanks.

src/test/scala/com/fulcrumgenomics/testing/SamRecordSetBuilder.scala
tfenne: Agreed - I fixed your test too though 😄

nh13: Not in this repo!

PR #56, opened 2016-07-16
Extracting re-usable parts of the consensus caller.

tfenne:
@nh13 This is still a work in progress, but I would like you to take a look please and see if you agree with the general direction this is taking.

A few notes about where I'm at and what else I'd like to do:

1. LogProbability.probabilityOfErrorTwoTrials needs independent tests
2. ConsensusCaller needs independent tests
3. Should SourceRead/ConsensusRead and methods to go from SourceRead -> ConsensusRead exist in ConsensusCaller?  I haven't done this for now because I really want to keep ConsensusCaller as pileup oriented as possible, given it's other potential uses.  I also don't really want to drag some of the other parameters in, like minReads, minMeanQuality etc.  I think that perhaps if we end up having multiple consensus callers that work simililarly to the Vanilla one, we could sub-class and add those methods.

In addition, I think maxReadLength should be computed taking into account minReads.  Because we also filter consensus reads based on their mean quality, I think that we should probably only build a consensus read up to the point where we have enough reads to keep going.  E.g. imagine minReads=3 and minMeanConsensusQual=20 and we have four reads, all high quality, 2 at 100bp each and 2 more at 150bp each.  The current implementation will create a consensus read 150bp long with 50bp of Ns at the end, and then discard the entire read for being too low quality.  What do you think?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2016 Fulcrum Genomics
+ * Copyright (c) 2016 Fulcrum Genomics LLC
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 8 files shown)


src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -20,448 +20,167 @@
...
+class ConsensusCaller(errorRatePreLabeling:  PhredScore,
+                      errorRatePostLabeling: PhredScore,
+                      rawBaseQualityShift:   PhredScore = 0.toByte,
+                      maxRawBaseQuality:     PhredScore = PhredScore.MaxValue
+                      ) {
+
+  assert(maxRawBaseQuality >= PhredScore.MinValue, s"maxBaseQuality must be >= ${PhredScore.MinValue}")
+  assert(maxRawBaseQuality <= PhredScore.MaxValue, s"maxBaseQuality must be <= ${PhredScore.MaxValue}")
+
+  private val LnErrorRatePreUmi  = LogProbability.fromPhredScore(errorRatePreLabeling)
```
nh13: `LnErrorRatePreLabeling` and `LnErrorRatePostLabeling`?

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -20,448 +20,167 @@
...
-
-    val qs = read.quals
-    var i = 0
-    while (i < len) {
-      val q = qs(i)
-      val newQ = Math.min(this.options.maxBaseQuality, Math.max(PhredScore.MinValue, q - this.options.baseQualityShift))
-      pErrors(i)   = this.phredToLogProbError(newQ)
-      pCorrects(i) = this.phredToLogProbCorrect(newQ)
-      i += 1
+  class ConsensusBaseBuilder () {
```
nh13: what's the purpose of `()` here ?

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -20,448 +20,167 @@
...
-    * unpaired reads, first of pair reads, and second of pair reads.
-    */
-  protected def subGroupRecords(records: Seq[SAMRecord]): (Seq[SAMRecord], Seq[SAMRecord],Seq[SAMRecord]) = {
-    val fragments    = records.filter { rec => !rec.getReadPairedFlag }
-    val firstOfPair  = records.filter { rec => rec.getReadPairedFlag && rec.getFirstOfPairFlag }
-    val secondOfPair = records.filter { rec => rec.getReadPairedFlag && rec.getSecondOfPairFlag }
-    (fragments, firstOfPair, secondOfPair)
-  }
+  /** Pre-computes the the log-scale probabilities of an not an error for each a phred-scaled base quality from 0-100. */
+  private val phredToAdjustedLogProbCorrect: Array[Double] = phredToAdjustedLogProbError.map(LogProbability.not)
```
nh13: `0-100` -> `0-127`

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
nh13: Is it meant to be private so you have to call the `builder` method?

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,359 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics
```
nh13: LLC

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,359 @@
...
+  val DefaultBaseQualityShift: PhredScore            = 10.toByte
+  val DefaultMinConsensusBaseQuality: PhredScore     = 13.toByte
+  val DefaultMinReads: Int                           = 1
+  val DefaultMinMeanConsensusBaseQuality: PhredScore = 13.toByte
+  val DefaultRequireConsensusForBothPairs: Boolean   = true
+}
+
+/** Holds the parameters/options for consensus calling. */
+case class ConsensusCallerOptions(tag: String                             = DefaultTag,
+                                  errorRatePreUmi: PhredScore             = DefaultErrorRatePreUmi,
```
nh13: I like the `Labeling` name for this variable, want to change this throughout?

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,359 @@
...
+  val progress: Option[ProgressLogger] = None
+) extends Iterator[SAMRecord] {
+  /** The type of consensus read to output. */
+  private object ReadType extends Enumeration {
+    val Fragment, FirstOfPair, SecondOfPair = Value
+  }
+  import ReadType._
+
+  private val DnaBasesUpperCase: Array[Byte] = Array('A', 'C', 'G', 'T').map(_.toByte)
+  private val LogThree      = LogProbability.toLogProbability(3.0)
```
nh13: spaces or alignment

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,359 @@
...
+  private val caller = new ConsensusCaller(errorRatePreLabeling=options.errorRatePreUmi,
+                                           errorRatePostLabeling=options.errorRatePostUmi,
+                                           maxRawBaseQuality = options.maxBaseQuality,
+                                           rawBaseQualityShift = options.baseQualityShift)
```
nh13: alignment

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,359 @@
...
+  val DefaultMinMeanConsensusBaseQuality: PhredScore = 13.toByte
+  val DefaultRequireConsensusForBothPairs: Boolean   = true
+}
+
+/** Holds the parameters/options for consensus calling. */
+case class ConsensusCallerOptions(tag: String                             = DefaultTag,
+                                  errorRatePreUmi: PhredScore             = DefaultErrorRatePreUmi,
+                                  errorRatePostUmi: PhredScore            = DefaultErrorRatePostUmi,
+                                  maxBaseQuality: PhredScore              = DefaultMaxBaseQuality,
+                                  baseQualityShift: PhredScore            = DefaultBaseQualityShift,
```
nh13: also `rawBaseQualityShift`

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,359 @@
...
+
+      // check that the mean base quality is high enough
+      if (MathUtil.mean(consensusQuals) >= options.minMeanConsensusBaseQuality)
+        Some(ConsensusRead(consensusBases, consensusQuals))
+      else
+        None
+    }
+  }
+
+
```
nh13: nl

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -20,448 +20,167 @@
...
+                      ) {
+
+  assert(maxRawBaseQuality >= PhredScore.MinValue, s"maxBaseQuality must be >= ${PhredScore.MinValue}")
+  assert(maxRawBaseQuality <= PhredScore.MaxValue, s"maxBaseQuality must be <= ${PhredScore.MaxValue}")
+
+  private val LnErrorRatePreUmi  = LogProbability.fromPhredScore(errorRatePreLabeling)
+  private val LnErrorRatePostUmi = LogProbability.fromPhredScore(errorRatePostLabeling)
+  private val DnaBasesUpperCase: Array[Base] = Array('A', 'C', 'G', 'T').map(_.toByte)
+  private val DnaBaseCount  = DnaBasesUpperCase.length
   private val LogThree      = LogProbability.toLogProbability(3.0)
```
nh13: Could use `LnThree` and `LnFour` in `NumericTypes`

nh13:
(1) I think `ConsensusCaller` should operate only a single pileup, and I would rename it to `SingleSiteConsensusCaller` or `PileupConsensusCaller` or something else to reflect that.  We can certainly have other methods that do filtering or call the entire read in another convenience class, but I think each implementation of a end-to-end consensus caller will be specialized, for example a graph based approach of grouping bases to call consensuses, or reads that have offsets relative to each other, or combining multiple of pileups.

(2) I do think you are right about stopping after falling below the minimum number of reads, as that is a bad unintended side effect to have a stretch of Ns and then falling below the mean base quality.  The alternative is to have the mean base quality calculation ignore `N`s.  Your call.

Looks good so far.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
tfenne: Good catch.  Thanks.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
tfenne: I had it as private, and then the outer class couldn't call it!  I'll kill the parens.  I think even though it's public the outer class is the only one that can construct it anyway.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: Doh.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: I'm not sure.  I think this class should have been renamed to `VanillaUmiConsensusCallingOptions`, in which case I think preUmi/postUmi makes most sense.  I wanted to pick another name as it's conceivable you have some other way to group reads without UMIs; e.g. you could have clones of a single molecule in a well and then make a library such that the entire library/RG is from one source molecule.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: Fixed.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: Done.

tfenne:
Getting there.  Still need to write a few more tests though.

tfenne:
@nh13 This should be good to go once the build passes.  I'm going to leave this to you to merge if you're happy or punt back to me if you want to re-review and find anything else.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -20,448 +20,166 @@
...
-      readType = FirstOfPair
-    )
-    // second end
-    createAndEnqueueSamRecord(
-      records  = secondRecords,
-      read     = secondRead.getOrElse(dummyConsensusRead(firstRead.get)),
-      readName = readName,
-      readType = SecondOfPair
-    )
+  def adjustedTruthProbability(qual: PhredScore): LogProbability = {
```
nh13: I don't see this used anymore.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,367 @@
...
+  private def consensusReadLength(reads: Seq[SourceRead]): Int = {
+    val n = this.options.minReads
+    if (reads.length < n) throw new IllegalArgumentException("Too few reads to create a consensus.")
+
+    reads.map(_.length).sortBy(len => -len).drop(n-1).head
+  }
+
+  /** Gets the longest common prefix of the given strings, None if there is only one string or if there is an empty string. */
+  @annotation.tailrec
+  private def longestCommonPrefix(strs: Iterable[String], accu: Option[String] = None): Option[String] = {
```
nh13: If this is not used, we should remove it.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,367 @@
...
+        rejectRecords(records=firstOfPair ++ secondOfPair)
+      case (Some(_), None) | (None, Some(_)) if needBothPairs => // reject
+        rejectRecords(records=firstOfPair ++ secondOfPair)
+      case (firstRead, secondRead)                            => // output
+        this.createAndEnqueueSamRecordPair(firstRecords=firstOfPair, firstRead=firstRead, secondRecords=secondOfPair, secondRead=secondRead)
+        success = true
+    }
+
+    if (success) true // consensus created
+    else if (this.iter.isEmpty) false // no more records, don't try again
```
nh13: Weird that the `else if` and `else` are not covered, is that true?

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,367 @@
...
+    * longest common prefix of all read names, and "<idx>" is the 1-based consensus read index.  If no prefix was found,
+    * "CONSENSUS" is used.  If no records are given, the empty string is returned.
+    */
+  private def nextReadName(records: Seq[SAMRecord]): String = {
+    if (records.isEmpty) ""
+    else this.options.tag + ":" + records.head.getStringAttribute(this.options.tag)
+//    val curIdx = readIdx
+//    readIdx += 1
+//    val prefix = readNamePrefix.getOrElse(longestCommonPrefix(records.map(_.getReadName)).getOrElse("CONSENSUS"))
+//    s"$prefix:$curIdx"
```
nh13: remove if not used.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -0,0 +1,367 @@
...
+    (fragments, firstOfPair, secondOfPair)
+  }
+
+  /** Returns the next read name with format "<prefix>:<idx>", where "<prefix>" is either the supplied prefix or the
+    * longest common prefix of all read names, and "<idx>" is the 1-based consensus read index.  If no prefix was found,
+    * "CONSENSUS" is used.  If no records are given, the empty string is returned.
+    */
+  private def nextReadName(records: Seq[SAMRecord]): String = {
+    if (records.isEmpty) ""
+    else this.options.tag + ":" + records.head.getStringAttribute(this.options.tag)
```
nh13: I am not sure the read name is going to be unique!

Also, the doc for this method needs updating.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -20,448 +20,166 @@
...
+                      ) {
+
+  assert(maxRawBaseQuality >= PhredScore.MinValue, s"maxBaseQuality must be >= ${PhredScore.MinValue}")
+  assert(maxRawBaseQuality <= PhredScore.MaxValue, s"maxBaseQuality must be <= ${PhredScore.MaxValue}")
+
+  private val LnErrorRatePreLabeling  = LogProbability.fromPhredScore(errorRatePreLabeling)
+  private val LnErrorRatePostLabeling = LogProbability.fromPhredScore(errorRatePostLabeling)
+  private val DnaBasesUpperCase: Array[Base] = Array('A', 'C', 'G', 'T').map(_.toByte)
+  private val DnaBaseCount  = DnaBasesUpperCase.length
+  private val LnFourThirds = LogProbability.normalizeByLogProbability(LnFour, LnThree)
```
nh13: Why not use the value in `LogProbability` (make it public)?

src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -0,0 +1,357 @@
...
+        consensus.quals      shouldBe outputQuals
+    }
+  }
+
+  it should "apply the post-umi-error-rate when it has probability greater than zero" in {
+    val caller = cc(cco(
+      errorRatePreUmi             = PhredScore.MaxValue,
+      errorRatePostUmi            = 10.toByte,
+      maxRawBaseQuality              = PhredScore.MaxValue,
+      rawBaseQualityShift            = 0.toByte,
```
nh13: What is this alignment blasphemy?

nh13:
@tfenne I realize that a few things I commented on weren't introduced by you, but it would be nice to clean up anyhow.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
tfenne: It's not - but I figured I would leave it to be symmetric with `adjustedErrorProbability` in case we did come up with a case where we want to apply adjustments to reads prior to feeding them into the pileup - which the API still supports.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: I'll update the doc - but the MI tags we're generating in GroupReadsByUmi are unique within the BAM file they are generated for.  Though you're right - not globally unique.  Perhaps we should also prepend something else to ensure global uniqueness?

tfenne: I've added something that either a) uses the input prefix or b) concatenates unique library names, or unique RG ids if library names are absent.  I think that's probably good enough.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
tfenne: This is unused, so I deleted it.

src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
tfenne: Heh, automated refactoring (adding `Raw` to those names).

tfenne:
@nh13 I've address the latest round and squashed ready for merging.  Your turn.

PR #60, opened 2016-07-19
Adding a simple NumericCounter.

nh13:


src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
```diff
@@ -0,0 +1,105 @@
...
+  * Super-simple class for counting occurrences of any [Numeric]. Will return
+  * zero for any item that has not been counted yet.  Implements some useful methods
+  * to compute statistics when the objects being counted are numeric types.
+  */
+class NumericCounter[T](implicit n: Numeric[T]) extends SimpleCounter[T] {
+  import n._
+
+  /** Returns the mean as a [Double]. */
+  def mean(): Double = {
+    var len = 0L
```
tfenne: Maybe just add an efficient `size` method to `SimpleCounter`?

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
```diff
@@ -0,0 +1,105 @@
...
+    val sum = iterator.map { case (k, v) => len += v; k.toDouble() * v }.sum
+    sum / len
+  }
+
+  /** Returns the standard deviation as a [Double]. */
+  def standardDeviation(): Double = {
```
tfenne: Two things:
1) Could we call this `stddev`?
2) How about adding a default parameter so it's `stddev(mean: Double = mean())` so that it works as it does now, but if the caller wants both they're not paying to generate the mean twice?

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  /** Returns the key with the greatest count (mode of the distribution). */
+  def mode(): T = iterator.maxBy(_._2)._1
+
+  /** Returns the median key.  If the number of values is even, consider using [medianAsDouble]. */
+  def median(): T = this.div(medianTimesTwo(), fromInt(2))
+
+  /** Returns the median key as a [Double].  This may be useful when the number of values is even and so the
+    * median is an average. */
+  def medianAsDouble(): Double = medianTimesTwo().toDouble() / 2.0
```
tfenne: I think I would just rename this to `median` and kill the other one if it were me.

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  /** Returns the median times two, so that methods can implement the divide as desired. */
+  private def medianTimesTwo(): T = {
```
tfenne: Holy method batman!  This seems complicated!  How about we change SimpleCounter to a) take a type tag and b) if the type extends Ordered we use a `SortedMap` instead of a `Map`?  That's probably a nice change to have in general, and this this method would boil down to something like:

```scala
val len = size
val sum = if (len % 2 == 1) {
    val item iterator.drop(len/2).head
    Iterator(item, item).sum
}
else {
    iterator.drop((len/2) - 1).take(2).sum
}

sum / 2
```

src/test/scala/com/fulcrumgenomics/util/NumericCounterTest.scala
```diff
@@ -0,0 +1,56 @@
...
+    val counter = new NumericCounter[Double]()
+    Stream.range(1, 10, 1).foreach { i => counter.count(i / 10.0, 1) }
+    counter.count(1 / 10.0)
+    counter.mean() shouldBe 0.46 +- 0.0001
+    counter.standardDeviation() shouldBe  0.2875181 +- 0.0001
+    counter.mode() shouldBe 0.1
+    counter.median() shouldBe 0.45
+    counter.medianAsDouble() shouldBe 0.45
+  }
+}
```
tfenne: Could you please add tests to make sure that the methods all do the "right thing" on counters that have:
  - zero counts added
  - a single count added
  - two counts added

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
tfenne: Realized that it's not quite so simple (counts vs. raw data) but I think keeping the map sorted if possible would still simplify things considerably.

tfenne:
Back to you @nh13

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
nh13: Added two methods to return the # of objects being tracked and the total sum of counts across all objects.

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
nh13: done

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
nh13: Since each each key is weighted, we can't perform a simple operation, and need to search through to calculate a running sum.

nh13:
@tfenne you're welcome for one more review.  Unfortunately the median needs to be complicated still.

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
```diff
@@ -39,17 +37,23 @@ object SimpleCounter {
   * zero for any item that has not been counted yet.  
   */
 class SimpleCounter[T] extends Iterable[(T, Long)] {
-  private val counts = mutable.Map[T,Long]()
+  import scala.collection.JavaConversions._
+
+  private val counts = new java.util.TreeMap[T,Long]()
```
tfenne: What happened to testing to see if `T` was `Comparable`?  This will now fail if given a non-`Comparable` type T.

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
```diff
@@ -39,17 +37,23 @@ object SimpleCounter {
...
+  private var _totalCounts: Long = 0L
 
   /** Increment the count of object T by 1. */
   def count(t: T): Unit = count(t, 1)
 
   /** Increment the count of object T by the specified value. */
-  def count(t: T, n: Long): Unit = counts(t) = counts.getOrElse(t, 0L) + n
+  def count(t: T, n: Long): Unit = { _totalCounts += n; counts(t) = counts.getOrDefault(t, 0L) + n }
```
tfenne: Is this going to cause multiple implicit conversions every time count is called?

src/test/scala/com/fulcrumgenomics/util/NumericCounterTest.scala
```diff
@@ -0,0 +1,97 @@
...
+  it should "handle when a single object is counted" in {
+    val counter = new NumericCounter[Double]()
+    counter.count(10.0)
+    counter.mean() shouldBe 10.0
+    counter.stddev() shouldBe 0.0
+    counter.mode() shouldBe 10.0
+    counter.median() shouldBe 10.0
+  }
+
+  it should "handle when a two objects are counted" in {
```
tfenne: `when a two` -> `when two`

src/test/scala/com/fulcrumgenomics/util/NumericCounterTest.scala
```diff
@@ -0,0 +1,97 @@
...
+    val counter = new NumericCounter[Double]()
+    counter.count(10.0)
+    counter.count(20.0)
+    counter.mean() shouldBe 15.0
+    counter.stddev() shouldBe 7.071 +- 0.0001
+    counter.mode() shouldBe 10.0
+    counter.median() shouldBe 15.0
+  }
+
+  it should "handle updating teh count and re-computing statistics" in {
```
tfenne: `teh -> the`

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
nh13: Only one, but I will change it to use `put` directly. Ctrl-Q on a Mac shows the implicit conversions (TIL).

nh13:
@tfenne thanks for the review so far.

PR #61, opened 2016-07-19
Move SamRecordSetBuilder so it is published in sources.

nh13:


nh13:
I am open to other ideas, but this is the most expedient at the moment, though this class lacks tests.

tfenne:
@nh13 Is there really no better way to do this?  Can we at least stick it in a `testing` package and exclude that package from coverage analysis?

PR #68, opened 2016-07-22
Updates to PickIlluminaIndices

nh13:
@tfenne, this fixes some issues with PickIlluminaIndices, though I still occasionally get negative values for the count, which is not addressed here to due to the difficulty to reproduce.

### DIFF
--- a/src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
+++ b/src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
@@ -240,6 +240,9 @@ class PickIlluminaIndicesCommand {
         }
         log.info("Retained " + indexes.size() + " indices after applying max homopolymer restriction.");
 
+        // Shuffle the indexes so there is no bias in nucleotide representation.
+        Collections.shuffle(indexes);
+
         if (!ALLOW_PALINDROMES) {
             filterForPalindromes(indexes);
             log.info("Retained " + indexes.size() + " indices after filtering for palindromes.");
--- a/src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
+++ b/src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
@@ -313,8 +313,9 @@ class PickIlluminaIndicesCommand {
                 log.info("There are " + count + " indices with a minimum edit distance of " + lastMinEdit);
             }
 
-            rankedIndexes.remove(first);
-            if (--count % 100 == 0) log.info("Down to " + count + " indices.");
+            if (rankedIndexes.remove(first)) count--;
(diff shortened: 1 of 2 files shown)


src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
```diff
@@ -184,6 +185,19 @@ Index reverseComplement() {
         }
 
         /**
+         * Checks to make sure tha this object is contained in this set prior to removing.  The object may be removed
+         * by another thread between the containment check and removal, so this method is not guaranteed to be atomic.
+         * If the object is not contained, an exception is thrown.  The value of the remove method is returned.
+         */
+        public boolean containsAndRemove(final Object o) {
```
tfenne: This is unnecessarily complicated.  The `remove()` method already returns a boolean which is true if the value was present and removed, and false if the value was not present.  No need to scan the set twice (and without atomicity).

src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
```diff
@@ -198,13 +212,20 @@ Index reverseComplement() {
...
-                        if (countdown.decrementAndGet() == 0) {
-                            synchronized (latch) { latch.notify(); }
+                        try {
+                            IndexSet.super.remove(other);
+                            other.removeRelated(index);
+                            other.recalculate();
+                            IndexSet.this.add(other);
+
+                            if (countdown.decrementAndGet() == 0) {
+                                synchronized (latch) {
```
tfenne: Would prefer to keep this on a single line please.

src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
```diff
@@ -310,8 +334,9 @@ protected int execute() {
                 log.info("There are " + count + " indices with a minimum edit distance of " + lastMinEdit);
             }
 
-            rankedIndexes.remove(first);
-            if (--count % 100 == 0) log.info("Down to " + count + " indices.");
+            if (rankedIndexes.containsAndRemove(first)) count--;
```
tfenne: I think this can just be `rankedIndices.remove(first)` again given the check of the return value and throwing if false.

src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
nh13: Done.

src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
nh13: why?  I will put the `synchronized on its own line, but move down the `}`.

PR #70, opened 2016-07-25
SimpleCounter uses a HashMap by default and exposes a method to

nh13:
sub-classes to make the internal counting object.

@tfenne you mentioned you hit a case where you were creating many counters and the reflection code was the bottleneck.  I also hit such a case.  Take a look.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
@@ -26,11 +26,11 @@
 package com.fulcrumgenomics.util
 
 import dagr.commons.CommonsDef._
-import scala.reflect.runtime.universe._
+import scala.collection.mutable
 
 object NumericCounter {
   /** Generates a counter that has counted all the numerics provided. */
-  def apply[T](ts: TraversableOnce[T])(implicit numeric: Numeric[T], tt: TypeTag[T]) : NumericCounter[T] = {
+  def apply[T](ts: TraversableOnce[T])(implicit numeric: Numeric[T]) : NumericCounter[T] = {
     val counter = new NumericCounter[T]
     ts.foreach(counter.count)
     counter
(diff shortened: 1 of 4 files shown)


src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
```diff
@@ -42,9 +41,13 @@ object NumericCounter {
   * zero for any item that has not been counted yet.  Implements some useful methods
   * to compute statistics when the objects being counted are numeric types.
   */
-class NumericCounter[T](implicit numeric: Numeric[T], tt: TypeTag[T]) extends SimpleCounter[T] {
+class NumericCounter[T](implicit numeric: Numeric[T]) extends SimpleCounter[T] {
+  import scala.collection.JavaConversions._
```
tfenne: Could you import just the 1 or 2 conversions that are needed?  I dislike importing all of them as they can have unintended consequences.

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
```diff
@@ -42,19 +40,14 @@ object SimpleCounter {
...
-      new java.util.TreeMap[T,Long]()
-    }
-    else {
-      new util.HashMap[T,Long]()
-    }
-  }
+  /** Creates the map in which we store the counts. */
+  protected def makeCounts(): scala.collection.mutable.Map[T, Long] = new util.HashMap[T,Long]()
+
+  private val counts = makeCounts()
```
tfenne: Can you add a type annotation here just to be super-clear?

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
```diff
@@ -42,19 +40,14 @@ object SimpleCounter {
...
-    val clazzComparable = ReflectionUtil.typeTagToClass[Comparable[T]]
-    if (clazzComparable.isAssignableFrom(clazzT)) {
-      new java.util.TreeMap[T,Long]()
-    }
-    else {
-      new util.HashMap[T,Long]()
-    }
-  }
+  /** Creates the map in which we store the counts. */
+  protected def makeCounts(): scala.collection.mutable.Map[T, Long] = new util.HashMap[T,Long]()
```
tfenne: `makeMap`?

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
```diff
@@ -69,6 +62,6 @@ class SimpleCounter[T](implicit tt: TypeTag[T]) extends Iterable[(T, Long)] {
   /** Iterates over all the objects and their counts. */
   override def iterator: Iterator[(T, Long)] = this.counts.entrySet().iterator().map(entry => (entry.getKey, entry.getValue))
 
-  /** Gets sum of counts. */
-  def totalCounts: Long = _totalCounts.toInt
+  /** Gets the number of counts stored in this counter (sum over all count). */
+  def sumOfCounts: Long = _totalCounts.toInt
```
tfenne: how about `totalObservations`?  Or just `total`?  I think the problem with these kinds of methods is finding good names.

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
```diff
@@ -112,5 +115,8 @@ class NumericCounter[T](implicit numeric: Numeric[T], tt: TypeTag[T]) extends Si
       }
     }
   }
+
+  /** Gets sum of values stored in this counter (sum over all (count * value)). */
+  def sumOfValues: T = totalSum
```
tfenne: `totalMass`?

tfenne:
@nh13 Naming is hard :(  A few minor comments.

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/SimpleCounter.scala
nh13: `total`

PR #71, opened 2016-07-26
Tool to compute the error rate by cycle.

nh13:
@tfenne FYI this exists.  I want to clean it up and add tests before a review.  If there is something like it already out there, let me know.


nh13:
@tfenne ready for review.  I have now used this tool multiple times.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+  *
+  * Gives the error rate for read one, read two, and across both reads, for the given position relative to the start of
+  * the read.  For paired end reads, the read position is relative to the first base sequenced for the given end of the
+  * read.  Soft-clipped bases, and unmapped, secondary, supplementary, duplicate, and vendor failing reads are ignored.
+  *
+  * @author Nils Homer
+  */
+@clp(description =
+  """
+    |Calculates the error rate by read position.
```
tfenne: More documentation please.  Does this require any particular sort order, work on mapped or unmapped bams?  What does it produce (text and PDF), does it require R etc.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+import htsjdk.samtools.{SAMRecord, SamReaderFactory}
+
+/**
+  * Calculates the error rate by position within each read.
+  *
+  * Gives the error rate for read one, read two, and across both reads, for the given position relative to the start of
+  * the read.  For paired end reads, the read position is relative to the first base sequenced for the given end of the
+  * read.  Soft-clipped bases, and unmapped, secondary, supplementary, duplicate, and vendor failing reads are ignored.
+  *
+  * @author Nils Homer
```
tfenne: Are we doing this now?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+  group = ClpGroups.SamOrBam)
+class ErrorRateByReadPosition
+( @arg(flag="i", doc="Input BAM file.") val input: PathToBam,
+  @arg(flag="o", doc="Output metrics prefix. If not given, will use the input BAM basename as a prefix.")
+  val output: Option[PathPrefix] = None
+) extends FgBioTool with LazyLogging {
+  import scala.collection.JavaConversions.asScalaIterator
+
+  Io.assertReadable(input)
+  output.foreach(out => Io.assertListable(out.getParent))
```
tfenne: I'm curious why this instead of just assertWritable?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  def execute(): Unit = {
+    val progress = new ProgressLogger(logger, verb="written", unit=1e6.toInt)
```
tfenne: You're not writing records, right?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+    val in       = SamReaderFactory.make.open(input.toFile)
```
tfenne: There should be an `open` that takes `Path` now, though you might need to bump the dependency version.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+    val iterator = in.iterator()
+    val counter  = new ErrorRateByReadPositionCounter
+
+    iterator.filterNot { rec =>
+      rec.getReadUnmappedFlag || rec.isSecondaryOrSupplementary ||
+        rec.getDuplicateReadFlag || rec.getReadFailsVendorQualityCheckFlag
```
tfenne: 1. I'm genuinely on the fence about whether duplicate reads should be excluded or not.
2. The set of filters should be documented in the usage.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+  * read.  Soft-clipped bases, and unmapped, secondary, supplementary, duplicate, and vendor failing reads are ignored.
+  *
+  * @author Nils Homer
+  */
+@clp(description =
+  """
+    |Calculates the error rate by read position.
+  """,
+  group = ClpGroups.SamOrBam)
+class ErrorRateByReadPosition
```
tfenne: Since we're calling this ErrorRateByPosition I think it would be really nice for it to take in an optional VCF and mask out positions that are present in the VCF.  Or call it mismatch rate by read position?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+( @arg(flag="i", doc="Input BAM file.") val input: PathToBam,
+  @arg(flag="o", doc="Output metrics prefix. If not given, will use the input BAM basename as a prefix.")
+  val output: Option[PathPrefix] = None
```
tfenne: How about parameters for mapQ and baseQ cutoffs?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+    val progress = new ProgressLogger(logger, verb="written", unit=1e6.toInt)
+    val in       = SamReaderFactory.make.open(input.toFile)
+    val iterator = in.iterator()
+    val counter  = new ErrorRateByReadPositionCounter
+
+    iterator.filterNot { rec =>
+      rec.getReadUnmappedFlag || rec.isSecondaryOrSupplementary ||
+        rec.getDuplicateReadFlag || rec.getReadFailsVendorQualityCheckFlag
+    }
+    .foreach { rec =>
```
tfenne: I feel like I want a `andThen` method on traversable so I can write:
```scala
recs.andThen(counter.accept).andThen(progress.record)
```

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+  *
+  * @param position the read position (1-based).
+  * @param error_rate_read_one the error rate of the first read.
+  * @param error_rate_read_two the error rate of the second read.
+  * @param error_rate_total the error rate across all reads at this read position.
+  * @param bases_read_one the total number of bases observed at this position for read one.
+  * @param bases_read_two the total number of bases observed at this position for read two.
+  * @param bases_total the total number of bases observed at this position across all reads.
+  */
+case class ErrorRateByReadPositionMetric
```
tfenne: How'd you feel about changing this to have a `read_number` field and drop the `_read_one` and `_read_two` stuff - i.e. collapse the duplicate columns and have twice the rows?  I think this is awkward if read 1 and read 2 are different lengths, and honestly I don't think it's good to calculate a total error rate across the two reads for the same position in read.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+  val r1ErrorBasesByReadPosition = new NumericCounter[Int]()
+  val r2TotalBasesByReadPosition = new NumericCounter[Int]()
+  val r2ErrorBasesByReadPosition = new NumericCounter[Int]()
+
+  def accept(rec: SAMRecord): Unit = {
+    val first: Boolean    = !rec.getReadPairedFlag || rec.getFirstOfPairFlag
+    val totalBasesByReadPosition = if (first) r1TotalBasesByReadPosition else r2TotalBasesByReadPosition
+    val errorBasesByReadPosition = if (first) r1ErrorBasesByReadPosition else r2ErrorBasesByReadPosition
+    val read              = maybeReverseComplement(rec.getReadBases.clone(), rc=rec.getReadNegativeStrandFlag)
+    val reference         = maybeReverseComplement(SequenceUtil.makeReferenceFromAlignment(rec, false), rc=rec.getReadNegativeStrandFlag)
```
tfenne: This depends on the MD tag right?  I would strongly prefer to just use the reference sequence directly than rely on the MD tag being present and correct.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+      }
+      readPosition += 1
+    }
+  }
+
+  def metrics: Seq[ErrorRateByReadPositionMetric] = {
+    if (r1TotalBasesByReadPosition.isEmpty && r2TotalBasesByReadPosition.isEmpty) return Seq.empty
+    // create the metrics
+    val maxReadPosition = (r1TotalBasesByReadPosition.map(_._1) ++ r2TotalBasesByReadPosition.map(_._1)).max
+    Seq.range(0, maxReadPosition+1).map { readPosition =>
```
tfenne: I would change this to either `Range.inclusive(0, maxReadPosition)` or `(0 to maxReadPosition)`

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+        error_rate_read_two = divide(r2Error, r2Total),
+        error_rate_total    = divide(r1Error + r2Error, r1Total + r2Total),
+        bases_read_one      = r1Total,
+        bases_read_two      = r2Total,
+        bases_total         = r1Total + r2Total
+      )
+    }
+  }
+
+  private def divide(num: Long, denom: Long): Double = {
```
tfenne: Just to be safe, num should also be zero any time denom is zero right?  Could be written as:
```scala
private def divide(num: Long, denom: Long): Double = (num, denom) match {
  case (0,0) => 0
  case (n,d) => n / d.toDouble
}
```
since I know you like matches 😄

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,170 @@
...
+    if (denom == 0) 0d
+    else num / denom.toDouble
+  }
+
+  private def validBase(base: Byte): Boolean = SequenceUtil.isValidBase(base)
```
tfenne: Could be replace with `import SequenceUtil.{isValidBase => validBase}`

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Auto-generated, sorry.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: done.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Nope, thx.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: I usually start with `FilterBam` and it uses `toFile`.  I've seen this comment enough times I will fix `FilterBam` too!

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Oops, the documentation was in the comments, not in the usage.  Fixed.

nh13: Also, adding an argument to include duplicates.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Agreed!  But I don't need it for my current purposes, so lets add this feature in when it is needed.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Sorry, but I have to do this, again I am really sorry:
https://www.youtube.com/watch?v=GKNX6dieVcc

nh13: Deferring.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: I'd feel pretty good about that.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Will not implement, since we skip over any metric with 0 bases observed.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Done.

nh13:
@tfenne one more look over?  I made a substantial # of changes to support using a reference sequence.  The compromise is that the reference FASTA is optional, and if it is not specified, we use the MD tag. (gasp).

nh13:
I also forgot to say that I am deferring having a list of positions or variants to ignore for read errors until it is needed.

nh13:
@tfenne assigning back to myself, since I want to test leading/trailing indels and soft clips

nh13:
@tfenne I have added a bunch more tests now, so I am fairly happy with it. I can see some discussion around how the `--include-insertions` works.  One thing to point out in addition to my comments above, is the bug I found in `SamRecordSetBuilder` where the minus strand had `isNegative` set to `false`!

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+
+@clp(description =
+  """
+    |Calculates the error rate by read position on mapped BAMs.
+    |
+    |If a reference is given, the sort order must be coordinate.  If no reference is given, the MD tag must be present
+    |for every mapped read and computed correctly, and no sort order is required.
+    |
+    |The output metrics text file will have the extension 'error_rate_by_read_position.txt'.  The output file will
+    |contain a row per read and per position in that read.  The 'read_number' will be 0 for fragments, 1 for the first
```
tfenne: per logical read (i.e. read 1, read 2).  For a second I thought you meant per physical read (i.e. many millions).

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+    |provided.  Soft-clipped bases, and unmapped, secondary, supplementary, and vendor failing reads are ignored.
+    |Optionally, duplicates are ignored.
+  """,
+  group = ClpGroups.SamOrBam)
+class ErrorRateByReadPosition
+( @arg(flag="i", doc="Input BAM file.") val input: PathToBam,
+  @arg(flag="o", doc="Output metrics prefix. If not given, will use the input BAM basename as a prefix.")
+  val output: Option[PathPrefix] = None,
+  @arg(flag="r", doc="Reference fasta. If not given, will use the MD tag.") val ref: Option[PathToFasta] = None,
+  @arg(flag="d", doc="Include duplicate reads, otherwise ignore.") val duplicates: Boolean = false,
```
tfenne: `includeDuplicates`?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+  @arg(flag="I", doc="Count insertions as errors, otherwise as matches.") val includeInsertions: Boolean = false,
```
tfenne: Maybe I'll find an explanation later, but I'm curious why insertions but not deletions get this handling.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+  @arg(flag="m", doc="The minimum mapping quality for a read to be included.") val minMappingQuality: Int = 0,
```
tfenne: I think it probably makes sense to have a higher default for MQ, since you want to make sure your reads are mapped to the right place and the errors you are seeing aren't mapping artifacts.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+  output.foreach(out => Io.assertCanWriteFile(out))
+
+  def execute(): Unit = {
+    val progress = new ProgressLogger(logger, verb="read", unit=1e6.toInt)
+    val in       = SamReaderFactory.make.open(input)
+    val iterator = in.iterator()
+    val counter  = ErrorRateByReadPositionCounter(ref=ref, includeInsertions=includeInsertions, minBaseQuality=minBaseQuality)
+
+    if (ref.exists(r => in.getFileHeader.getSortOrder != SortOrder.coordinate)) {
+      throw new IllegalArgumentException("Input BAM must be coordinate sorted when the reference is given.")
```
tfenne: `fail()`

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+        (!duplicates && rec.getDuplicateReadFlag) || rec.getReadFailsVendorQualityCheckFlag ||
+        (rec.getMappingQuality < minMappingQuality)
+    }
+    .foreach { rec =>
+      counter.accept(rec)
+      progress.record(rec)
+    }
+
+    val path = PathUtil.pathTo(output.getOrElse(PathUtil.removeExtension(input)) + ErrorRateByReadPositionMetric.FileExtension)
+    Metric.write[ErrorRateByReadPositionMetric](metrics=counter.metrics, path=path)
```
tfenne: Is the type annotation needed here?  If `counter.metrics` has a declared return type I would think it could be inferred.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+    |
+    |If a reference is given, the sort order must be coordinate.  If no reference is given, the MD tag must be present
+    |for every mapped read and computed correctly, and no sort order is required.
+    |
+    |The output metrics text file will have the extension 'error_rate_by_read_position.txt'.  The output file will
+    |contain a row per read and per position in that read.  The 'read_number' will be 0 for fragments, 1 for the first
+    |end of a pair, and 2 for the second end of a pair. Both the error rate and number of bases observed will be
+    |provided.  Soft-clipped bases, and unmapped, secondary, supplementary, and vendor failing reads are ignored.
+    |Optionally, duplicates are ignored.
+  """,
```
tfenne: Would be nice to describe here what constitutes an error.  Specifically how are the following handled:
1. No-calls
2. Insertions
3. Deletions

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+  private val totalByReadOffset  = List.tabulate(3)(x => new NumericCounter[Int]())
+
+  /** Gets the list of metrics sorted by read number and then position. */
+  def metrics: Seq[ErrorRateByReadPositionMetric] = {
+    if (totalByReadOffset.forall(_.isEmpty)) return Seq.empty
+    // create the metrics
+    errorsByReadOffset.zip(totalByReadOffset).zipWithIndex.filter {
+      case ((errCounter, totalCounter), idx) => totalCounter.nonEmpty
+    }.flatMap { case ((errCounter, totalCounter), idx) =>
+      val masOffset = totalCounter.map(_._1).max
```
tfenne: maxOffset?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+      Range.inclusive(0, masOffset).map { offset =>
+        new ErrorRateByReadPositionMetric(
+          read_number = idx,
+          position    = offset + 1,
+          error_rate  = errCounter.countOf(offset) / totalCounter.countOf(offset).toDouble,
```
tfenne: I _think_ there are pathological situations where this could cause a divide by zero error.  E.g. if the read is 100 bases long, but everything at cycle 99 (for example) had Q < `minBaseQuality`, you could have zero in the totalCounter.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+    // Assert the bases must be valid
+    if (validBase(refBase) && validBase(readBase) && minBaseQuality <= qual) {
+      increment(readIndex=readIndex, readOffset=readOffset, qual=qual, isError = !SequenceUtil.basesEqual(readBase, refBase))
+    }
+  }
+
+  /** Increments the error count if `isError` is true. Also increments the total bases observed. The read index is
+    * 0 for fragment reads, 1 for the first of a pair, and 2 for second of a pair.  The read offset is zero-based. */
+  protected def increment(readIndex: Int, readOffset: Int, qual: Byte, isError: Boolean): Unit = {
+    if (minBaseQuality <= qual) {
```
tfenne: Is there a reason this is checked both here and in `incrementIfValid`?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+      if (isError) {
```
tfenne: I would single-line this `if` if it where me.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+    if (rc) SequenceUtil.reverse(qualities, 0, qualities.length)
+    qualities
+  }
+}
+
+/** Computes the error rate using a reference FASTA.  Does not rely on the MD tag. */
+private class ErrorRateByReadPositionWithReferenceCounter
+( ref: PathToFasta,
+  includeInsertions: Boolean,
+  minBaseQuality: Int = 0
```
tfenne: Maybe kill the default here since this is a private class and the CLP always provides a value?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+      var readOffset   = if (rec.getReadNegativeStrandFlag) readLength - readPosition - 1 else readPosition
+
+      // Matches/mismatches
+      while (readPosition < readEnd) {
+        val readBase = read(readPosition)
+        val qual     = quals(readPosition)
+        val refBase  = referenceBases(refPosition)
+        incrementIfValid(readIndex=readIndex, readOffset=readOffset, readBase=readBase, qual=qual, refBase=refBase)
+        refPosition += 1
+        readPosition   += 1
```
tfenne: Funky alignment/spacing.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+
+    // Handle insertions
+    var readOffset = if (rec.getReadNegativeStrandFlag) rec.getReadLength - 1 else 0
+    rec.getCigar.getCigarElements.foreach { elem =>
+      val op: CigarOperator = elem.getOperator
+      if (op == CigarOperator.INSERTION) {
+        var i = 0
+        while (i < elem.getLength) {
+          val qual = quals(readOffset)
+          increment(readIndex = readIndex, readOffset = readOffset, qual=qual, isError = includeInsertions)
```
tfenne: I think it's a little strange to always count insertions in the denominator, but only in the numerator if `includeInsertions` is set.  I would be more tempted to wrap this whole block with `if (includeInsertions)`

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+          readOffset += readOffsetIncrement
+          i += 1
+        }
+      }
+      else {
+        readOffset += readOffsetIncrement * elem.getLength
```
tfenne: This probably wants to have `else if (op.consumeReadBases)`.  Otherwise, e.g. padding and deletions will do strange things.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -0,0 +1,280 @@
...
+  /** Implement me! */
+  protected def updateErrors(rec: SAMRecord, readIndex: Int): Unit
+
+  def accept(rec: SAMRecord): Unit = {
+    val readIndex: Int = if (!rec.getReadPairedFlag) 0 else if (rec.getFirstOfPairFlag) 1 else 2
+    updateErrors(rec=rec, readIndex=readIndex)
+  }
+
+  protected def maybeReverseComplement(bases: Array[Byte], rc: Boolean): Array[Byte] = {
+    if (rc) SequenceUtil.reverseComplement(bases)
```
tfenne: I don't like that this (and the following) method reverse in place but return a value as it seems misleading.  Why not push the `.clone()` calls down into these methods and then a) it won't be misleading to return a value and b) you won't clone unless you need to.

src/main/scala/com/fulcrumgenomics/testing/SamRecordSetBuilder.scala
```diff
@@ -40,7 +40,7 @@ import scala.collection.JavaConversions._
 object SamRecordSetBuilder {
   sealed trait Strand { val isNegative: Boolean }
   object Plus  extends Strand { val isNegative = false }
-  object Minus extends Strand { val isNegative = false }
+  object Minus extends Strand { val isNegative = true }
```
tfenne: Oopsie.  Thanks for fixing this.

tfenne:
I didn't review the tests, let me know if you really want me to.  The rest looks good with a few minor comments.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: To which base in the read does a deletion belong?  What if we have a 10bp deletion?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: OK

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Ah, `fail` is in `fgbio`, not dagr commons.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: True.  Scala is good at that.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: 👍

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: 👍

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: I'll be you.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: 👍 here and elsewhere

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: 👍

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: 👍

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Since the following use of it doesn't need a clone, and thus would have an _extra_ clone:
```
maybeReverseComplement(SequenceUtil.makeReferenceFromAlignment(rec, false), rc=rec.getReadNegativeStrandFlag)
```

PR #72, opened 2016-07-28
Tool to assess phased variant calls

nh13:
@tfenne a WIP, but there are a few preceding commits that may useful to some of your work, or not.


src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,735 @@
...
+import htsjdk.variant.vcf.VCFFileReader
+
+import scala.annotation.tailrec
+import scala.collection.JavaConversions._
+import scala.collection.mutable.ListBuffer
+
+@clp(
+  description =
+    """
+      |Assess variant calls for phasing.
```
tfenne: ``Assess the accuracy of phasing for a set of variants`?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,735 @@
...
+      |
+      |All phased genotypes should be annotated with the "PS" (phase set) FORMAT tag, which by convention is the
+      |position of the first variant in the phase set (see the VCF specification).  Furthermore, the alleles of a phased
+      |genotype should use the '|' separator instead of the '/' separator, where the latter indicates the genotype is
+      |unphased.
+    """,
+  group=ClpGroups.VcfOrBcf
+)
+class AssessPhasing
+( @arg(flag="c", doc="The VCF with called phased variant.") val callVcf: PathToVcf,
```
tfenne: variant -> variants

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,735 @@
...
+  @arg(flag="t", doc="The VCF with known phased variant.") val truthVcf: PathToVcf,
```
tfenne: Ditto

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,735 @@
...
+  @arg(flag="o", doc="The output prefix for all output files.") val output: PathPrefix,
+  @arg(flag="l", doc="The interval list over which known calls should be used.") val intervals: Option[PathToIntervals] = None,
```
tfenne: This description doesn't make a ton of sense to me.  Isn't this just the set of regions over which the assessment is done?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,735 @@
...
+  @arg(flag="m", doc="Allow missing fields in the VCF header.") val allowMissingFieldsInVcfHeader: Boolean = true,
```
tfenne: Is there any reason not to hard code this to true?  I f'ing hate the VCF code.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,735 @@
...
+  @arg(flag="s", doc="Skip sites where the truth and call are both called but do not share the same alleles.") val skipMismatchingAlleles: Boolean = true
+  //@arg(doc="Ignore the allele separator and use the PS tag exclusively to detect sets of phased variants") val usePhaseSetTagOnly: Boolean = false // TODO: implement me
```
tfenne: Do you still plan to implement this?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,735 @@
...
+      truthReader.close()
+      (callReader.getFileHeader.getSequenceDictionary, Seq(truthReader, callReader).map(_.getFileHeader.getSampleNamesInOrder.head))
+    }
+
+    val intervalList = intervals.map { intv => IntervalList.fromFile(intv.toFile) }
+
+    // get the phased blocks
+    val callPhaseBlockDetector: OverlapDetector[PhaseBlock] = {
+      val callReader  = new VCFFileReader(callVcf.toFile, true)
+      val detector = PhaseBlock.toOverlapDetector(callIterator(callReader), dict=callReader.getFileHeader.getSequenceDictionary)
```
tfenne: This means reading each VCF twice, right?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,735 @@
...
+
+    callReader.safelyClose()
+    truthReader.safelyClose()
+  }
+
+  private def callIterator(callReader: VCFFileReader): Iterator[VariantContext] = {
+    callReader.filter(v => v.isSNP && v.isBiallelic).iterator
+  }
+
+  private def truthIterator(truthReader: VCFFileReader, intervalList: Option[IntervalList]): Iterator[VariantContext] = {
```
tfenne: Since the interval list is optional here, why not just have one method instead of `callIterator` and `truthIterator` ?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+      val truthReader = new VCFFileReader(truthVcf.toFile, true)
+      val detector = PhaseBlock.toOverlapDetector(truthIterator(truthReader, intervalList=intervalList), dict=truthReader.getFileHeader.getSequenceDictionary)
+      truthReader.close()
+      detector
+    }
+
+    // get an iterator of the pairs
+    val callReader  = new VCFFileReader(callVcf.toFile, true)
+    val truthReader = new VCFFileReader(truthVcf.toFile, true)
+    val pairedIterator = {
```
tfenne: Braces here are unnecessary.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+      (callReader.getFileHeader.getSequenceDictionary, Seq(truthReader, callReader).map(_.getFileHeader.getSampleNamesInOrder.head))
+    }
+
+    val intervalList = intervals.map { intv => IntervalList.fromFile(intv.toFile) }
+
+    // get the phased blocks
+    logger.info("Gathering the phased blocks for the call sample")
+    val callPhaseBlockDetector: OverlapDetectorByContig = {
+      val callReader  = new VCFFileReader(callVcf.toFile, true)
+      val detector = PhaseBlock.toOverlapDetector(callIterator(callReader), dict=callReader.getFileHeader.getSequenceDictionary)
```
tfenne: `dict=dict`?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+      callReader.close()
+      detector
+    }
+    logger.info("Gathering the phased blocks for the truth sample")
+    val truthPhaseBlockDetector: OverlapDetectorByContig = {
+      val truthReader = new VCFFileReader(truthVcf.toFile, true)
+      val detector = PhaseBlock.toOverlapDetector(truthIterator(truthReader, intervalList=intervalList), dict=truthReader.getFileHeader.getSequenceDictionary)
```
tfenne: same here (`dict=dict`)

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Perhaps more context for its use: NIST usually gives a VCF and a set of confident regions; this interval list would be the confident region interval list over which to filter "known phased" variants.  We don't filter out the calls initially, since we care about the % of SNPs phased, even outside the confident regions, but when comparing to truth, we compare only _within_ the confident regions if this interval list is given.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    /////////////////////////////////////////////////////////////////////////////////////////////////////////
+    // Create the phasing cigar
+    /////////////////////////////////////////////////////////////////////////////////////////////////////////
+    logger.info("Creating the phasing CIGAR")
+    val metric = new AssessPhasingMetric()
+    val cigar = toCigar(
+      pairedIterator          = pairedIterator,
+      truthPhaseBlockDetector = truthPhaseBlockDetector,
+      callPhaseBlockDetector  = callPhaseBlockDetector,
+      metric                  = metric,
```
tfenne: I missed this the first time through the code, and was wondering where all the other metrics were set.  I'm sure I'll see shortly, but am wondering if it makes sense to have Cigar generation and metric calculation intermixed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    callReader.safelyClose()
+    truthReader.safelyClose()
+  }
+
+  private def callIterator(callReader: VCFFileReader): Iterator[VariantContext] = {
+    import scala.collection.JavaConversions.asScalaIterator
+    callReader.iterator.filter(v => v.isSNP && v.isBiallelic)
+  }
+
+  private def truthIterator(truthReader: VCFFileReader, intervalList: Option[IntervalList]): Iterator[VariantContext] = {
```
tfenne: I said this earlier but my comment got eaten.  Since the interval list is optional, I don't think you need separate methods for `callIterator` and `truthIterator`, and it would ensure they use the same filtering if it was one method.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+      case Some(intv) => ByIntervalListVariantContextIterator(truthReader, intv, useIndex=false)
+      case None =>
+        import scala.collection.JavaConversions.asScalaIterator
+        truthReader.iterator.toIterator
+    }).filter(v => v.isSNP && v.isBiallelic)
+  }
+}
+
+private object AssemblyStatistics {
+  def apply(counter: NumericCounter[Long]): AssemblyStatistics = {
```
tfenne: It would help readability to either call the parameter something more specific (e.g. blockLengthCounter) or typedef Long -> BlockLength.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    val fiftyPercent  = numBases / 2.0
+    val ninetyPercent = numBases * 0.9
+
+    // N50/N90
+    var n50: Option[Long] = None
+    var n90: Option[Long] = None
+    var i = 0
+    var blockLengthSum = 0d
+    val countAndLengths = counter.toSeq.sortBy(_._1)
+    while (i < countAndLengths.length && (n50.isEmpty || n90.isEmpty)) {
```
tfenne: You could use the new forloop here:
`forloop (0) (_ < countAndLengths.length && (n50.isEmpty || n90.isEmpty)) (_ + 1) { ... }`

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    val numBases = counter.map { case (length, count) => length * count }.sum
+    val fiftyPercent  = numBases / 2.0
+    val ninetyPercent = numBases * 0.9
+
+    // N50/N90
+    var n50: Option[Long] = None
+    var n90: Option[Long] = None
+    var i = 0
+    var blockLengthSum = 0d
+    val countAndLengths = counter.toSeq.sortBy(_._1)
```
tfenne: Don't you want to invert this sort so that you've got the largest blocks first?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    while (i < countAndLengths.length && (n50.isEmpty || n90.isEmpty)) {
+      blockLengthSum += countAndLengths(i)._1 * countAndLengths(i)._2
```
tfenne: Given the repeated usage I'd be tempted to start the loop with:
`val (length, count) = countAndLengths(i)`
It's also a bit confusing that the variable is called `countAndLengths` but it's (length, count).

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+      }
+      if (n90.isEmpty && ninetyPercent <= blockLengthSum) {
+        n90 = Some(countAndLengths(i)._1)
+      }
+
+      i += 1
+    }
+
+    // L50
+    i = countAndLengths.length - 1
```
tfenne: A few thoughts here:
1. I don't like the re-use of the `i` var .. that just seems like it'll cause errors later.  Using `forloop` would obviate that.
2. Why not just compute the L50 at the same time as the N50?
3. I think technically your L50 is not as accurate as it could be.  While for N50 reporting the length of block where the cutover happens is correct, for L50 if there are many blocks at the N50 size, you likely would only need a subset of them to get to having sum(lengths) >= 0.5 * sum(all lengths).

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    blockLengthSum = 0
+    var l50: Long = 0
+    while (0 <= i && blockLengthSum < fiftyPercent) {
+      blockLengthSum += countAndLengths(i)._1 * countAndLengths(i)._2
+      l50 += countAndLengths(i)._2
+      i -= 1
+    }
+
+    (n50, n90) match {
```
tfenne: Why the belt and suspenders?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    }
+  }
+}
+
+/**
+  * @param n50 the smallest block length such that the bases covered by all blocks this length and shorter are at least
+  * 50% of the # of bases covered by all blocks.
+  * @param n90 the smallest block length such that the bases covered by all blocks this length and shorter are at least
+  * 90% of the # of bases covered by all blocks.
+  * @param l50 the smallest number of blocks that whose sum produces the n50 value.
```
tfenne: that whose

tfenne: I think more correctly the definition is _the smallest number of blocks such that the sum of the lengths of the blocks is >= 50% of the sum of the lengths of all blocks_.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  * @param length the length of the phased block
+  * @param count the number of phased blocks of the given length.
+  */
+case class PhaseBlockLengthMetric
+( is_truth: Boolean,
+  length: Long = 0,
+  count: Long = 0
+) extends Metric
+
+object PhasingUtils {
```
tfenne: Seems so little to declare an object for ;)

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  def getPhasingSetId(ctx: VariantContext): Int = {
+    ctx.getGenotype(0).getAttributeAsInt("PS", -1) // Integer.valueOf(ctx.getGenotype(0).getExtendedAttribute("PS", -1).toString)
+  }
+}
+
+private[vcf] class OverlapDetectorByContig(dict: SAMSequenceDictionary) {
```
tfenne: I don't understand the need for this.  OverlapDetector already partitions internally by Contig, and I can't figure out what this adds.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  }
+}
+
+object PhaseBlock extends LazyLogging {
+  import scala.collection.mutable
+
+  /** Creates an overlap detector for blocks of phased variants.  Variants from the same block are found using the
+    * "PS" tag.
+    */
+  private[vcf] def toOverlapDetector(iterator: Iterator[VariantContext], dict: SAMSequenceDictionary): OverlapDetectorByContig = {
```
tfenne: `PhaseBlock.toOverlapDetector` makes me think it should be turning a `PhaseBlock` into an `OverlapDetector`.  How about `buildOverlapDetector`?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    val detector = new OverlapDetectorByContig(dict=dict)
+    val progress = new ProgressLogger(logger)
+
+    logger.info("Creating the overlap detector")
+    // create the blocks
+    var curContig = ""
```
tfenne: Oh my ... not `None`?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+        if (ctx.getContig != curContig) {
+          detector.addAll(phaseBlocks.values.toIterator)
+          phaseBlocks.clear()
+          curContig = ctx.getContig
+        }
+        val phaseBlock = phaseBlocks.get(phaseSetId) match {
+          case Some(block) =>
+            require(block.getContig == ctx.getContig)
+            require(block.getStart <= ctx.getStart)
+            new PhaseBlock(contig=ctx.getContig, start=block.getStart, end=Math.max(block.getEnd, ctx.getEnd)) // extend the block
```
tfenne: Why not `block.copy(end= Math.max(block.getEnd, ctx.getEnd))`?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+
+    // add any remaining
+    detector.addAll(phaseBlocks.values.toIterator)
+    phaseBlocks.clear()
+
+    detector
+  }
+}
+
+case class PhaseBlock private (contig: String, start: Int, end: Int) extends Locatable {
```
tfenne: I keep doing this in places too.  I think we should create a utility trait something like:
```scala
trait LocatableAdapter { // better name needed
  def contig: String
  def start: Int
  def end: Int
  def length: Int = end - start + 1

 def getContig: String = contig
 def getStart: Int = start
 def getEnd: Int = end
}
```

What do you think?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+
+/** Some counts about phasing
+  *
+  * @param num_called the number of variants called.
+  * @param num_phased the number of variants called with phase.
+  * @param num_truth the number of variants with known truth genotypes.
+  * @param num_truth_phased the number of variants with known truth genotypes with phase.
+  * @param num_called_with_truth_phased the number of variants called that had a known phased genotype.
+  * @param num_phased_with_truth_phased the number of variants called with phase that had a known phased genotype.
+  * @param num_truth_phased_in_called_block the number of known phased variants that were in a called phased block.
```
tfenne: Does this imply that the variant was actually called?  Or just that the position of the truth variant is within the boundary of a called phase block?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  * @param num_both_phased_in_called_block the number of called phase variants that had a known phased genotype in a called phased block.  TODO: is this the same as `num_phased_with_truth_phased`?
```
tfenne: phase -> phased

tfenne: Also, your own TODO?  I'm pretty sure I can't understand what this is.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  * @param num_short_switch_errors the number of short switch errors (isolated switch errors).
+  * @param num_long_switch_errors the number of long switch errors (# of runs of consecutive switch errors).
+  * @param num_long_switch_sites the number of sites that could be long switch errors.
```
tfenne: I get switch errors, but switch sites are new to me.  Could you add a one sentence explanation here and/or a link to some definitions please?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  * @param num_illumina_point_switch_errors the number of point switch errors (defined in http://dx.doi.org/10.1038%2Fng.3119).
+  * @param num_illumina_long_switch_errors the number of long switch errors (defined in http://dx.doi.org/10.1038%2Fng.3119).
+  * @param num_illumina_long_switch_sites the number of sites that could be long switch errors (defined in http://dx.doi.org/10.1038%2Fng.3119).
+  * @param frac_phased the fraction of called variants with phase.
```
tfenne: I feel like I really also want some kind of either error rate or accuracy here.  Are there conventions for, say, accuracy being something like:
```
short_accuracy = (1-short_switch_errors) / num_phased_with_truth_phased
long_accuracy = (1-long_switch_errors) / num_phased_with_truth_phased
```

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  var n50_truth_block_length: Double = 0,
+  var n90_truth_block_length: Double = 0,
+  var l50_truth_block_length: Double = 0
+) extends Metric {
+  private def divide(a: Double, b: Double): Double = {
+    if (b == 0) 0
+    else a / b
+  }
+  def finalizeMetric(): this.type = {
+    this.frac_phased                                   = this.divide(this.num_phased, this.num_called)
```
tfenne: Generally shouldn't put `this.` infront of method invocations.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    this.frac_phased_with_truth_phased                 = this.divide(this.num_phased_with_truth_phased, this.num_called_with_truth_phased)
+    this.frac_truth_phased_in_called_block             = this.divide(this.num_truth_phased_in_called_block, this.num_truth_phased)
+    this.frac_phased_with_truth_phased_in_called_block = this.divide(this.num_both_phased_in_called_block, this.num_truth_phased_in_called_block)
+    this
+  }
+}
+
+/** The cigar Elements */
+private object CigarType extends Enumeration {
```
tfenne: So on the one hand I get that this has as much right to the name Cigar/CigarType as does the cigar in SAM/BAM, but given how pervasive the latter is, I wonder if it wouldn't be better to prefix these classes and objects with `Phase`, i.e. `CigarType` -> `PhaseCigarType`?

Also, if the cigar is the name of the string of pairwise relations, I think `CigarType` is more correctly `CigarOp` or `CigarRelation`.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  Call,           // 3
+  TruthEnd,       // 4
+  CallEnd,        // 5
+  BothEnd = Value // 6
+}
+
+private[vcf] object Cigar {
+  import CigarType._
+
+  type Cigar = Seq[CigarType]
```
tfenne: You have a bunch of methods in the object here that take a single `Cigar` as their first (sometimes only) parameter.  I think you'd be better off implementing a trivial Cigar _class_ that has a single member that is a `Seq[CigarType]` and has those methods on it.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    * disagree between the two.
+    *
+    * Examples of short and long errors:
+    *   CALL :  MMMMXXXMMM   MMMM|FFFFFFF|MMMMMFMMMM|F|MMFFFMMM
+    *   TRUTH:  MMMM   MMMMMMMMMM-MMMMMMM|MMMMMMMMMM|F|MMMMMMMM
+    *             correct       long        short     long
+    *   CIGAR:  00003330002220000511111116000001000060600111000
+    * shows one known block split
+    */
+  def toCigar(pairedIterator: JointVariantContextIterator,
```
tfenne: Does this make one massive genome wide cigar?  That seems a little strange to me, given that there cannot be connectivity between chromosomes.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+              truthPhaseBlockDetector: OverlapDetectorByContig,
+              callPhaseBlockDetector: OverlapDetectorByContig,
+              metric: AssessPhasingMetric,
+              skipMismatchingAlleles: Boolean): Cigar = {
+    val cigar: Cigar = pairedIterator
+      .filter { // ensure the alleles are the same when both truth and call are called
+        case Seq(Some(t: VCtx), Some(c: VCtx)) =>
```
tfenne: Why does a PairedIterator return a Seq and not a Tuple2?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+          import scala.collection.JavaConversions.iterableAsScalaIterable
+          !skipMismatchingAlleles || t.getAlleles.toSet == c.getAlleles.toSet
+        case _                                 => true
+      }
+      .flatMap { case Seq(t, c) =>
+        val ctxs                 = Seq(t, c).flatten // NB: can be 1 or 2 contexts here
+        val inTruthPhaseBlock    = ctxs.exists(truthPhaseBlockDetector.overlapsAny)
```
tfenne: Won't the two ctxs have the same position, and the querying exists here and below will double-query when they variant is not in a block?  Why not just check the head?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+        val inCallPhaseBlock     = ctxs.exists(callPhaseBlockDetector.overlapsAny)
+        val isTruthVariant       = t.isDefined
+        val isCallVariant        = c.isDefined
+        val truthPhasedVariant   = t.filter { ctx => PhasingUtils.getPhasingSetId(ctx) > 0 }
+        val callPhasedVariant    = c.filter { ctx => PhasingUtils.getPhasingSetId(ctx) > 0 }
+        val isTruthVariantPhased = truthPhasedVariant.isDefined
+        val isCallVariantPhased  = callPhasedVariant.isDefined
+
+        // DEBUG: make sure we overlap at most one phase block
+        require(ctxs.forall(truthPhaseBlockDetector.getOverlaps(_).size <= 1))
```
tfenne: Given that you're just dealing with SNPs, you could just check this when constructor the detectors, that no pair of adjacent block overlap each other, instead of re-quering for each variant.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+        case _                                 => true
+      }
+      .flatMap { case Seq(t, c) =>
+        val ctxs                 = Seq(t, c).flatten // NB: can be 1 or 2 contexts here
+        val inTruthPhaseBlock    = ctxs.exists(truthPhaseBlockDetector.overlapsAny)
+        val inCallPhaseBlock     = ctxs.exists(callPhaseBlockDetector.overlapsAny)
+        val isTruthVariant       = t.isDefined
+        val isCallVariant        = c.isDefined
+        val truthPhasedVariant   = t.filter { ctx => PhasingUtils.getPhasingSetId(ctx) > 0 }
+        val callPhasedVariant    = c.filter { ctx => PhasingUtils.getPhasingSetId(ctx) > 0 }
```
tfenne: My brain keeps tripping up on these names.  I think it would be easier to read if you used `called` instead of `call` in places like this where your usage is as an adjective on a thing (variant, block).

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+            if (isCallVariantPhased) metric.num_both_phased_in_called_block += 1
+          }
+        }
+
+        // NB: only create the cigar from phased variants.
+        if (isCallVariantPhased || isTruthVariantPhased) {
+          contextsToCigar(truthPhasedVariant, callPhasedVariant)
+        }
+        else Seq.empty
+      }.toList match { // remove leading `*End`
```
tfenne: But why?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    }
+    // Make sure to ge the last one
+    if (lastCigar.nonEmpty) cigarsToReturn.append(lastCigar)
+
+    cigarsToReturn.toList
+  }
+
+  /** Returns the cigar for the two variant contexts. */
+  private[vcf] def contextsToCigar(t: Option[VariantContext], c: Option[VariantContext]): Cigar = (t, c) match {
+    case (None, Some(c: VCtx))          if isStartOfPhaseBlock(c)                             => Seq(CallEnd, Call)
```
tfenne: You're shadowing `t` and `c` left and right here.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    case (None, Some(c: VCtx))                                                                => Seq(Call)
+    case (Some(t: VCtx), None)          if isStartOfPhaseBlock(t)                             => Seq(TruthEnd, Truth)
+    case (Some(t: VCtx), None)                                                                => Seq(Truth)
+    case (Some(t: VCtx), Some(c: VCtx)) if isStartOfPhaseBlock(t)  && isStartOfPhaseBlock(c)  => Seq(BothEnd, cigarTypeForVariantContexts(t, c))
```
tfenne: Doing these four cases as a single case that does another `match (isStart(t), isStart(c))` would prevent you from invoking those checks so frequently.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    case _ => unreachable()
+  }
+
+  /** True if the phasing set id is the same as the start position of the given variant, false otherwise. */
+  def isStartOfPhaseBlock(ctx: VariantContext): Boolean = ctx.getStart == PhasingUtils.getPhasingSetId(ctx)
+
+  /** Computes the cigar for two variant contexts.  Returns [[Match]] if they share the same alleles in the same order,
+    * [[Mismatch]] otherwise.
+    */
+  private[vcf] def cigarTypeForVariantContexts(truth: VariantContext, call: VariantContext): CigarType = {
```
tfenne: Does this work?  It could be my lack of experience with the VCF phasing code, but is there some kind of convention that will make sure this is always true?  What if my truth and called variants are phased, but the phasing started at a different variant, couldn't I end up with an entire block where the allele ordering is inverted for the two VCFs?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+ /** Gets all the indices in the cigar where there is a short switch error.
+  *
+  * Short switches are sites where the alleles are misphased (flipped), and both upstream and downstream, either the
+  * haplotype block ends, or the next variant has the same phase.
+  */
+  def toShortSwitchErrorIndices(cigar: Cigar): Seq[Int] = {
+    // check if we can find a match or the end of a block at the given idx.  If we cannot, move to the next idx.
+    @tailrec
+    def checkNext(cigar: Seq[CigarType], idx: Int, by: Int): Boolean = {
+      if (idx < 0 || cigar.length <= idx) return true // no more
```
tfenne: This is a little belt-and-suspenders again.  You stick a `BothEnd` at each end of the Cigar, so this should never happen.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+      cigar(idx) match {
+        case Match | CallEnd | BothEnd | TruthEnd => true
+        case Mismatch     => false
+        case Call | Truth => checkNext(cigar=cigar, idx=idx+by, by=by)
```
tfenne: I think calling these two `CallOnly` and `TruthOnly` would make this a little clearer.  Also a comment that you need to scan for either the end of a block or where both samples are called would help.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  }
+
+  /** Gets the number of long switch errors and the number of sites examined.
+    *
+    * The number of sites examined are places where there are both a truth and called variant.
+    *
+    * Long switches are a stretch of two or more consecutive sites where the alleles are misphased (flipped), and
+    * both upstream and downstream, either the haplotype block ends, or the next variant has the same phase.
+    */
+  def toLongSwitchErrorsAndSites(cigar: Cigar): (Int, Int) = {
```
tfenne: Can you document what the two values in the return tuple are please.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+    // check if we can find a match or the end of a block at the given idx.  If we cannot, move to the next idx.
+    @tailrec
+    def checkNext(cigar: Seq[CigarType], idx: Int, by: Int, stretchLength: Int = 0): Int = {
```
tfenne: Since this does more than check the next thing, maybe rename it to `numberOfContiguousMismatches` or something like that?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+      cigar(idx) match {
+        case Match | CallEnd | BothEnd | TruthEnd => stretchLength
+        case Mismatch     => checkNext(cigar=cigar, idx=idx+by, by=by, stretchLength=stretchLength+1)
+        case Call | Truth => checkNext(cigar=cigar, idx=idx+by, by=by, stretchLength=stretchLength)
+      }
+    }
+
+    // Gets all the indices in the cigar where there is a long switch error. Currently returns the index for all cigar
+    // mismatches that contribute to a long switch error, and tries to maximize such a stretch.
+    val longSwitchErrorIndices = cigar.indices
```
tfenne: This seems like it's potentially quite inefficient if there are many switch errors of length > 1.  I think I'd be tempted to change this to:
a) have `checkNext` return a pair of numbers: `(stretchLength, cigarsConsumed)` where the second element also counts any `Call` or `Truth` that were skipped over
b) Count the long switch errors using a while/for loop over the cigar indices that, when it finds a long switch errors, jumps forward by `cigarsConsumed` instead of one.

tfenne: You could also, I think, implement this with a [better]buffered iterator over the Cigar, something like:

```scala
val iter = cigar.iterator.bufferBetter
var count = 0
while (iter.hasNext) {
  if (iter.head == MisMatch) {
    val len = iter.takeWhile(c => c == Mismatch || c == Call || c == Truth).filter(_ == Mismatch).length
    if (len > 0) count += 1
  }
  iter.dropWhile(_ != Mismatch)
}
```

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+      .filter { i => cigar(i) == Mismatch }
+      .filter { i => 1 + checkNext(cigar, i-1, -1) + checkNext(cigar, i+1, 1) > 1 }
+
+    val numLongSwitchErrors = longSwitchErrorIndices.foldLeft((-2, 0)) { case ((lastIdx, count), idx) =>
+      if (lastIdx + 1 == idx) (idx, count)
+      else (idx, count+1)
+    }._2
+
+    val numLongSwitchSites = cigar.count {
```
tfenne: I don't get this.  Is this supposed to be the number of sites that are out of phase in long switch errors?  Because it appears to just be all sites that can be compared between the two samples.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -0,0 +1,780 @@
...
+  }
+
+  /** Computes the number of point and long switch error rates as described in http://dx.doi.org/10.1038%2Fng.3119
+    *
+    * The HMM with a simple scoring system and two hidden states, one for each parental haplotype.  The emission
+    * probability is scored as follows: 0 for being on the correct haplotype, otherwise `pointPenalty`.  The transition
+    * probability is scored as follows: 0 for staying on the same haplotype, otherwise `transitionPenalty`.
+    *
+    * We prefer a long switch error over a sequence of point errors (i.e. when both produce have the same score). */
+  private[vcf] def toIlluminaSwitchErrorsHmm(cigar: Cigar, pointPenalty: Int = -1, transitionPenalty: Int = -5): IlluminaSwitchErrors = {
```
tfenne: @nh13 I feel like to properly review this method i need to read and fully understand the method from the paper, and that's going to take a while.  If you think that's necessary/worth the investment let me know and I'll do it, otherwise I'm skipping this method for now.

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
```diff
@@ -0,0 +1,91 @@
...
+package com.fulcrumgenomics.vcf
+
+import java.util.NoSuchElementException
+
+import htsjdk.samtools.util.{CloseableIterator, Interval, IntervalList}
+import htsjdk.variant.variantcontext.VariantContext
+import htsjdk.variant.vcf.VCFFileReader
+import htsjdk.samtools.util.OverlapDetector
+
+
```
tfenne: All the newlines!

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
```diff
@@ -0,0 +1,91 @@
...
+
+object ByIntervalListVariantContextIterator {
```
tfenne: What do you think about turning this into an `implicit class RichVariantContext` that provides an enhanced `query()` method that does what you're doing below?

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
```diff
@@ -0,0 +1,91 @@
...
+            intervalList: IntervalList,
+            useIndex: Boolean = true): Iterator[VariantContext] = {
+    if (useIndex) {
+      new IndexQuery(reader, intervalList)
+    }
+    else {
+      import scala.collection.JavaConversions.asScalaIterator
+      val detector = {
+        import scala.collection.JavaConverters._
+        val intervals = intervalList.uniqued().iterator().toList.asJava
```
tfenne: Why convert it back and forth?  `intervalList.uniqued.getIntervals()` gives you what you want.

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
```diff
@@ -0,0 +1,91 @@
...
+        d.addAll(intervals, intervals)
+        d
+      }
+      reader.iterator().toIterator.filter { ctx => detector.overlapsAny(ctx) }
+    }
+  }
+}
+
+private class IndexQuery(reader: VCFFileReader,
+                         intervalList: IntervalList)
```
tfenne: No need to line-wrap for such a short param list.

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
```diff
@@ -0,0 +1,91 @@
...
+  /**
+    * Creates an iterator over variant contexts that overlap any interval in an interval list.
+    *
+    * If `useIndex` is true, this will query the VCF when moving to the next variant context, and so may be quite slow
+    * if we jump around the VCF a lot.  Otherwise, all variants will be read in and filtered appropriately.
+    */
+  def apply(reader: VCFFileReader,
+            intervalList: IntervalList,
+            useIndex: Boolean = true): Iterator[VariantContext] = {
+    if (useIndex) {
```
tfenne: Maybe unique the intervalList up here since you need to do it for both branches, and then you don't need to also do it in IndexQuery?

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
```diff
@@ -0,0 +1,91 @@
...
+        val d = new OverlapDetector[Interval](0, 0)
+        d.addAll(intervals, intervals)
+        d
+      }
+      reader.iterator().toIterator.filter { ctx => detector.overlapsAny(ctx) }
+    }
+  }
+}
+
+private class IndexQuery(reader: VCFFileReader,
```
tfenne: `reader` should probably be a `private val`

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
```diff
@@ -0,0 +1,91 @@
...
+                         intervalList: IntervalList)
+  extends Iterator[VariantContext] {
+  private var iterator: Option[CloseableIterator[VariantContext]] = None
+  private val intervals = intervalList.uniqued().iterator()
+
+  def hasNext(): Boolean = {
```
tfenne: I know this is ported over (and probably from something I wrote!!), but can we do the following refactor please:
1) In the constructor call advance() once to prime the iterator
2) remove the advance() call from hasNext()
3) change the next() method to:
```scala
def next(): VariantContext = {
  this.iterator match {
    case None => throw new NoSuchElementException("Called next when hasNext is false")
    case Some(iter) => yieldAndThen(iter.next()) { advance() }
  }
}
```

src/main/scala/com/fulcrumgenomics/vcf/JointVariantContextIterator.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import htsjdk.samtools.SAMSequenceDictionary
+import htsjdk.variant.variantcontext.{VariantContext, VariantContextComparator}
+
+/**
+  * Iterates over multiple variant context iterators such that we return a list of contexts for the union of sites
+  * across the iterators.  If samples is given, we subset each variant context to just that sample.
+  */
+class JointVariantContextIterator(iters: Seq[Iterator[VariantContext]],
+                                  dict: SAMSequenceDictionary,
```
tfenne: Would it make sense to have a single parameter `Either[SAMSequenceDictionary,VariantContextComparator]` since it looks like you really need one or the other?

src/main/scala/com/fulcrumgenomics/vcf/JointVariantContextIterator.scala
```diff
@@ -0,0 +1,65 @@
...
+  private val iterators = iters.map(_.buffered)
+  private val comparator = comp.getOrElse(new VariantContextComparator(dict))
+
+  if (sampleNames.nonEmpty && iters.length != sampleNames.length) {
+    throw new IllegalArgumentException("Samples should have the same length as the number of iterators.")
+  }
+
+  def hasNext(): Boolean = iterators.exists(_.nonEmpty)
+
+  def next(): Seq[Option[VariantContext]] = {
```
tfenne: This seems fine, but it would probably be more efficient (and maybe more complicated) to keep the iterators in a sorted data structure  (e.g. TreeSet) using the comparator on the head for each one, then when you consume an element, pop the iterator out and back into the set.  That way you don't need to do a full sort each time through, but rather just re-sort those that have changed.

src/main/scala/com/fulcrumgenomics/vcf/JointVariantContextIterator.scala
```diff
@@ -0,0 +1,65 @@
...
+
+package com.fulcrumgenomics.vcf
+
+import htsjdk.samtools.SAMSequenceDictionary
+import htsjdk.variant.variantcontext.{VariantContext, VariantContextComparator}
+
+/**
+  * Iterates over multiple variant context iterators such that we return a list of contexts for the union of sites
+  * across the iterators.  If samples is given, we subset each variant context to just that sample.
+  */
```
tfenne: The use of sampleNames seems to me to be driven by a very specific use case that is a lot less general than the iterator itself (i.e. I want a single sample per iterator).  I think I would tend to exclude that from this class and either:
a) Just have the caller map the iterators ahead of time `(iters.zipWith(samples).map { case (i, s) => i. subContextFromSample(s) }`
b) Provide an apply method in a companion object that does the mapping of the iterators before providing them to the `JointVariantContextIterator`

src/test/resources/com/fulcrumgenomics/vcf/intervalList.1-10.assess_phasing_metrics.block_lengths.txt
```diff
@@ -0,0 +1,6 @@
+is_truth	length	count
```
tfenne: Just thinking aloud - would it make sense to have the first column be `dataset` or something like that with values being `truth` and `called`?  I think that might be easier to consume, and also leaves room if you wanted to add more in future (e.g. the intersection of the truth and called phase blocks).

src/test/scala/com/fulcrumgenomics/vcf/AssessPhasingTest.scala
```diff
@@ -0,0 +1,787 @@
...
+import com.fulcrumgenomics.util.{Metric, NumericCounter}
+import com.fulcrumgenomics.vcf.Cigar.IlluminaSwitchErrors
+import dagr.commons.io.PathUtil
+import htsjdk.samtools.SAMFileHeader
+import htsjdk.samtools.util.{Interval, IntervalList}
+import htsjdk.variant.variantcontext.writer.{Options, VariantContextWriterBuilder}
+import htsjdk.variant.variantcontext.{GenotypeBuilder, VariantContext, VariantContextBuilder}
+import htsjdk.variant.vcf.VCFHeader
+
+object AssessPhasingTest {
```
tfenne: I'm going to skip reviewing these tests for now because I'm exhausted after reviewing everything so far, and I'm not sure how much value it will add.  Let me know if you want me to take a look later.

src/test/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIteratorTest.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  private val dict = new VariantContextSetBuilder().header.getSequenceDictionary
+
+  private def emtpyIntervalList(): IntervalList = {
+    val header = new SAMFileHeader
+    header.setSequenceDictionary(this.dict)
+    new IntervalList(header)
+  }
+
+  "ByIntervalListVariantContextIterator" should "return no variants if the interval list is empty" in {
```
tfenne: A couple more tests I would think to add:
1. How about Indels that span the start or end of an interval (i.e. a deletion from 495-505 and an interval from 500-1000)
2. I think that the using-index mechanism will return the same variant twice if a variant overlaps two intervals that can't be merged because they are more than 1bp apart:
    Deletion spanning 495-505
    Interval1 is 400-498
    Interval2 is 502-599

src/test/scala/com/fulcrumgenomics/vcf/JointVariantContextIteratorTest.scala
```diff
@@ -0,0 +1,111 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) $year Fulcrum Genomics
```
tfenne: $year

src/test/scala/com/fulcrumgenomics/vcf/JointVariantContextIteratorTest.scala
```diff
@@ -0,0 +1,111 @@
...
+      .addVariant(refIdx=0, start=10, variantAlleles=List("A"))
+      .addVariant(refIdx=0, start=30, variantAlleles=List("A"))
+    val builderRight = new VariantContextSetBuilder()
+      .addVariant(refIdx=0, start=10, variantAlleles=List("A"))
+      .addVariant(refIdx=0, start=20, variantAlleles=List("A"))
+    val iterator = new JointVariantContextIterator(iters=Seq(builderLeft.iterator, builderRight.iterator), dict=dict)
+    // pos: 10 status: both
+    iterator.hasNext shouldBe true
+    iterator.next().flatten match {
+      case Seq(left, right) => compareVariantContexts(left, right)
```
tfenne: Maybe add an explicit `case _ => fail("....")` here?

src/test/scala/com/fulcrumgenomics/vcf/JointVariantContextIteratorTest.scala
```diff
@@ -0,0 +1,111 @@
...
+    }
+    // pos: 20 status: right
+    iterator.hasNext shouldBe true
+    iterator.next() match {
+      case Seq(None, Some(right)) => compareVariantContexts(right, builderRight.last)
```
tfenne: Ditto

tfenne:
That's a _lot_ of code @nh13.  I've reviewed where I can, and left comments where I've skipped sections.  Would love to understand a little more the design philosophy behind the cigar stuff, but maybe we can talk about that in person.

A few more general comments:

- Should document in the CLP that the VCFs are expected to both be single-sample VCFs.
- Doc should be really clear that it's only dealing with biallelic SNPs
- All the licenses seem to have an extra blank line at the end

And back to you.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Note to self: how about a version that takes in `Seq[(Long, Long)]`?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Changing to `The interval list over which known phased variants should be kept`

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Removed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: We read over each VCF multiple unfortunately.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done, good catch.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Removed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Both need the same information at the site level, so it is convenient to both at the same time.   Perhaps the method name should be changed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Why not both?  Fixed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: For loops, for loops everywhere.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: For N50, I want to start with the shortest block.  The definition is the smallest block length such that the sum of all lengths of blocks shorter than N50 are 50% of the sum of all block lengths.  So I start with the shortest block(s) and add blocks until the mass is at least 50%.  This is also why I start with the _largest_ block for L50.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: yikes, fixed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: 1. I'll start warming up to `foorloop`, I swear.
2. Since the order of traversal is opposite.
3. Good catch.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: They computation may fail.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: 👍

nh13: Correctly so...

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: So it is.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Has been removed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: okey dokey.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: removed in later commit.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: So many places, that would be nice, I will add it.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: It implies that this known variant overlaps the span of a phase block.  The variant does not need to be called.

nh13: Also, I could really use help naming.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Removed.  I was wondering if the values would be the same.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Added `(i.e. the # of sites with both known and called phased variants)`

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: I wanted to avoid having a ton of metrics (I failed).  But it would be
```
short_accuracy = 1 - (num_short_switch_errors / num_switch_sites)
long_accuracy = 1 - (num_long_switch_errors / num_switch_sites)
illumina_point_accuracy = 1 - (num_illumina_point_switch_errors / num_illumina_switch_sites )
illumina_long_accuracy = 1 - (num_illumina_long_switch_errors / num_illumina_switch_sites )
```

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: I generally wont do it anymore.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: First mover advantage.  Fixed.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: I have changed the tool to operate by chromosome (for parallelization) so I fixed this for free!

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: `JointVariantContextIterator` can handle more than two samples!  I can map it to a tuple, then the parameter's signature would be `Iterator[Tuple2[Option[VariantContext], Option[VariantContext]]]`

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: changed to `ctxs.headOption.exists(callPhaseBlockDetector.overlapsAny)`

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Added in the check to the overlap constructors.  Fun story, it turns out some tools will provide overlapping blocks, so I added some functionality to resolve them if the user wants.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Comment is now ` remove leading `*End` since we are going to prepend Seq(BothEnd)`

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: :/

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: This is a detail neither very well encoded in the VCF spec nor in the VCF code in `htsjdk`, but the order of alleles defines the phase!

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: It happens on the first index for the cigar!  Not belt and suspenders.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: It would, thanks!

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: This is the # of sites that _could_ be long switch errors, or the denominator if we were to compute an error rate.

nh13: Changing this to `numSites`.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: I have some good tests, so I don't think it is worth it.

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
nh13: NA after new code.

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
nh13: NA in new code.

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIterator.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/JointVariantContextIterator.scala
nh13: Will make an either, make this constructor private, and add apply methods.

src/main/scala/com/fulcrumgenomics/vcf/JointVariantContextIterator.scala
nh13: Adding a note.  I am not seeing this as the bottleneck at the moment.

src/main/scala/com/fulcrumgenomics/vcf/JointVariantContextIterator.scala
nh13: I will go with a version of a)

src/test/resources/com/fulcrumgenomics/vcf/intervalList.1-10.assess_phasing_metrics.block_lengths.txt
nh13: I will make it  generic.

src/test/scala/com/fulcrumgenomics/vcf/ByIntervalListVariantContextIteratorTest.scala
nh13: I ran into 2) in Picard's version of this class, so good idea.

src/test/scala/com/fulcrumgenomics/vcf/JointVariantContextIteratorTest.scala
nh13: Fixed.

src/test/scala/com/fulcrumgenomics/vcf/JointVariantContextIteratorTest.scala
nh13: I'll just leave it as match exceptions!

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
tfenne: I generally appreciate that ;)

tfenne:
I think it would be good for us to talk through N50 in person.  Given that it's `50` it's probably not super important, but I've always thought of N50 as inverted from the way you describe it - i.e. as the _longest_ `length` of contig where all contigs >= `length` contain >= 50% of the bases in the assembly (or phased genome).

Confusingly [Wikipedia](https://en.wikipedia.org/wiki/N50,_L50,_and_related_statistics) describes N50 the way you do (i.e. _shorter_ than), and then completely inverts it when discussing N90.   Sigh.

PR #73, opened 2016-07-29
Adding VariantContextSetBuilder.

nh13:
VariantContextSetBuilder helps tests create variant contexts and VCF
files.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
+++ b/src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
@@ -1,0 +1,190 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+package com.fulcrumgenomics.testing
+
+import java.nio.file.Files
+
+import dagr.commons.CommonsDef._
+import htsjdk.samtools.SAMSequenceDictionary
+import htsjdk.variant.variantcontext._
+import htsjdk.variant.variantcontext.writer.{Options, VariantContextWriterBuilder}
+import htsjdk.variant.vcf.{VCFFileReader, VCFHeader, VCFHeaderLine}
+
+import scala.collection.JavaConversions._
+import scala.collection.mutable.ListBuffer
+
+object VariantContextSetBuilder {
+  def apply(sampleName: String): VariantContextSetBuilder = {
+    new VariantContextSetBuilder(List(sampleName))
+  }
+}
+
+/**
+  * A class to make artificial variant contexts and VCFs just a little bit easier to generate.
+  *
+  * This builder uses the default sequence dictionary from [SamRecordSetBuilder] by default.
+  */
+class VariantContextSetBuilder(sampleNames: Seq[String] = List("Sample")) extends Iterable[VariantContext] {
+
+  if (sampleNames.isEmpty) throw new IllegalArgumentException("At least one sample name must be given")
+
+  private var _header = {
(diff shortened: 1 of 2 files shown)


nh13:
@tfenne I can add a few more tests, but this is useful else where and in #72, so I would like to get it into master soon.

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/**
+  * A class to make artificial variant contexts and VCFs just a little bit easier to generate.
+  *
+  * This builder uses the default sequence dictionary from [SamRecordSetBuilder] by default.
+  */
+class VariantContextSetBuilder(sampleNames: Seq[String] = List("Sample")) extends Iterable[VariantContext] {
+
+  if (sampleNames.isEmpty) throw new IllegalArgumentException("At least one samle name must be given")
```
tfenne: samle -> sample

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  private var _header = {
+    val h = new VCFHeader(Set.empty[VCFHeaderLine], sampleNames.toList)
+    h.setSequenceDictionary(new SamRecordSetBuilder().header.getSequenceDictionary)
+    h
+  }
+
+  private var variants = new ListBuffer[VariantContext]()
```
tfenne: mutable collection that's also a var?

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  def header: VCFHeader = this._header
+
+  /** Sets the VCF header for this builder.  This should be set before adding variants. */
+  def setHeader(header: VCFHeader): this.type = {
+    this._header = header
+    this
```
tfenne: I think I said this elsewhere, but anywhere where the majority of the functions return `this` I would create a function `yieldingThis(f: => Unit): this.type = { f(); this}` and use it to tidy things up.

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  /** Adds the header line to the header for this builder.  This should be set before adding variants. */
+  def addMetaDataLine(headerLine: VCFHeaderLine): this.type = {
+    this._header.addMetaDataLine(headerLine)
+    this
+  }
+
+  /** Add a variant. */
+  def addVariant(refIdx: Int = 0,
```
tfenne: This could use some documentation since it's not clear to me how this works if your builder has > 1 sample without digging into the code, since you pass in `genotypeAlleles` and `sampleName`.

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -0,0 +1,163 @@
...
+                 genotypeAlleles: List[String] = List.empty,
+                 sampleName: Option[String] = None,
+                 phased: Boolean = false): this.type = {
+    val contig          = this.dict.getSequence(refIdx).getSequenceName
+    val alleles         = toAlleles(variantAlleles)
+    val stop            = VariantContextUtils.computeEndFromAlleles(alleles, start.toInt, -1)
+    val (ctxBuilder, prevGenotypes) = this.variants.find { ctx =>
+      ctx.getContig == contig &&
+        ctx.getStart == start &&
+        ctx.getEnd == stop &&
```
tfenne: I think VCFs are only _supposed_ to have one entry for a given ref:start, so isn't the comparison of stop unnecessary?

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -0,0 +1,163 @@
...
+  def toTempFile(deleteOnExit: Boolean = true): PathToVcf = {
+    val path = Files.createTempFile("VariantContextSet.", ".vcf.gz")
+    if (deleteOnExit) path.toFile.deleteOnExit()
+    write(path)
+    path
+  }
+
+  /** Writes the contents to a temporary file that will be deleted when the JVM exits and
+    * provides a [VCFFileReader]. */
+  def toVCFFileReader(deleteOnExit: Boolean = true): VCFFileReader = {
```
tfenne: I know the type is this way, but can we have the method be `toVcfFileReader` please?

tfenne:
@nh13 Looks nice, and super useful.  Maybe just a bit more scaladoc so we're not scratching our heads when we come to use it after a few weeks' absence?

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
nh13: Done.

nh13: Actually, [VCF 4.2](https://samtools.github.io/hts-specs/VCFv4.2.pdf) says "_It is permitted to have multiple records with the
same POS._" in section 1.4.1

nh13:
@tfenne one more look at the `addVariant` doc?

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
tfenne: TIL

PR #75, opened 2016-08-02
Modifications to ConsensusCaller to add some more detailed tracking m…

tfenne:
…etrics and expose all four likelihoods.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
@@ -26,7 +26,6 @@ package com.fulcrumgenomics.umi
 
 import java.util
 
-import com.fulcrumgenomics.umi.ConsensusCaller.Base
 import com.fulcrumgenomics.util.MathUtil
 import com.fulcrumgenomics.util.NumericTypes._
 import htsjdk.samtools.util.SequenceUtil
@@ -35,6 +34,13 @@ import scala.math.{max, min}
 
 object ConsensusCaller {
   type Base = Byte
+
+  private val DnaBasesUpperCase: Array[Base] = Array('A', 'C', 'G', 'T').map(_.toByte)
+  private val DnaBaseCount  = DnaBasesUpperCase.length
+
+  /** Returns a copy of the DNA bases in the order they are used in all internal arrays. */
+  def DnaBases: Array[Base] = DnaBasesUpperCase.clone()
+
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -141,6 +159,22 @@ class ConsensusCaller(errorRatePreLabeling:  PhredScore,
...
+
+    /**
+      * Returns the calculated likelihoods of the four bases given the data and the model. The likelihoods
+      * returned factor in both the base observations and the probability of error prior to applying the
+      * labels.
+      */
+    def logLikelihoods: Array[LogProbability] = {
+      // get the sum of the likelihoods
+      // pick the base with the maximum posterior
+      val lls  = likelihoods
```
nh13: I thought all `=`s had to be aligned?

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -141,6 +159,22 @@ class ConsensusCaller(errorRatePreLabeling:  PhredScore,
...
     }
+
+    /**
+      * Returns the calculated likelihoods of the four bases given the data and the model. The likelihoods
+      * returned factor in both the base observations and the probability of error prior to applying the
+      * labels.
+      */
+    def logLikelihoods: Array[LogProbability] = {
+      // get the sum of the likelihoods
+      // pick the base with the maximum posterior
```
nh13: one line comment `get the sum of the likelihoods and pick the base with the maximum posterior`

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -141,6 +159,22 @@ class ConsensusCaller(errorRatePreLabeling:  PhredScore,
...
+      val lls  = likelihoods
+      val likelihoodSum   = LogProbability.or(lls)
+      val posteriors      = lls.map(LogProbability.normalizeByLogProbability(_, likelihoodSum))
```
nh13: when the `_` is far away from the `map` I'd prefer to use a named var.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -141,6 +159,22 @@ class ConsensusCaller(errorRatePreLabeling:  PhredScore,
...
+      val errors          = posteriors.map(LogProbability.not)
+      val errorsTwoTrials = errors.map(LogProbability.probabilityOfErrorTwoTrials(LnErrorRatePreLabeling, _))
```
nh13: same here.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -119,7 +128,16 @@ class ConsensusCaller(errorRatePreLabeling:  PhredScore,
...
+    def contributions: Int = this.observations.sum
+
+    /** Gets the number of observations of the base in question. */
+    def observations(base: Base): Int = base match {
+      case 'A' => this.observations(0)
+      case 'C' => this.observations(1)
+      case 'G' => this.observations(2)
+      case 'T' => this.observations(3)
+      case x   => throw new IllegalArgumentException("Unsupported base: " + x.toChar)
+    }
```
nh13: Could also just return `this.observations` since folks have access to `DnaBases` above.  Perhaps both since this is a nice convenience method.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -141,6 +159,22 @@ class ConsensusCaller(errorRatePreLabeling:  PhredScore,
       val base = DnaBasesUpperCase(maxLlIndex)
       (base, q)
     }
+
+    /**
+      * Returns the calculated likelihoods of the four bases given the data and the model. The likelihoods
```
nh13: Rather than `Returns` perhaps `Gives`.  For `def`s that have no `()` you should write their comments like they are `val`s.

nh13:
@tfenne Minor stuff, I'll trust you to fixup and merge.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
tfenne: It's just that observations is an array, so I would want to clone it to return the whole thing.  I figured accessing either the total (which is provided by `contribution`) or the per-base values would avoid that.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
tfenne: Changed, although I'm not sure I agree with your assertion.  If the computation is expensive then I think I'd prefer the documentation to make it clear that there's a computation happening every time and not a simple access.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
tfenne: Comment was a copy/paste error anyway!  I've killed it.

PR #78, opened 2016-08-03
Fixed consensus caller to output N at Q2 if there is no single base...

tfenne:
…with max likelihood.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
@@ -35,6 +35,7 @@ import scala.math.{max, min}
 object ConsensusCaller {
   type Base = Byte
 
+  private val NoCall = 'N'.toByte
   private val DnaBasesUpperCase: Array[Base] = Array('A', 'C', 'G', 'T').map(_.toByte)
   private val DnaBaseCount  = DnaBasesUpperCase.length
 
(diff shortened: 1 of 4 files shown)


tfenne:
@nh13 I considered making the test be if the max posterior is below `log(0.5)` which would be a more efficient check, though marginally less correct since a base with ~0.250001 posterior is technically more likely than the rest if the error is equally distributed.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -146,18 +147,24 @@ class ConsensusCaller(errorRatePreLabeling:  PhredScore,
...
-      // Pr(error) = Pr(any pre-UMI error AND correct consensus) + Pr(no pre-UMI error AND any error in consensus)
-      //               + Pr(pre-UMI error AND error in consensus, that do not give us the correct bases)
-      // The last term tries to capture the case where a pre-UMI error modifies the base (ex. A->C) but a sequencing
-      // error calls it the correct base (C->A).  Only 2/3 times will the two errors result in the incorrect base.
-      val p = LogProbability.probabilityOfErrorTwoTrials(LnErrorRatePreLabeling, pConsensusError)
-      val q = PhredScore.cap(PhredScore.fromLogProbability(p))
-      val base = DnaBasesUpperCase(maxLlIndex)
-      (base, q)
+
+      lls.count(_ == maxLikelihood) match {
```
nh13: Or could have `MathUtil.maxWithIndex` return -1 if there is more than one value with the max.

nh13:
@tfenne back to you, decide what you want, and merge at will.

PR #9, opened 2016-03-04
Adding a tool to demultiplex FASTQs using an Illumina Sample Sheet

nh13:

PR #79, opened 2016-08-04
Various tools and classes as pre-requisites for DemuxFastqs

nh13:
This is an attempt to split apart some of the functionality in https://github.com/fulcrumgenomics/fgbio/pull/9

I have made separate commits for each major addition.


src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
```diff
@@ -0,0 +1,65 @@
...
+             val sampleName: String,
+             val libraryId: Option[String],
+             val project: Option[String],
+             val description: Option[String],
+             val lane: Option[Int] = None,
+             val r1SampleBarcodeBases: Option[String] = None,
+             val r2SampleBarcodeBases: Option[String] = None,
+             val i7IndexBases: Option[String] = None,
+             val i5IndexBases: Option[String] = None,
+             var sampleBarcode: Option[SampleBarcode] = None) {
```
nh13: I could use a few opinions here: 

1. should we have a `val` for each thing we could ever imagine? should we store optional values in a `Map[String, Option[Any]]` using the header names in `SampleSheet`?
2. should this be a case class?

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
```diff
@@ -0,0 +1,65 @@
...
+  * Optionally contains information about dual indexes: i7 and i5.
+  *
+  * @author Nils Homer
+  *
+  * @param sampleOrdinal the sample ordinal if this sample belongs to a sample set.
+  * @param sampleId the unique sample identifier.
+  * @param sampleName the sample name.
+  * @param libraryId the library identifier.
+  * @param project the project identifier.
+  * @param description the sample description.
```
nh13: needs `@param lane`.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -150,24 +141,64 @@ object ReadSegment {
...
-
   /** Creates a new read segment of the same type but with a new length */
   def apply(segment: ReadSegment, length: Int): ReadSegment = segment match {
     case seg: Template         => Template(seg.offset, length)
     case seg: SampleBarcode    => SampleBarcode(seg.offset, length)
     case seg: MolecularBarcode => MolecularBarcode(seg.offset, length)
     case seg: Skip             => Skip(seg.offset, length)
   }
+
+  def structureRead(segments: Seq[ReadSegment], bases: String, useOffset: Boolean = false): String = {
```
nh13: I am not in love with this.  The reason for this method and its `useOffset` parameter, is if we the sample barcode as one contiguous sequence specified in the sample sheet, but the sample barcode is in multiple places across a given read (say inline at the start and end of read one).  I think an early commit in `ReadStructure` I had `ReadStructure. structureRead ` be able to only use a specific type of `ReadSegment`, which would be ideal.  Thoughts on how I can implement this more cleanly?

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
```diff
@@ -0,0 +1,152 @@
...
+  val SampleName: String     = "Sample_Name"
+  val LibraryId: String      = "Library_ID"
+  val SampleProject: String  = "Sample_Project"
+  val Description: String    = "Description"
+  /** Optional header names for standard fields in the Illumina Experiment Manager Sample Sheet. */
+  val Lane: String           = "Lane"
+  val R1BarcodeBases: String = "R1_Barcode_Bases"
+  val R2BarcodeBases: String = "R2_Barcode_Bases"
+  val I7Bases: String        = "Index"
+  val I5Bases: String        = "Index2"
```
nh13: I could break this out to a separate object, but it seems natural to live here, though folks may want to extend that class to add more header names.  Thoughts?

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
```diff
@@ -0,0 +1,152 @@
...
+      case Some(str) if str.isEmpty => None
+      case Some(str) => Some(str.trim)
+    }
+  }
+
+  private def getIntField(sampleDatum: Map[String, String], name: String): Option[Int] = {
+    getStringField(sampleDatum, name).map(_.toInt)
+  }
+
+  private[miseq] val SplitRegex = ",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)"
```
nh13: Sometimes commas are in quotes (ex. `A,B,"50,000",C`).

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -0,0 +1,129 @@
...
+/**
+  * Base trait for metrics.
+  *
+  * All classes extending this class should be a case class.
+  */
+trait Metric {
+  private lazy val reflectiveBuilder = new ReflectiveBuilder(this.getClass)
+
+  /** Get the names of the arguments in the order they were defined. */
+  def argNames: Seq[String] = {
```
nh13: Maybe `metricNames`?

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -0,0 +1,129 @@
...
+    this.reflectiveBuilder.argumentLookup.ordered.map(_.name)
+  }
+
+  /** Get the values of the arguments in the order they were defined. */
+  def argValues: Seq[String] = {
```
nh13: Maybe `metricValues`?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -150,24 +141,64 @@ object ReadSegment {
...
+    var idx = 0
+    val barcodeBases = new StringBuilder
+    while (idx < segments.length) {
+      val seg = segments(idx)
+      barcodeBases.append(if (useOffset) seg.structureRead(bases, offset = -1) else seg.structureRead(bases, offset = barcodeBases.length))
+      idx += 1
+    }
+    barcodeBases.toString()
```
nh13: Note to self: needs a test

src/main/scala/com/fulcrumgenomics/util/SampleBarcodeMetric.scala
```diff
@@ -0,0 +1,112 @@
...
+package com.fulcrumgenomics.util
+
+import scala.collection.mutable
+
+object SampleBarcodeMetric {
+  def apply(barcodeName: String, libraryName: String, barcode: String): SampleBarcodeMetric = {
+    new SampleBarcodeMetric(BARCODE_NAME=barcodeName, LIBRARY_NAME=libraryName, BARCODE=barcode)
+  }
+
+  def finalizeMetrics(barcodeToMetrics: mutable.Map[String, SampleBarcodeMetric],
```
nh13: Note to self: needs a doc and tests.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -150,24 +141,64 @@ object ReadSegment {
...
   def offset: Int
   /** The length of this segment. */
   def length: Int
   /** The single-character symbol representing this segment. */
   def symbol: Char
+
+  /** Gets first and last bases (exclusive, 0-based) associated with this read segment.  If strict is false then only
+    * bound the last base to `basesLen`, otherwise if the end is too big, throw an exception.
+    **/
+  private[util] def getStartAndEnd(basesLen: Int, offset: Int = -1, strict: Boolean): (Int, Int) = {
```
tfenne: Probably either drop the `get` from the name or call it `range` or something like that?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
tfenne: Honestly I'm having a really hard time figuring out what it is this method is trying to do.  I think the name should probably be `extractBarcode` but even that's not clear to me.  A concise doc comment of what it is this method is doing would help me.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -150,24 +141,64 @@ object ReadSegment {
...
+      if (strict) throw new IllegalStateException("Not enough bases for read segment: " + this)
+      else (start, basesLen)
+    }
+    else (start, end)
+  }
+
+  /** Gets the bases associated with this read segment.  If strict is false then only return
+    * the sub-sequence for which we have bases in `bases`, otherwise throw an exception.
+    **/
+  def structureRead(bases: String, offset: Int = -1, strict: Boolean = true): String = {
```
tfenne: I know this mirrors the name in `ReadStructure` but I don't think that's appropriate.  Can we call these `extractBases` and `extractBasesAndQualities` or something similar?  I think that's a more accurate description of what they do.

tfenne:
In the PR on ReadStructure I'm a bit confused by all the extra offsets being passed around and what exactly it is you're trying to accomplish. Sorry.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: Consider a read structure of "10B50T10B".  A read will be 70bp, but the sample barcode bases to match against in the sample sheet will be 20bp (missing the template bases obviously).  So the input to this method will be a string of length 20 (sample barcode bases read) but the three read segments ("10B", "50T", "10B") have offsets relative to a 70bp read.

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
```diff
@@ -0,0 +1,65 @@
...
+  * @param sampleBarcode the full sample barcode for optimized access.
+  */
+class Sample(val sampleOrdinal: Int,
+             val sampleId: String,
+             val sampleName: String,
+             val libraryId: Option[String],
+             val project: Option[String],
+             val description: Option[String],
+             val lane: Option[Int] = None,
+             val r1SampleBarcodeBases: Option[String] = None,
```
tfenne: I don't understand why we have four things for barcodes.  My understanding is that in a standard illumina sample sheet there can be up to two (i7 and i5) and that anything else is site-specific.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleBarcode.scala
```diff
@@ -0,0 +1,46 @@
...
+/**
+  * A little class to store the sample barcode as bytes and concatenated with `IlluminaUtil.barcodeSeqsToString`.
+  */
+class SampleBarcode(sampleBarcodes: Seq[String]) {
+  import SampleBarcode.SampleBarcodeDelimiter
+
+  val barcodeBytes: Array[Byte]   = sampleBarcodes.flatMap(_.getBytes).toArray
+  val concatenatedBarcode: String = sampleBarcodes.filter(_.nonEmpty).mkString(SampleBarcodeDelimiter)
+
+  override def equals(o: Any): Boolean = o.isInstanceOf[SampleBarcode] && this.hashCode == o.hashCode
```
tfenne: This doesn't seem right - multiple concatenated barcodes could produce the same hashcode.  Why not `this.concatenatedBarcodes == that.asInstanceOf[SampleBarcode]concatenatedBarcodes`

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
```diff
@@ -0,0 +1,152 @@
...
+object SampleSheet {
+  /** Required header names for standard fields in the Illumina Experiment Manager Sample Sheet. */
+  val SampleId: String       = "Sample_ID"
+  val SampleName: String     = "Sample_Name"
+  val LibraryId: String      = "Library_ID"
+  val SampleProject: String  = "Sample_Project"
+  val Description: String    = "Description"
+  /** Optional header names for standard fields in the Illumina Experiment Manager Sample Sheet. */
+  val Lane: String           = "Lane"
+  val R1BarcodeBases: String = "R1_Barcode_Bases"
```
tfenne: I double-checked the MiSeq Sample Sheet Quick Reference and these are not standard fields.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
tfenne: I think it's fine here.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
tfenne: Ugh.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
```diff
@@ -0,0 +1,152 @@
...
+  private def getIntField(sampleDatum: Map[String, String], name: String): Option[Int] = {
+    getStringField(sampleDatum, name).map(_.toInt)
+  }
+
+  private[miseq] val SplitRegex = ",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)"
+}
+
+/** Create a sample sheet from the given file (typically 'SampleSheet.csv'). If the library id is omitted, the
+  * sample name will be used in its stead. */
+class SampleSheet(sampleSheet: Path, lane: Option[Int] = None) extends Iterable[Sample] {
```
tfenne: Looking at this again, it feels like it would be nicer to have the primary constructor take the set of samples in, and have this code in a `SampleSheet.apply` method - would make testing much simpler.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  def iterator: Iterator[Sample] = this.samples.iterator
+
+  override def size: Int = this.samples.size
+
+  def get(index: Int): Sample = this.samples(index)
+
+  /** Reads in the the Sample data only, one per row */
+  private def getSampleData(sampleSheet: Path): List[Map[String, String]] = {
```
tfenne: I thought `get` was verboten!  `parseSampleData`?

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
```diff
@@ -0,0 +1,152 @@
...
+      case None => throw new IllegalArgumentException("Could not find the header for sample data.")
+    }).map(_.toUpperCase.trim)
+
+    // get the rows (values), so skip "[Data]" and the header
+    val lineNumber = preData.size + 2 // 0-based
+    val sampleDatums = postData.drop(2).zipWithIndex.map { case (line, rowNumber) =>
+      val values = line.split(SplitRegex, -1)
+      // check we have the correct # of columns
+      if (values.size != header.length) {
+        throw new IllegalArgumentException(s"# of columns in the header and current row do not match ('${header.length}' != '${values.size}'): row #${rowNumber + 1} line number #${lineNumber + rowNumber + 1}")
```
tfenne: Would probably be easier on most folks to just spit out the offending line in the error rather than the line number.  Not to mention it would make your life simpler here.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
```diff
@@ -0,0 +1,152 @@
...
+  /** Reads in the the Sample data only, one per row */
+  private def getSampleData(sampleSheet: Path): List[Map[String, String]] = {
+    // ignore data pre-"[Data]"
+    val (preData, postData) = Source.fromFile(sampleSheet.toFile).getLines().toList.span(!_.startsWith("[Data]"))
+
+    // get the header (keys) (skip "[Data]")
+    val header = (postData.drop(1).headOption match {
+      case Some(line) => line.split(SplitRegex, -1) // NB: include trailing empty strings
+      case None => throw new IllegalArgumentException("Could not find the header for sample data.")
+    }).map(_.toUpperCase.trim)
```
tfenne: I'm surprised by the upper casing here since all the column name constants are in their as-used mixed case format.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
tfenne: Document that providing lane here causes the set of samples to be filtered to just that lane's samples.  That wasn't clear to me up front.

src/test/scala/com/fulcrumgenomics/util/miseq/SampleBarcodeTest.scala
```diff
@@ -0,0 +1,44 @@
...
+/**
+  * Tests for SampleBarcode.
+  */
+class SampleBarcodeTest extends UnitSpec {
+  "SampleBarcode.hashCode" should "be the hash code of the concatenated barcode sequences" in {
+    new SampleBarcode(Seq("GATTACA")).hashCode shouldBe "GATTACA".hashCode
+    new SampleBarcode(Seq("A", "C", "G", "T")).hashCode shouldBe "A-C-G-T".hashCode
+  }
+
+  "SampleBarcode.equals" should "use the hash code of the concatenated barcode sequences" in {
```
tfenne: But why?

src/test/scala/com/fulcrumgenomics/util/miseq/SampleBarcodeTest.scala
```diff
@@ -0,0 +1,44 @@
...
+    new SampleBarcode(Seq("GATTACA")).equals(2) shouldBe false
```
tfenne: You can use ` == 2` instead of `.equals(2)` it amounts to the same thing.

tfenne:
@nh13 The sample sheet commit looks good.  A few minor comments, and two fields that crept in but are non-standard.

tfenne:
Woah, just noticed this is enough new code to drop our test coverage overall by 3%.  I saw some todos for extra tests - would be nice to keep coverage up.

nh13:
@tfenne yes, I am going to add those tests, since codecov showed me what I was not covering.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -0,0 +1,129 @@
...
+    if (lines.isEmpty) Seq.empty
+    else {
+      lines.map { line =>
+        val arguments = line.split(Delimiter)
+        build[T](arguments)
+      }.toSeq
+    }
+  }
+
+  def read[T <: Metric](path: Path)(implicit tt: ru.TypeTag[T]): Seq[T] = read[T](Io.readLines(path), Some(path.toString))
```
tfenne: Public method needs doc.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
tfenne: I think probably even just `names` would be fine.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
tfenne: And here `values`?  Also, I think given the requirement that subclasses be case classes, if there's some way in the type system to require that then this could be implemented as `productIterator.map(_.toString).toSeq`

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  /** Get the values of the arguments in the order they were defined. */
+  def argValues: Seq[String] = {
+    this.reflectiveBuilder.argumentLookup.ordered.map { arg =>
+      this.getClass.getDeclaredMethod(arg.name).invoke(this).toString
+    }
+  }
+
+  override def toString: String = argValues.mkString(Metric.Delimiter)
```
tfenne: Is this just for debugging?  In which case it might be nice to do `names.zip(values).toMap.toString`

src/main/scala/com/fulcrumgenomics/util/SampleBarcodeMetric.scala
```diff
@@ -0,0 +1,112 @@
...
+        updateBarcodeMetrics(barcode,  metric.copy(PF_NORMALIZED_MATCHES = metric.PF_READS / mean))
+      }
+    }
+  }
+}
+
+/**
+  * Metrics for matching reads to sample barcodes.
+  */
+case class SampleBarcodeMetric(
```
tfenne: Maybe just make things vars instead of all the copying?  

Also, let's talk about naming conventions here ... do we want to stick with all caps and underscores in the classes?  Something more-scalaish?  And/or do name translation on writing to disk?

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  /** Get the names of the arguments in the order they were defined. */
+  def argNames: Seq[String] = {
+    this.reflectiveBuilder.argumentLookup.ordered.map(_.name)
+  }
+
+  /** Get the values of the arguments in the order they were defined. */
+  def argValues: Seq[String] = {
+    this.reflectiveBuilder.argumentLookup.ordered.map { arg =>
+      this.getClass.getDeclaredMethod(arg.name).invoke(this).toString
```
tfenne: I think it would be nice if instead of toString'ing here this invoked a method in the trait like `format(thing: Any)` that either did something a little smarter, or allowed subclasses to override.  Perhaps using htsjdk's FormatUtil for starters?

tfenne:
@nh13 On the metrics stuff - my only real concern is the rigidity of it.  I think actually that if you didn't manually enforce that the column headers match the names in the class constructor exactly and in order, it would all still work and it would allow for classes to grow (new values in same class - so long as they had default values or were Options), and for re-ordering of fields too.  What do you think?

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -0,0 +1,129 @@
...
+    if (lines.isEmpty) throw new IllegalArgumentException(s"No header found in metrics" + source.map(" in source: " + _).getOrElse(""))
+    val names = lines.next().split(Delimiter)
+    argNames[T].zip(names).foreach { case (expectedName, actualName) =>
+      if (expectedName != actualName) {
+        throw new IllegalStateException(
+          s"Header out of order (found '$actualName' but expected '$expectedName' in metrics" +
+          source.map(" in source: " + _).getOrElse("")
+        )
+      }
+    }
```
nh13: remove

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: Done.

nh13: this trait extends `Product` now.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/SampleBarcodeMetric.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/SampleBarcodeMetric.scala
nh13: I have been conditioned to be able to read all caps and underscores.

nh13: So thinking more, my vote is caps and underscores, you?

tfenne: My temptation would be to go all-lower, either with underscores or with camel case.  Other than re-parsing programatically with scala, my primary use case would be to pull stuff into R or similar, and using all-caps names is (albeit a minor) annoyance.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: Doe.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleBarcode.scala
nh13: copy paste, will update

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
nh13: I will take these out of the standard definition, and make sub-classes of `Sample` and `SampleSheet` in `DemuxFastqs` to support in-line (in R1/R2) sample barcodes.

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
nh13: Not making a case class so I can sub-class.  I could make a trait->case class but pattern matching is not needed yet.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
nh13: Done.  I am going to keep the tests as is.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
nh13: Some folks use `index` and some folks `Index`.  I think we should be case insensitive.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
nh13: Done.

src/test/scala/com/fulcrumgenomics/util/miseq/SampleBarcodeTest.scala
nh13: Then I have to add brackets.

src/test/scala/com/fulcrumgenomics/util/miseq/SampleBarcodeTest.scala
nh13: fixed.

nh13:
@tfenne thanks for the review, I have fixed how I want it.  A few things:

1. I am open to the convention for naming the fields for metrics, but I vote for upper-case and underscores, due to familiarity.
2. I think that the column names in the sample sheet should be case insensitive, since we have seen sample sheets in the wild that have different cases and work just fine.  Your thoughts?
3. Ripping out the support for inline sample barcodes in the "standard" implementation of a sample sheet will require me to sub-class `Sample` and `SampleSheet` in `DemuxFastqs`, as I want to support them there.  Having the relevant methods public or protected will facilitate that. The other option is to add them back in.

@tfenne back to you.

nh13:
@tfenne we agree on underscore and lowercase for (1), I would like a 👍  / 👎  on (2), and I think (3) can be ignored as I found a way around it.

If I get a 👍  I will squash, wait for tests to pass, then merge.

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
tfenne: Ok.  I was just noting my surprise.  Maybe a doc comment where all the keys are stored would help just noting that they are stored in the case they are usually used in the sample sheet, but used case-insensitive?

tfenne:
I'm with you on 2.  On 3 - I just worry that not everyone will do this the same way, and will use different fields in the sample sheet.  We talked about supporting a Map[String,String] for parameters that are non-standard - you could even go so far as on sample to add a `def extraField[T](name: String) : T` that uses the code in ReflectUtil to parse the value and return the correct type?  Then in DemuxFastqs it could take optional parameters (with defaults?) for the column names for inline barcodes?  

If you don't like that, then maybe just having a way in SampleSheet for the column names to be parameterized?

PR #80, opened 2016-08-05
Adding a tool to demultiplex FASTQs using an Illumina Sample Sheet.

nh13:


nh13:
@tfenne I think this PR should be started in-person.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+
+  val UnmatchedSampleId: String = "unmatched"
+
+  /** Custom columns in the Illumina Experiment Manager Sample Sheet. */
+  object CustomSampleSheetHeaderNames {
+    val R1SampleBarcodeBases = "R1_Barcode_Bases"
+    val R2SampleBarcodeBases = "R2_Barcode_Bases"
+  }
+
+  /** Creates the sample output BAM for the given sample.  [setSampleBarcode] should be called first. */
```
tfenne: I would reword the comment to state that `Sample` must have it's barcode populated first.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+  def sampleOutputBam(output: Path, sample: Sample): Path = {
```
tfenne: I realize this (and other methods) don't need access to any state, but I think it's nicer to have them inside the class instead of the object for two reasons: 1) it keeps this together a little more, 2) it makes it much easier to extend DemuxFastqs and modify behaviour a little bit (e.g. introduce _another_ custom column).

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      SampleSheet(sampleSheet, lane=lane)
+        .map(new CustomSample(_))
+        .toSeq
+    )
+  }
+
+  private val solexaQualityConverter: SolexaQualityConverter = SolexaQualityConverter.getSingleton
+
+  /** A custom sample class to support inline sample barcodes for read one and read two. */
+  private[fastq] class CustomSample(sample: Sample)
```
tfenne: It feels like there ought to be a way to do this with a simple trait that you could mix in at construction time, and maybe override how SampleSheet constructs it's Sample instances?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      i7IndexBases       = sample.i7IndexBases,
+      i5IndexBases       = sample.i5IndexBases,
+      sampleBarcode      = sample.sampleBarcode,
+      extendedAttributes = sample.extendedAttributes
+    ) {
+    val r1SampleBarcodeBases: Option[String] = extendedAttribute(CustomSampleSheetHeaderNames.R1SampleBarcodeBases)
+    val r2SampleBarcodeBases: Option[String] = extendedAttribute(CustomSampleSheetHeaderNames.R2SampleBarcodeBases)
+
+    /** Returns the sample barcodes in order of sequencing. */
+    override def sampleBarcodeBases: Seq[Option[String]] = {
```
tfenne: Wouldn't it be nicer to return `Seq[String]` and flatten here?  Or do you need to know which ones are Some and which are None?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      Seq(i7IndexBases, i5IndexBases, r1SampleBarcodeBases, r2SampleBarcodeBases)
+    }
+  }
+
+  /** A custom sample barcode class that stores the reverse complement of the sample barcode. */
```
tfenne: I see what it is ... but why is it needed?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+    lazy val reverseComplementedBarcodeBytes: Array[Byte] = {
+      val bytes = barcodeBytes.clone()
+      SequenceUtil.reverseComplement(bytes)
+      bytes
+    }
+  }
+
+
+  /** Counts the nucleotide mismatches between two strings of the same length.  Ignores no calls in expectedBases.
+    * Observed base qualities less than the minimum base quality are counted as mismatches if not a no call. Observed
```
tfenne: Base qualities are not used in the method.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+    * qualities may not be null. */
+  private[fastq] def countMismatches(observedBases: Array[Byte], expectedBases: Array[Byte]): Int = {
+    var idx = 0
+    var count = 0
+    while (idx < observedBases.length) {
```
tfenne: Could be replaced with `forloop` now.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+        count += 1
+      }
+      idx += 1
+    }
+    count
+  }
+
+  /** Reads multiple FASTQs at the same time, ensuring their read names match along the way.  This is
+    * useful for when we don't have a single interleaved FASTQ. */
+  private[fastq] class ParallelFastqReader(fastq: Seq[Path], fastqs: Seq[Path]*) extends Iterator[Seq[FastqRecord]] {
```
tfenne: This seems generally useful - maybe it should be a top-level public class?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+
+    val readers: Seq[SequentialFastqSource] = (fastq +: fastqs.toSeq).map { otherFastqs: Seq[Path] =>
```
tfenne: private

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
tfenne: Though it's slightly confusing that it's taking many `Seq`s.  Since you can do things like `iterator ++ iterator` in scala, I wonder if this class shouldn't just deal with reading one set of correlated fastqs (i.e. R1, R2, I7, I5) and then outside you could just combine the iterators?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+    private def readersHaveNext(desiredValue: Boolean): Boolean = readers.forall(_.hasNext == desiredValue)
+
+    private def verifySameReadNames(records: Seq[FastqRecord]): Unit = {
+      val readName = records.head.name
+      if (records.exists(_.name != readName)) {
+        throw new IllegalArgumentException(String.format("Mismatching read names in FASTQS:\n%s\n%s\n", readName, records.find(_.name != readName).get))
+      }
+    }
+
+    def close() = readers.foreach(_.close())
```
tfenne: `safelyClose` now?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+
+  /** Helper class to convert fastq records into SAM records and write them out.
+    *
+    * @param input the iterator through groups of FASTQ records (r1, r2, i7, and i5).
+    * @param headers the headers, one per sample
+    * @param sampleSheet the sample sheet describing all the samples
+    * @param readStructures the read structures info, one for each expected fastq record that will be passed
+    *                           into `convertFastqsAndWriteRecords`.
+    */
+  private class FastqConverter(val input: Iterator[Seq[FastqRecord]],
```
tfenne: One thing I'd like to do in the near future is write a replacement for Picard's FastqToSam which has become overly complicated and doesn't support on-the-fly extraction of UMIs.  How hard would it be to refactor this class so that we had one per sample instead of one per sample sheet?  Given that it would be fairly trivial to write the single sample FastqToSam replacement.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+                               val qualityFormat: FastqQualityFormat,
+                               val minQ: Int,
+                               val maxQ: Int,
+                               val maxMismatches: Int,
+                               val minMismatchDelta: Int,
+                               val maxNoCalls: Int,
+                               val matchReverseComplement: Boolean,
+                               val readStructures: ReadStructure*) extends Iterator[AssignedRecords] {
+    /** The barcode string for reads that do not match. */
+    private val noMatchBarcode: String = {
```
tfenne:
```scala
readStructures.flatMap { rs => rs.filter { case SampleBarcode(n) => true; case _ => false }
    .map(s => "N"*s.length)
    .mkString(SampleBarcodeDelimiter)

File: src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala [hunk 13]
```diff
@@ -0,0 +1,663 @@
...
+          if (rs.isEmpty) None
+          else Some(rs)
+        }
+        .map { readStructure =>
+          StringUtil.repeatCharNTimes('N', readStructure.map(_.length).sum)
+        }.mkString(SampleBarcodeDelimiter)
+    }
+
+    /** Read structures for the template bases, one read structure per read (ex. r1, r2, i7, and i5). */
+    private val templateReadStructures = readStructures.map { readStructure =>
```
tfenne: All of these methods can be replaced now with `rs.template` or equivalent.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+          case None    => throw new IllegalArgumentException(s"Sample with id '${sample.sampleId}' did not have a sampleBarcode")
+        }
+        val libraryId: String = sample.libraryId match {
+          case Some(id) => id
+          case None     => throw new IllegalArgumentException(s"Sample with id '${sample.sampleId}' did not have a library id")
+        }
+        metricMap.put(barcode, SampleBarcodeMetric(barcodeName=sample.sampleName, libraryName=libraryId, barcode=barcode))
+      }
+      metricMap.put(noMatchBarcode, SampleBarcodeMetric(UnmatchedSampleId, UnmatchedSampleId, noMatchBarcode))
+      metricMap.toMap
```
tfenne: You're using a LinkedHashMap above - does it's `toMap` preserve ordering?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    def hasNext: Boolean = input.hasNext
+
+    def next(): AssignedRecords = {
+      if (!hasNext) throw new NoSuchElementException("Calling next() when hasNext is false")
+
+      val fastqRecords                        = input.next()
+      val sampleBarcodeReadBases: Seq[String] = getSampleBarcode(fastqRecords).filter(_.nonEmpty)
```
tfenne: Why do you need to filter here?  I would expect `getSampleBarcode()` to return only useful stuff?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      val sampleOrdinal: Int                  = getSampleOrdinalFromSampleBarcode(sampleBarcodeReadBases.mkString, sampleSheet).getOrElse(headers.length)
+      val header: SAMFileHeader               = headers(sampleOrdinal-1)
```
tfenne: I don't love the fact that the headers are stored in a Seq and found by `ordinal - 1`.  What do you think about just keeping them in a map by barcode?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      val molecularBarcodes: Seq[String]      = getMolecularBarcodes(fastqRecords).filter(_.nonEmpty)
+      val pairedEnd: Boolean                  = readStructures.count(_.template.nonEmpty) > 1
+
+      val records = fastqRecords.zipWithIndex.flatMap { case (record, recordIndex) =>
+        val readStructure = templateReadStructures(recordIndex)
+        if (readStructure.isEmpty) {
+          None
+        }
+        else {
+          if (1 < recordIndex) throw new IllegalArgumentException("Found more than two ends with template bases")
```
tfenne: This is a little tenuous, since it's only testing if you found template bases in recordIndex 2 or higher - but maybe it's possible (though irregular) that one of the first two reads didn't contain template bases but a later read did?

tfenne: Maybe filter on line 240 for structures that have a template segment and then get rid of your `if` on 242?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    def hasNext: Boolean = input.hasNext
+
+    def next(): AssignedRecords = {
+      if (!hasNext) throw new NoSuchElementException("Calling next() when hasNext is false")
+
+      val fastqRecords                        = input.next()
+      val sampleBarcodeReadBases: Seq[String] = getSampleBarcode(fastqRecords).filter(_.nonEmpty)
+      val sampleOrdinal: Int                  = getSampleOrdinalFromSampleBarcode(sampleBarcodeReadBases.mkString, sampleSheet).getOrElse(headers.length)
```
tfenne: The `get` prefix here made me think this is a simple lookup, when this is actually where all the matching takes place.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+        if (readStructure.isEmpty) {
+          None
+        }
+        else {
+          if (1 < recordIndex) throw new IllegalArgumentException("Found more than two ends with template bases")
+          val samRecord: SAMRecord = makeSAMRecord(header, record, readStructure, sampleId, pairedEnd, recordIndex == 0)
+          
+          // set attributes: sample and molecular barcodes
+          samRecord.setAttribute("BC", sampleBarcodeReadBases.mkString(SampleBarcodeDelimiter))
+          val umiValue = molecularBarcodes.mkString(SampleBarcodeDelimiter)
```
tfenne: Perhaps we should just call it `BarcodeDelimiter` then?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+          if (umiValue.nonEmpty) samRecord.setAttribute(umiTag, molecularBarcodes.mkString(SampleBarcodeDelimiter))
```
tfenne: Re-use `umiValue` instead of making it again.

tfenne: Also maybe `umiTag` could be a constant now that it's nearly in the spec as `RX`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+
+    def getBarcodeMetrics: Iterable[SampleBarcodeMetric] = {
+      SampleBarcodeMetric.finalizeMetrics(this.barcodeToMetrics, this.noMatchBarcode)
+      this.barcodeToMetrics.values
+    }
+
+    private def noMatchBarcodeMetric: SampleBarcodeMetric = this.barcodeToMetrics(noMatchBarcode)
+
+    /** Gets the # of mismatches for the given sample and observed sample barcode read bases. */
+    private def sampleOrdinalAndMismatches(sample: Sample, sampleBarcodeReadBases: String): (Int, Int) = {
```
tfenne: Return type needs to be documented.  Would suggest also either making it a case class, or adding a type def for `Ordinal = Int` and using that everywhere to make it clearer.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      val sampleBarcode = sample.sampleBarcode match {
+        case Some(barcode) => barcode
+        case None => throw new IllegalArgumentException(s"Sample barcode required for sample with id '${sample.sampleId}'")
+      }
+      val forwardMismatches = countMismatches(sampleBarcodeReadBases.getBytes, sampleBarcode.barcodeBytes)
+      val numMismatches = if (matchReverseComplement) {
+        if (!sampleBarcode.isInstanceOf[CustomSampleBarcode]) unreachable("Cannot match the reverse complement unless the barcode is of type 'CustomSampleBarcode'.")
```
tfenne: 1) Is this truly unreachable?
2) The error message would be meaningless to a user 
3) Why limit it to `CustomSampleBarcode` only?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+        val reverseMismatches = countMismatches(sampleBarcodeReadBases.getBytes, sampleBarcode.asInstanceOf[CustomSampleBarcode].reverseComplementedBarcodeBytes)
+        require(reverseMismatches != 0 || forwardMismatches != 0) // otherwise they are reverse complements of each other
```
tfenne: But ... what if one of your barcodes is designed this way?  E.g. `ACCGGT`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+        if (forwardMismatches < reverseMismatches) forwardMismatches
```
tfenne: `Math.min(forwardMismatches, reverseMismatches)`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      }
+      else {
+        forwardMismatches
+      }
+      (sample.sampleOrdinal, numMismatches)
+    }
+
+    /** Searches for a matching sample given the observed barcode.  Returns -1 if no match is found, given the
+      * sample sheet and matching parameters */
+    private def getSampleOrdinalFromSampleBarcode(sampleBarcodeReadBases: String, sampleSheet: SampleSheet): Option[Int] = {
```
tfenne: `matchBarcode`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+        if (forwardMismatches < reverseMismatches) forwardMismatches
+        else reverseMismatches
+      }
+      else {
+        forwardMismatches
+      }
+      (sample.sampleOrdinal, numMismatches)
+    }
+
+    /** Searches for a matching sample given the observed barcode.  Returns -1 if no match is found, given the
```
tfenne: Returns `None` if no match is found.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      * sample sheet and matching parameters */
+    private def getSampleOrdinalFromSampleBarcode(sampleBarcodeReadBases: String, sampleSheet: SampleSheet): Option[Int] = {
+      val numNoCalls: Int = sampleBarcodeReadBases.getBytes.count(base => SequenceUtil.isNoCall(base))
```
tfenne: This is, unfortunately, going to be slow due to `byte -> Byte -> byte` conversions.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      val (bestSampleOrdinal, bestMismatches, secondBestMismatches) = if (numNoCalls <= maxNoCalls) {
+        sampleSheet.map { sample => sampleOrdinalAndMismatches(sample, sampleBarcodeReadBases) }
+          .toList.sortBy(_._2).take(2) match {
+            case Nil                              => (-1,           Integer.MAX_VALUE, Integer.MAX_VALUE)
+            case List(bestTuple)                  => (bestTuple._1, bestTuple._2,      Integer.MAX_VALUE)
+            case List(bestTuple, secondBestTuple) => (bestTuple._1, bestTuple._2,      secondBestTuple._2)
+          }
+      }
+      else {
+        (-1, Integer.MAX_VALUE, Integer.MAX_VALUE)
```
tfenne: Why not just restructure so that the check for `numNoCalls > maxNoCalls` is at the top level and has returns `None` instead of use magic values here?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      }
+      // Make sure we are within the parameter limits and update barcode metrics if necessary
+      if (maxMismatches < bestMismatches || maxNoCalls < numNoCalls || (secondBestMismatches - bestMismatches) < minMismatchDelta) {
```
tfenne: I don't like that this method which sounds like it's just getting/matching, is also updating metrics :/  Maybe it could instead return a `MatchResult` that says a) whether it matched, b) which barcode to and c) with how many mismatches?  And then the updates could happen outside?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+        }
+        else if (bestMismatches == 1) {
+          bestBarcodeMetric.one_mismatch_matches += 1
+          bestBarcodeMetric.pf_one_mismatch_matches += 1
+        }
+        Some(bestSampleOrdinal)
+      }
+    }
+
+    /** Creates a SAM record (always paired) */
```
tfenne: Can we not handle unpaired reads?  It seems like we really should if this is going to be a generally useful tool.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+    private def makeSAMRecord(header: SAMFileHeader, fastqRecord: FastqRecord, readStructure: ReadStructure,
```
tfenne: `makeSamRecord`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+            throw new IllegalStateException(s"Base quality $uQual is not in the range $minQ ... $maxQ for read ${fastqRecord.header}")
+          }
+      }
+      record.setBaseQualities(baseQualities)
+
+      record
+    }
+
+    /** Based on the type of quality scores coming in, converts them to a numeric byte[] in phred scale. */
+    private[fastq] def convertQuality(quals: Array[Byte], version: FastqQualityFormat) {
```
tfenne: Return type please since it's hard to infer.

tfenne: I think I've figure out that it's actually `Unit` and all the other methods modify the byte[] in place?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      record.setReadUnmappedFlag(true)
+      if (pairedEnd) {
+        record.setReadPairedFlag(true)
+        record.setMateUnmappedFlag(true)
+        if (firstOfPair) record.setFirstOfPairFlag(true)
+        else record.setSecondOfPairFlag(true)
+      }
+      record.setAttribute(ReservedTagConstants.READ_GROUP_ID, readGroupId)
+
+      convertQuality(baseQualities, qualityFormat)
```
tfenne: I would move this up under the `val baseQualities` line so it's clear they're being modified.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      baseQualities.foreach {
```
tfenne: I would also put this down into `convertQualities()`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+        qual =>
+          val uQual: Int = qual & 0xff
```
tfenne: And since your bytes are already int scaled, I have no idea why you need to do this.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+        case FastqQualityFormat.Standard => SAMUtils.fastqToPhred(quals)
+        case FastqQualityFormat.Solexa => solexaQualityConverter.convertSolexaQualityCharsToPhredBinary(quals)
+        case FastqQualityFormat.Illumina => solexaQualityConverter.convertSolexa_1_3_QualityCharsToPhredBinary(quals)
+      }
+    }
+
+    /** Gets the sample barcode bases */
+    private def getSampleBarcode(records: Seq[FastqRecord]): Seq[String] = {
+      records.zip(sampleBarcodeReadStructures).map {
+        case (rec, rs) => rs.structureRead(rec.bases).map(_.bases).mkString
```
tfenne: Two things (applies to `getMolecularBarcodes` too:
1) I think you should filter out empty values here - it's weird to return empties
2) I think this will do the wrong thing _if_ there are two barcodes of the same type in the same read.  Imagine a read structure of `5M8S5M` - the two molecular barcodes (one each side of the sample barcode) will get concatenated and returned as a single 10-base string instead of being concatenated with the delimiter between.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+  * Possible Future Improvements:
+  * - adapter trimming
+  * - more metadata stored in the SampleSheet.csv
+  */
+@clp(
+  description = """|Demultiplexes FASTQs based on the given Illumina Experiment Manager Sample Sheet.
+                   |
+                   |Fastqs and read structures for read one, read two, i7 read, and i5 read should be given.  The read structures
+                   |may contain sample barcode bases ('B'), molecular identifier bases ('M'), template bases ('T'), and bases to
+                   |skip ('S'). Template bases should only be found on read one or read two.  Any molecular identifiers will be
```
tfenne: on -> in

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+                   |
+                   |The sample barcode for each sample in the sample sheet will be compared against the sample barcode bases in
+                   |each read, to assign each read to a sample.  Reads that to not match any sample within the given error tolerance
+                   |will be placed in the 'unmatched' file.
+                   |
+                   |The output directory will contain one BAM file per sample in the sample sheet, plus a BAM for reads that could
+                   |not be assigned to a sample given the criteria.  The output files will be the concatenation of sample id, sample
+                   |name, and sample barcode bases (expected not observed), delimited by ".".  A metrics file will also be output
+                   |for sample barcodes.  More information about these metrics can be found here:
+                   |  https://broadinstitute.github.io/picard/picard-metric-definitions.html#SampleBarcodeMetric
```
tfenne: Wait, what?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+                   |
+                   |The read group's sample id, sample name, and library id all correspond to the similarly named values in the sample sheet.
+                   |Library id will be the sample id if not found, and the platform unit will be the sample name concatenated with the sample
```
tfenne: Perhaps we should allow the input of a PU or PU prefix so it can be the run id?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+                   |
+                   |The sample section of the sample sheet should contain information related to each sample with the following keys and values:
+                   |  - Sample Identifier:  Sample_ID
+                   |  - Sample Name:        Sample_Name
+                   |  - Library Identifier: Library_ID
+                   |  - Sample Project:     Sample_Project
+                   |  - Description:        Description
+                   |
+                   |The following are optional values may included in the sample sheet to specify sample barcode bases in any read:
+                   |  - Read One Inline Sample Barcode Bases*: R1_Barcode_Bases
```
tfenne: Should we use `Index` to match the standard fields?  E.g. `R1Index` and `R2Index`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+                   |  - i5 Sample Barcode Bases:              Index2
+                   |* Note that these are custom extensions and not standard.
+                   |
+                   |The read structures will be used to extract the observed sample barcode and molecular identifiers from each
+                   |read.  The observed sample barcode will be matched to the sample barcodes extracted from the bases in the sample sheet
+                   |and associated read structures.  Please see the following link for the read structure:
+                   |  https://broadinstitute.github.io/picard/javadoc/picard/picard/illumina/parser/ReadStructure.html
+                   |
+                   |""",
+  group = ClpGroups.Utilities)
```
tfenne: I think we should strive not to have a `utilities` group.  Maybe put this in the `Fastq` group or add an `Illumina` group?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+class DemuxFastqs
```
tfenne: Should this really be `DemuxIlluminaFastqs` since it relies on an Illumina sample sheet?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+(
+  @arg(flag="f1", doc="Input fastq file (optionally gzipped) for the first read of paired end data.")
```
tfenne: Input fastq file(s) (optionally gzipped) for the first template read of paired end data.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+  val fastq1: List[PathToFastq],
+  @arg(flag="f2", doc="Input fastq file (optionally gzipped) for the second read of paired end data.", minElements = 0)
+  val fastq2: List[PathToFastq] = List.empty,
+  @arg(flag="i7", doc="Input fastq file (optionally gzipped) for the index read of the Illumina i7 sequencing primer. This is typically the I1 FASTQ (index read one for dual-indexed data).", minElements = 0)
```
tfenne: `index read of the Illumina i7 sequencing primer` -> `i7 index read`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
tfenne: Lots of multi-character flags here that we should eliminate.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+
+    // Verify that the template bases are only on read one or read two
+    if (r1ReadStructure.template.isEmpty)           throw new IllegalArgumentException("No template bases found in the read structure for read one.")
+    if (r2ReadStructure.exists(_.template.isEmpty)) throw new IllegalArgumentException("No template bases found in the read structure for read two.")
+    if (i7ReadStructure.exists(_.template.nonEmpty)) throw new IllegalArgumentException("Template bases not allowed in the i7 read.")
+    if (i5ReadStructure.exists(_.template.nonEmpty)) throw new IllegalArgumentException("Template bases not allowed in the i5 read.")
+
+    val fastqs = Seq(fastq1, fastq2, fastq7, fastq5)
+    val readStructures: Seq[ReadStructure] = Seq(Some(r1ReadStructure), r2ReadStructure, i7ReadStructure, i5ReadStructure).flatten
+    if (readStructures.length != fastqs.count(_.nonEmpty)) throw new IllegalArgumentException("The # of read structures did not match the number of fastqs.")
```
tfenne: This check shoud probably zip the fastqs with the readStructure and assert `f.isEmpty == rs.isEmpty` otherwise you will do odd things if a user messes up and provides the same number but different fastqs and read structures.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+
+  override def execute(): Unit = {
+    // Make some data structures to help us on our way
+    val r1ReadStructure          = ReadStructure(rs1)
+    val r2ReadStructure          = rs2.map(ReadStructure(_))
+    val i7ReadStructure          = rs7.map(ReadStructure(_))
+    val i5ReadStructure          = rs5.map(ReadStructure(_))
+
+    // Verify that the template bases are only on read one or read two
+    if (r1ReadStructure.template.isEmpty)           throw new IllegalArgumentException("No template bases found in the read structure for read one.")
```
tfenne: All these throws could be `fail()`s.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      )
+    }
+
+    // Read the sample sheet, convert to a custom sample type, and set the barcode bases for each sample
+    val sampleSheet: SampleSheet = DemuxFastqs.sampleSheet(sampleSheet=this.sampleSheet, lane=lane).sampleBarcodes(readStructures:_*)
+
+    // If the sample barcodes are in-line in the read (i.e. in r1 or r2) then they may be reverse complemented (ex. amplification).
+    if (matchReverseComplement) {
+      // check that we only have in-line sample barcodes.
+      if (Seq(i7ReadStructure, i5ReadStructure).exists(_.exists(_.sampleBarcode.nonEmpty))) {
```
tfenne: Again, this seems totally arbitrary - that you can't even _have_ i5 or i7 indices if you have inline indices which might be RC'd.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      qualityFormat.getOrElse(unreachable("Quality format not set")),
+      minQ,
+      maxQ,
+      maxMismatches,
+      minMismatchDelta,
+      maxNoCalls,
+      matchReverseComplement,
+      readStructures:_*)
+
+    val progress = new com.fulcrumgenomics.util.ProgressLogger(logger)
```
tfenne: Why the full package name?  Can't you just import it?  Or import it as `=> FgProgressLogger` if you need to?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+    fastqConverter.foreach { assignedRec =>
+      val writer = writers(assignedRec.sampleOrdinal-1)
```
tfenne: Still not keen on the whole match-by-ordinal stuff.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      assignedRec.records.foreach { rec => writer.addAlignment(rec) }
+      progress.record(assignedRec.records:_*)
+    }
+
+    // close them all
+    reader.close()
+    writers.foreach(_.close())
+
+    // write the metrics
+    Metric.write[SampleBarcodeMetric](metrics=fastqConverter.getBarcodeMetrics.toSeq, path=metrics)
```
tfenne: Is the type parameter needed here, or can scala infer it from the `metrics` parameter type?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+    this.qualityFormat = Some(format)
+  }
+
+  /** Creates SAM file writers for each sample in the Sample Sheet, returning them in the same order as specified
+    * by each sample's ordinal.  An extra writer is appended for records that do not match a sample barcode. */
+  private def createSamFileWriters(sampleSheet: SampleSheet): List[SAMFileWriter] = {
+    val writers: List[SAMFileWriter] = sampleSheet.map { sample =>
+      val outputBam: Path       = DemuxFastqs.sampleOutputBam(this.output, sample)
+      val header: SAMFileHeader = createSamFileHeader(sample)
+      val writer: SAMFileWriter = new SAMFileWriterFactory().makeSAMOrBAMWriter(header, false, outputBam.toFile)
```
tfenne: Can drop the assignment and the next line, and just yield the value of `new ...`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,663 @@
...
+      writer
+    }.toList
+    val unmatchedWriter = {
+      val outputFile: File      = new File(this.output.toFile, this.unmatched)
+      val header: SAMFileHeader = createSamFileHeader(new Sample(writers.length - 1, DemuxFastqs.UnmatchedSampleId, DemuxFastqs.UnmatchedSampleId, Some(DemuxFastqs.UnmatchedSampleId), None, None))
+      val writer: SAMFileWriter = new SAMFileWriterFactory().makeSAMOrBAMWriter(header, false, outputFile)
```
tfenne: Ditto on yielding `new ...`

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -139,10 +139,14 @@ class ReadStructure(segs: Seq[ReadSegment], resetOffsets: Boolean = false) exten
...
 
   private def collectSegments[T <: ReadSegment: ClassTag]: Seq[T] = this.collect { case t: T => t }
-  def template: Seq[Template] = this.collectSegments[Template]
-  def sampleBarcode: Seq[SampleBarcode] = this.collectSegments[SampleBarcode]
+  def template: Seq[Template]                 = this.collectSegments[Template]
+  def sampleBarcode: Seq[SampleBarcode]       = this.collectSegments[SampleBarcode]
   def molecularBarcode: Seq[MolecularBarcode] = this.collectSegments[MolecularBarcode]
-  def skip: Seq[Skip] = this.collectSegments[Skip]
+  def skip: Seq[Skip]                         = this.collectSegments[Skip]
+  def templateStructure(resetOffsets: Boolean = false): ReadStructure         = new ReadStructure(this.template, resetOffsets=resetOffsets)
```
tfenne: I didn't see any usages of this with `reset=true`, so I'm not sure it's needed vs. just the `def template`.  Same for the others below.

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
```diff
@@ -26,13 +26,15 @@
   */
 package com.fulcrumgenomics.util.miseq
 
+import com.fulcrumgenomics.util.ReadStructure
+import dagr.commons.CommonsDef._
+
 /**
   * Represents information about a sample within an Illumina Experiment Manager sample sheet.
   *
   * Optionally contains information about dual indexes: i7 and i5.
```
tfenne: Seems an odd thing to say.  It optionally contains many things ;)

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
```diff
@@ -26,13 +26,15 @@
   */
 package com.fulcrumgenomics.util.miseq
```
tfenne: Perhaps we should move this out of `miseq` into a `com.fulcrumgenomics.illumina` package?

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
```diff
@@ -71,5 +73,26 @@ class Sample(val sampleOrdinal: Int,
       case Some(str) => Some(str.trim)
     }
   }
+
+  /** Sets the sample barcode based on the read structure.
+    *
+    * @param sampleBarcodeReadStructures a read structure per read (ex. i7 and i5) that contains only sample barcode segments.
+    */
+  def setSampleBarcode(sampleBarcodeReadStructures: Seq[ReadStructure]): this.type = {
```
tfenne: This name is really confusing.  It's _updating_ the existing set of barcode bases, based on what we think we actually sequenced given the read-structure.  Maybe call it `updateSampleBarcodeBasedOnReadStructure`?

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
```diff
@@ -71,5 +73,26 @@ class Sample(val sampleOrdinal: Int,
...
+    import com.fulcrumgenomics.util.{SampleBarcode => ReadSegmentSampleBarcode}
```
tfenne: SampleBarcodeSegment reads better in my opinion

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
tfenne: Also it seems like the correlation between read and index is well known, but between read and ReadStructure is inferred based on position in the Seq?  That seems dangerous to me.

src/main/scala/com/fulcrumgenomics/util/miseq/Sample.scala
```diff
@@ -71,5 +73,26 @@ class Sample(val sampleOrdinal: Int,
...
+    if (sampleBarcodeReadStructures.exists(rs => !rs.forall { case s: ReadSegmentSampleBarcode => true; case _ => false } )) {
```
tfenne: If you're going to bother doing this check here ... why not just allow any read structure in and do the filtering yourself?

src/main/scala/com/fulcrumgenomics/util/miseq/SampleSheet.scala
```diff
@@ -159,4 +162,21 @@ class SampleSheet(samples: Seq[Sample]) extends Iterable[Sample] {
...
 
   def get(index: Int): Sample = this.samples(index)
+
+  /** Sets the sample barcode for the samples using the given read structures and bases from the sample sheet.
+    *
+    * The sample barcode for each read segment (ex. i7, and i5) are matched against the read structures for each read
+    * that contains a sample barcode (ex. i7, and i5).  There should be the same # of read structures as reads with
+    * sample barcodes.
+    */
+  def sampleBarcodes(readStructure: ReadStructure*): this.type = {
```
tfenne: Again, I don't think the implicit ordering of the ReadStructures is safe/clear.

PR #81, opened 2016-08-08
Add an option to ExtractUmisFromBam to update the clipping position SAM

nh13:
tag, when specified.

@tfenne I ran into a problem where I would run `ExtractUmisFromBam` after running Picard's `MarkIlluminaAdapters`. The former would remove bases from the read (i.e. molecular barcodes) and the the clipping position (usually stored in the `XT` tag) would be past the end of the read.  This would cause Picard's `SamToFastq` tool to throw an exception when clipping the reads using the clipping position.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
@@ -41,7 +41,7 @@ import scala.collection.JavaConversions._
   """
     |Extracts unique molecular indexes from reads in a BAM file into tags.
     |
-    |Currently only paired end read data are supported.
+    |Currently only unmapped paired end read data are supported.
     |
     |Only template bases will be retained as read bases (stored in the SEQ field) as specified by the read structure.
     |
@@ -76,7 +76,8 @@ class ExtractUmisFromBam
   @arg(flag = "r", doc = "The read structure, one per read in a template.")      val readStructure: Seq[String],
   @arg(flag = "b", doc = "SAM tags in which to store the molecular barcodes (one-per segment).")
                                                                                  val molecularBarcodeTags: Seq[String] = Seq.empty,
-  @arg(flag = "a", doc = "Annotate the read names with the molecular barcodes. See usage for more details.") val annotateReadNames: Boolean = false
+  @arg(flag = "a", doc = "Annotate the read names with the molecular barcodes. See usage for more details.") val annotateReadNames: Boolean = false,
+  @arg(flag = "c", doc = "The SAM tag with the position in read to clip adapters (e.g. XT as produced by Picard's MarkIlluminaAdapters).") val clippingAttribute: Option[String] = None
 ) extends FgBioTool with LazyLogging {
 
   val progress = new ProgressLogger(logger, verb="written", unit=5e6.toInt)
@@ -131,8 +132,8 @@ class ExtractUmisFromBam
         if (!r2.getSecondOfPairFlag) throw fail(s"Read ${r2.getReadName} was not the second end of a pair")
         if (!r1.getReadName.equals(r2.getReadName)) throw fail(s"Read names did not match: '${r1.getReadName}' and '${r2.getReadName}'")
         // now do some work
-        val bases1 = ExtractUmisFromBam.annotateRecord(record=r1, readStructure=rs1, molecularBarcodeTags=molecularBarcodeTags1)
-        val bases2 = ExtractUmisFromBam.annotateRecord(record=r2, readStructure=rs2, molecularBarcodeTags=molecularBarcodeTags2)
+        val bases1 = ExtractUmisFromBam.annotateRecord(record=r1, readStructure=rs1, molecularBarcodeTags=molecularBarcodeTags1, clippingAttribute=clippingAttribute)
+        val bases2 = ExtractUmisFromBam.annotateRecord(record=r2, readStructure=rs2, molecularBarcodeTags=molecularBarcodeTags2, clippingAttribute=clippingAttribute)
         if (annotateReadNames) {
           // Developer Note: the delimiter must be an ascii character less than what is usually in the read names.  For
           // example "|" doesn't work.  I am not completely sure why.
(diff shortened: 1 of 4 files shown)


src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -76,7 +76,8 @@ class ExtractUmisFromBam
   @arg(flag = "r", doc = "The read structure, one per read in a template.")      val readStructure: Seq[String],
   @arg(flag = "b", doc = "SAM tags in which to store the molecular barcodes (one-per segment).")
                                                                                  val molecularBarcodeTags: Seq[String] = Seq.empty,
-  @arg(flag = "a", doc = "Annotate the read names with the molecular barcodes. See usage for more details.") val annotateReadNames: Boolean = false
+  @arg(flag = "a", doc = "Annotate the read names with the molecular barcodes. See usage for more details.") val annotateReadNames: Boolean = false,
+  @arg(flag = "x", doc = "The SAM tag with clipping information.") val clippingAttribute: Option[String] = None
```
tfenne: A few quick things:
1. `-x` seems like an odd choice for flag.  How about either `-c` for clipping, or just omit the flag altogether?
2. A little more documentation might be nice, e.g. "The SAM tag with the position in read to clip adapters (e.g. XT as produced by Picard's MarkIlluminaAdapters".
3. Are there ever other tags anyone would want to do the same with such that this should take a `Seq` and adjust multiple?

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -191,9 +193,44 @@ object ExtractUmisFromBam {
...
+
+  /**
+    * Update the clipping information produced by Picard's MarkIlluminaAdapters to account for any non-template bases
+    * that will be removed from the read.
+    */
+  private[umi] def updateClippingInformation(record: SAMRecord,
+                                             clippingAttribute: Option[String],
+                                             readStructureBases: Seq[(String, String, ReadSegment)]): Unit = {
+    clippingAttribute.map(tag => (tag, record.getIntegerAttribute(tag))) match {
+      case None =>
```
tfenne: I think it's nice to put `Unit` on the right hand side of these cases to indicate clear intention of doing nothing.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -191,9 +193,44 @@ object ExtractUmisFromBam {
...
+      case Some((tag, null)) =>
+      case Some((tag, clippingPosition)) =>
+        var i = 0
+        var basesSoFar = 0
+        var newClippingPosition = 0
+        while (i < readStructureBases.length && basesSoFar < clippingPosition) {
+          val bases   = readStructureBases(i)._1
```
tfenne: This code makes me think that `ReadStructure.structure...` should return a case class with named values instead of tuples.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -191,9 +193,44 @@ object ExtractUmisFromBam {
...
+                                             clippingAttribute: Option[String],
+                                             readStructureBases: Seq[(String, String, ReadSegment)]): Unit = {
+    clippingAttribute.map(tag => (tag, record.getIntegerAttribute(tag))) match {
+      case None =>
+      case Some((tag, null)) =>
+      case Some((tag, clippingPosition)) =>
+        var i = 0
+        var basesSoFar = 0
+        var newClippingPosition = 0
+        while (i < readStructureBases.length && basesSoFar < clippingPosition) {
```
tfenne: Maybe I'm missing something, but couldn't this whole loop be replaced with something like (assuming you pass in the ReadStructure as rs):
```scala
val minus = rs.takeWhile(_.offset < clippingPosition).filter(_.symbol != 'T').map(_.length).sum
val newClipplingPosition = clippingPosition - minus
```

tfenne:
A few minor comments.  Address as you see fit and then merge.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
nh13: 1 & 2 - Done.
3 - I think this tool should be run right after `SamToFastq` and `MarkIlluminaAdapters`, and I don't see any other tags that should be adjusted.  My only thought is that the default should be "XT", but right now I don't know of a way to clear a command line `Option` option.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
nh13: I am going with: 
```
        val minus = readStructure.takeWhile(_.offset < clippingPosition).collect { case t: Template => t.length }.sum
        val newClippingPosition = clippingPosition - minus
```

nh13: Both are wrong, since the clipping position can happen in the middle of a stretch of template bases.  Will show on the next commit.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
nh13: Good idea, let me see what it would take.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -64,6 +64,9 @@ object ReadStructure {
     rs
   }
 
+  /** Contains the bases and optionally base qualities that correspond to the given read segment. */
+  case class ReadSegmentAndBases(bases: String, quals: Option[String] = None, segment: ReadSegment)
```
nh13: @tfenne this shows how bad I am at naming, any preference?

nh13:
@tfenne one question about the naming of `ReadSegmentAndBases` and then I will squash and merge.

nh13:
`SubRead` it is.  Merging.

PR #83, opened 2016-08-10
Remove the PS tag when the variant will not be phased.

nh13:
If the input VCF is already phased, then we should unset these values when the output variant is not phased.


tfenne:
Merge away!

PR #84, opened 2016-08-11
WIP: delimited file parsing with type conversion

tfenne:
DO NOT MERGE

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
@@ -34,7 +34,7 @@ import scala.reflect.runtime.{universe => ru}
   * Represents a row of parsed data.  Provides methods for accessing values in a type-safe
   * way either via apply() methods for non-optional fields or via get for optional fields.
   */
-class Row (private val headerIndices: Map[String,Int], private val fields: Array[String], val trim: Boolean) {
+class Row private[util] (private val headerIndices: Map[String,Int], private val fields: Array[String], val trim: Boolean) {
   /* Internal method to pull a value out of the field array by index and trim it if requested. */
   private def value(index: Int) = {
     if (index > fields.length-1) throw new IndexOutOfBoundsException(s"Invalid column index supplied: ${index}")
@@ -42,14 +42,14 @@ class Row (private val headerIndices: Map[String,Int], private val fields: Array
     if (trim) string.trim else string
   }
 
-  /** Fetches a value of the desired type by 0-based column index. */
+  /** Fetches a value of the desired type by the 0-based column index. */
   def apply[A](index: Int)(implicit tag: ru.TypeTag[A]): A = {
     val c = ReflectionUtil.typeTagToClass(tag)
     ReflectionUtil.constructFromString(c, c, value(index)).get.asInstanceOf[A]
   }
 
   /** Fetches a value of the desired type by column name. */
-  def apply[A](column: String)(implicit tag: ru.TypeTag[A]): A = apply(headerIndices.apply(column))
+  def apply[A](column: String)(implicit tag: ru.TypeTag[A]): A = apply(headerIndices(column))
 
   /** Gets a value from the column with the specified index. If the value is null or empty returns None. */
   def get[A](index: Int)(implicit tag: ru.TypeTag[A]): Option[A] = {
@@ -59,19 +59,18 @@ class Row (private val headerIndices: Map[String,Int], private val fields: Array
     }
     else {
       val c = ReflectionUtil.typeTagToClass(tag)
-      Some(ReflectionUtil.constructFromString(c, c, value(index)).get.asInstanceOf[A])
+      Some(ReflectionUtil.constructFromString(c, c, string).get.asInstanceOf[A])
     }
   }
 
   /** Gets a value from the specified column. If the value is null or empty returns None. */
-  def get[A](column: String)(implicit tag: ru.TypeTag[A]): Option[A] = get(headerIndices.apply(column))
+  def get[A](column: String)(implicit tag: ru.TypeTag[A]): Option[A] = get(headerIndices(column))
 }
 
 
 object DelimitedDataParser {
-  val DefaultDelimiter: Char = '\t'
   val DefaultBlankPolicy: Boolean = true
-  val DefaultTrim : Boolean = true
+  val DefaultTrim :       Boolean = true
 
   /** Constructs a DelimitedDataParser for a path. */
   def apply(path: FilePath, delimiter: Char): DelimitedDataParser =
@@ -91,7 +90,7 @@ object DelimitedDataParser {
   * @param trimFields whether individual fields should have their String values trimmed
   */
 class DelimitedDataParser(lines: TraversableOnce[String],
-                          val delimiter: Char = DelimitedDataParser.DefaultDelimiter,
+                          val delimiter: Char,
                           val ignoreBlankLines: Boolean = DelimitedDataParser.DefaultBlankPolicy,
                           val trimFields: Boolean = DelimitedDataParser.DefaultTrim) extends Iterator[Row] with LazyLogging {
 
@@ -103,10 +102,13 @@ class DelimitedDataParser(lines: TraversableOnce[String],
     else IndexedSeq.empty
   }
 
+  // A temporary array used for parsing each line into fields
+  private val tmp = new Array[String](headers.size * 2)
+
   // A map of header name to column index
   private val headerIndices: Map[String,Int] = _headers.zipWithIndex.toMap
 
-  /** Returns an iterator over the headers encountered in the file, in order. */
+  /** Returns the headers encountered in the file, in order. */
   def headers: Seq[String] = this._headers
 
   /** True if there is another line available for parsing. */
@@ -115,7 +117,7 @@ class DelimitedDataParser(lines: TraversableOnce[String],
   /** Retrieves the next row. */
   override def next(): Row = {
     val line   = this._lines.next()
-    val fields = line.split(delimiter)
+    val fields = split(line)
     if (fields.length != _headers.length) {
       logger.error(s"Line has incorrect number of fields. Header length is ${_headers.length}, row length is ${fields.length}")
       logger.error(s"Headers: ${_headers.mkString(delimiter.toString)}")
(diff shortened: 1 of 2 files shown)


tfenne:
@nh13 I think this is ready for a first pass review.  I'd welcome ideas for other things that should be tested.

A couple of deficiencies I don't want to try and fix just yet:
1. I found through trial and error that Dagr's reflective string to type doesn't support `Char` which would be nice
2. It will currently blow up if you use `row[T].apply("col")` on a row that has no value in the column and the target type can't be constructed from an empty string.  Maybe that's ok?  The workaround is to use `row[T].get("col")` which will return a None if the field is empty.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+import dagr.commons.reflect.ReflectionUtil
+import dagr.commons.util.LazyLogging
+
+import scala.reflect.runtime.{universe => ru}
+
+/**
+  * Represents a row of parsed data.  Provides methods for accessing values in a type-safe
+  * way either via apply() methods for non-optional fields or via get for optional fields.
+  */
+class Row (private val headerIndices: Map[String,Int], private val fields: Array[String], val trim: Boolean) {
```
nh13: Should the constructor be private, since it is only used in one place?  If not, probably a good idea to provide documentation for the parameters.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+  /* Internal method to pull a value out of the field array by index and trim it if requested. */
+  private def value(index: Int) = {
+    if (index > fields.length-1) throw new IndexOutOfBoundsException(s"Invalid column index supplied: ${index}")
+    val string = fields(index)
+    if (trim) string.trim else string
+  }
+
+  /** Fetches a value of the desired type by 0-based column index. */
```
nh13: by the

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+  def apply[A](index: Int)(implicit tag: ru.TypeTag[A]): A = {
+    val c = ReflectionUtil.typeTagToClass(tag)
+    ReflectionUtil.constructFromString(c, c, value(index)).get.asInstanceOf[A]
+  }
+
+  /** Fetches a value of the desired type by column name. */
+  def apply[A](column: String)(implicit tag: ru.TypeTag[A]): A = apply(headerIndices.apply(column))
```
nh13: why not `apply(headerIndicies(column))`? why the extra `apply`?

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  /** Gets a value from the column with the specified index. If the value is null or empty returns None. */
+  def get[A](index: Int)(implicit tag: ru.TypeTag[A]): Option[A] = {
+    val string = value(index)
+    if (string == null || string.isEmpty) {
+      None
+    }
+    else {
+      val c = ReflectionUtil.typeTagToClass(tag)
+      Some(ReflectionUtil.constructFromString(c, c, value(index)).get.asInstanceOf[A])
```
nh13: 1. `value(index)` could be `string`, or you can just call `this.apply(index)`.  
2. This needs some test coverage.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  /** Gets a value from the specified column. If the value is null or empty returns None. */
+  def get[A](column: String)(implicit tag: ru.TypeTag[A]): Option[A] = get(headerIndices.apply(column))
```
nh13: ditto about `apply`

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+}
+
+
+object DelimitedDataParser {
+  val DefaultDelimiter: Char = '\t'
+  val DefaultBlankPolicy: Boolean = true
+  val DefaultTrim : Boolean = true
```
nh13: I've gotten to like aligning the equal signs, how about here?

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  /** Constructs a DelimitedDataParser for a path. */
+  def apply(path: FilePath, delimiter: Char): DelimitedDataParser =
```
nh13: `delimiter: Char = DefaultDelimiter`?

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+    new DelimitedDataParser(Io.toSource(path).getLines(), delimiter=delimiter)
+
+  /** Constructs a DelimitedDataParser for a path. */
+  def apply(lines: Traversable[String], delimiter: Char): DelimitedDataParser =
```
nh13: ditto about delimiter.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+  // An array of the headers
+  private val _headers = {
+    if (this._lines.hasNext) this._lines.next().split(delimiter).map(h => if (trimFields) h.trim else h).toIndexedSeq
+    else IndexedSeq.empty
+  }
+
+  // A map of header name to column index
+  private val headerIndices: Map[String,Int] = _headers.zipWithIndex.toMap
+
+  /** Returns an iterator over the headers encountered in the file, in order. */
```
nh13: returns a `Seq`, not an `Iterator`.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+    val line   = this._lines.next()
+    val fields = line.split(delimiter)
+    if (fields.length != _headers.length) {
+      logger.error(s"Line has incorrect number of fields. Header length is ${_headers.length}, row length is ${fields.length}")
+      logger.error(s"Headers: ${_headers.mkString(delimiter.toString)}")
+      logger.error(s"Line: ${line}")
+      throw new IllegalStateException(s"Incorrect number of values in line.")
+    }
+
+    new Row(headerIndices, fields, trimFields)
```
nh13: name boolean parameters please: `trim=trimFields`.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -0,0 +1,128 @@
...
+  * @param lines the lines from the file
+  * @param delimiter the delimiter between columns
+  * @param ignoreBlankLines whether blank lines should be ignored
+  * @param trimFields whether individual fields should have their String values trimmed
+  */
+class DelimitedDataParser(lines: TraversableOnce[String],
+                          val delimiter: Char = DelimitedDataParser.DefaultDelimiter,
+                          val ignoreBlankLines: Boolean = DelimitedDataParser.DefaultBlankPolicy,
+                          val trimFields: Boolean = DelimitedDataParser.DefaultTrim) extends Iterator[Row] with LazyLogging {
+
```
nh13: It may be useful if the delimiter were a regex, for example, if we we want to ignore commas within quotes we could use `"\t(?=([^\"]*\"[^\"]*\")*[^\"]*$)".r`.

nh13:
@tfenne minor stuff, but I do think it would be useful to support a `Pattern` for parsing the file.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
tfenne: I'll make it `private[util]`.  I wanted to make it private but then the parser couldn't call it.  I wanted to make `Row` an inner of the parser, but then the parser couldn't extend `Iterator[Row`.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
tfenne: Fixed.  I'd made a bunch of "exploratory" changes because the compiler was giving me an inscrutable error message at some point so I was being more explicit everywhere.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
tfenne: I left it un-defaulted because you can't have default values for two methods of the same name, so I can't put defaults on both apply methods, and I would like them to be symmetric.  Maybe I'll just remove the default delimiter everywhere.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
tfenne: Cannot, sadly.

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
tfenne: Good catch, thanks.

tfenne:
@nh13 I made most of your suggested changes.  I also implemented my own string splitting as when I added a test with just delimiters on a line, or trailing multiple delimiters together _none_ of the available methods worked (not's scala's `String.split(Char)`, not java's `String.split(Pattern)`, nor htsjdk's `StringUtil.split`).  They all returned less than the expected number of fields.

I don't think pattern based splitting is the way to go here.  I think if we want to handle quoting (which I'm deferring altogether since I don't need it currently) the right thing to do is implement some state-fulness in the splitting.  I'm not convinced that a regex can handle the full set of necessary things including quoting, escaping of quotes and escaping of escape characters properly.

nh13:
👍 squash and merge at will.

PR #85, opened 2016-08-17
Added an option to include non-PF reads while grouping, off by default.

tfenne:


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -394,6 +394,7 @@ class GroupReadsByUmi
   @arg(flag="t", doc="The tag containing the raw UMI.")  rawTag: String    = "RX",
   @arg(flag="T", doc="The output tag for UMI grouping.") assignTag: String = "MI",
   @arg(flag="m", doc="Minimum mapping quality.")         minMapQ: Int      = 30,
+  @arg(flag="p", doc="Include non-PF reads.")            includeNonPfReads: Boolean = false,
```
nh13: perhaps `f` instead of `p`?  `f` is for fail, and `p` is for pass?

PR #88, opened 2016-08-25
Making the rejects BAM file optional in CallMolecularConsensusReads.

tfenne:


tfenne:
@nh13 Simple change here so that the CLP also treated outputting the rejects as optional during consensus building.

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -125,15 +127,15 @@ class CallMolecularConsensusReads
...
       readGroupId    = readGroupId,
       options        = options,
-      rejects        = Some(rej),
+      rejects        = rej,
       progress       = Some(progress)
     )
 
     consensusCaller.foreach { rec => out.addAlignment(rec) }
 
     CloserUtil.close(in)
```
nh13: want to use `safelyClose` instead?

nh13:
👍, your call on cleaning up that one line.

PR #97, opened 2016-08-31
New tool to assist in reviewing reads that contributed to variant cal…

tfenne:
…ls from consensus reads.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
@@ -1,0 +1,139 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.umi
+
+import java.nio.file.Path
+
+import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.util.Io
+import dagr.sopt.{arg, clp}
+import htsjdk.samtools.reference.{ReferenceSequenceFile, ReferenceSequenceFileFactory}
+import htsjdk.samtools.util.IOUtil.{VCF_EXTENSIONS => VcfExtensions}
+import htsjdk.samtools.util.{IntervalList, SequenceUtil}
+import htsjdk.samtools.{QueryInterval, SAMFileWriterFactory, SAMRecord, SamReaderFactory}
+import htsjdk.variant.vcf.VCFFileReader
+
+import scala.collection.JavaConversions.{asScalaIterator, iterableAsScalaIterable}
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+@clp(group=ClpGroups.Umi, description=
+  """
+    |Extracts data to make reviewing of variant calls from consensus reads easier. Creates
+    |a list of variant sites from the input VCF or IntervalList and then extracts all the
+    |consensus reads that do not contain a reference allele at the the variant sites, and
+    |all raw reads that contributed to those consensus reads.  This will include consensus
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -0,0 +1,139 @@
...
+    |Extracts data to make reviewing of variant calls from consensus reads easier. Creates
+    |a list of variant sites from the input VCF or IntervalList and then extracts all the
+    |consensus reads that do not contain a reference allele at the the variant sites, and
+    |all raw reads that contributed to those consensus reads.  This will include consensus
+    |reads that carry the alternate allele, a third allele, a no-call or a spanning
+    |deletion at the variant site.
+    |
+    |Both input BAMs must be coordinate sorted and indexed.
+    |
+    |A pair of output BAMs named {output}.consensus.bam and {output}.grouped.bam
```
nh13: `<output>` instead of `{output}`?

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -0,0 +1,139 @@
...
+    def toQueryInterval: QueryInterval = new QueryInterval(dict.getSequenceIndex(chrom), start, start)
+  }
+
+  override def execute(): Unit = {
+    def f(ext: String): Path = output.getParent.resolve(output.getFileName + ext)
+    val consensusIn = SamReaderFactory.make().open(consensusBam)
+    val groupedIn   = SamReaderFactory.make().open(groupedBam)
+    val factory = new SAMFileWriterFactory().setCreateIndex(true)
+    val consensusOut = factory.makeWriter(consensusIn.getFileHeader, true, f(".consensus.bam").toFile, ref.toFile)
+    val groupedOut   = factory.makeWriter(groupedIn.getFileHeader, true, f(".grouped.bam").toFile, ref.toFile)
```
nh13: align `=`?

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -0,0 +1,139 @@
...
+
+    // Pull out the consensus reads
+    val variants        = loadPositions(path=input, refFile=refFile)
+    val variantsByChrom = variants.groupBy(v => v.chrom)
+    val queries         = variants.map(_.toQueryInterval).toArray
+    val sources         = mutable.HashSet[String]()
+    consensusIn.query(queries, false).filter(r => nonReferenceAtAnyVariant(r, variantsByChrom)).foreach { rec =>
+      consensusOut.addAlignment(rec)
+      Option(rec.getStringAttribute(tag)).map(mi => mi.takeWhile(_ != '/')).foreach(sources.add)
```
nh13: Add a doc in the usage that all characters after the first `/` will be ignored in the `MI` tag?

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -0,0 +1,139 @@
...
+    groupedOut.close()
+  }
+
+  /** Loads the set of locations for review. */
+  private[umi] def loadPositions(path: FilePath, refFile: ReferenceSequenceFile): Seq[LocAndRef] = {
+    val buffer = ListBuffer[LocAndRef]()
+
+    if (VcfExtensions.exists(ext => path.getFileName.toString.endsWith(ext))) {
+      val in = new VCFFileReader(input.toFile, false)
+      in.filter(_.isSNP).foreach(v => buffer += LocAndRef(v.getContig, v.getStart, v.getReference.getBases()(0).toChar.toUpper))
```
nh13: add to the usage that if we have a VCF only SNP sites will be loaded?

nh13: Do we care about it being biallelic? Namely, do we want to output homozygous alts?

nh13: Can you confirm that MNPs are filtered?

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -0,0 +1,139 @@
...
+    }
+    else {
+      val list = IntervalList.fromFile(path.toFile).uniqued(false)
+      for (i <- list; pos <- i.getStart to i.getEnd) {
+        buffer += LocAndRef(i.getContig, pos, refFile.getSubsequenceAt(i.getContig, pos, pos).getBases()(0).toChar.toUpper)
```
nh13: `getSubsequenceAt` is likely slow, how about `refFile.getSequence(i.getContig).getBases()(pos-1).toChar.toUpper`?

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -0,0 +1,139 @@
...
+
+    * @param rec the SAMRecord of interest
+    * @param variantsByChromAndPos the set of variants we care about, grouped by chromosome
+    * @return true if the read is non-reference at any variant base, otherwise false
+    */
+  private def nonReferenceAtAnyVariant(rec: SAMRecord, variantsByChromAndPos: Map[String, Seq[LocAndRef]]): Boolean = {
+    !rec.getReadUnmappedFlag && variantsByChromAndPos(rec.getContig).exists { v =>
+      if (v.start >= rec.getAlignmentStart && v.start <= rec.getAlignmentEnd) {
+        val readPos = rec.getReadPositionAtReferencePosition(v.start)
+        readPos == 0 || !SequenceUtil.basesEqual(rec.getReadBases()(readPos-1), v.refBase.toByte )
```
nh13: extra space ` )`.

nh13:
Not much to change, merge when happy.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
tfenne: Looking at the (nasty) VCF code this does the right thing - it'll return any site at which there is at least one non-ref allele, all alleles are length 1.  So indels and MNPs are excluded, but homozygous alt sites will be included.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
tfenne: So long as there's an `.fai` file beside the fasta this is going to be doing indexed retrievals from the fasta.  There's definitely a tipping point where if you're reviewing many thousands of variants it would be faster to load the whole chromosome, and index into it, but for moderate amounts this should be faster.

PR #99, opened 2016-09-02
A new tool for more efficiently picking sets of long indices.

tfenne:


tfenne:
@nh13 It's possible that we should just deprecate/remove `PickIlluminaIndices` in favor of this which is much simpler.  It could probably also do with a better name and command documentation, but I wanted to see if you think we should keep both before finalizing that.

tfenne:
Also, I'm sure there are opportunities for further optimization here, but it runs pretty quickly as is.  I can pick sets of 11mers in < 1 minute, and sets of 12mers in 1-2 minutes.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -225,33 +225,25 @@ class PickLongIndices
...
   private[util] def homopolymerLengthOk(index: Index, longest: Int): Boolean = {
-    var i = 0
     var end = index.length - longest
     var ok = true
 
     // Check for a homopolymer starting at each base up to longest+1 from the end of the index
-    while (i < end && ok) {
+    forloop (0)(ok && _ < end)(_ + 1) { i =>
       val base = index(i)
       var j = i+1
```
nh13: remove `var j = i + 1`?

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -41,6 +41,47 @@ class FgBioDef extends CommonsDef {
...
+    * via foreach(), and especially with zipWithIndex.foreach().
+    *
+    * Equivalent to: for(int i=from; i<until; i+=by) f(i)
+    *
+    * @param from an initial integer values
+    * @param until a value one higher than the last value that should be accepted
+    * @param by a step value to increment by each iteration
+    * @param f a function that takes the index and is called each iteration
+    */
+  @inline def forloop(from: Int, until:Int, by:Int=1)(f: Int => Unit): Unit = {
```
nh13: Space after `:`.

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -41,6 +41,47 @@ class FgBioDef extends CommonsDef {
...
+    val comp: (Int,Int) => Boolean = if (by > 0) (_ < _) else (_ > _)
```
nh13: I am curious, is calling this function slower than having an if/else for the two cases?

```
if (by > 0) {
  var i = from 
  while (i < until) {
     ...
}
else {
}

File: src/main/scala/com/fulcrumgenomics/FgBioDef.scala [hunk 4]
```diff
@@ -41,6 +41,47 @@ class FgBioDef extends CommonsDef {
...
+    var i = from
+    while (comp(i, until)) {
+      f(i)
+      i += by
```
nh13: What if by is `0`?

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -41,6 +41,47 @@ class FgBioDef extends CommonsDef {
...
+    *
+    *  for (i=from; check(i); i=next(i)) f(i)
+    *
+    * @param from an initial value
+    * @param check a function that checks to see if a value should be evaluated (not cause termination)
+    * @param next a function that takes a value and produced the next value
+    * @param f a function called on all values
+    * @tparam T the type of the index
+    */
+  @inline @specialized def forloop[T](from: T)(check: T => Boolean)(next: T => T)(f: T => Unit): Unit = {
```
nh13: I am having a hard time understanding if this is the way to use the `@specialized` annotation, for example, see the use of the annotation _within_ the type symbol: http://stackoverflow.com/questions/13173181/specialization-of-generic-functions-in-scala-or-java.  So would it be:
```
@inline def forloop[@specialized(Char, Byte, Int, Long, Double) T](from: T)(check: T => Boolean)(next: T => T)(f: T => Unit): Unit
```
?

src/test/scala/com/fulcrumgenomics/ForLoopTesting.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @specialized def for2[T](from: T)(next: T => T)(check: T => Boolean)(f: T => Unit): Unit = {
+    var t = from
+    while (check(t)) {
+      f(t)
+      t = next(t)
+    }
+  }
+}
```
nh13: Is this file meant to be part of the PR?  If so, why not use the functions in `FgBioDef`?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+  def contains(index: Array[Byte]): Boolean = contains(new String(index))
+
+  /** Checks to see whether an index is contained in the set. */
+  def contains(index: String): Boolean = this.set.contains(index)
+}
+
+@clp(group=ClpGroups.Utilities, description =
+    """
+      |Picks a set of molecular indices that have at least a given number
+      |of mismatches between them.
```
nh13: I think it makes sense to have more documentation here both for how the program works, and for the various options.

For example, how are the indices inserted into the provided adapter sequences?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+(
+  @arg(flag="l", doc="The length of each barcode sequence.")                           val length: Int = 8,
+  @arg(flag="n", doc="The number of indices desired.")                                 val numberOfIndices: Int,
+  @arg(flag="e", doc="The minimum edit distance between two indices in the set.")      val editDistance: Int = 3,
+  @arg(flag="o", doc="File to write indices to.")                                      val output: FilePath,
+  @arg(          doc="Allow indices that are lexical reverses of one another")         val allowReverses: Boolean = false,
+  @arg(          doc="Allow indices that are reverse complements of one another")      val allowReverseComplements: Boolean = false,
+  @arg(          doc="Allow indices that are palindromic (index == revcomp(index)).")  val allowPalindromes: Boolean = false,
+  @arg(          doc="Reject indices with a homopolymer of greater than this length.") val maxHomopolymer: Int = 2,
+  @arg(flag="g", doc="The minimum GC fraction for a barcode to be accepted.")          val minGc: Double = 0.2,
```
nh13: barcode to index.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+  @arg(flag="G", doc="The maximum GC fraction for a barcode to be accepted.")          val maxGc: Double = 0.8,
```
nh13: ditto

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+  @arg(          doc="File of existing index sequences to integrate, one per line.")   val existing: Option[FilePath] = None,
+  @arg(flag="s", doc="Random seed value.")                                             val seed: Int = 1,
+  @arg(flag="a", doc="Attempts to pick the next index before quitting.")               val attempts: Int = 1e6.toInt,
+  @arg(          doc="The installation directory for ViennaRNA.")                      val viennaRnaDir: Option[DirPath] = None,
+  @arg(flag="t", doc="The temperature at which to predict secondary structure.")       val temperature: Double = 25d,
+  @arg(          doc="The lowest acceptable secondary structure deltaG.")              val minDeltaG: Double = -10,
+  @arg(          doc="Adapter sequence(s) indices will be inserted into.", minElements=0) val adapters: Seq[String] = Seq(),
+  @arg(          doc="Sequences that should be avoided.  Any kmer of 'length' that appears in these sequences and their " + "reverse complements will be thrown out.")
```
nh13: Why the `+` on one line?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+  @arg(          doc="Reject indices with a homopolymer of greater than this length.") val maxHomopolymer: Int = 2,
+  @arg(flag="g", doc="The minimum GC fraction for a barcode to be accepted.")          val minGc: Double = 0.2,
+  @arg(flag="G", doc="The maximum GC fraction for a barcode to be accepted.")          val maxGc: Double = 0.8,
+  @arg(          doc="File of existing index sequences to integrate, one per line.")   val existing: Option[FilePath] = None,
+  @arg(flag="s", doc="Random seed value.")                                             val seed: Int = 1,
+  @arg(flag="a", doc="Attempts to pick the next index before quitting.")               val attempts: Int = 1e6.toInt,
+  @arg(          doc="The installation directory for ViennaRNA.")                      val viennaRnaDir: Option[DirPath] = None,
+  @arg(flag="t", doc="The temperature at which to predict secondary structure.")       val temperature: Double = 25d,
+  @arg(          doc="The lowest acceptable secondary structure deltaG.")              val minDeltaG: Double = -10,
+  @arg(          doc="Adapter sequence(s) indices will be inserted into.", minElements=0) val adapters: Seq[String] = Seq(),
```
nh13: Adapter sequence(s) into which indices will be inserted?  (end with a preposition?)

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+  if (minGc >= maxGc)            invalid("Min GC must be greater than max GC.")
+  if (editDistance < 0)          invalid("Edit distance must be >= 0.")
+  if (editDistance > length)     invalid("Edit distance must be <= length.")
+  if (viennaRnaDir.isDefined && adapters.isEmpty) invalid("Specifying Vienna RNA dir has no effect if no adapters are provided.")
+  if (viennaRnaDir.isEmpty && adapters.nonEmpty)  invalid("Specifying adapters has no effect if Vienna RNA dir is not provided.")
+
+  /** Parse and check any set of existing indices. */
+  private val ExistingIndices = existing match {
+    case None    => Seq.empty[Index]
+    case Some(p) => Io.readLines(p).map(_.trim.toUpperCase).filter(_.nonEmpty).map(_.getBytes).toSeq
```
nh13: check that the existing indices are of the same length?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+) extends FgBioTool with LazyLogging {
+  // A few "constants"
+  private val Bases  = Array[Byte]('A', 'C', 'G', 'T')
+  private val random = new Random(seed)
+
+  // Input checking
+  existing.foreach(Io.assertReadable)
+  Io.assertCanWriteFile(output)
+  if (length < 1)                invalid("Length must be a positive integer.")
+  if (numberOfIndices < 1)       invalid("Number of indices to pick must be a positive integer.")
```
nh13: how about having less than 1 be pick as many as possible?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+  if (maxHomopolymer < 0)        invalid("Max homopolymer length must be >= 0.")
+  if (minGc >= maxGc)            invalid("Min GC must be greater than max GC.")
+  if (editDistance < 0)          invalid("Edit distance must be >= 0.")
+  if (editDistance > length)     invalid("Edit distance must be <= length.")
+  if (viennaRnaDir.isDefined && adapters.isEmpty) invalid("Specifying Vienna RNA dir has no effect if no adapters are provided.")
+  if (viennaRnaDir.isEmpty && adapters.nonEmpty)  invalid("Specifying adapters has no effect if Vienna RNA dir is not provided.")
```
nh13: Probably good to document this somewhere.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
nh13: Doh, it's right below!

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+
+  /** Parse and check any set of existing indices. */
+  private val ExistingIndices = existing match {
+    case None    => Seq.empty[Index]
+    case Some(p) => Io.readLines(p).map(_.trim.toUpperCase).filter(_.nonEmpty).map(_.getBytes).toSeq
+  }
+
+  if (ExistingIndices.exists(_.length != length)) invalid(s"Existing indices are not all of length ${length}")
+  if (ExistingIndices.exists(idx => idx.exists(b => !Bases.contains(b)))) invalid("One or more existing indices contains a non-[ACGT] character.")
```
nh13: Do we check if the existing indices satisfy the various criteria (mismatches, temperature)?  I'll find out as I go, but I am curious now.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+
+  /** The set of KMERs from the avoid sequences, that should not be selected. */
+  private val AvoidKmers = avoidSequence.flatMap { seq =>
```
nh13: Why are these upper case variables in a `class`.  I would be fine if they were in an `object`.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+    (0 to seq.length - length).flatMap { i =>
+      val sub = seq.substring(i, i + length).toUpperCase
+      Seq(sub, SequenceUtil.reverseComplement(sub))
+    }
+  }.toSet
+
+  /* An Option[DnaFoldPredictor] to use to predict secondary structure. */
+  private val dnaFoldPredictor = viennaRnaDir.map(dir => new DnaFoldPredictor(dir.toFile, temperature))
+  private[util] val AdapterRegex = Pattern.compile("^[ACGT]*N+[ACGT]*$")
```
nh13: Upper case again!

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+  adapters.map(_.toUpperCase).filterNot(AdapterRegex.matcher(_).matches()) match {
```
nh13: Why not use `unapply` on a pattern?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+    case Seq() => Unit
+    case xs    => invalid(s"Adapters must be all [ACGT] with a single contiguous block of Ns where the index goes: ${xs}")
+  }
+
+  /** The main method! */
+  override def execute(): Unit = {
```
nh13: is override necessary?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+    val picks = pickIndices
+    logger.info(s"Picked ${picks.size} indices.")
+    writeOutput(picks, output)
+  }
+
+  /** Picks a set of indices. */
+  private[util] def pickIndices: IndexSet = {
```
nh13: No side effects? Nice.

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
tfenne: I don't know .. but it seems fast enough for now 😉

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
tfenne: Then you've written a non-terminating loop and it doesn't really matter which way the condition is written.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+      // when making many attempts to find a suitable next index, the time spent structure checking
+      // all the indices we're going to discard for too-few-edits anyway explodes.
+      if ( !picks.contains(string) &&
+           (allowReverses || !picks.contains(string.reverse)) &&
+           (allowReverseComplements || !picks.contains(SequenceUtil.reverseComplement(string))) &&
+           picks.forall(p => mismatches(index, p) >= minEdits) &&
+           worstDeltaG(index).forall(_ >= minDeltaG)) {
+        pick = Some(index)
+      }
+      attempt += 1
```
nh13: use a `forloop` here ;)

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    pick
+  }
+
+  /** Generates the next random index which is acceptable as a standalone index, prior to any secondary structure checking. */
+  @tailrec private def randomAcceptableIndex(length: Int): Index = {
+    val index  = randomIndex(length)
+    val string = new String(index)
+    val gc = SequenceUtil.calculateGc(index)
```
nh13: alignment

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
tfenne: Good catch - seems like it should be on type parameter.

src/test/scala/com/fulcrumgenomics/ForLoopTesting.scala
tfenne: It got caught in a `commit -a`.  I've deleted it now.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+
+    if (gc >= minGc && gc <= maxGc &&
+        homopolymerLengthOk(index, maxHomopolymer) &&
+        !AvoidKmers.contains(string) &&
+        (allowPalindromes || !isPalindrome(index))
+    ) {
+      index
+    }
+    else {
+      randomAcceptableIndex(length)
```
nh13: any chance this could have infinite recursion?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: Yup.  Did you read my comment(s) before starting?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+    // Check for a homopolymer starting at each base up to longest+1 from the end of the index
+    forloop (0)(ok && _ < end)(_ + 1) { i =>
+      val base = index(i)
+      var j = i+1
+      var inHomoPolymer = true
+      forloop (i+1)(inHomoPolymer && _ < i+longest+1)(_ + 1) { j =>
+        if (base != index(j)) inHomoPolymer = false
+      }
+
+      ok = !inHomoPolymer
```
nh13: If the longest hp is `5` and you find a hp of `4`, you can skip over the next `4` bases in the loop.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: Global search/replace done.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: That was weird.  I want to blame that one on an IntelliJ reformat.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+    case class AnnotatedIndex(index: Index, existing: Boolean, distances: NumericCounter[Int] = new NumericCounter[Int])
+    val annotated = indices.map(index => AnnotatedIndex(index, existing=ExistingIndices.contains(index)))
+      .toList.sortBy(a => (!a.existing, new String(a.index)))
+
+    annotated.tails.foreach {
+      case x :: ys => ys.foreach { y =>
+        val n = mismatches(x.index, y.index)
+        x.distances.count(n)
+        y.distances.count(n)
+      }
```
nh13: clever

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+      case _ => Unit
+    }
+
+    val Sep = "\t"
+    val lines = ListBuffer[String]()
+    val dfmt = new DecimalFormat("0.000")
+
+    lines += Seq("barcode", "source", "min_mismatches", "barcodes_at_min_mismatches", "gc", "min_deltag").mkString(Sep)
```
nh13: How about a `Metric` class?
```
/**
* @param barcode <desc>
* ...
*/
class BarcodeMetric(
  val barcode: String,
  val source: ...
} extends Metric
*/

File: src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala [hunk 27]
```diff
@@ -0,0 +1,291 @@
...
+    lines ++= annotated.map(ann => Seq(
+      new String(ann.index),
+      if (ann.existing) "existing" else "novel",
+      ann.distances.headOption.map { case (distance, count) => distance }.getOrElse(ann.index.length),
+      ann.distances.headOption.map { case (distance, count) => count    }.getOrElse(0),
+      dfmt.format(SequenceUtil.calculateGc(ann.index)),
+      worstDeltaG(ann.index).getOrElse("n/a")
+    ).mkString(Sep))
```
nh13: You can then use `Metric.write` here.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: I'm not terribly keen on this.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: Yup.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -0,0 +1,291 @@
...
+
+    Io.writeLines(output, lines)
+  }
```
nh13: If you use a `Metric` class, other folks can read the output of this tool easily!

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
nh13: What's the output format (hopefully a metrics class)?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: We don't.  I figure the use case if you've already invested in some, or been given some, so those are ones you're stuck with.

nh13:
@tfenne back to you

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: I named them that way because they behave mostly like constants once initialized.  I can change them if you really object.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: I think this is probably more concise if I just want to know "do they all match" and don't need to pull out any groupings.  Am I missing something?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: It's not necessary, but I think it is good form, since it'll generate a compiler error if you use it and then mis-spell a method name, or change the name in the parent and miss during refactoring.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: I think that's possible, yes, if you fed in pathalogical parameters.  But it would be super unlikely.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
nh13: Your first comment was `No description provided.` and I went straight to the file view from there, so I missed the last ones.

nh13:
@tfenne I think keeping both tools are useful until we are convinced that we can use one and not the other.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
nh13: I would change them to lower case for style purposes.  They are vals so they are constants.  Either that or move them to an object.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: It's true, but then I think I'd have to track homopolyerLength through the second loop.  I think I prefer the simpler method and marginally slower runtime.

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: Ooh, I didn't even think of that.

tfenne:
@nh13 I've fixed everything I agreed with.  Want to take a final look?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -64,31 +64,62 @@ private class IndexSet extends Iterable[Index] {
   def contains(index: String): Boolean = this.set.contains(index)
 }
 
+/** A Metric class for information about each picked index. */
+case class IndexMetric(index: String,
+                       source: String,
+                       min_mismatches:Int,
```
nh13: space after `:`

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -64,31 +64,62 @@ private class IndexSet extends Iterable[Index] {
...
+                       indices_at_min_mismatches: Int,
+                       gc: Double,
+                       min_delta_g: Option[Double]) extends Metric
```
nh13: I am not sure if reading and writing a metric with an option parameter works.  It's probably a good idea to add a test to the metric class?

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -272,20 +301,15 @@ class PickLongIndices
...
-      dfmt.format(SequenceUtil.calculateGc(ann.index)),
-      worstDeltaG(ann.index).getOrElse("n/a")
-    ).mkString(Sep))
-
-    Io.writeLines(output, lines)
+    val metrics = annotated.map(ann => IndexMetric(
+      index                     = new String(ann.index),
+      source                    = if (ann.existing) "existing" else "novel",
+      min_mismatches            = ann.distances.headOption.map { case (distance, count) => distance }.getOrElse(ann.index.length),
+      indices_at_min_mismatches =  ann.distances.headOption.map { case (distance, count) => count.toInt }.getOrElse(0),
```
nh13: space after `=`

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -75,7 +75,7 @@ class FgBioDef extends CommonsDef {
     * @param f a function called on all values
     * @tparam T the type of the index
     */
-  @inline @specialized def forloop[T](from: T)(check: T => Boolean)(next: T => T)(f: T => Unit): Unit = {
+  @inline def forloop[@specialized T](from: T)(check: T => Boolean)(next: T => T)(f: T => Unit): Unit = {
```
nh13: Do you have to name the primitive types?
`@specialied(Byte, Char ,,,) T`?

nh13:
@tfenne a few more minor comments.

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
tfenne: Only if you want it specialized for a sub-set of primitive types.  As is it will cause it to be specialized for all primitives.

PR #100, opened 2016-09-02
Adding fragment read support to ExtractUmisFromBam.

nh13:


src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -87,7 +87,8 @@ class ExtractUmisFromBam
   Io.assertCanWriteFile(output)
 
   val (rs1, rs2) = readStructure match {
-    case Seq(readStructure1, readStructure2) => (ReadStructure(readStructure1), ReadStructure(readStructure2))
+    case Seq(readStructure1, readStructure2) => (ReadStructure(readStructure1), Some(ReadStructure(readStructure2)))
+    case Seq(readStructure1) => (ReadStructure(readStructure1), None)
     case _ => throw new ValidationException("Two read structures must be given (one per end of the template).")
```
tfenne: Error message is no longer accurate - "Exactly 1 or 2 read structures..."

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -148,15 +160,40 @@ class ExtractUmisFromBam
...
+        // now do some work
+        val bases1 = ExtractUmisFromBam.annotateRecord(record=r1, readStructure=rs1, molecularBarcodeTags=molecularBarcodeTags1, clippingAttribute=clippingAttribute)
+        if (annotateReadNames) {
+          // Developer Note: the delimiter must be an ascii character less than what is usually in the read names.  For
+          // example "|" doesn't work.  I am not completely sure why.
+          r1.setReadName(r1.getReadName + "+" + bases1)
+        }
+
+        // If we have duplicate tags, then concatenate them in the same order across the read
+        val tagAndValues = molecularBarcodeTags1.map { tag => (tag, r1.getStringAttribute(tag)) }
```
tfenne: Isn't this whole next ~5 lines unnecessary?

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
```diff
@@ -159,6 +184,24 @@ class ExtractUmisFromBamTest extends UnitSpec with OptionValues {
     }
   }
 
+  it should "accept duplicate molecular barcode SAM tags and concatenate the values for fragment reads" in {
+    val output  = newBam
+    val builder = new SamRecordSetBuilder(readLength=100)
+    builder.addFrag(name="Frag1", start=1, unmapped=true).map { rec => rec.setReadString("A" * 100); rec }
+    builder.addFrag(name="Frag2", start=1, unmapped=true).map { rec => rec.setReadString("C" * 100); rec }
+
+    new ExtractUmisFromBam(input=builder.toTempFile(), output=output, readStructure=Seq("10M80T10M"), molecularBarcodeTags=Seq("A1", "A1")).execute()
```
tfenne: I see what you're doing here, but how does this differ from just providing `A1` once?  Won't they still get concatenated and placed in a single tag?

tfenne:
A few minor comments.  Merge when you're happy.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
nh13: Fixed.

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
nh13: You may want to specify ("A1", "B1", "A1") etc.  It's overkill but I am trying to be consistent with the paired end case.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
nh13: If the same tag is given more than once, then no.

PR #101, opened 2016-09-06
takeWhile in BetterBufferedIterator should not read all the elements.

nh13:
Currently, `BetterBufferedIterator.takeWhile` will store all the elements that match the predicate in a `ListBuffer` before returning an iterator.  This causes OOM for large numbers of elements that match the predicate.  This fix will create an iterator such that calling `next` on the iterator returned by `BetterBufferedIterator.takeWhile` is on-demand/lazy/whatever-the-buzz-word-is-today.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/BetterBufferedIterator.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/BetterBufferedIterator.scala
@@ -51,9 +51,11 @@ class BetterBufferedIterator[A](private val iterator: Iterator[A]) extends Buffe
     * Returns an iterator over contiguous elements that match the predicate.
     */
   override def takeWhile(p: (A) => Boolean): Iterator[A] = {
-    val xs = new ListBuffer[A]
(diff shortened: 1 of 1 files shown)


tfenne:
@nh13 this seems reasonable to me .. we just need to be careful in usage.  I.e. the following would work with the previous implementation, the latter not:
```scala
val x = iterator.next
val ys = iterator.takeWhile(y => y.foo == x.foo)
val zs = iterator.takeWhile(z => z.foo == x.foo + "/2")

if (zs.length > ys.length) ...
```
since evaluation of `zs`' iterator will not happen prior to `ys`.  Not a dealbreaker.

nh13:
@tfenne From `Iterator.scala`:
> It is of particular importance to note that, unless stated otherwise, ''one should never use an iterator after calling a method on it''. The two most important exceptions are also the sole abstract methods: `next` and `hasNext`.

I can understand what you are trying to do above but I think having such behavior in a class that make it completely obvious is a better idea.

PR #104, opened 2016-09-08
Support optional values in Metric.

### LINKED ISSUE #102: Metric does not support optional values
The Metric class does not support optional values via `Option[Type]`.  This would be nice as there are definitely cases where some values are optional, and not necessarily present.

nh13:
Fixes #102 but depends on https://github.com/fulcrumgenomics/dagr/pull/197

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/Metric.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/Metric.scala
@@ -28,6 +28,7 @@ package com.fulcrumgenomics.util
 import java.io.Writer
 import java.nio.file.Path
 
+import dagr.commons.CommonsDef.unreachable
 import dagr.commons.reflect.{ReflectionUtil, ReflectiveBuilder}
 import htsjdk.samtools.util.FormatUtil
 
@@ -35,7 +36,8 @@ import scala.reflect.runtime.{universe => ru}
 import scala.util.{Failure, Success}
 
 object Metric {
-  val Delimiter = "\t"
+  val Delimiter: Char = '\t'
+  val DelimiterAsString: String = s"$Delimiter"
 
   /** Get the names of the arguments in the order they were defined for the type [T]. */
   def names[T <: Metric](implicit tt: ru.TypeTag[T]): Seq[String] = {
@@ -63,24 +65,23 @@ object Metric {
       }
     }
     // build it.  NB: if arguments are missing values, then an exception will be thrown here
-    reflectiveBuilder.build()
+    // Also, we don't use the default "build()" method since if a collection or option is empty, it will be treated as
+    // missing.
+    reflectiveBuilder.build(reflectiveBuilder.argumentLookup.ordered.map(arg => arg.value getOrElse unreachable(s"Arguments not set: ${arg.name}")))
   }
 
   /** Reads metrics from a set of lines.  The first line should be the header with the field names.  Each subsequent
     * line should be a single metric. */
   def read[T <: Metric](lines: Iterator[String], source: Option[String] = None)(implicit tt: ru.TypeTag[T]): Seq[T] = {
     if (lines.isEmpty) throw new IllegalArgumentException(s"No header found in metrics" + source.map(" in source: " + _).getOrElse(""))
-    val names = lines.next().split(Delimiter)
-    if (lines.isEmpty) Seq.empty
-    else {
-      lines.zipWithIndex.map { case (line, lineNumber) =>
-        val values = line.split(Delimiter)
-        if (names.length != values.length) {
-          throw new IllegalArgumentException(s"Line ${lineNumber+1} did not have the same length (${values.length}) as the header (${names.length}")
-        }
-        build[T](names.zip(values))
-      }.toSeq
-    }
+    val parser = new DelimitedDataParser(lines=lines, delimiter=Delimiter, ignoreBlankLines=false, trimFields=true)
+    val names  = parser.headers
+    parser.zipWithIndex.map { case (row, lineNumber) =>
+      val values = names.indices.map { i => row[String](i) }.map { value =>
+        if (value.nonEmpty) value else ReflectionUtil.SpecialEmptyOrNoneToken
+      }
+      build[T](names.zip(values))
+    }.toSeq
   }
 
   /** Reads metrics from the given path.  The first line should be the header with the field names.  Each subsequent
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -70,17 +70,17 @@ object Metric {
...
-        if (names.length != values.length) {
-          throw new IllegalArgumentException(s"Line ${lineNumber+1} did not have the same length (${values.length}) as the header (${names.length}")
-        }
-        build[T](names.zip(values))
-      }.toSeq
-    }
+    val parser = new DelimitedDataParser(lines=lines, delimiter=Delimiter, ignoreBlankLines=false, trimFields=true)
+    val names  = parser.headers
+    parser.zipWithIndex.map { case (row, lineNumber) =>
+      val values = names.indices.map { i => row[String](i) }.map { value =>
```
nh13: @tfenne I am torn.  I like forcing the API to be type-safe in `DelimitedDataParser` but sometimes I just want the fields.

tfenne: I think it's probably totally reasonable to add some methods to DelimitedDataParser to access the `String` values more efficiently.  I would just like to do it in a way that prevents the caller from modifying the underlying values.  Two possibilities:

1) Provide a `Row.field(name: String): String` which does it more efficiently.  I'm struggling to come up with a good name for the method though.
2) Provide a `toMap: Map[String,String]` that just puts all the non-None values into a map by field name and returns them?

tfenne: I'm also confused that you're using the row.apply() here which doesn't return an option.  I think this is really a bug in DelimitedDataParser as I would expect that to throw an exception if there's no actual value for a field.  I think it would be better to use `row[String].get(i).getOrElse(ReflectionUtil.SpecialEmptyOrNoneToken)`

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -70,17 +70,17 @@ object Metric {
...
+        if (value.nonEmpty) value else ReflectionUtil.SpecialEmptyOrNoneToken
+      }
+      if (names.length != values.length) {
```
tfenne: I think this check is unnecessary as DelimitedDataParser enforces this for you now.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -91,10 +91,10 @@ object Metric {
     * line is a single metric.
     */
   def write[T <: Metric](metrics: Seq[T], writer: Writer)(implicit tt: ru.TypeTag[T]): Unit = {
-    writer.write(names[T].mkString(Delimiter))
+    writer.write(names[T].mkString(s"$Delimiter"))
```
tfenne: Maybe it's worth having `Delimiter` and `DelimiterAsString` as constants?

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -117,7 +117,13 @@ object Metric {
   */
 trait Metric extends Product {
   private lazy val reflectiveBuilder = new ReflectiveBuilder(this.getClass)
-  private val formatter = new FormatUtil
+  private val formatter = new FormatUtil {
+    override def format(value: Any): String = value match {
+      case option: Option[_] => this.format(option)
+      case _                 => super.format(value)
+    }
+    def format(option: Option[_]): String = format(option.getOrElse(""))
```
tfenne: Wouldn't it be clearer just to have three cases?
```scala
override def format(value: Any): String = value match {
  case None    => ""
  case Some(x) => super.format(x)
  case y       => super.format(y)
}
```

src/test/scala/com/fulcrumgenomics/util/MetricTest.scala
```diff
@@ -31,6 +31,8 @@ import com.fulcrumgenomics.testing.UnitSpec
 
 private case class TestMetric(foo: String, bar: Int, car: String = "default") extends Metric
 
+private case class TestMetricWithOption(foo: String, bar: Int, option: Option[String]) extends Metric
```
tfenne: Could you add an `Option[Int]` or other non-String type as well please?  Would be good to test that it works for things that need type converting too.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: @tfenne an empty value is only invalid if the type does not have a `String` constructor that handles an empty `String`.  Also, I am exploiting the apply method with type `String` since any value can be represented as string.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: I had a sneaky suspicion you'd say this, and I was right. 👏

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: Fixed.

src/test/scala/com/fulcrumgenomics/util/MetricTest.scala
nh13: Fixed.

nh13:
@tfenne thanks for the comments, back to you for a final look, and then I will await fulcrumgenomics/dagr#197

src/main/scala/com/fulcrumgenomics/util/Metric.scala
tfenne: @nh13 The point I was making is that I think you've highlighted a bug in `DelimitedDataParser` that I want to fix, but you're relying on what I see as _invalid_ behaviour I want to correct.  I think probably `Row[T].apply(x)` should be implemented as `Row.get[T](x).get` to ensure consistency.  Let's take a look at these two things together.

nh13: It's part of the API, you can't change it now!

tfenne:
@nh13 I rebased this on top of master, but the tests for the optional fields are failing :(

PR #107, opened 2016-09-08
Sam record clipping code and a CLP for trimming primers from aligned reads.

tfenne:


src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+    * Adds clipping of _at least_ numberOfBasesToClip to the start (left hand end)
+    * of an alignment.
+    *
+    * @param rec the record to be clipped
+    * @param numberOfBasesToClip the number of additional bases to clip beyond any clipping
+    *                            already applied to the read
+    */
+  def clipStartOfAlignment(rec: SAMRecord, numberOfBasesToClip: Int, mode: ClippingMode) : Unit = {
```
nh13: It would be nice to support hard-clipping of unmapped reads.

nh13: For hard-clipping and soft-cliping-with-mask, it may be useful to return the bases and qualities that were removed or masked respectively.

tfenne:
@nh13 I think this one is ready for a real review whenever you're ready.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.bam
+
+import java.util
+
+import com.fulcrumgenomics.FgBioDef._
+import htsjdk.samtools.{Cigar, CigarElement, SAMRecord, SAMUtils, CigarOperator => Op}
```
nh13: `CigarOp` instead of `Op`?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+
+import scala.collection.JavaConversions.iterableAsScalaIterable
+import scala.collection.mutable.ListBuffer
+
+/**
+  * Provides a suite of methods for clipping (soft- and hard-) bases from the beginnings
+  * and ends of reads in various ways.  Cigar strings, bases, qualities and alignment
```
nh13: `from both the beginning and end of reads`?  beginnings and ends of reads reads funny to me </2cents>

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+    * of an alignment. If clipping already exists at the start of the read it is
+    * preserved, and numberOfBasesToClip more clipping is added.
+    *
+    * If is unmapped or the numberOfBasesToClip is < 1 nothing is done.
+    *
+    * If the read has fewer clippable bases than requested clipping, the read is
+    * unmapped.
+    *
+    * If hard-clipping is requested and the read has existing soft-clipping at the start
+    * it is converted to hard-clipping before adding further clipping.
```
nh13: This is surprising, since if I request 5bp to be hard-clipped at the start for `10S90M` I would expect `5H5S90M` not `15H85M`.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+  object ClippingMode extends Enumeration { val Soft, Hard, SoftWithMask = Value }
+  type ClippingMode = ClippingMode.Value
+
+  private val NoCallBase = 'N'.toByte
+  private val NoCallQual = 2.toByte
+
+  /**
+    * Adds clipping of _at least_ numberOfBasesToClip to the start (left hand end)
+    * of an alignment. If clipping already exists at the start of the read it is
+    * preserved, and numberOfBasesToClip more clipping is added.
```
nh13: I am surprised that we add more clipping rather than it being the desired # of bases to clip.  My thought was using these methods to ensure that all alignments had their first `N` bases clipped.

nh13: Coming back to this, and my comment below, I suppose the methods are named such that they apply clipping to the _alignment_ and not the _read_.  Is this the desired behavior?  Namely ask to clip an additional 5bp at the start of each alignment, regardless of how many bases aligned?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+    else if (numberOfClippableBases(rec) <= numberOfBasesToClip) {
+      SAMUtils.makeReadUnmapped(rec)
+    }
+    else {
+      val oldElems = rec.getCigar.toSeq
+      val oldBases = rec.getReadBases
+      val oldQuals = rec.getBaseQualities
+      val (newElems, readBasesClipped, refBasesClipped, bases, quals) = clip(oldElems, numberOfBasesToClip, oldBases, oldQuals, mode)
+      rec.setAlignmentStart(rec.getAlignmentStart + refBasesClipped)
+      rec.setCigar(new Cigar(util.Arrays.asList(newElems:_*)))
```
nh13: Or you can use `toList` and an implicit ;)

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+    * @param bases the array of bases for the read to be clipped
+    * @param quals the array of quality scores for the read to be clipped
+    * @param mode  the clipping mode
+    * @return a tuple of (newCigarElems, readBasesClipped, refBasesClipped, newBases, newQuals)
+    */
+  private def clip(originalElems: Seq[CigarElement],
+                   numberOfBasesToClip: Int,
+                   bases: Array[Byte],
+                   quals: Array[Byte],
+                   mode: ClippingMode): (Seq[CigarElement], Int, Int, Array[Byte], Array[Byte]) = {
```
nh13: It may be clearer to return a case class.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+    * @param numberOfBasesToClip the number of additional bases to clip beyond any clipping
+    *                            already applied to the read
+    * @param mode the kind of clipping to be performed
+    */
+  def clipStartOfAlignment(rec: SAMRecord, numberOfBasesToClip: Int, mode: ClippingMode) : Unit = {
+    if (rec.getReadUnmappedFlag || numberOfBasesToClip < 1) {
+      Unit
+    }
+    else if (numberOfClippableBases(rec) <= numberOfBasesToClip) {
+      SAMUtils.makeReadUnmapped(rec)
```
nh13: tests for making the read unmapped?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+    }
+    else {
+      val oldElems = rec.getCigar.toSeq
+      val oldBases = rec.getReadBases
+      val oldQuals = rec.getBaseQualities
+      val (newElems, readBasesClipped, refBasesClipped, bases, quals) = clip(oldElems, numberOfBasesToClip, oldBases, oldQuals, mode)
+      rec.setAlignmentStart(rec.getAlignmentStart + refBasesClipped)
+      rec.setCigar(new Cigar(util.Arrays.asList(newElems:_*)))
+      rec.setReadBases(bases)
+      rec.setBaseQualities(quals)
```
nh13: It probably makes sense to invalidate some of the tags that now could be wrong (ex. `MD`, `NM`, and `UQ` come to mind).

nh13: Here and elsewhere of course.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+  /**
+    * Attempts to clip an additional numberOfBasesToClip from the 5' end of the read. For
+    * details see [[clipStartOfAlignment()]] and [[clipEndOfAlignment()]].
+    *
+    * @param rec the record to be clipped
+    * @param numberOfBasesToClip the number of additional bases to be clipped
+    * @param mode the clipping mode
+    */
+  def clip5PrimeEndOfAlignment(rec: SAMRecord, numberOfBasesToClip: Int, mode: ClippingMode) = {
+    if (!rec.getReadUnmappedFlag && rec.getReadNegativeStrandFlag) clipEndOfAlignment(rec, numberOfBasesToClip, mode)
```
nh13: `!rec.getReadUnmappedFlag` not necessary.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+  /**
+    * Attempts to clip an additional numberOfBasesToClip from the 5' end of the read. For
+    * details see [[clipStartOfAlignment()]] and [[clipEndOfAlignment()]].
+    *
+    * @param rec the record to be clipped
+    * @param numberOfBasesToClip the number of additional bases to be clipped
+    * @param mode the clipping mode
+    */
+  def clip3PrimeEndOfAlignment(rec: SAMRecord, numberOfBasesToClip: Int, mode: ClippingMode) = {
+    if (!rec.getReadUnmappedFlag && rec.getReadNegativeStrandFlag) clipStartOfAlignment(rec, numberOfBasesToClip, mode)
```
nh13: ditto

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
nh13: And way down there are the methods I am talking about, doh! Perhaps in the documentation of the object it is describe clipping _reads_ versus _alignments_?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+    * of the read, _including_ any existing soft and hard clipping.  Calculates any
+    * additional clipping and delegates to [[clipStartOfAlignment()]].
+    *
+    * @param rec the record to be clipped
+    * @param clipLength the total amount of clipping desired, including any existing clipping
+    * @param mode the clipping mode
+    */
+  def clipStartOfRead(rec: SAMRecord, clipLength: Int, mode: ClippingMode) = {
+    val existingClipping = rec.getCigar.getCigarElements.takeWhile(_.getOperator.isClipping).map(_.getLength).sum
+    if (clipLength > existingClipping) clipStartOfAlignment(rec, clipLength - existingClipping, mode)
```
nh13: What if we want to hard-clip bases that are currently soft-clipped?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+    * of the read, _including_ any existing soft and hard clipping.  Calculates any
+    * additional clipping and delegates to [[clipStartOfAlignment()]].
+    *
+    * @param rec the record to be clipped
+    * @param clipLength the total amount of clipping desired, including any existing clipping
+    * @param mode the clipping mode
+    */
+  def clipEndOfRead(rec: SAMRecord, clipLength: Int, mode: ClippingMode) = {
+    val existingClipping = rec.getCigar.getCigarElements.toSeq.reverse.takeWhile(_.getOperator.isClipping).map(_.getLength).sum
+    if (clipLength > existingClipping) clipEndOfAlignment(rec, clipLength - existingClipping, mode)
```
nh13: Ditto.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+    val existingHardClip = originalElems.takeWhile(_.getOperator == Op.HARD_CLIP).map(_.getLength).sum
+    val existingSoftClip = originalElems.dropWhile(_.getOperator == Op.HARD_CLIP).takeWhile(_.getOperator == Op.SOFT_CLIP).map(_.getLength).sum
+    val postClipElems = originalElems.dropWhile(e => e.getOperator == Op.HARD_CLIP || e.getOperator == Op.SOFT_CLIP).iterator.bufferBetter
+    var readBasesClipped = 0
+    var refBasesClipped  = 0
+    val newElems = ListBuffer[CigarElement]()
+
+    // At the end of the loop newElems is either:
+    //   a) Empty
+    //   b) Contains a single element which is the remainder of an element that had to be split
```
nh13: Is this true? A cigar of `====` or a cigar of `10M1D10M` (with 15 clipping) would have `newElems.length > 1` right?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -0,0 +1,303 @@
...
+    while (readBasesClipped < numberOfBasesToClip ||
+      (readBasesClipped == numberOfBasesToClip && postClipElems.hasNext && postClipElems.head.getOperator == Op.DELETION)) {
```
nh13: A comment about keeping deletions would be nice for our future selves.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -0,0 +1,195 @@
...
+    override def getStart: Int = leftStart
+    override def getEnd: Int = rightEnd
+  }
+
+  override def execute(): Unit = {
+    val in = SamReaderFactory.make().open(input)
+    val outSortOrder = sortOrder.getOrElse(in.getFileHeader.getSortOrder)
+    val outHeader = in.getFileHeader.clone()
+    outHeader.setSortOrder(outSortOrder)
+    val out = new SAMFileWriterFactory().setCreateIndex(true).makeWriter(outHeader, outSortOrder == SortOrder.queryname, output.toFile, null)
```
nh13: Why do we assume presorted when it's queryname?  Isn't the data mapped?  Is this meant to be used after alignment but before sorting?

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -0,0 +1,195 @@
...
+  """
+    |Trims primers from reads post-alignment.  Takes in a BAM file of aligned reads
+    |and a tab-delimited file with five columns (chrom, left_start, left_stop, right_start,
+    |and right_stop) which provide the 1-based inclusive start and end positions of the
+    |primers for each amplicon.
+    |
+    |Paired end reads that map to a given amplicon position are trimmed so that the
+    |alignment no-longer includes the primer sequences.
+    |
+    |All other aligned reads have the maximum primer length trimmed!
```
nh13: Probably a good idea to state that sorting will occur if the input is not `queryname` sorted.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -0,0 +1,195 @@
...
+          // FR mapped pairs get the full treatment
+          if (!r1.getReadUnmappedFlag && !r2.getReadUnmappedFlag &&
+            r1.getReferenceName == r2.getReferenceName &&
+            SamPairUtil.getPairOrientation(r1) == PairOrientation.FR) {
+
+            val (left, right) = if (r1.getReadNegativeStrandFlag) (r2, r1) else (r1, r2)
+            val (start, end ) = (left.getUnclippedStart, right.getUnclippedEnd)
+            val insert = new Interval(left.getContig, start, end)
+            val amp = detector.getOverlaps(insert).find(amp => abs(amp.leftStart - start) <= slop && abs(amp.rightEnd - end) <= slop)
+            amp match {
```
nh13: Could just match directly.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -0,0 +1,195 @@
...
+  private def nextTemplate(iterator: BufferedIterator[SAMRecord]): Seq[SAMRecord] = {
+    val first    = iterator.next()
+    val template = first.getReadName
+    first :: iterator.takeWhile(_.getReadName == template).toList
+  }
+
+  /** Creates an overlap detector for all the amplicons from the input file. */
+  private def loadPrimerFile(path: FilePath): OverlapDetector[Amplicon] = {
+    val parser = DelimitedDataParser(path, '\t')
+    val detector = new OverlapDetector[Amplicon](0,0)
```
nh13: could add the `slop` to the left-hand side: `val detector = new OverlapDetector[Amplicon](-slop,0)`

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -0,0 +1,195 @@
...
+            val (left, right) = if (r1.getReadNegativeStrandFlag) (r2, r1) else (r1, r2)
+            val (start, end ) = (left.getUnclippedStart, right.getUnclippedEnd)
+            val insert = new Interval(left.getContig, start, end)
+            val amp = detector.getOverlaps(insert).find(amp => abs(amp.leftStart - start) <= slop && abs(amp.rightEnd - end) <= slop)
+            amp match {
+              case Some(amplicon) =>
+                val leftClip  = amplicon.leftPrimerLength
+                val rightClip = amplicon.rightPrimerLength
+                reads.foreach { rec =>
+                  val toClip = if (rec.getFirstOfPairFlag == left.getFirstOfPairFlag) leftClip else rightClip
```
nh13: My apologies if I am confused, but if the first end of a read is on the reverse strand, then the second end is on the forward strand, and so the second end's alignment start is _before_ the first end's alignment end, so we should be using the strand to condition on `leftClip` vs. `rightClip`.  I need a whiteboard to verify.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -0,0 +1,195 @@
...
+        case (Some(r1), Some(r2)) =>
+          // FR mapped pairs get the full treatment
+          if (!r1.getReadUnmappedFlag && !r2.getReadUnmappedFlag &&
+            r1.getReferenceName == r2.getReferenceName &&
+            SamPairUtil.getPairOrientation(r1) == PairOrientation.FR) {
+
+            val (left, right) = if (r1.getReadNegativeStrandFlag) (r2, r1) else (r1, r2)
+            val (start, end ) = (left.getUnclippedStart, right.getUnclippedEnd)
+            val insert = new Interval(left.getContig, start, end)
+            val amp = detector.getOverlaps(insert).find(amp => abs(amp.leftStart - start) <= slop && abs(amp.rightEnd - end) <= slop)
```
nh13: find the one with the smallest `slop`? overkill?

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -0,0 +1,195 @@
...
+    */
+  private def clipFullyOverlappedFrReads(r1: SAMRecord, r2: SAMRecord): Unit = {
+    val (plus, minus) = if (r1.getReadNegativeStrandFlag) (r2,r1) else (r1, r2)
+
+    if (plus.getAlignmentStart < minus.getAlignmentEnd) {
+      val plusTrim  = plus.getAlignmentEnd   - minus.getAlignmentEnd
+      val minusTrim = plus.getAlignmentStart - minus.getAlignmentStart
+
+      if (plusTrim  > 0) SamRecordClipper.clip3PrimeEndOfAlignment(plus, plusTrim, mode)
+      if (minusTrim > 0) SamRecordClipper.clip3PrimeEndOfAlignment(minus, minusTrim, mode)
```
nh13: invalidate `MD`, `NM`, and `UQ` tags?

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -0,0 +1,195 @@
...
+    first :: iterator.takeWhile(_.getReadName == template).toList
+  }
+
+  /** Creates an overlap detector for all the amplicons from the input file. */
+  private def loadPrimerFile(path: FilePath): OverlapDetector[Amplicon] = {
+    val parser = DelimitedDataParser(path, '\t')
+    val detector = new OverlapDetector[Amplicon](0,0)
+    parser.foreach { row =>
+      val amp = Amplicon(
+        chrom = row[String]("chrom"),
```
nh13: align it with the rest below 📟

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
```diff
@@ -0,0 +1,374 @@
...
+
+    val rec3 = r(10, "50M", strand=Plus)
+    SamRecordClipper.clip3PrimeEndOfRead(rec3, 10, ClippingMode.Soft)
+    rec3.getCigarString shouldBe "40M10S"
+
+    val rec4 = r(10, "40M10S", strand=Plus)
+    SamRecordClipper.clip3PrimeEndOfRead(rec4, 10, ClippingMode.Soft)
+    rec4.getCigarString shouldBe "40M10S"
+  }
+}
```
nh13: SO many tests 👍

nh13:
@tfenne I think it looks really good, my two major thoughts were (1) around adding versus overwriting clipping (ex turn a subset of the leading soft-clipped bases into hard-clipped bases), and (2) verifying primer trimming was on a read 1 versus read 2 basis and not orientation.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: I think you might have skipped over the object level doc, because it includes:
```
* Note that there are several "flavours" of method:
*   - [snip]
*   - Ones that attempt to clip an additional N bases beyond any clipping already
*     provided (clip[Start|End|5PrimeEnd|3PrimeEnd]OfAlignment) and ones that
*     attempt to make it so that the read has N bases clipped, including any existing
*     clipping (clip[Start|End|5PrimeEnd|3PrimeEnd]OfRead)
```

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: I'm assuming this is less surprising once you've read through all the other methods?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: I'm happy being explicit :p

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: It's almost always used in a context that makes it super clear, so unless you hate it I'm going to stick with `Op`.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: I think I'd rather leave that outside of this code - e.g. TrimPrimers should probably do that since it's writing a BAM file and you don't want an invalid BAM.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: Huh, ok.  I'm so used to SAMRecord pitching a fit if I try to retrieve a flag that it feels is in appropriate given the state of the read, I'm overly defensive.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: Yes, it's true.  The loop doesn't append all the other elements into the `newElems` buffer, only the remainder of a single element that has to be split (if splitting is necessary), which I realize is slightly unintuitive, hence the comment!

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: Except that we're purposefull _discarding_ deletions at the boundary.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
nh13: I actually disagree, since now the `SAMRecord` itself is inconsistent.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
tfenne: It's because I'm iterating over the input in queryname order (either because it is, or because I sorted it), so when writing the output I need to re-sort if the desired sort isn't queryname.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
tfenne: Yeah, I thought about that and figured it's overkill until proven otherwise.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
tfenne: It's confusing, but it's doing the right thing.  I'm trying to handle the case where I have either secondary or supplemental records, and I want them all clipped the same as the primary...  So above I figure out which is the `left` and `right` read using position/orientation, and then I go over _all_ reads saying "if you're the same read number as my left read, trim by the `leftClip` amount, else by the `rightClip` amount".  There's an extra level of indirection there.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
tfenne: I could, but I think it's probably clearer to keep it where the matching happens.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: Let's chat about this one.  I sort of agree, but don't really want to take this on right now.

PR #108, opened 2016-09-12
Switch logging to stderr instead of stdout.

nh13:
This precluding piping out of tools.


tfenne:
@nh13 This seems ok, but why not just fix it in dagr so stderr is the default?

nh13:
@tfenne I thought there was some reason it was `stdout` in Dagr.  I guess not, I'll change it there.

PR #109, opened 2016-09-13
Fixing a bug where we did not move to the next reference in ErrorRate…

nh13:
…ByReadPosition.

This affected running the tool when a reference was supplied.


src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -205,6 +205,11 @@ private class ErrorRateByReadPositionWithReferenceCounter
   var referenceSequence = referenceFile.getSequence(referenceFile.getSequenceDictionary.getSequence(0).getSequenceName)
 
   protected def updateErrors(rec: SAMRecord, readIndex: Int): Unit = {
+    // Move to the correct reference.
+    if (rec.getReferenceIndex != referenceSequence.getContigIndex) {
+      referenceSequence = referenceFile.getSequence(referenceFile.getSequenceDictionary.getSequence(rec.getReferenceIndex).getSequenceName)
```
tfenne: It might be easier/cleaner to have a `private val ref: ReferenceSequenceFileWalker` which you just request the appropriate reference from each time.  It's less code and then it will yell if it tries to reload a contig it's already gone past.  Up to you though - this works fine.

PR #106, opened 2016-09-08
Making configuration parallel collections easier.

nh13:

PR #111, opened 2016-09-14
Making configuration parallel collections easier.

tfenne:
@nh13 an alternative to #106.  This version has the following advantages IMHO:
a) Adds the methods directly to all parallelizable collections, so you can call `Seq(1,2,3).parWith(parallelism=8)`, instead of `Seq(1,2,3).par.with...`
b) With some type parameter/inference magic, the return type appropriate subclass (e.g.`ParSeq` instead of `ParIterable`)
c) Has tests ;)


nh13:
Tests pass then 👍

Issue #113, opened 2016-09-19
error in HapCutToVcf


jaredo:
Hello! I am getting the following error when trying to run HapCutToVcf. I can share some test data but it is quite large.

```
$ java -jar /illumina/thirdparty/joconnell/fgbio/target/scala-2.11/fgbio-0.1.2-SNAPSHOT.jar HapCutToVcf -v ../NA12878.1000G.vcf -i na12878.output_haplotype_file  -o test.vcf
[2016/09/19 20:52:24 | FgBioMain | Info] Executing HapCutToVcf from fgbio version 0.1.2-SNAPSHOT as joconnell@ukch-dev-lnt8-jo.local on JRE 1.8.0_40-b26
[2016/09/19 20:52:25 | FgBioMain | Info] HapCutToVcf failed. Elapsed time: 0.07 minutes.
Exception in thread "main" java.lang.NumberFormatException: For input string: "1|0"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
	at scala.collection.immutable.StringLike$class.toInt(StringLike.scala:272)
	at scala.collection.immutable.StringOps.toInt(StringOps.scala:30)
	at com.fulcrumgenomics.vcf.HapCutCall$$anonfun$9.apply(HapCutToVcf.scala:503)
	at com.fulcrumgenomics.vcf.HapCutCall$$anonfun$9.apply(HapCutToVcf.scala:503)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:186)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
	at scala.collection.mutable.ArrayOps$ofRef.map(ArrayOps.scala:186)
	at com.fulcrumgenomics.vcf.HapCutCall.toVariantContext(HapCutToVcf.scala:503)
	at com.fulcrumgenomics.vcf.HapCutAndVcfMergingIterator.next(HapCutToVcf.scala:269)
[...]

PR #114, opened 2016-09-19
Support when the input VCF to HapCut has phased genotypes.

nh13:
Fixes the bug reported in https://github.com/fulcrumgenomics/fgbio/issues/113.


nh13:
@tfenne care to take a brief look?

PR #118, opened 2016-09-23
Added optional re-calculation of NM/UQ/MD to TrimPrimers.

tfenne:


src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -85,71 +91,117 @@ class TrimPrimers
...
+    // Setup the outputs depending on whether or not we have a reference file
+    val (sorter: Option[SortingCollection[SAMRecord]], write: (SAMRecord => Unit), out: SAMFileWriter) = ref match {
+      case Some(path) =>
+        val sorter = buildSorter(SortOrder.coordinate, outHeader)
+        val out = new SAMFileWriterFactory().setCreateIndex(true).makeWriter(outHeader, outSortOrder == SortOrder.coordinate, output.toFile, null)
+        (Some(sorter), sorter.add _, out)
+      case None =>
+        val out = new SAMFileWriterFactory().setCreateIndex(true).makeWriter(outHeader, outSortOrder == SortOrder.queryname, output.toFile, null)
+        (None, out.addAlignment _, out)
+    }
```
nh13: There's a lot of conditional sorting going on, which now I understand reading through the commits, but perhaps a few more comments?

1. We need the reads to be queryname sorted prior to trimming.
2. If the reference is given, then we need to sort by coordinate after trimming to reset MD/NM.
3. If the reference is given, and the output sort order is coordinate (either specified via clp, or unspecified and in the input), then the records will be "presorted" when adding to the output after resetting MD/NM, otherwise we will need the output to sort.
4. If no reference is given, then the output is "presorted" if it is queryname.

So complex!

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -85,71 +91,117 @@ class TrimPrimers
...
-            val mate = if (rec.getFirstOfPairFlag) r2 else r1
-            SamPairUtil.setMateInformationOnSupplementalAlignment(rec, mate, true);
-          }
-        case _ =>
-          // Just trim each read independently
-          reads.foreach(r => SamRecordClipper.clip5PrimeEndOfRead(r, maxPrimerLength, mode))
-      }
+    // If we had a reference and re-sorted above, reset the NM/UQ/MD tags as we push to the final output
+    (sorter, ref) match {
+      case (Some(recs: SortingCollection[SAMRecord]), Some(path)) =>
```
nh13: recs -> sorter

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -85,71 +91,117 @@ class TrimPrimers
...
+        val walker = new ReferenceSequenceFileWalker(path.toFile)
+        val progress = new ProgressLogger(this.logger, "Written")
```
nh13: do we need to call `doneAdding` on the sorter?

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -85,71 +91,117 @@ class TrimPrimers
...
 
-      reads.foreach(out.addAlignment)
+        recs.foreach { rec =>
+          recalculateTags(rec, walker)
+          out.addAlignment(rec)
+          progress.record(rec)
+        }
```
nh13: do we need to call `sorter.close()` to get rid of temporary files?

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -85,71 +91,117 @@ class TrimPrimers
...
+      case _ => Unit
     }
 
     out.close()
   }
 
+  /** Recalculates the NM, UQ and MD tags on aligned records. */
```
nh13: tag names in alphabetical order, though even in Picard it is not: `SetNmMdAndUqTags`.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -85,71 +91,117 @@ class TrimPrimers
...
+        // Pairs without both reads mapped in FR orientation are just maximally clipped
+        else {
+          reads.foreach(r => SamRecordClipper.clip5PrimeEndOfRead(r, maxPrimerLength, mode))
+        }
+
+        SamPairUtil.setMateInfo(r1, r2, true)
+        reads.filter(_.getSupplementaryAlignmentFlag).foreach { rec =>
+          val mate = if (rec.getFirstOfPairFlag) r2 else r1
+          SamPairUtil.setMateInformationOnSupplementalAlignment(rec, mate, true);
+        }
```
nh13: I was going to suggest using `SetMateInfoIterator` but I don't think it handles frags.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
tfenne: Yeah - I wish it were less so, but I really don't want to force an extra sort if not necessary, and that got me to where I am :(    I'll add more comments.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
tfenne: You're the one who yells at me any time I shadow a variable name, and that's exactly what that would do!

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
tfenne: `iterator()` calls `doneAdding` internally.  You only need to call it manually if you want to sort the last chunk and write it out immediately because you're not going to use the iterator for a while.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
tfenne: In theory it should all get cleaned up at exit, but yes probably a good idea to close() when done.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
tfenne: And who's fault is that 😉 .  Fixed.

PR #121, opened 2016-10-17
AsssessPhasing can now only evaluate the given chromosomes.

nh13:
- remove the "-c" to specify an individual chromosome.
- change the "-l" option to "-k" to specify the known set of
  intervals; added to the usage.
- add the "-l" option to specify the chromosomes to evaluate; added to
  the usage.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
+++ b/src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
@@ -58,6 +58,12 @@ import scala.collection.JavaConverters._
       |The input VCFs are assumed to be single sample: the genotype from the first sample is used.
       |
       |Only bi-allelic heterozygous SNPs are considered.
+      |
+      |The input known phased variants can be subsetted using the known interval list, for example to keep only variants
+      |from high-confidence regions.
+      |
+      |If the intervals argument is supplied, only the set of chromosomes specified will be analyzed.  Note that the full
+      |chromosome will be analyzed and start/stop positions will be ignored.
     """,
   group=ClpGroups.VcfOrBcf
 )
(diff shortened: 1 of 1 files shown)


nh13:
@tfenne a simple PR.  The idea is that we may want to evaluate phasing on only the autosomal chromosomes and also ignore any alternate or decoy sequences (I am looking at you hs38DH).

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -56,17 +56,23 @@ import scala.collection.mutable.ListBuffer
       |The input VCFs are assumed to be single sample: the genotype from the first sample is used.
       |
       |Only bi-allelic heterozygous SNPs are considered.
+      |
+      |The input known phased variants can be subsetted using the known interval list, for example to keep only variants
+      |from high-confidence regions.
+      |
+      |If an interval list is given, only the set of chromosomes specified will be analyzed.  Note that the full
```
tfenne: "If the intervals argument is supplied" would be clearer since the prior line is discussing an interval list for the known variants.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -56,17 +56,23 @@ import scala.collection.mutable.ListBuffer
...
 )
 class AssessPhasing
 ( @arg(flag="c", doc="The VCF with called phased variants.") val calledVcf: PathToVcf,
   @arg(flag="t", doc="The VCF with known phased variants.") val truthVcf: PathToVcf,
   @arg(flag="o", doc="The output prefix for all output files.") val output: PathPrefix,
-  @arg(flag="l", doc="The interval list over which known phased variants should be kept.") val knownIntervals: Option[PathToIntervals] = None,
+  @arg(flag="k", doc="The interval list over which known phased variants should be kept.") val knownIntervals: Option[PathToIntervals] = None,
   @arg(flag="m", doc="Allow missing fields in the VCF header.") val allowMissingFieldsInVcfHeader: Boolean = true,
   @arg(flag="s", doc="Skip sites where the truth and call are both called but do not share the same alleles.") val skipMismatchingAlleles: Boolean = true,
-  @arg(flag="C", doc="Analyze only the given chromosome.") val chromosome: Option[String] = None,
```
tfenne: This still seems like it would be useful on occasion no?  Why not keep it and make it mutex with `intervals` and then if specific, use it go generate an `IntervalList`?  Up to you though since I don't really ever use this tool.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -56,17 +56,23 @@ import scala.collection.mutable.ListBuffer
...
+  @arg(flag="l", doc="Analyze only the given chromosomes in the interval list.") val intervals: Option[PathToIntervals] = None,
```
tfenne: It seems strange to me to use an interval list, and then use it only to grab the chromosome names you want to use.  Why not just use the regions as described?  At the very least I think the documentation for this option needs to be _much_ clearer as it's too easy to assume that by `chromosome` you meant `region`.  Maybe repeat the warning from the general usage?

nh13:
I really do want a way to specify a set of chromosomes easily from an interval list!  I know it may be confusing, but I am open to other ways. I updated the docs and added a warning to the user as requested.

tfenne:
Please rebase on top of master and ensure tests pass, then merge.

nh13:
Already rebased onto master, will squash and merged.

PR #128, opened 2016-11-17
Fixed column names in TrimPrimers usage.

tfenne:


tfenne:
@nh13 trivial PR here for you

PR #129, opened 2016-11-17
Fixed case where TrimPrimers would erroneously remove some deletions from CIGAR strings.

tfenne:
A small but important change.  TrimPrimers was incorrectly removing some deletions from CIGAR strings that it should not have been.  

The intent has always been to remove deletions that would be adjacent to the clipping in the cigar string (e.g. `20M5D50M` -> `20S50M` not `20S5D50M`.   The problem was that the way it was implemented caused it to do the wrong thing when splitting a cigar operator with a following deletion.  Examples:
`10M5D90M` trim 10 should go to `10S90M`
`50M5D50M` trim 10 should go to `10S40M5D50M` but was going to `10S40M50M`


nh13:
That `while` loop is gnarly.  So long as it is tested, I'm happy.

PR #133, opened 2016-11-22
Bug fix: truncating overlapping blocks may require re-sorting of the

nh13:
blocks.

The blocks are sorted by start position then length.  If we update a block's start position, it cannot be assumed to have the smallest start position in the list of remaining blocks, so we need to re-insert it and re-sort the list of remaining blocks.  This is in the part of the code that truncates overlapping phased blocks.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
+++ b/src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
@@ -26,6 +26,7 @@ package com.fulcrumgenomics.vcf
 
 import java.nio.file.Paths
 import java.util
+import java.util.Comparator
 
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.util.{GenomicSpan, Metric, NumericCounter, ProgressLogger}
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -404,49 +405,65 @@ object PhaseBlock extends LazyLogging {
       progress.record(ctx.getContig, ctx.getStart)
     }
 
+    val blockSet = new util.TreeSet[PhaseBlock](new Comparator[PhaseBlock] {
```
tfenne: Scala has a mutable TreeSet in `scala.collecitons.mutable`, why use the `java.util` one instead?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -404,49 +405,65 @@ object PhaseBlock extends LazyLogging {
...
+      def compare(a: PhaseBlock, b: PhaseBlock): Int = {
+        if (a.start < b.start) -1
+        else if (a.start == b.start) b.length - a.length
```
tfenne: Could you comment that this is to return the shorter block first (I think)?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -404,49 +405,65 @@ object PhaseBlock extends LazyLogging {
...
-          else {
-            blocks.append(left)
-            left = right
-          }
+
+    var left = blockSet.pollFirst()
+    while (!blockSet.isEmpty) {
+      val right = blockSet.pollFirst()
+
+      if (right.getStart < left.getStart) {
```
tfenne: Checks like this can be written as just:
`require(left.getStart <= right.getStart, s"left: $left right: $right")`

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -404,49 +405,65 @@ object PhaseBlock extends LazyLogging {
...
+        throw new IllegalArgumentException(s"left: $left right: $right")
+      }
+      require(left.getStart <= right.getStart)
```
tfenne: Oh, I see you know that.  Why not just put the message in here and scrap the above?

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -404,49 +405,65 @@ object PhaseBlock extends LazyLogging {
...
+      if (left.overlaps(right)) {
+        if (!modifyBlocks) {
```
tfenne: Could also be a `require(modifyBlocks, "....")`

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -404,49 +405,65 @@ object PhaseBlock extends LazyLogging {
...
+          throw new IllegalStateException(s"Block $left overlaps $right")
+        }
+        if (left.encloses(right)) { // keep the left block only
+          logger.info(s"Removing $right enclosed in $left")
+          // do nothing
```
tfenne: `// do nothing (keep left == left) because left may overlap subsequent blocks too!`

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -404,49 +405,65 @@ object PhaseBlock extends LazyLogging {
...
+          left = right
         }
-        // don't forget the last one
+        else { // truncate right
+          logger.info(s"Truncating $right which overlaps $left");
+          blocks.append(left)
+          // Since the start position of the right block is changed, and we are going to use it in the next iteration,
+          // then re-insert it into the set and poll.
+          blockSet.add(right.copy(start=left.getEnd+1))
+          left = blockSet.pollFirst()
```
tfenne: I think even if you truncate right, you should assign `left = right` here instead of `left = blockSet.pollFirst()`, since presumably `right` could also overlap with the subsequent block, but you'll never do that comparison with the code as is.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Since it has `pollFirst`, which retrieves the first element and removes it all in one shot!

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
nh13: This is the part that was buggy.  We require that `left.start <= right.start` at the start of the loop. If we assign `left = right`, then in the requirement many not hold.  An example, which I will add to a test, would be: `(1, 10), (8, 12), (9, 13)`.  If we do as you suggested, then in the first iteration we would truncate the right to `(11, 12)` and assign `left = (11,12)`, and in the next loop we would assign `right = (9, 13)` (`right = blockSet.pollFirst()`), and now our requirement that `left.start <= right.start` at the start of the loop fails.  What we do instead is re-insert `right = (11, 12)` into the tree-set, then call `pollFirst` which assigns `left = (9, 13)`, and so in the next loop, `right = (11, 12)` and the requirement holds.  This is all to do with sorting by start position.

nh13:
@tfenne I am going to leave the `if/else` as is.  I added two tests that should cover out the corner case I was trying to solve.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
tfenne: I stand by my earlier comment that I find this loop super-confusing 😛

tfenne: I see.  The only other suggestion I have for making this clearer would to rename `blockSet` to `blocksIn` and `blocks` to `blocksOut` since you push things back into both of them.

nh13: Done.  And I added more explanation to the loop for our future selves (or myself).

PR #138, opened 2016-12-12
Added output of the longest homopolymer and also the seq/dot-bracket …

tfenne:
…structure of the worst structure when picking long indices.

### DIFF
--- a/src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
+++ b/src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
@@ -71,9 +71,9 @@ case class IndexMetric(index: String,
                        gc: Double,
                        longest_homopolymer: Int,
                        worst_structure_seq: Option[String],
-                       worst_structure: Option[String],
+                       worst_structure_dbn: Option[String],
                        worst_structure_delta_g: Option[Double]) extends Metric {
-  
+
   override protected def formatValues(value: Any): String = value match {
     case None    => ""
     case Some(x) => super.formatValues(x)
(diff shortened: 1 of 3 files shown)


src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
```diff
@@ -70,7 +69,10 @@ case class IndexMetric(index: String,
                        min_mismatches: Int,
                        indices_at_min_mismatches: Int,
                        gc: Double,
-                       min_delta_g: Option[Double]) extends Metric {
+                       longest_homopolymer: Int,
+                       worst_structure_seq: Option[String],
+                       worst_structure: Option[String],
```
nh13: I don't want `worst_structure_structure`, but I kind of do.  It wasn't obvious what should be stored in this string until I went and found where it was set.  Could you add in some scaladoc for this metric?

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -37,4 +38,33 @@ object Sequences {
 
     s1.zip(s2).count(pair => pair._1.toUpper != pair._2.toUpper)
   }
+
+  /**
+    * Returns the 0-based index of and the length of the longest homopolymer in a string. If there are
```
nh13: _non-empty_

src/main/java/com/fulcrumgenomics/util/PickLongIndices.scala
tfenne: how about `worst_structure_dbn`?

PR #141, opened 2016-12-16
Tool to make a VCF with genotypes constructed by mixing the genotypes…

tfenne:
… of two other samples.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
+++ b/src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
@@ -1,0 +1,207 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.vcf
+
+import java.util
+import java.util.Collections
+
+import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.util.Io
+import com.fulcrumgenomics.vcf.MakeTwoSampleMixtureVcf._
+import dagr.sopt
+import dagr.sopt.{arg, clp}
+import htsjdk.samtools.util.IntervalList
+import htsjdk.variant.variantcontext._
+import htsjdk.variant.variantcontext.writer.{Options, VariantContextWriter, VariantContextWriterBuilder}
+import htsjdk.variant.vcf._
+
+import scala.collection.JavaConversions.asJavaCollection
+import scala.collection.JavaConversions.iterableAsScalaIterable
+import scala.collection.mutable.ListBuffer
+
+object MakeTwoSampleMixtureVcf {
+  val AlleleFractionField = "AF"
+  val GermlineFilter      = "alt_allele_in_normal"
+  val UnknownGtFilter     = "unknown_gt"
+  val MultiAllelicFilter  = "multi_allelic"
+
+  val TumorSampleName  = "tumor"
+  val NormalSampleName = "normal"
+}
+
+/**
+  * Creates a VCF by mixing two germline samples at a given proportion.
+  *
+  * @author Tim Fennell
+  */
+@clp(group=ClpGroups.VcfOrBcf, description=
+  """
+    |Creates a simulated tumor or tumor/normal VCF by in-silico mixing genotypes from two samples.
+    |The tumor genotypes are created by mixing the incoming genotypes for the for the
+    |'tumor' sample and the incoming genotypes for the 'normal' samples with the 'tumor'
+    |alleles accounting for 'tumorFraction' of the resulting mixture and the 'normal' alleles
+    |accounting for '1 - tumorFraction' of the resulting mixture.  E.g. if the 'tumor' genotype
+    |is A/C and the 'normal' genotype is C/C and 'tumorFraction' is set at 0.5 the resulting
+    |tumor genotype will be A/C with an allele fraction of 0.75.  The resulting allele fraction
+    |is written to the AF info field in the VCF.
+    |
+    |In tumor-only mode only tumor genotypes are output.  In tumor/normal mode genotypes for
+    |the 'normal' samples are also emitted, and match the genotypes from the input sample.
+    |
+    |All loci (potentially restricted by intervals) that are variant in one or both samples are written
+    |to the output VCF, though in several cases the variants will be filtered:
+    |  - If either of the tumor or normal sample is no-called the resulting locus will have the
+    |    'unknown_gt' filter applied
+    |  - If the tumor and the normal have more than one alternative allele between them the
+    |    'multi_allelic' filter will be applied
+    |  - In tumor/normal mode (as opposed to tumor-only) loci that have an alt allele in the normal
+    |    sample will have the 'alt_allele_in_normal' filter applied
+  """
+)
+class MakeTwoSampleMixtureVcf
+( @arg(flag="i", doc="Input VCF file.")  val input: PathToVcf,
+  @arg(flag="o", doc="Output VCF file.") val output: PathToVcf,
+  @arg(flag="t", doc="Name of the 'tumor' sample in the input VCF.")  val tumor: String,
+  @arg(flag="n", doc="Name of the 'normal' sample in the input VCF.") val normal: String,
+  @arg(flag="f", doc="What fraction of the mixture comes from the 'tumor' sample.") val tumorFraction: Double = 0.5,
+  @arg(flag="T", doc="Tumor only mode - only output tumor genotypes and don't filter sites.") val tumorOnly: Boolean = false,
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
```diff
@@ -0,0 +1,200 @@
...
+    |    'multi_allelic' filter will be applied
+    |  - In tumor/normal mode (as opposed to tumor-only) loci that have an alt allele in the normal
+    |    sample will have the 'alt_allele_in_normal' filter applied
+  """
+)
+class MakeTwoSampleMixtureVcf
+( @arg(flag="i", doc="Input VCF file.")  val input: PathToVcf,
+  @arg(flag="o", doc="Output VCF file.") val output: PathToVcf,
+  @arg(flag="t", doc="Name of the 'tumor' sample in the input VCF.")  val tumor: String,
+  @arg(flag="n", doc="Name of the 'normal' sample in the input VCF.") val normal: String,
```
nh13: Can the names be optional and if not present we just use the first name or ensure there is only one sample?

nh13: Oh, I see, the input VCF is multi-sample.  This was not clear that both input samples should be in the single input VCF, rather than providing two input single-sample VCFs.

src/main/scala/com/fulcrumgenomics/vcf/MultiIntervalVcfIterator.scala
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package com.fulcrumgenomics.vcf
+
+import com.fulcrumgenomics.FgBioDef._
+import htsjdk.samtools.util.{CloseableIterator, IntervalList}
+import htsjdk.variant.variantcontext.VariantContext
+import htsjdk.variant.vcf.VCFFileReader
+
+object MultiIntervalVcfIterator {
```
nh13: If a variant (indel) overlaps multiple intervals it will be returned twice.  Also, we already have an implementation: See `ByIntervalListVariantContextIterator`.  This implementation is quite nice as it has two different underlying implementations: (1) using an overlap detector and (2) using the VCF index.

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
```diff
@@ -0,0 +1,200 @@
...
+      .setReferenceDictionary(in.getSequenceDictionary)
+      .setOption(Options.ALLOW_MISSING_FIELDS_IN_HEADER)
+      .build()
+
+    writer.writeHeader(header)
+    writer
+  }
+
+  /** Generates an iterator over the whole file, or over the intervals if provided. */
+  def iterator(in: VCFFileReader, intervals: Option[PathToIntervals]): Iterator[VariantContext] = {
```
nh13: this looks like something that would be nice in `FgbioDefs` as an implicit method on the VCFFileReader. What do you think?

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
nh13: Also, having both samples in one VCF file removes the need to handle merging INFO/FILTER/FORMAT fields across VCFs.

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
```diff
@@ -0,0 +1,200 @@
...
+
+  /** Gets the genotype of the sample, substituting in a HomRef genotype for no-calls if required. */
+  private def genotypeOf(sample: String, ctx: VariantContext): Genotype = {
+    val actual = ctx.getGenotype(sample)
+    if (actual.isNoCall && this.noCallIsHomRef) new GenotypeBuilder(sample, util.Arrays.asList(ctx.getReference)).make()
+    else actual
+  }
+
+  /** Makes a no-call genotype object with the provided sample name. */
+  private def noCall(sampleName: String) = {
```
nh13: Add return type.

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
```diff
@@ -0,0 +1,200 @@
...
+        }
+        else {
+          val refAllele = ctx.getReference
+          val altAllele = ctx.getAlternateAllele(0)
+
+          // Build the Tumor genotype
+          val builder = new GenotypeBuilder(TumorSampleName)
+          val alleles = new util.ArrayList[Allele]()
+          if (oldGts.forall(_.isCalled)) {
+            if (oldGts.map(_.countAllele(refAllele)).sum + oldGts.count(_.isNoCall) > 0) alleles.add(refAllele)
```
nh13: I am a bit confused by the `_.isNoCall` here given the command line option `noCallIsHomRef` isn't conditioned upon.  Are you thinking of a genotype with allele a No-call and the other a base?

src/main/scala/com/fulcrumgenomics/vcf/MultiIntervalVcfIterator.scala
nh13: Since I think this class is going to be removed, I am not going to review it further.

src/test/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcfTest.scala
```diff
@@ -0,0 +1,136 @@
...
+    builder.addVariant(start=pos, variantAlleles=alleles, sampleName=Some("s1"), genotypeAlleles=List(s1Allele1, s1Allele2))
+    builder.addVariant(start=pos, variantAlleles=alleles, sampleName=Some("s2"), genotypeAlleles=List(s2Allele1, s2Allele2))
+  }
+
+  addVariant(10, _A, _A, _A, _A, _A)  // Monomorphic, should not come out
+  addVariant(20, _A, _A, _C, _A, _C)
+  addVariant(30, _A, _C, _C, _A, _A)
+  addVariant(40, _A, _A, _A, _C, _C)
+  addVariant(50, _A, _C, _C, _C, _A)
+  addVariant(60, _A, _C, _C, _C, _C)
```
nh13: What do these do `addVariant(?, _A, _C, _C, _C, _T)`,  `addVariant(?, _A, _A, _C, _C, _C)`,  `addVariant(?, _A, _C, _C, _G, _G)`?  I think there are some more cases to test.  Basically, how do I get a mixture of two non-reference alternate alleles?

nh13: Also, can you add comments to each `addVariant` as to their result?

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
```diff
@@ -0,0 +1,200 @@
...
+  * Creates a VCF by mixing two germline samples at a given proportion.
+  *
+  * @author Tim Fennell
+  */
+@sopt.clp(group=ClpGroups.VcfOrBcf, description=
+  """
+    |Creates a VCF by in-silico mixing genotypes from two samples. Writes a new VCF for a 'tumor' and
+    |optionally a 'normal' where the tumor is a mixture of two input samples, and the normal
+    |is one of the same input samples. Only outputs loci that are variant in the 'tumor'.
+    |Writes the 'AF' attribute in the format field with the expected allele fraction in the tumor.
```
nh13: I think it should be stated that if the input normal has AF=0.0 and input tumor has AF=0.5, the output tumor will have the specified AF not 0.5 times the specified AF.  I would have expected the output tumor AF to be `tumorFraction * X` where `X` is the AF in the input tumor sample.

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
```diff
@@ -0,0 +1,200 @@
...
+  val TumorSampleName  = "tumor"
+  val NormalSampleName = "normal"
+}
+
+/**
+  * Creates a VCF by mixing two germline samples at a given proportion.
+  *
+  * @author Tim Fennell
+  */
+@sopt.clp(group=ClpGroups.VcfOrBcf, description=
```
tfenne: I don't t know why, but ever since upgrade to the latest IntelliJ it keeps making this change.

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
tfenne: I've re-written the usage, but I thought it was clear from the name "MakeTwoSampleMixture" that the tumor genotypes were generated as a mixture of the two sets of incoming genotypes.

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
tfenne: Yeah - I think it's probably clearer when trying to use the program as there's only one input VCF you can supply, which is easier to miss just reading the code.

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
tfenne: I think this was just a holdover from before that was optional.  I've removed the count of no calls.

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
tfenne: It's private ;)

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
tfenne: That' probably a good idea, but I don't really want to tackle it right now.

src/main/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcf.scala
```diff
@@ -57,12 +57,19 @@ object MakeTwoSampleMixtureVcf {
...
-    |is one of the same input samples. Only outputs loci that are variant in the 'tumor'.
-    |Writes the 'AF' attribute in the format field with the expected allele fraction in the tumor.
+    |Creates a simulated tumor or tumor/normal VCF by in-silico mixing genotypes from two samples.
+    |The tumor genotypes are created by mixing the incoming genotypes for the for the
+    |'tumor' sample and the incoming genotypes for the 'normal' samples with the 'tumor'
+    |alleles accounting for 'tumorFraction' of the resulting mixture and the 'normal' alleles
+    |accounting for '1 - tumorFraction' of the resulting mixture.  E.g. if the 'tumor' genotype
+    |is A/C and the 'normal' genotype is C/C and 'tumorFraction' is set at 0.5 the resulting
+    |tumor genotype will be A/C with an allele fraction of 0.75.  The resulting allele fraction
+    |is written to the AF info field in the VCF.
```
nh13: 👍

tfenne:
I agree that having it work in the case where the samples carry two alts but no ref allele would be nice.  It shouldn't even be that hard, except that there's a ton of screwing around with the VCF code to push the reference allele into the context but then make sure the two samples have the right alleles set.  I'd accept a PR to add that functionality ;)

PR #142, opened 2016-12-16
Add option to ignore Ns as alternates.

nh13:


src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -140,7 +141,7 @@ class ReviewConsensusVariants
     !rec.getReadUnmappedFlag && variantsByChromAndPos(rec.getContig).exists { v =>
       if (v.start >= rec.getAlignmentStart && v.start <= rec.getAlignmentEnd) {
         val readPos = rec.getReadPositionAtReferencePosition(v.start)
-        readPos == 0 || !SequenceUtil.basesEqual(rec.getReadBases()(readPos-1), v.refBase.toByte)
+        readPos == 0 || !SequenceUtil.basesEqual(rec.getReadBases()(readPos-1), v.refBase.toByte) && (!this.ignoreN || !SequenceUtil.isNoCall(rec.getReadBases()(readPos-1)))
```
tfenne: Probably should extract `val base = rec.getReadBases()(readPos-1)` since it's being used twice now.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -67,7 +67,8 @@ class ReviewConsensusVariants
   @arg(flag="g", doc="BAM file of grouped raw reads used to build consensuses.") val groupedBam : PathToBam,
   @arg(flag="r", doc="Reference fasta file.") val ref: PathToFasta,
   @arg(flag="o", doc="Basename of output files to create.") val output : PathPrefix,
-  @arg(flag="t", doc="The SAM/BAM tag used to uniquely identify source molecules.") val tag: String = "MI"
+  @arg(flag="t", doc="The SAM/BAM tag used to uniquely identify source molecules.") val tag: String = "MI",
+  @arg(flag="N", doc="Ignore N base calls.") val ignoreN: Boolean = false
```
tfenne: Ignore N bases in consensus reads?

nh13:
@tfenne I want this when I want to extract just the non-N alternate bases.  I do find both to be quite interesting.

PR #143, opened 2016-12-21
Added code to filter reads going into consensus calling by their ciga…

tfenne:
…r strings to ensure that only reads that are correctly aligned with respect to one another are used.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
@@ -30,6 +30,7 @@ import com.fulcrumgenomics.umi.ConsensusCaller.Base
 import com.fulcrumgenomics.umi.VanillaUmiConsensusCallerOptions._
 import com.fulcrumgenomics.util.NumericTypes._
 import com.fulcrumgenomics.util.{MathUtil, ProgressLogger}
+import dagr.commons.util.LazyLogging
 import htsjdk.samtools._
 import htsjdk.samtools.util.{Murmur3, SequenceUtil}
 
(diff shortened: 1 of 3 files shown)


tfenne:
@nh13 This is a first attempt at filtering reads going into consensus calling based on their cigar, to make sure that there aren't indel artefacts between reads being used to generate a consensus.  Perhaps look at the tests first to understand the intent a little better, and then the main code?  

Also, I'm very open to feedback about how much information we want to capture and log when doing this.  The default is that at the end of running it'll report how many reads were filtered by cigar, and that if you specify -D to turn on debug info, it'll spit out a line every time a tag family is made smaller by this filtering.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -117,6 +151,11 @@ class VanillaUmiConsensusCaller
   private val NotEnoughReadsQual: PhredScore = 0.toByte // Score output when masking to N due to insufficient input reads
   private val TooLowQualityQual: PhredScore = 2.toByte  // Score output when masking to N due to too low consensus quality
 
+  // var to track how many reads meet various fates
+  private var _totalReads: Long = 0
+  private var _filteredForMinReads: Long = 0
+  private var _filteredForMinorityAlignment: Long = 0
```
nh13: underscores are ugly, perhaps just make simple class to store them?

src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -376,4 +376,48 @@ class VanillaUmiConsensusCallerTest extends UnitSpec {
...
+    // These should not be!
+    (1 to  2).foreach { i => builder.addFrag(start=100, cigar="25M2D25M") }
+    (1 to  2).foreach { i => builder.addFrag(start=100, cigar="25M1I24M") }
+    (1 to  2).foreach { i => builder.addFrag(start=100, cigar="20M1D5M1D25M") }
+
+    val recs = VanillaUmiConsensusCaller.filterToMostCommonAlignment(builder.toSeq)
+    recs should have size 11
+    recs.map(_.getCigarString).distinct.sorted shouldBe Seq("25M1D25M", "5S20M1D25M", "5S20M1D20M5H", "25M1D20M5S").sorted
+  }
+
```
nh13: nl

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: Agreed, but a class seems overkill.  I think this is more or less convention when defining read-only vars since the var and the method name cannot match.

PR #144, opened 2016-12-22
Attempt at estimating pooling fractions for a pooled sequencing sample.

tfenne:


tfenne:
@yfarjoun: take a look here for a first cut at an implementation.  I think that samples that are very low in the pool are going to cause a problem unless we have a) sufficient sites at which they are the only variant sample and b) sufficient sequencing depth to see that sample.  Otherwise those samples float around 0 but can become negative.

I'm scaling the pooling fractions ... and also the stderrs, which I'm not sure is appropriate.  If there's a better way to do that please let me know!

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+  @arg(flag="s", minElements=0, doc="Optional subset of samples from VCF to use.") val samples: Seq[String] = Seq.empty,
+  @arg(flag="g", doc="Minimum genotype quality. Use -1 to disable.") val minGenotypeQuality: Int = 30,
+  @arg(flag="c", doc="Minimum sequencing coverage per sample.") val minCoveragePerSample: Int = 12,
+  @arg(flag="m", doc="Minimum mapping quality.") val minMappingQuality: Int = 30,
+  @arg(flag="q", doc="Minimum base quality.") val minBaseQuality:Int = 30
+) extends FgBioTool with LazyLogging {
+  Io.assertReadable(vcf :: bam :: intervals.toList)
+
+  /* Class to hold information about a single locus. */
+  case class Locus(chrom: String, pos: Int, ref: Char, alt: Char, fractions: Array[Double], var observedFraction: Option[Double] = None)
```
yfarjoun: could you name  fractions and observedFractions in a way that makes it clearer which is from the genotypes and which is from the reads?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+      val list = IntervalList.fromFile(this.intervals.head.toFile)
+      val dict = list.getHeader.getSequenceDictionary
+      this.intervals.tail.foreach { f =>
+        val tmp = IntervalList.fromFile(f.toFile)
+        if (!SequenceUtil.areSequenceDictionariesEqual(dict, tmp.getHeader.getSequenceDictionary))
+          fail(s"Sequence dictionaries differ between ${this.intervals.head} and ${f}")
+        else
+          list.addall(tmp.getIntervals)
+      }
+
```
yfarjoun: hey look a space space! 😄

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+    // Get the expected fractions from the VCF
+    val vcfIterator = constructVcfIterator(vcfReader, intervals, sampleNames)
+    val loci = vcfIterator.map { v => Locus(
+      chrom = v.getContig,
+      pos = v.getStart,
+      ref = v.getReference.getBaseString.charAt(0),
+      alt = v.getAlternateAllele(0).getBaseString.charAt(0),
+      fractions = sampleNames.map { s => val gt = v.getGenotype(s); if (gt.isHomRef) 0 else if (gt.isHet) 0.5 else 1.0 }
+    )}.toArray
```
yfarjoun: are we guaranteed that the order that these genotypes will be in is the order of the sorted sample names?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /** Generates an iterator over non-filtered bi-allelic SNPs where all the required samples are genotyped. */
+  def constructVcfIterator(in: VCFFileReader, intervals: Option[IntervalList], samples: Array[String]): Iterator[VariantContext] = {
+    val iterator = intervals match {
+      case None     => in.toIterator
+      case Some(is) => ByIntervalListVariantContextIterator(in, is)
+    }
+
+    val samplesAsUtilSet = CollectionUtil.makeSet(samples:_*)
```
yfarjoun: why the funny name? why not `samplesSet`?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+          list.addall(tmp.getIntervals)
+      }
+
+      Some(list.uniqued(false))
+    }
+  }
+
+  /** Generates an iterator over non-filtered bi-allelic SNPs where all the required samples are genotyped. */
+  def constructVcfIterator(in: VCFFileReader, intervals: Option[IntervalList], samples: Array[String]): Iterator[VariantContext] = {
+    val iterator = intervals match {
```
yfarjoun: can this be named VcIterator? to remind me what type it is?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+      case None     => in.toIterator
+      case Some(is) => ByIntervalListVariantContextIterator(in, is)
+    }
+
+    val samplesAsUtilSet = CollectionUtil.makeSet(samples:_*)
+
+    iterator
+      .filterNot(_.isFiltered)
+      .map(_.subContextFromSamples(samplesAsUtilSet, true))
+      .filter(v => (v.isSNP && v.isBiallelic) || v.isMonomorphicInSamples)
```
yfarjoun: Rather pointless to keep monomorphic variants...why are you doing this?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+      .filter(_.getNoCallCount == 0)
+      .filter(v => v.getGenotypesOrderedByName.forall(gt => gt.getGQ >= this.minGenotypeQuality))
+  }
+
+  /** Constructs a SamLocusIterator that will visit every locus in the input. */
+  def constructBamIterator(loci: Traversable[Locus]): Iterator[LocusInfo] = {
+    val in = SamReaderFactory.make().open(this.bam)
+    val intervals = new IntervalList(in.getFileHeader)
+    loci.foreach(l => intervals.add(new Interval(l.chrom, l.pos, l.pos)))
+    val iterator = new SamLocusIterator(in, intervals.uniqued())
```
yfarjoun: I don't remember if this will include soft-clipped bases...I think not, but better be sure.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+    iterator.setEmitUncoveredLoci(true)
+    iterator.setIncludeNonPfReads(false)
+    iterator.setMappingQualityScoreCutoff(this.minMappingQuality)
+    iterator.setQualityScoreCutoff(this.minBaseQuality)
```
yfarjoun: remove duplicate reads?

yfarjoun:
a test would be good :-)

yfarjoun:
I'm not sure where would be a good place to record this...so I'll use this PR.

There are two things that are missing from the plain linear regression solution:

1. The constraint that the sum of the parameters is 1.
2. The constraint that the parameters are all positive.

These two seem to be solvable:

1. Can be solved by transforming the solution as shown here: https://inst.eecs.berkeley.edu/~ee127a/book/login/l_ols_cls_def.html
which keeps the solution fast and linear. It requires a simple transformation of the inputs, and of-course of the output.  
2. Can be solved by "Non Linear Least Squares" as here: 
https://en.wikipedia.org/wiki/Non-negative_least_squares 
though that solution is for the original, not transformed by (1), so a little bit of work will be required to mashup these two solutions. This solution is "convex" which means that the iterations are bound to converge after a finite number of iterations (the number of parameters in the problem), so it will not take a long time to compute.

src/main/scala/com/fulcrumgenomics/bam/BaseCounts.scala
```diff
@@ -0,0 +1,70 @@
...
+    */
+  def apply(locus: LocusInfo): BaseCounts = apply(locus.getRecordAndOffsets)
+
+  /**
+    * Generates a BaseCounts object from a collection of RecordAndOffest objects
+    * representing a pileup.  Ensures that each template is only counted once.
+    */
+  def apply(recs: Traversable[SamLocusIterator.RecordAndOffset]): BaseCounts = {
+    val countsByBase = recs.groupBy(r => Character.toUpperCase(r.getReadBase.toChar))
+      .map { case (ch, rs) => (ch, rs.map(_.getRecord.getReadName).toSeq.distinct.size) }
```
tfenne: Yes, this is a way to only count a given template once to avoid double-counting over-lapping reads.  The filtering by base quality and mapping quality is done internally by the `SamLocusIterator`.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+    | --min-base-quality filters out bases in the BAM with Q < n
+  """)
+class EstimatePoolingFractions
+( @arg(flag="v", doc="VCF of individual sample genotypes.")  val vcf: PathToVcf,
+  @arg(flag="b", doc="Path to BAM file of sequencing data.") val bam: PathToBam,
+  @arg(flag="o", doc="Output file to write with pooling fractions.") val output: FilePath,
+  @arg(flag="l", minElements=0, doc="Zero or more set of regions to restrict analysis to.") val intervals: Seq[PathToIntervals] = Seq.empty,
+  @arg(flag="s", minElements=0, doc="Optional subset of samples from VCF to use.") val samples: Seq[String] = Seq.empty,
+  @arg(flag="g", doc="Minimum genotype quality. Use -1 to disable.") val minGenotypeQuality: Int = 30,
+  @arg(flag="c", doc="Minimum sequencing coverage per sample.") val minCoveragePerSample: Int = 12,
```
tfenne: I figure it's just easier to do it that way.  Whether I'm looking at a pool of 20 equimolar samples or 4, it's the same number.  Though I guess it might be more complicated for non-equimolar pools.  I'm not tied to it.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: Yes.  sampleNames is an Array (so an ordered collection), and so it's guaranteed that `.map()` will produce a collection in the same order.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,211 @@
...
+      alt = v.getAlternateAllele(0).getBaseString.charAt(0),
+      fractions = sampleNames.map { s => val gt = v.getGenotype(s); if (gt.isHomRef) 0 else if (gt.isHet) 0.5 else 1.0 }
+    )}.toArray
+
+    logger.info(s"Loaded ${loci.length} bi-allelic SNPs from VCF.")
+
+    // And fill in the observed fractions from the BAM
+    val locusIterator = constructBamIterator(loci)
+    locusIterator.zip(loci.iterator).foreach { case (locusInfo, locus) =>
+      if (locusInfo.getSequenceName != locus.chrom || locusInfo.getPosition != locus.pos) fail("VCF and BAM iterators out of sync.")
```
tfenne: Yes, entirely a sanity check.  I figure it's easy to check and if they went out of sync due to some kind of bug, whether here or down in htsjdk, that would be very bad.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: seriously?  It's in a method called `constructVcfIterator` 😛

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: Because the set is not a `scala.collection.Set` but a `java.util.Set`.  This is the convention I've seen used elsewhere for denoting that.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: I couldn't quite decide if it was pointless, or not.  I figure that if you've not included the right samples in the VCF or sample list, including variants that are supposed to be monomorphic in those samples will still make the R^2 more accurate.  If you think it's not worthwhile though I can pull it out.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: It absolutely won't.  The only places I've seen soft-clipping being undone is inside variant callers like the GATK, VarDict etc.  SamLocusIterator is a modern, well-behaved class that only reports pileups of _aligned_ bases 😉

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: It's perhaps not clear, but the standard set of `SamRecordFilter`s on a `SamLocusIterator` already filter out duplicate reads and secondary/supplementary reads.

tfenne:
Tests will be forthcoming - the goal of the WIP (work in progress) review was to make sure you agreed on the general approach before investing the effort in testing!

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
yfarjoun: and that's simply because you need to pass it into subContextFromSamples which takes a Java Set, right?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
yfarjoun: Good thing it's well behaved!

yfarjoun:
👍 LGTM

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,219 @@
...
+      case None     => in.toIterator
+      case Some(is) => ByIntervalListVariantContextIterator(in, is)
+    }
+
+    val samplesAsUtilSet = CollectionUtil.makeSet(samples:_*)
+
+    vcfIterator
+      .filterNot(_.isFiltered)
+      .map(_.subContextFromSamples(samplesAsUtilSet, true))
+      .filter(v => (v.isSNP && v.isBiallelic) || v.isMonomorphicInSamples)
```
yfarjoun: I think that sites should actually be polymorphic. I'd change this line to
```
 .filter(v => (v.isSNP && v.isBiallelic && !v.isMonomorphicInSamples)
```

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: Yes.

tfenne:
@yfarjoun & @nh13 Can you both take a look at this PR when you get a chance please?  A few things worth noting:

1. This PR enables [git-lfs](https://git-lfs.github.com/) usage in fgbio for storage of BAMs, gzipped VCFs and their indices.  This means that going forward we'll have to have git-lfs installed in order to run the tests.  Assuming we're ok with that I'll add info to the README.md before I merge.
1. The tests for this are rather slow.  I've limited the amount of samples and data, but even so they take ~15s.  Not the end of the world, but since it's a regression that's based on allele frequencies in BAMs and the implementation is likely to evolve, I'm not sure that there's a better way to do the tests.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+    |BAM of sequencing data derived from the pool.  Performs a multiple regression
+    |for the alt allele fractions at each SNP locus, using as inputs the individual
+    |sample's genotypes.  Only SNPs that are bi-allelic within the pooled samples are
+    |used.
+    |
+    |Various filtering parameters can be used to control which loci are used:
+    | --intervals will restrict analysis to variants within the described intervals
+    | --min-genotype-quality will filter out any site with any genotype with GQ < n
+    | --min-coverage-per-sample requires that the coverage of a site in the BAM be
+    |     >= min-coverage-per-sample * n_samples
```
nh13: I think the name of this option is confusing, since `min-coverage-per-sample` to me implies that _each_ sample needs to have at least this coverage, whereas you want to filter based on the _mean_ coverage across samples.  Perhaps call this `mean-coverage-per-sample`? Nonetheless, the option name should be changed.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+    | --min-mapping-quality filters out reads in the BAM with MQ < n
+    | --min-base-quality filters out bases in the BAM with Q < n
+  """)
+class EstimatePoolingFractions
+( @arg(flag="v", doc="VCF of individual sample genotypes.")  val vcf: PathToVcf,
+  @arg(flag="b", doc="Path to BAM file of sequencing data.") val bam: PathToBam,
+  @arg(flag="o", doc="Output file to write with pooling fractions.") val output: FilePath,
+  @arg(flag="l", minElements=0, doc="Zero or more set of regions to restrict analysis to.") val intervals: Seq[PathToIntervals] = Seq.empty,
+  @arg(flag="s", minElements=0, doc="Optional subset of samples from VCF to use.") val samples: Seq[String] = Seq.empty,
+  @arg(flag="n", doc="Non-autosomal chromosomes to avoid.") val nonAutosomes: Seq[String] = Sequences.CommonNonAutosomalContigNames,
```
nh13: Alternate haplotypes?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+) extends FgBioTool with LazyLogging {
+  Io.assertReadable(vcf :: bam :: intervals.toList)
+
+  private val Ci99Width = 2.58 // Width of a 99% confidence interval in units of std err
+
+  /* Class to hold information about a single locus. */
+  case class Locus(chrom: String, pos: Int, ref: Char, alt: Char, expectedSampleFractions: Array[Double], var observedFraction: Option[Double] = None)
+
+  override def execute(): Unit = {
+    val vcfReader = new VCFFileReader(vcf.toFile)
```
nh13: align the `=`

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+      if (missingSamples.nonEmpty) fail(s"Samples not present in VCF: ${missingSamples.mkString(", ")}")
+      else samples.toArray.sorted
+    }
+    else {
+      vcfReader.getFileHeader.getSampleNamesInOrder.toSeq.toArray.sorted // toSeq.toArray is necessary cos util.ArrayList.toArray() exists
+    }
+  }
+
+  /** Loads up and merges all the interval lists provided. Returns None if no intervals were specified. */
+  private def loadIntervals: Option[IntervalList] = {
```
nh13: As a style comment, methods without parameters in my mind should look like `val`s, and so should be named as such.  What do you think?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+        ci99_high          = min(1, fractions(index) + stderrs(index)*Ci99Width))
+    }
+
+    Metric.write(metrics, path=output)
+
+    if (regression.estimateRegressionParameters().exists(_ < 0)) {
+      logger.error("#################################################################################")
+      logger.error("# One or more samples is estimated to have fraction < 0. This is likely due to  #")
+      logger.error("# insufficient coverage at some loci and/or too few SNPs being used.            #")
+      logger.error("# Please try again with a higher min-coverage-per-sample.                       #")
```
nh13: Wont increasing `--min-coverage-per-sample` reduce the number of loci?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+      else samples.toArray.sorted
+    }
+    else {
+      vcfReader.getFileHeader.getSampleNamesInOrder.toSeq.toArray.sorted // toSeq.toArray is necessary cos util.ArrayList.toArray() exists
+    }
+  }
+
+  /** Loads up and merges all the interval lists provided. Returns None if no intervals were specified. */
+  private def loadIntervals: Option[IntervalList] = {
+    if (this.intervals.isEmpty) {
```
nh13: Could match on `this.intervals` to get head and tail directly.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+
+      Some(list.uniqued(false))
+    }
+  }
+
+  /** Generates an iterator over non-filtered bi-allelic SNPs where all the required samples are genotyped. */
+  def constructVcfIterator(in: VCFFileReader, intervals: Option[IntervalList], samples: Array[String]): Iterator[VariantContext] = {
+    val vcfIterator = intervals match {
+      case None     => in.toIterator
+      case Some(is) => ByIntervalListVariantContextIterator(in, is)
```
nh13: Just so you aware, there are two underlying implementations, and the one you are going to use skips around the vcf using the index, if that's ok.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+
+  /**
+    * Fills in the observedFraction field for each locus that meets coverage and then returns
+    * the subset of loci that met coverage.
+    */
+  def fillObserveredFractionAndFilter(loci: Array[Locus], minCoverage: Int): Array[Locus] = {
+    val locusIterator = constructBamIterator(loci)
+    locusIterator.zip(loci.iterator).foreach { case (locusInfo, locus) =>
+      if (locusInfo.getSequenceName != locus.chrom || locusInfo.getPosition != locus.pos) fail("VCF and BAM iterators out of sync.")
+      if (locusInfo.getRecordAndOffsets.size() > minCoverage) {
```
nh13: Are we concerned about overlapping read pairs in this coverage calculation?  If so, we could just create `BaseCounts` first then query the latter.  Edit: perhaps the second check below handles this?  If so, add a comment there please.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+        val counts = BaseCounts(locusInfo)
+        val (ref, alt) = (counts(locus.ref), counts(locus.alt))
+
+        // Somewhat redundant with check above, but this protects against a large fraction of Ns or other alleles
```
nh13: Or overlapping reads I think.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+        if (ref + alt >= minCoverage) {
+          locus.observedFraction = Some(alt / (ref + alt).toDouble)
+        }
+      }
+    }
```
nh13: Why not just use a `flatMap` instead of the `foreach` and subsequent filtering below?

src/main/scala/com/fulcrumgenomics/bam/BaseCounts.scala
```diff
@@ -0,0 +1,70 @@
...
+      g=countsByBase.getOrElse('G', 0),
+      t=countsByBase.getOrElse('T', 0),
+      n=countsByBase.getOrElse('N', 0))
+  }
+}
+
+/** Utility class to count the number of observation of each kind of base. */
+case class BaseCounts(a:Int , c: Int, g: Int, t: Int, n: Int) {
+  /** Returns the total number of counted bases. */
+  def total: Int = a + c + g + t + n
```
nh13: Not used anywhere

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+        singletons         = singletons,
+        estimated_fraction = fractions(index),
+        standard_error     = stderrs(index),
+        ci99_low           = max(0, fractions(index) - stderrs(index)*Ci99Width),
+        ci99_high          = min(1, fractions(index) + stderrs(index)*Ci99Width))
+    }
+
+    Metric.write(metrics, path=output)
+
+    if (regression.estimateRegressionParameters().exists(_ < 0)) {
```
nh13: Add a simple test that filters out all loci to test that regression fails?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -0,0 +1,242 @@
...
+  /** Loads up and merges all the interval lists provided. Returns None if no intervals were specified. */
+  private def loadIntervals: Option[IntervalList] = {
+    if (this.intervals.isEmpty) {
+      None
+    }
+    else {
+      val list = IntervalList.fromFile(this.intervals.head.toFile)
+      val dict = list.getHeader.getSequenceDictionary
+      this.intervals.tail.foreach { f =>
+        val tmp = IntervalList.fromFile(f.toFile)
```
nh13: Multiple interval lists not tested?

.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
```
nh13: I am not a fan of this, since this moves us to the full-VM environment from the container service: https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments.   Do we need sudo to install git-lfs?  Can it be done locally?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: Huh?  This has nothing to do with haplotypes, it is just trying to avoid chromosomes/contigs that have ploidy != 2.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: This is actually testing for a different condition that is somewhat hard to induce via a test.  It is testing to see if one or more of the samples is estimated to be a negative fraction of the pool, which _can_ but won't necessary happen if you list one or more samples that aren't actually in the pool

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: I'll get rid of the please try again line.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: I'm in two minds about this.  When they do something relatively lightweight, I agree.  When they do things like load files etc. that you wouldn't want to do multiple times, I like to have a verb in the method name to make it clear.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: Nice

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: I don't see a problem with that.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: Yes, will comment to that effect.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: I'll add a comment.  The idea is to do an easy/gross check here to avoid doing a bunch of work if the coverage is obviously too low, then check the real depth later.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: I think this is probably cleaner.  With a `flatMap` I'd have to add 2 more `else { None }` branches.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: I don't disagree but I'm struggling to come up with a better name.  I want it to be (coverage / samples) so that a sensible default can be set, and you don't have to think about changing it for the number of samples in the pool.   `min-mean-coverage-per-sample` is that best I've got, and I still don't like it.

tfenne: Going with `min-mean-sample-coverage` for now.

tfenne:
@nh13 I've addressed what I can, including parallelizing some test suites to bring the testing time back down.  In had never really occurred to me before that you can have scalatest tests in a loop, e.g.: 
```scala
  Range.inclusive(3, Samples.size-1).foreach { n =>
    it should s"accurately estimate a mixof $n samples" in {
        // test goes here
    }
  }
```

Want to take a final quick look through?

nh13:
👍  thanks for the edits.

Issue #145, opened 2016-12-23
Proposal for improvements to consensus calling


tfenne:
There are really three parts to this proposal:

1. Simplify the consensus caller (particularly the usage thereof) by removing parameters that we have found are not useful and are overlapped with other parameters

2. Add tags to the consensus reads that are generated so that downstream tools can know a lot more about the properties of each consensus read and it's supporting raw reads

3. Move filtering of consensus reads and bases _largely_ into a separate program that can be run after consensus calling, removing the need to call consensuses multiple times to experiment with different cutoffs

# Simplifying the consensus caller

To recap, we have a lot of parameters to consensus calling:

```scala
errorRatePreUmi: PhredScore
errorRatePostUmi: PhredScore
minInputBaseQuality: PhredScore
maxRawBaseQuality: PhredScore
rawBaseQualityShift: PhredScore
minConsensusBaseQuality: PhredScore
minReads: Int
minMeanConsensusBaseQuality: PhredScore
requireConsensusForBothPairs: Boolean
```

I'm finding that in general there are two sets of highly overlapped parameters where I favor one of the set, and would advocate for removal of the others:

1. `errorRatePostUmi`, `maxRawBaseQuality` and `rawBaseQualityShift` all work to effectively reduce the probability we construct from the raw read base qualities, albeit in slightly different ways.  I'm finding that I tend to use `errorRatePostUmi` exclusively and then set `maxRawBaseQuality` to something high that will have little to no effect, and `rawBaseQualityShift` to 0.  I propose removing `maxRawBaseQuality` and `rawBaseQualityShift`.

2. `minConsensusBaseQuality` and `minMeanConsensusBaseQuality`.  While not exactly redundant I find that I set minConsensusBaseQuality high enough (i.e. close enough to `errorRatePreUmi` that then the `minMeanConsensusBaseQuality` becomes meaningless.  I would propose getting rid of `minMeanConsensusBaseQuality`, perhaps in favor of introducing a filter on the number or proportion of no-calls generated in a consensus read (more in the filtering section).

# Adding Tags to Consensus Reads

It would help significantly to encode information about the consensus reads into tags that are added to the consensus reads in the BAM file for three reasons:

1. One could generate a single consensus BAM, re-align it, and then filter it different ways to experiment with filtering settings much more easily and quickly

[...]

PR #147, opened 2017-01-04
Consensus calling improvements

tfenne:
@n13 This is a _work in progress_ for #145.  I'm trying to keep each of the three areas in their own independent commits.  Making this PR in case you are interested in looking as I go along.


src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+                    r1Supplementals: Seq[SAMRecord] = Nil,
+                    r2Supplementals: Seq[SAMRecord] = Nil,
+                    r1Secondaries:   Seq[SAMRecord] = Nil,
+                    r2Secondaries:   Seq[SAMRecord] = Nil) {
+
+  /** Gets the query name of the template. */
+  val name: String = {
+    if (r1.isDefined) r1.get.getReadName
+    else if (r2.isDefined) r2.get.getReadName
+    else Seq(r1Supplementals, r2Supplementals, r1Secondaries, r2Secondaries).find(_.nonEmpty).map(_.head.getReadName)
```
nh13: Could also use `.flatMap(_.headOption).headOption.map(_.getReadName)` for added complexity :1st_place_medal:

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+  * Class that represents all reads from a template within a BAM file.
+  */
+case class Template(r1: Option[SAMRecord],
+                    r2: Option[SAMRecord],
+                    r1Supplementals: Seq[SAMRecord] = Nil,
+                    r2Supplementals: Seq[SAMRecord] = Nil,
+                    r1Secondaries:   Seq[SAMRecord] = Nil,
+                    r2Secondaries:   Seq[SAMRecord] = Nil) {
+
+  /** Gets the query name of the template. */
```
nh13: Add "Does not validate that all the records have the sam name"?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+    else Seq(r1Supplementals, r2Supplementals, r1Secondaries, r2Secondaries).find(_.nonEmpty).map(_.head.getReadName)
+      .getOrElse(throw new IllegalStateException("Template created with no reads!"))
+  }
+
+  /** Returns an iterator over all records that are not the primary r1 and r2. */
+  def allSupplementaryAndSecondary: Iterator[SAMRecord] =
+    r1Supplementals.iterator ++ r2Supplementals.iterator ++ r1Secondaries.iterator ++ r2Secondaries.iterator
+
+  /** The total count of records for the template. */
+  def readCount: Int = r1.size + r2.size + r1Supplementals.size + r2Supplementals.size + r1Secondaries.size + r2Secondaries.size
```
nh13: If you are guarding against an expensive `size` operation in making this a `val`, why not just make this a `lazy val`?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+    * Generates a Template for the next template in the buffered iterator. Assumes that the
+    * iterator is queryname sorted or grouped.
+    */
+  def apply(recs: BetterBufferedIterator[SAMRecord]): Template = {
+    require(recs.hasNext)
+    val name = recs.head.getReadName
+    apply(recs.takeWhile(_.getReadName == name))
+  }
+
+
```
nh13: nl

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+  def apply(recs: Iterator[SAMRecord]): Template = {
+    var r1: Option[SAMRecord] = None
+    var r2: Option[SAMRecord] = None
+    val r1Supp = ListBuffer[SAMRecord]()
+    val r2Supp = ListBuffer[SAMRecord]()
+    val r1Sec  = ListBuffer[SAMRecord]()
+    val r2Sec  = ListBuffer[SAMRecord]()
+
+    recs foreach { r =>
+      val isR1 = !r.getReadPairedFlag || r.getFirstOfPairFlag
```
nh13: Validate the read names are the same.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+      if (isR1) {
+        if      (r.getNotPrimaryAlignmentFlag) r1Sec += r
```
nh13: Can you shift over `r1Sec` a few spaces :)

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+    * @return a SortingCollection
+    */
+  def sortingCollection(order: SortOrder,
+                        header: SAMFileHeader,
+                        maxInMemory: Int = 1e6.toInt,
+                        tmpDir: Option[DirPath] = None): SortingCollection[SAMRecord] = {
+    SortingCollection.newInstance(
+      classOf[SAMRecord],
+      new BAMRecordCodec(header),
+      if (order == SortOrder.queryname) new SAMRecordQueryNameComparator else new SAMRecordCoordinateComparator,
```
nh13: These are not the only sort orders!  Can you use `order.getComparator()`?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+      maxInMemory,
+      tmpDir.map(_.toFile).orNull
+    )
+  }
+
+
```
nh13: nl

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+                        maxInMemory: Int = 1e6.toInt,
+                        tmpDir: Option[DirPath] = None): SortingCollection[SAMRecord] = {
+    SortingCollection.newInstance(
+      classOf[SAMRecord],
+      new BAMRecordCodec(header),
+      if (order == SortOrder.queryname) new SAMRecordQueryNameComparator else new SAMRecordCoordinateComparator,
+      maxInMemory,
+      tmpDir.map(_.toFile).orNull
+    )
+  }
```
nh13: This is really nice!  Likely I will want to pass in my own comparator class instead of sort order in the future, so this will make that easy.  Do you want to check that the order passed in is the same as the order set in the header?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -0,0 +1,207 @@
...
+    if (in.getFileHeader.getSortOrder == SortOrder.queryname || in.getFileHeader.getGroupOrder == GroupOrder.query) {
+      in.iterator().bufferBetter
+    }
+    else {
+      logger.info("Sorting into queryname order.")
+      val progress = new ProgressLogger(this.logger, "Queryname sorted")
+      val sorter = sortingCollection(SortOrder.queryname, in.getFileHeader, maxInMemory, tmpDir)
+      in.foreach { rec =>
+        sorter.add(rec)
+        progress.record(rec)
```
nh13: I wish we had a foreach that could take a list of methods to apply.  I see us doing this a lot.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+
+import scala.collection.JavaConversions.iterableAsScalaIterable
+
+@clp(
+  group = ClpGroups.Umi, description =
+    """
+      |Filters a BAM file of consensus reads generated by CallMolecularConsensusReads. Two kinds of
+      |filtering are performed:
+      |  1. Masking/filtering of individual bases in reads
+      |  2. Filtering out of reads
```
nh13: Add `(i.e. they will not be present in the output)`

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+      |
+      |Base-level filtering of masking is only applied if per-base tags are present. Read-level
```
nh13: ` filtering of masking` -> `masking/filtering` or `filtering or masking`

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+      |  1. Bases with quality scores below min-base-quality are masked to Ns
+      |  2. Bases with fewer than min-reads contributing raw reads are masked to Ns
+      |  3. Bases with a consensus error rate (defined as the fraction of contributing reads that
+      |     voted for a different base than the consensus call) higher than max-base-error-rate
+      |     are masked to Ns
+      |  4. Reads with a proportion of Ns higher than max-no-call-fraction *after* per-base
+      |     filtering are filtered out.
+      |  5. Reads with fewer than min-reads contributing reads are filtered out
+      |  6. Reads with an average consensus error rate higher than max-read-error-rate are
+      |     filtered out
```
nh13: Can 5 & 6 be applied *before* the per-base filters, for efficiency?   If they do come afterwards, state that the min-reads and average consensus error rate are not affected by 1-4 but simply come from the corresponding tag value.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+      |Filters a BAM file of consensus reads generated by CallMolecularConsensusReads. Two kinds of
+      |filtering are performed:
+      |  1. Masking/filtering of individual bases in reads
+      |  2. Filtering out of reads
+      |
+      |Base-level filtering of masking is only applied if per-base tags are present. Read-level
+      |filtering is always applied.  When filtering reads, secondary alignments and supplementary
+      |records may be removed independently if they fail one or more filters; if either R1 or R2
+      |primary alignments fail a filter then all records for the template will be filtered out.
+      |
```
nh13: Add in some doc (or better a link to the wiki) where the tags are defined.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+      |  4. Reads with a proportion of Ns higher than max-no-call-fraction *after* per-base
+      |     filtering are filtered out.
+      |  5. Reads with fewer than min-reads contributing reads are filtered out
+      |  6. Reads with an average consensus error rate higher than max-read-error-rate are
+      |     filtered out
+      |
+      |In order to correctly filter reads in or out by template, if the input BAM is not queryname
+      |sorted or grouped it will be sorted into queryname order.  Likewise, in order to emit
+      |records with corrected NM, UQ and MD tags the reads are re-sorted into coordinate order
+      |for output.
```
nh13: This is good.  I would also like an option to set the final output sort order so I don't have to have this conversation in my head: "When I corrected the NM tag, I get coordinate, but if I don't I get queryname sorted, but only if the input wasn't queryname grouped." but instead can think: "Wait, what output sort order am I getting?  Oh, I can specify it?  Lets do that".

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+class FilterConsensusReads
+( @arg(flag="i", doc="The input SAM or BAM file of consensus reads.") val input: PathToBam,
+  @arg(flag="o", doc="Output SAM or BAM file.") val output: PathToBam,
+  @arg(flag="r", doc="Reference fasta file.") val ref: PathToFasta,
+  @arg(flag="R", doc="Reverse [complement] per base tags on reverse strand reads.") val reversePerBaseTags: Boolean = true,
+  @arg(flag="N", doc="Mask (make 'N') consensus bases with quality less than this threshold.") val minBaseQuality: PhredScore = DefaultMinConsensusBaseQuality,
+  @arg(flag="M", doc="The minimum number of reads supporting a consensus base/read.") val minReads: Int = DefaultMinReads,
+  @arg(flag="E", doc="The maximum raw-read error rate across the entire consensus read.") val maxReadErrorRate: Double = 0.025,
+  @arg(flag="e", doc="The maximum error rate for a single consensus base.") val maxBaseErrorRate: Double = 0.1,
+  @arg(flag="n", doc="Maximum fraction of no-calls in the read after filtering.") val maxNoCallFraction: Double = 0.2
```
nh13: Could you make sure the order of these parameters match the order in which they are applied in the usage doc?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+    """
+)
+class FilterConsensusReads
+( @arg(flag="i", doc="The input SAM or BAM file of consensus reads.") val input: PathToBam,
+  @arg(flag="o", doc="Output SAM or BAM file.") val output: PathToBam,
+  @arg(flag="r", doc="Reference fasta file.") val ref: PathToFasta,
+  @arg(flag="R", doc="Reverse [complement] per base tags on reverse strand reads.") val reversePerBaseTags: Boolean = true,
+  @arg(flag="N", doc="Mask (make 'N') consensus bases with quality less than this threshold.") val minBaseQuality: PhredScore = DefaultMinConsensusBaseQuality,
+  @arg(flag="M", doc="The minimum number of reads supporting a consensus base/read.") val minReads: Int = DefaultMinReads,
+  @arg(flag="E", doc="The maximum raw-read error rate across the entire consensus read.") val maxReadErrorRate: Double = 0.025,
```
nh13: Should the default values be stored somewhere else here and below, similar to `DefaultMinReads`?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+  @arg(flag="e", doc="The maximum error rate for a single consensus base.") val maxBaseErrorRate: Double = 0.1,
+  @arg(flag="n", doc="Maximum fraction of no-calls in the read after filtering.") val maxNoCallFraction: Double = 0.2
+) extends FgBioTool with LazyLogging {
+  // Baseline input validation
+  Io.assertReadable(input)
+  Io.assertReadable(ref)
+  Io.assertCanWriteFile(output)
+  if (maxReadErrorRate < 0 && maxReadErrorRate > 1) fail("max-read-error-rate must be between 0 and 1.")
+  if (maxBaseErrorRate < 0 && maxBaseErrorRate > 1) fail("max-base-error-rate must be between 0 and 1.")
+  if (maxNoCallFraction < 0 && maxNoCallFraction > 1) fail("max-no-call-fraction must be between 0 and 1.")
```
nh13: What about a `failif` method that takes in a boolean and a message, similar to `require`?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+  private var totalBases:  Long = 0
+
+  // Case class to track results of filtering a single read
+  private[umi] case class FilterResult(keepRead: Boolean, maskedBases: Int)
+
+  override def execute(): Unit = {
+    val in       = SamReaderFactory.make().open(input)
+    val header   = in.getFileHeader.clone()
+    header.setSortOrder(SortOrder.coordinate)
+    val sorter    = Bams.sortingCollection(SortOrder.coordinate, header)
```
nh13: Seeing how `sortingCollection` is used, isn't `order` redundant here? I.e. we could inspect the sort order in the provided header?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+        totalBases  += r1.getReadLength + template.r2.map(_.getReadLength).getOrElse(0)
+        maskedBases += r1Result.maskedBases + r2Result.maskedBases
+        sorter.add(r1)
+        progress1.record(r1)
+        template.r2.foreach { r => sorter.add(r); progress1.record(r) }
+
+        template.allSupplementaryAndSecondary.foreach { r =>
+          val result = filterRecord(r)
+          if (result.keepRead) {
+            totalBases += r.getReadLength
```
nh13: Line it up.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+          }
+        }
+      }
+    }
+
+    // Then iterate the reads in coordinate order and re-calculate key tags
+    logger.info("Filtering complete; fixing tags and writing coordinate sorted reads.")
+    val progress2 = new ProgressLogger(logger, verb="Wrote")
+    val walker = new ReferenceSequenceFileWalker(ref.toFile)
+    sorter.doneAdding()
```
nh13: not necessary since the subsequent call to `iterator()` will do this for you.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+    header.setSortOrder(SortOrder.coordinate)
+    val sorter    = Bams.sortingCollection(SortOrder.coordinate, header)
+    val out       = new SAMFileWriterFactory().setCreateIndex(true).makeWriter(header, true, output.toFile, null)
+    val progress1 = new ProgressLogger(logger, verb="Filtered and masked")
+
+    // Go through the reads by template and do the filtering
+    Bams.templateIterator(in).foreach { template =>
+      val r1 = template.r1.getOrElse(throw new IllegalStateException(s"${template.name} had no R1."))
+      if (r1.getReadPairedFlag) require(template.r2.isDefined, s"Paired read missing R2: ${template.name}")
+      totalReads += template.readCount
```
nh13: My vote would be for `totalReads`, `totalBases`, and `maskedBases` to only count primary consensus reads.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+    *         in the read that were masked to Ns as a result of per-base filtering
+    */
+  private[umi] def filterRecord(rec: SAMRecord): FilterResult = {
+    val maxDepth = rec.getIntegerAttribute(ConsensusTags.PerRead.RawReadCount)
+    val errRate = rec.getFloatAttribute(ConsensusTags.PerRead.RawReadErrorRate)
+    if (maxDepth == null || errRate == null) fail(s"Read ${rec.getReadName} does not appear to have consensus calling tags present.")
+
+    // Only bother looking at reads where per-read criteria are met
+    if (maxDepth < this.minReads || errRate > maxReadErrorRate) {
+      FilterResult(false, 0)
```
nh13: This disagrees with the usage.  I like this better.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+      val errors = rec.getSignedShortArrayAttribute(ConsensusTags.PerBase.RawReadErrors)
+
+      if (rec.getReadNegativeStrandFlag && this.reversePerBaseTags) {
+        reverse(depths)
+        reverse(errors)
+        rec.setAttribute(ConsensusTags.PerBase.RawReadCount, depths)
+        rec.setAttribute(ConsensusTags.PerBase.RawReadErrors, errors)
+      }
+
+      val pb = depths != null && errors != null
```
nh13: Why do we need both?  Can't we just have one or the other?  Seems simple to support.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+    *
+    * @param  rec the read to filter
+    * @return a tuple of (Boolean, Int) where the first value is true if the read should be kept
+    *         and false if it should be filtered out, and the second value is the count of bases
+    *         in the read that were masked to Ns as a result of per-base filtering
+    */
+  private[umi] def filterRecord(rec: SAMRecord): FilterResult = {
+    val maxDepth = rec.getIntegerAttribute(ConsensusTags.PerRead.RawReadCount)
+    val errRate = rec.getFloatAttribute(ConsensusTags.PerRead.RawReadErrorRate)
+    if (maxDepth == null || errRate == null) fail(s"Read ${rec.getReadName} does not appear to have consensus calling tags present.")
```
nh13: Should we support missing per-read values but present per-base values?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -0,0 +1,217 @@
...
+    if (maxDepth < this.minReads || errRate > maxReadErrorRate) {
+      FilterResult(false, 0)
+    }
+    else {
+      val bases = rec.getReadBases
+      val quals = rec.getBaseQualities
+      val depths = rec.getSignedShortArrayAttribute(ConsensusTags.PerBase.RawReadCount)
+      val errors = rec.getSignedShortArrayAttribute(ConsensusTags.PerBase.RawReadErrors)
+
+      if (rec.getReadNegativeStrandFlag && this.reversePerBaseTags) {
```
nh13: This is almost too cute, since reverse will work when depths/errors is null (not present).

src/test/scala/com/fulcrumgenomics/bam/BamsTest.scala
```diff
@@ -0,0 +1,175 @@
...
+class BamsTest extends UnitSpec {
+  /* Dummy implementation of a reference sequence file walker that always returns chr1 w/2000 As. */
+  private object DummyRefWalker extends ReferenceSequenceFileWalker(null.asInstanceOf[ReferenceSequenceFile]) {
+    private val bases = new Array[Byte](2000)
+    util.Arrays.fill(bases, 'A'.toByte)
+    private val ref = new ReferenceSequence("chr1", 0, bases)
+    override def get(sequenceIndex: Int): ReferenceSequence = ref
+  }
+
+  "Bams.queryGroupedIterator" should "work on a reader that is already query sorted already" in {
```
nh13: too many already.

src/test/scala/com/fulcrumgenomics/bam/BamsTest.scala
```diff
@@ -0,0 +1,175 @@
...
+      }
+    }
+  }
+
+  it should "handle reads with secondary and supplementary records" in {
+    val builder = new SamRecordSetBuilder(sortOrder=SortOrder.coordinate)
+    Range.inclusive(1, 5).foreach { i=>
+      builder.addPair(s"p$i", start1=i*100,  start2=i*100 + 500)
+      builder.addPair(s"p$i", start1=i*1000, start2=i*1000+500 ).foreach(_.setNotPrimaryAlignmentFlag(true))
+      builder.addPair(s"p$i", start1=i*10,   start2=i*10+500   ).foreach(_.setSupplementaryAlignmentFlag(true))
```
nh13: Funky spacing around the `500`s

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -114,6 +124,7 @@ class CallMolecularConsensusReads
       minInputBaseQuality          = minInputBaseQuality,
       minConsensusBaseQuality      = minConsensusBaseQuality,
       minReads                     = minReads,
+      producePerBaseTags           = outputPerBaseTags,
```
tfenne: I'm on the fence.  I worded them differently because one _produces_ tags and the other _outputs_ them to a file on disk.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -60,19 +68,69 @@ case class VanillaUmiConsensusCallerOptions
...
+  * names for consistency if/when we add duplex calling:
+  *     Value                 AB  BA  Final
+  *     ===================== ==  ==  =====
+  *     bases                 ac  bc  bases
+  *     quals                 aq  bq  quals
+  *     per-read-depth        aD  bD  cD
+  *     per-read-min-depth    aM  bM  cM
+  *     per-read-error-count  aE  bE  cE
+  *     per-base-depth        ad  bd  cd
+  *     per-base-error-count  ae  be  ce
```
tfenne: I was waiting to see if you agreed with it!  Yes, let's add an update to the wiki page to the todo list.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -60,19 +68,69 @@ case class VanillaUmiConsensusCallerOptions
...
+  object PerBase {
+    /** The per-base number of raw-reads contributing to the consensus (stored as a short[]). */
+    val RawReadCount    = "cd" // consensus depth
+    /** The number of bases at each position that disagreed with the final consensus call (stored as a short[]). */
+    val RawReadErrors   = "ce" // consensus errors
+  }
+
+  object PerRead {
+    /** The number of raw reads that contributed to the consensus. I.e. the maximum of the per-base raw-read counts. */
+    val RawReadCount    = "cD" // consensus Depth
```
tfenne: Likely noone but us will be looking at the long names for the tags - so I don't know that it matters here so much as wherever we explain in more detail.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -262,15 +322,22 @@ class VanillaUmiConsensusCaller
 
           }
         }
-        consensusBases(positionInRead) = base
-        consensusQuals(positionInRead) = qual
+
+        consensusBases(positionInRead)  = base
+        consensusQuals(positionInRead)  = qual
```
tfenne: To see if you're paying attention ;)

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -262,15 +322,22 @@ class VanillaUmiConsensusCaller
...
+
+        // Generate the values for depth and count of errors
+        val depth  = builder.contributions
+        val errors = if (base == NoCall) depth else depth - builder.observations(base)
+        consensusDepths(positionInRead) = if (depth  > Short.MaxValue) Short.MaxValue else depth.toShort
+        consensusErrors(positionInRead) = if (errors > Short.MaxValue) Short.MaxValue else errors.toShort
```
tfenne: Added in a couple of places.

src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -375,6 +375,56 @@ class VanillaUmiConsensusCallerTest extends UnitSpec {
...
+    val caller = new VanillaUmiConsensusCaller(input=builder.toIterator, header=builder.header, options=cco(producePerBaseTags=false))
+
+    caller.hasNext shouldBe true
+    val consensus = caller.next()
+    caller.hasNext shouldBe false
+    consensus.getSignedShortArrayAttribute(ConsensusTags.PerBase.RawReadCount)  shouldBe null
+    consensus.getSignedShortArrayAttribute(ConsensusTags.PerBase.RawReadErrors) shouldBe null
+    consensus.getIntegerAttribute(ConsensusTags.PerRead.RawReadCount)     shouldBe 2
+    consensus.getIntegerAttribute(ConsensusTags.PerRead.MinRawReadCount)  shouldBe 2
+    consensus.getFloatAttribute(ConsensusTags.PerRead.RawReadErrorRate) shouldBe 0.toFloat
```
tfenne: I think about this all the time - I do wonder if we shouldn't have a wrapper for SAMRecord, and an implicit to convert Iterator[SAMRecord] to Iterator[BetterSamRecord] so that the conversion is one-time, not everytime an enhanced method is used.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
tfenne: LOL

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
tfenne: It's currently a `def` not a `val`, but making it a `lazy val` is a good idea anyway.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
tfenne: They're the only ones in the spec ;)  But yes, for you, I'll use `order.getComparatorInstance()` instead.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
tfenne: I don't think it matters what's in the header - I think that's only uses so that it can write out the sequence name/index mapping information.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
tfenne: Agreed.  I'm sure we could come up with something via implicits that gave us what we wanted.  It's just not clear to me that that is good long term since it'd likely confuse others coming into the project.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: 5 & 6 are applied first.  I'll move them up the list to make it clearer.  I wanted to keep the base and read level filtering steps together in the list, but 4 is last and is dependent on 1-3...

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: For now I'll add a "see CallMolecularConsensusReads".  Once we have an updated wiki page I'll add that here.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: Output sort order is always `coordinate`, I'll make that clearer but would prefer not to make that an option for now.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: I was actually tempted to go the other way and get rid of the centralized defaults because I think that `CallMolecularConsensusReads` and `FilterConsensusReads` probably want different defaults, with the former being quite low bars (e.g. min=2 q=30) and filter being higher.  What do you think of that?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: Done, although `minRead` is used for both per-read and per-base but can't appear twice.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: I'm happy enough with this as it is, I think.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: In this usage it is .. but not always.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: I'm on the fence, but since you have a preference I can make that change.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: 'Tis gone!

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: Trust but verify?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: We currently have no way to create that situation, so I think that would be un-needed complexity.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: _almost_

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
tfenne: In theory ... but it would make the code more complicated as one would need to check just depths, the other depths & errors.  And our tools only write _both_ or _neither_.

src/test/scala/com/fulcrumgenomics/bam/BamsTest.scala
tfenne: alright already!

nh13:
0.1.2 release is done, so rebase on master and you should be unblocked on that front at least.

PR #148, opened 2017-01-05
Added ability to output all UMIs to a single tag as well as individua…

tfenne:
…l tags in ExtractUmisFromBam.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
@@ -238,7 +238,10 @@ object ExtractUmisFromBam {
     val readStructureBases = readStructure.structureReadWithQualities(bases, qualities, strict = false)
     // get the molecular index segments
     val molecularIndexBases = readStructureBases.collect { case SubRead(b: String, _, m: MolecularBarcode) => b }
+
     // set the index tags
+    // TODO: when we remove the deprecated molecularBarcodeTags option, consider whether or not we still
+    //       need to have support for specifying a single tag via molecularIndexTags.
     molecularIndexTags match {
       case Seq(tag) => record.setAttribute(tag, molecularIndexBases.mkString(UmiDelimiter))
       case _ =>
--- a/src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
@@ -101,15 +101,15 @@ class ExtractUmisFromBam
   // This can be removed once the @deprecated molecularBarcodeTags is removed
   private val perIndexTags = if (molecularIndexTags.nonEmpty) molecularIndexTags else molecularBarcodeTags
 
-  // validate the read structure versus the molecular barcode tags
+  // validate the read structure versus the molecular index tags
   {
     // create a read structure for the entire template
     val rs = ReadStructure(readStructure.mkString(""))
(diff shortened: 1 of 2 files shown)


tfenne:
@nh13 There are really two changes in one here.  The functional change is to allow the tool to output a single tag with concatenated UMIs (e.g. `RX`) in addition to per-UMI tags (e.g. `ZA` and `ZB` or similar).  It's only really useful when there are multiple UMIs but it can be used whether or not there are multiple.

In addition it was driving me slightly crazy that the tool was called `ExtractUmisFromBam` where `umi` == "unique molecular _index_" but that a lot of the documentation and variables used `molecularBarcode`.  I changed everything over to use _index_ instead, and deprecated the prior command line parameter.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -204,26 +225,26 @@ object ExtractUmisFromBam {
...
-    // get the molecular barcode segments
-    val molecularBarcodeBases = readStructureBases.collect { case SubRead(b: String, _, m: MolecularBarcode) => b }
-    // set the barcode tags
-    molecularBarcodeTags match {
-      case Seq(tag) => record.setAttribute(tag, molecularBarcodeBases.mkString(UmiDelimiter))
+    // get the molecular index segments
+    val molecularIndexBases = readStructureBases.collect { case SubRead(b: String, _, m: MolecularBarcode) => b }
+    // set the index tags
+    molecularIndexTags match {
+      case Seq(tag) => record.setAttribute(tag, molecularIndexBases.mkString(UmiDelimiter))
```
nh13: I think when the `molecularBarcodesTag` option is removed, we'll want to not have the behavior of having the values concatenated if the  `molecularBarcodesTag` has only one tag.  Could you add a note to that effect here?

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
```diff
@@ -266,7 +266,41 @@ class ExtractUmisFromBamTest extends UnitSpec with OptionValues {
...
+    an[ValidationException] shouldBe thrownBy {
+      new ExtractUmisFromBam(input=in, output=out, molecularIndexTags=Seq("RX", "RZ"), singleTag=Some("RX"), readStructure=Seq("5M5T", "5M5T"))
+    }
+
+    an[ValidationException] shouldBe thrownBy {
+      new ExtractUmisFromBam(input=in, output=out, molecularIndexTags=Seq("RX", "RZ"), singleTag=Some("RXX"), readStructure=Seq("5M5T", "5M5T"))
+    }
+  }
+
+
```
nh13: nl

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
tfenne: Sure.  Though I think we may still want it....  Without it there's no way to get all the values contatenated in a single tag _without_ also writing per-UMI tags.  The latest version supports three modes:
1) Just write per umi tags by providing the appropriate number of `molecular-index-tag`s
2) Just write a single tag with all values by providing a single `molecular-index-tag`
3) Write per-umi tags _and_ a concatenated tag by providing the appropriate number of `molecular-index-tag`s _and_ `single-tag`.

tfenne:
Agreed on the deprecation making it a bit messy, but I didn't want to just change the option name without warning.  Requested changes are made, will merge once tests pass.

PR #149, opened 2017-01-05
All parameters in order are required for @deprecated annotations.

nh13:
Related: https://issues.scala-lang.org/browse/SI-7656


nh13:
@tfenne looks like we have a broken tool in the wild, could you take a look at this PR?

PR #151, opened 2017-01-06
Tool to clip overlapping reads so that reads from the same insert don't overlap

tfenne:


src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -201,8 +199,8 @@ object Bams extends LazyLogging {
     }
     else {
       val refBases = ref.get(rec.getReferenceIndex).getBases
-      SequenceUtil.calculateMdAndNmTags(rec, refBases, rec.getAttribute(MD.name()) != null, rec.getAttribute(NM.name()) != null)
-      if (rec.getAttribute(SAMTag.UQ.name) != null && rec.getBaseQualities != null && rec.getBaseQualities.length != 0) {
+      SequenceUtil.calculateMdAndNmTags(rec, refBases, true, true)
```
tfenne: I realized I couldn't do the "replace only if it's there strategy" as we decided that the clipping code should forcefully remove tags that are invalid ... so I need to fill them in on reads that don't have them.  Oh well.

src/main/scala/com/fulcrumgenomics/bam/ClipOverlappingReads.scala
```diff
@@ -0,0 +1,134 @@
...
+    |If the input BAM is neither queryname sorted nor query-grouped, it will be sorted into queryname
+    |order so that clipping can be performed on both ends of a pair simultaneously and so that mate
+    |pair information can be reset across all reads for the template.  Post-clipping the reads are
+    |resorted into coordinate order, any existing NM, UQ and MD tags are repaired, and the output is
+    |written in coordinate order.
+  """)
+class ClipOverlappingReads
+( @arg(flag="i", doc="Input SAM or BAM file of aligned reads in coordinate order.") val input: PathToBam,
+  @arg(flag="o", doc="Output SAM or BAM file.") val output: PathToBam,
+  @arg(flag="s", doc="Soft clip reads instead of hard clipping.") val softClip: Boolean = false,
```
nh13: I am curious as to why the default is false.

src/main/scala/com/fulcrumgenomics/bam/ClipOverlappingReads.scala
```diff
@@ -0,0 +1,134 @@
...
+        case (Some(r1), Some(r2)) =>
+          clip(r1, r2)
+          SamPairUtil.setMateInfo(r1, r2, true)
+          template.r1Supplementals.foreach(s => SamPairUtil.setMateInformationOnSupplementalAlignment(s, r2, true))
+          template.r2Supplementals.foreach(s => SamPairUtil.setMateInformationOnSupplementalAlignment(s, r1, true))
+        case _ => Unit
+      }
+
+      // TODO: add an allReads iterator to Template
+      (template.r1.iterator ++ template.r2.iterator ++ template.allSupplementaryAndSecondary) foreach { r =>
```
nh13: Period before foreach?  Mainly to match the style of foreach above.

src/main/scala/com/fulcrumgenomics/bam/ClipOverlappingReads.scala
```diff
@@ -0,0 +1,134 @@
...
+        sorter.add(r)
+        progress.record(r)
+      }
+    }
+
+    // Then go through the coordinate sorted reads and fix up tags
+    logger.info("Resorting into coordinate order and writing output.")
```
nh13: Re-sorting.

src/test/scala/com/fulcrumgenomics/bam/ClipOverlappingReadsTest.scala
```diff
@@ -0,0 +1,186 @@
...
+      val Seq(nm, uq, md) = Seq("NM", "UQ", "MD").map(r.getAttribute)
+      SequenceUtil.calculateMdAndNmTags(r, chr1, true, true)
+      r.setAttribute(SAMTag.UQ.name, SequenceUtil.sumQualitiesOfMismatches(r, chr1, 0))
+      val Seq(expNm, expUq, expMd) = Seq("NM", "UQ", "MD").map(r.getAttribute)
+
+      nm shouldBe expNm
+      uq shouldBe expUq
+      md shouldBe expMd
+    }
+
```
nh13: I found one!

src/main/scala/com/fulcrumgenomics/bam/ClipOverlappingReads.scala
tfenne: Because I don't trust variant callers not to try and un-clip reads and re-align them, undoing all our hard work.  GATK definitely does this, and I believe VarDict also does.

PR #153, opened 2017-01-09
Added handling for the case where one or both reads gets unmapped due…

tfenne:
… to being fully-clipped.

### DIFF
--- a/src/test/scala/com/fulcrumgenomics/bam/ClipOverlappingReadsTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/bam/ClipOverlappingReadsTest.scala
@@ -27,11 +27,9 @@ package com.fulcrumgenomics.bam
 import com.fulcrumgenomics.fasta.ReferenceSetBuilder
 import com.fulcrumgenomics.testing.SamRecordSetBuilder._
 import com.fulcrumgenomics.testing.{SamRecordSetBuilder, UnitSpec}
-import htsjdk.samtools.{SAMTag, SamReaderFactory}
-import htsjdk.samtools.SAMTag.{MD, NM}
 import htsjdk.samtools.reference.ReferenceSequenceFileFactory
-
 import htsjdk.samtools.util.SequenceUtil
+import htsjdk.samtools.{SAMTag, SamReaderFactory}
 
 import scala.collection.JavaConversions.iterableAsScalaIterable
 import scala.util.Random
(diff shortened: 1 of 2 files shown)


src/test/scala/com/fulcrumgenomics/bam/ClipOverlappingReadsTest.scala
```diff
@@ -129,6 +127,21 @@ class ClipOverlappingReadsTest extends UnitSpec {
...
+    val builder = new SamRecordSetBuilder(readLength=50)
+    val clipper = new ClipOverlappingReads(input=dummyBam, output=dummyBam, ref=ref)
+    val Seq(r1a, r2a) = builder.addPair(name="q1", start1=100, start2=100, cigar1="3M47S", cigar2="48S2M")
+    val Seq(r1b, r2b) = builder.addPair(name="q2", start1=100, start2=100, cigar1="2M48S", cigar2="47S3M")
+
+    for ((r1, r2) <- Seq((r1a, r2a), (r1b, r2b))) {
+      clipper.clip(r1, r2)
+      val ok = r1.getReadUnmappedFlag || r2.getReadUnmappedFlag || r1.getAlignmentEnd < r2.getAlignmentStart
+      ok shouldBe true
+    }
```
nh13: Could you make sure that one of them always is unmapped?  I understand how line 140 will work currently, but just to make sure, we could have `Seq(r1a, r2a, r1b, r2b).count(_.getReadUnmappedFlag) shouldBe 2`.

nh13:
@tfenne one small request in the test, and after adding the change, feel free to merge.

src/test/scala/com/fulcrumgenomics/bam/ClipOverlappingReadsTest.scala
tfenne: That's not actually the case though, and so I prefer the general test that they are either non-overlapping or at least one read got unmapped (hence not overlapping).

The current algorithm sizes the overlap (2 in this case) then attempts to clip half that from each read.  In this case, since the read with `3M` extends 1-base past the end of the other read, that's not sufficient.  So it again looks to clip 1 more base, and allocates the extra base to the R strand read, so ends up unmapping the R strand read in the first case (r1a, r2a), but simply shortening the R strand read in the second case.

I think it's better to test the desired outcome (either one gets unmapped or the overlap is clipped off) than testing specifics of the existing algorithm (which could potentially be improved).

nh13: The test states that there will be a read that is unmapped.  If the records get incorrectly modified by the clipping utility such that `r1a.end < r2a.start` and `r1b.end < r2b.start` then this test would still pass!

PR #154, opened 2017-01-09
Fixed silly requirement to supply _both_ mutex tag arguments.

tfenne:


tfenne:
@nh13 I'm merging this as it's broken in it's current state, but feel free to review post-facto and I'll address any concerns.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -100,6 +100,7 @@ class ExtractUmisFromBam
 
   // This can be removed once the @deprecated molecularBarcodeTags is removed
   private val perIndexTags = if (molecularIndexTags.nonEmpty) molecularIndexTags else molecularBarcodeTags
+  if (perIndexTags.isEmpty) invalid("At least one molecular-index-tag must be specified.")
```
nh13: This seems odd to me.  I would have thought that I could not specify both  `molecularIndexTags` and `molecularBarcodeTags` but instead use `singleTag`.  In this case, `perIndexTags` would be empty.

PR #155, opened 2017-01-09
Propogated the ID field into the two sample mixture VCF.

tfenne:
@nh13 Two one-line changes.  Hopefully nothing controversial!


nh13:
Interesting... 👍

PR #156, opened 2017-01-13
Added automatic clipping of extended attrs when hard clipping.

tfenne:


tfenne:
@nh13 I may still tweak this a little and/or add more tests, but this is most of it if you wanted to take a look.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -48,6 +48,13 @@ import scala.collection.mutable.ListBuffer
 trait SamRecordClipper {
   import SamRecordClipper._
 
+  /**
+    * Should extended attributes that are the same length as the bases and qualities
+    * be automatically clipped in the same way as the bases and qualities.
+    * @return
+    */
+  def autoClipAttributes: Boolean
```
nh13: I am now thinking that `ClippingMode` should also be a `def clippingMode: ClippingMode` just like this, since when are you going to want different clipping modes?  Or am I wrong about that?

nh13: Looking at a few of the tools, it looks like `mode` is always either hard-clip or soft-clip, so why not the same treatment?

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
```diff
@@ -281,13 +302,34 @@ class SamRecordClipperTest extends UnitSpec {
...
+        println("Hello")
+      }
+      clipper.clipEndOfAlignment(rec, 5, mode=mode)
+      rec.getAttribute("A1") shouldBe { if (cut) "ABABABABABABABA" else "AB"*10 }
+      rec.getAttribute("A2") shouldBe Range.inclusive(1, if (cut) 15 else 20).toArray
+      rec.getAttribute("B1") shouldBe  "A" * 10
+      rec.getAttribute("B2") shouldBe Range.inclusive(1, 10).toArray
+    }
+  }
+
```
nh13: extra nl

tfenne:
@nh13 Good call on the handling of mode vs. autoClip.  I've refactored the trait+object into a more traditional class that just takes both as constructor parameters, and removed the mode parameter from each method.  Changes the tests was a bit of a chore, but I think this is better overall.  Want to take one last quick look?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -27,12 +27,32 @@ package com.fulcrumgenomics.bam
...
+  type ClippingMode = ClippingMode.Value
+
+  /** The set of tags that should be invalidated if a read undergoes clipping. */
+  val TagsToInvalidate = Seq("MD", "NM", "UQ")
+
+  private val NoCallBase = 'N'.toByte
+  private val NoCallQual = 2.toByte
+}
+
+
```
tfenne: I know there are two newlines here ... but I like the separation between the object and the class 😛

tfenne:
Forgot to re-assign this to you @nh13 when I pushed the review feedback.  It's ready whenever you are for a second pass.

src/main/scala/com/fulcrumgenomics/bam/ClipOverlappingReads.scala
```diff
@@ -67,8 +67,7 @@ class ClipOverlappingReads
   Io.assertReadable(ref)
   Io.assertCanWriteFile(output)
 
-  private val clippingMode = if (softClip) ClippingMode.Soft else ClippingMode.Hard
-  private val clipper = new SamRecordClipper { override val autoClipAttributes = ClipOverlappingReads.this.autoClipAttributes }
+  private val clipper = new SamRecordClipper(mode=if (softClip) ClippingMode.Soft else ClippingMode.Hard, autoClipAttributes=autoClipAttributes)
```
nh13: You could just add an apply method takes a boolean for `softClip`

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -44,17 +64,16 @@ import scala.collection.mutable.ListBuffer
   *     provided (clip[Start|End|5PrimeEnd|3PrimeEnd]OfAlignment) and ones that
   *     attempt to make it so that the read has N bases clipped, including any existing
   *     clipping (clip[Start|End|5PrimeEnd|3PrimeEnd]OfRead)
+  *
+  * @param autoClipAttributes if true attributes that are the same length as the bases
+  *                           and qualities be automatically clipped in the same way as
+  *                           the bases and qualities, otherwise attributes are not
+  *                           touched.
+  * @param mode how should clipping be performed (hard, soft, soft with masking)
```
nh13: can you re-order the scaladoc?

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -70,8 +70,7 @@ class TrimPrimers
   @arg(flag="t", doc="Temporary directory to use when sorting.") val tmp: Option[DirPath] = None,
   @arg(flag="a", doc="Automatically trim extended attributes that are the same length as bases.") val autoTrimAttributes: Boolean = false
 )extends FgBioTool with LazyLogging {
-  private val mode = if (hardClip) ClippingMode.Hard else ClippingMode.Soft
-  private val clipper = new SamRecordClipper { override val autoClipAttributes = autoTrimAttributes }
+  private val clipper = new SamRecordClipper(mode=if (hardClip) ClippingMode.Hard else ClippingMode.Soft, autoClipAttributes=autoTrimAttributes)
```
nh13: Could use that new `apply` method here too!

src/main/scala/com/fulcrumgenomics/bam/ClipOverlappingReads.scala
tfenne: I'd prefer not to do that because a) there are three clipping modes, and b) in the two tools where I use this I have different defaults, and hence different booleans (one is `softClip` the other is `hardClip`).

PR #157, opened 2017-01-16
Add a tool to compute insert size for RNA-Seq experiments.

nh13:
@tfenne I am looking for an early review before writing tests for how I am computing insert size in transcript space.  I went a little nuts with `@tailrec` methods, and decided to punt on handling overlapping reads (just use `ClipOverlappingReads`).


src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+      fail(s"Record overlaps with its mate, consider using ClipOverlappingReads.  Record: $rec")
+    }
+
+    // Get the overlapping genes
+    val leftMostBase     = Math.min(rec.getAlignmentStart, rec.getMateAlignmentStart)
+    val rightMostBase    = Math.max(rec.getAlignmentEnd, mateAlignmentEnd)
+    val recInterval      = new Interval(rec.getReferenceName, leftMostBase, rightMostBase)
+    val overlappingGenes = overlapDetector.getOverlaps(recInterval)
+
+    if (overlappingGenes.size() == 1) {
```
nh13: I could enforce the read is enclosed in the gene.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+      // calculate the # of bases overlapping the lower read
+      val lower = numReadBasesOverlappingTranscript(lowerReadAlignmentBlocks.toIterator.bufferBetter, exons)
+
+      // calculate the # of bases overlapping the upper read
+      val middle = {
+        val start = upperReadAlignmentBlocks.head.getReferenceStart // don't move past the start of the upper blocks
+        exons.takeWhile(_.end < start).map { e => e.end - e.start + 1 }.sum
+      }
+
+      // calculate the # of bases in exons in-between
```
nh13: comments here and 171 swapped.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+  """Computes the insert size for RNA-Seq experiments.
+    |
+    |Computes the insert size be counting the # of bases sequenced in transcript space.  Skips reads that overlap
+    |multiple genes, and reads where the insert size would disagree across transcripts from the same gene.
+    |
+    |Also skips unpaired, failed QC, secondary, supplementary, pairs without both ends mapped, duplicates, and pairs whose
+    |reads map to different chromosomes.
+    |
+    |This tool requires each mapped pair to have the mate cigar (MC) tag and for read pairs not to overlap.  Use Picard's
+    |RevertOriginalBaseQualitiesAndAddMateCigar to add the former and Fgbio's ClipOverlappingReads to ensure the latter.
```
tfenne: `fgbio SetMateInformation` will  set the mate cigar.  And in fact, `ClipOverlappingReads` will generate `MC` too   as it goes through and does it's work.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+  """,
+  group = ClpGroups.SamOrBam)
+class ComputeRnaSeqInsertSize
```
tfenne: We should just decide if we want all our metrics programs (going forward) to start with "Compute" before we merge this.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+(@arg(flag="i", doc="Input BAM file.") val input: PathToBam,
+ @arg(flag="r", doc="Input gene annotations in refFlat form.  Format described here: http://genome.ucsc.edu/goldenPath/gbdDescriptionsOld.html#RefFlat")
```
tfenne: How hard would it be to support refFlat _or_ GTF?

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
tfenne: Thinking out loud, I think _maybe_ you can get away without the MC tag in some cases.  If we accept for a minute that we're ok with not handling `RR` mapped reads and that `FF` mapped reads are representative of all tandem reads (which I think is reasonable) you could just calculate the insert size from the `R` read for each pair, where you have the cigar string to get the 5' position, and use the mate alignment position as the 5' end of the other read (since it's `F` and you know alignment start == 5' end).

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+
+    val metrics = PairOrientation.values().toList.map { pairOrientation =>
+      val counter = counters(pairOrientation)
+      val mean    = counter.mean()
+      InsertSizeMetric(
+        pairOrientation    = pairOrientation,
+        read_pairs         = counter.total,
+        mean               = mean,
+        median             = counter.median(),
+        min                = counter.map(_._2).min,
```
tfenne: Can I have `median_absolute_deviation` as well please?

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+
+    Metric.write(metrics=metrics, path=output)
+  }
+
+  /** Calculates the insert size in transcript space. */
+  private def calculateInsertSize(rec: SAMRecord, overlapDetector: OverlapDetector[Gene]): Option[Int] = {
+    if (!rec.getReadPairedFlag || rec.getReadFailsVendorQualityCheckFlag || rec.isSecondaryOrSupplementary) return None
+    else if (rec.getReadUnmappedFlag || rec.getMateUnmappedFlag) return None
+    else if (rec.getDuplicateReadFlag || rec.getReferenceIndex != rec.getMateReferenceIndex) return None
+    else if (rec.getSecondOfPairFlag) return None
```
tfenne: A couple of thoughts:
1) You might want to just use a `FilteringSamIterator` to do a bunch of this for you and not have to have guards with returns here.
2) Do you want to also screen out reads with low mapq?

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+
+    val mateCigar        = SAMUtils.getMateCigar(rec)
+    require(mateCigar != null, s"Mate CIGAR (Tag MC) not found for $rec.")
+    val mateAlignmentEnd = CoordMath.getEnd(rec.getMateAlignmentStart, mateCigar.getReferenceLength)
+
+    // Check for overlapping reads
+    if (CoordMath.overlaps(rec.getAlignmentStart, rec.getAlignmentEnd, rec.getMateAlignmentStart, mateAlignmentEnd)) {
```
tfenne: I don't understand why reads that overlap are a problem.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
tfenne: I think that would be a good thing to check/enforce.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+      fail(s"Record overlaps with its mate, consider using ClipOverlappingReads.  Record: $rec")
+    }
+
+    // Get the overlapping genes
+    val leftMostBase     = Math.min(rec.getAlignmentStart, rec.getMateAlignmentStart)
```
tfenne: I think we should discuss what the right pair of coordinates to use are.  My initial reaction is that I'd prefer to stick with the Picard/htsjdk style of using the 5' coordinate of each read, and the distance between them as the insert size.  My take is that:

1. Definitely makes the most sense for `FR` reads
2. Is the most predictable/interpretable thing to do for `RF` reads
3. Makes as much sense as anything else for `TANDEM` reads
4. Side-steps any issues with improperly clipped ends of reads

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+    }
+  }
+
+  /**
+    * Calculates the insert size for all remaining transcripts ensuring that the results are the same.  If a transcript
+    * is encountered that yields an insert size different than previously calculated, computation is stopped, and None
+    * is returned.
+    */
+  @tailrec
+  private def calculateInsertSizeFromRemainingTranscripts(gene: Gene,
```
tfenne: `gene` isn't used except in the recursive call; can be eliminated.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+    if (remainingTranscripts.isEmpty) { // no more transcripts, return the result
+      result
+    }
+    else {
+      // calculate the result for this transcript
+      val transcript = remainingTranscripts.next()
+      val exons      = transcript.toIterator.bufferBetter
+
+      // calculate the # of bases overlapping the lower read
+      val lower = numReadBasesOverlappingTranscript(lowerReadAlignmentBlocks.toIterator.bufferBetter, exons)
```
tfenne: I was thinking this whole calculation could be a little simpler.  My thinking is that you could do:
1) Take the 5' end position of both reads in the pair
2) Calculate the transcript coordinate of the two 5' ends
3) Calculate the length in transcript space as the distance between the two ends on the transcript

Admittedly that doesn't account for indels in the reads themselves, but generally speaking other insert size tools don't either, I think.

tfenne: Looking at your code I realize now that this is a bit of an oversimplification as it only works for reads that are from mature (i.e. post-splicing) transcripts.  Still, I think this is probably a good route but with a check up front that:
a) Both 5' ends fall within exons
b) Some very large fraction of bases in the read (95%?) are overlapped with exons; I'd suggest allowing a few bases to not be, in case the mapper gets it wrong around splice junctions and leaves a couple of dangling bases.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,330 @@
...
+
+      // calculate the # of bases overlapping the upper read
+      val middle = {
+        val start = upperReadAlignmentBlocks.head.getReferenceStart // don't move past the start of the upper blocks
+        exons.takeWhile(_.end < start).map { e => e.end - e.start + 1 }.sum
```
tfenne: This confused me because it wasn't clear until I got there, that `numReadBasesOverlappingTranscript` consumed out of the `exons` iterator and left it in the right place.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
nh13: I can change it to `EstimateRnaSeqInsertSize` as currently we have no convention on tools.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
nh13: I started going down the road of writing a GTF->refFlat parser, but it was quite tedious with all the book keeping.  It's not needed so I will defer, but if you have a case for needing it, I can share my unfinished code.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
nh13: We are counting the # of aligned bases on *both* ends below.  The calculation below doesn't handle  the case where they overlap.

nh13:
@tfenne want to take another look before I write tests and ask for a full review?

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,362 @@
...
+    |Computes the insert size by counting the # of bases sequenced in transcript space.  The insert size is defined
+    |as the distance between the first bases sequenced of each pair respectively (5' sequencing ends).
+    |
+    |This tool skips reads that overlap multiple genes, reads that aren't fully enclosed in a gene, and reads where the
+    |insert size would disagree across transcripts from the same gene.  Also skips reads that are unpaired, failed QC,
+    |secondary, supplementary, pairs without both ends mapped, duplicates, and pairs whose reads map to different
+    |chromosomes. Finally, skips transcripts where the too few mapped read bases overlap exonic sequence.
+    |
+    |This tool requires each mapped pair to have the mate cigar (MC) tag and for read pairs not to overlap.  Use
+    |SetMateInformation to add the former or Fgbio's ClipOverlappingReads to add both.
```
tfenne: Some words (also doesn't it now support overlapping reads?

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
tfenne: Fair enough.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,362 @@
...
+  """,
+  group = ClpGroups.SamOrBam)
+class ComputeRnaSeqInsertSize
+(@arg(flag="i", doc="Input BAM file.") val input: PathToBam,
+ @arg(flag="r", doc="Input gene annotations in refFlat form.  Format described here: http://genome.ucsc.edu/goldenPath/gbdDescriptionsOld.html#RefFlat")
+ val refFlat: FilePath,
+ @arg(flag="o", doc="Output metrics file to which to write.") val output: FilePath,
```
tfenne: Perhaps make this a prefix, so if we want to spit out the full distributions and plot, we can still do that without changing the parameter.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,362 @@
...
+ @arg(flag="d", doc="Include duplicates") val includeDuplicates: Boolean = false,
+ @arg(flag="D", doc= "This is done because insert size data typically includes enough anomalous values from chimeras" +
```
tfenne: This == what?  😉   Also, I'm curious if the filtering is still necessary with all the transcript filtering going on.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,362 @@
...
+    val mateCigar        = SAMUtils.getMateCigar(rec)
+    require(mateCigar != null, s"Mate CIGAR (Tag MC) not found for $rec, consider using SetMateInformation or ClipOverlappingReads.")
+    val mateAlignmentEnd = CoordMath.getEnd(rec.getMateAlignmentStart, mateCigar.getReferenceLength)
+
+    // Get the overlapping genes
+    val leftMostBase     = Math.min(rec.getAlignmentStart, rec.getMateAlignmentStart)
+    val rightMostBase    = Math.max(rec.getAlignmentEnd, mateAlignmentEnd)
+    val recInterval      = new Interval(rec.getReferenceName, leftMostBase, rightMostBase)
+    val overlappingGenes = overlapDetector.getOverlaps(recInterval)
+    val firstGene        = overlappingGenes.headOption
```
tfenne: I don't think this helps you.  In the `if` below you can just `&& overlappingGenes.forall`, and on line 146 you have to unpack the option anway, so might as well be `overlappingGenes.head`.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,362 @@
...
+      val gene                    = firstGene.get
+      val transcripts             = gene.iterator
+      var insertSize: Option[Int] = None
+      var sameInsertSize: Boolean = true
+      while (transcripts.hasNext && sameInsertSize) {
+        val transcript = transcripts.next()
+
+        // Ignore transcripts with too little overlap
+        val overlap    = numReadBasesOverlappingTranscript(recBlocks, transcript) + numReadBasesOverlappingTranscript(mateBlocks, transcript)
+        val mappedBases = recBlocks.map(_.getLength).sum + mateBlocks.map(_.getLength).sum
```
tfenne: This line could be outside the `while` so you only compute it once.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
nh13: Fixed.

src/main/scala/com/fulcrumgenomics/bam/ComputeRnaSeqInsertSize.scala
nh13: Also making it optional, and use the input if not given (sans the input's file extension)

nh13:
@tfenne OK, ready for review, thanks!

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+      val mappedBases             = recBlocks.map(_.getLength).sum + mateBlocks.map(_.getLength).sum
+      val gene                    = overlappingGenes.head
+      val transcripts             = gene.iterator
+      var insertSize: Option[Int] = None
+      var sameInsertSize: Boolean = true
+      while (transcripts.hasNext && sameInsertSize) {
+        val transcript = transcripts.next()
+
+        // Ignore transcripts with too little overlap
+        val overlap    = numReadBasesOverlappingTranscript(recBlocks, transcript) + numReadBasesOverlappingTranscript(mateBlocks, transcript)
```
tfenne: You could write
```scala
Seq(recBlocks, mateBlocks).map(numReadBasesOverlappingTranscript(_, transcript).sum
```

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+        if (overlap / mappedBases.toDouble >= minimumOverlap) {
+          // get the 5' position of the record and its mate
+          val rec5Prime   = if (rec.getReadNegativeStrandFlag) rec.getAlignmentEnd else rec.getAlignmentStart
+          val mate5Prime  = if (rec.getMateNegativeStrandFlag) mateAlignmentEnd    else rec.getMateAlignmentStart
+
+          // calculate the distance in transcript space
+          val lower5Prime = Math.min(rec5Prime, mate5Prime)
+          val upper5Prime = Math.max(rec5Prime, mate5Prime)
+          val transcriptInsertSize = transcript.filter { exon => CoordMath.overlaps(lower5Prime, upper5Prime, exon.start, exon.end) }
+            .map { exon =>
```
tfenne: Could this `map` go all on one line please?

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+
+          // Make sure that the insert size is the same as any previous result
+          insertSize match {
+            case None        => insertSize = Some(transcriptInsertSize)
+            case Some(isize) => sameInsertSize = isize == transcriptInsertSize
+          }
+        }
+      }
+
+      if (sameInsertSize) {
```
tfenne: Could these few lines be collapsed down into:
```
      if (sameInsertSize) insertSize
      }
    None
```
to remove both dangling elses?

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+  private def numReadBasesOverlappingTranscript(alignmentBlocks: List[AlignmentBlock], transcript: Transcript): Int = {
+    val blocks  = alignmentBlocks.iterator.bufferBetter
+    val exons   = transcript.iterator.bufferBetter
+    var overlap = 0
+
+    while (blocks.hasNext && exons.hasNext) {
+      val block               = blocks.head
+      val exon                = exons.head
+      val blockReferenceStart = block.getReferenceStart
+      val blockReferenceEnd   = blockReferenceStart + block.getLength
```
tfenne: I'm fairly sure this should have a `-1` on the end.

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+    }
+    else {
+      None
+    }
+  }
+
+  /** Gets the number of read bases that overlap the exons. */
+  private def numReadBasesOverlappingTranscript(alignmentBlocks: List[AlignmentBlock], transcript: Transcript): Int = {
+    val blocks  = alignmentBlocks.iterator.bufferBetter
+    val exons   = transcript.iterator.bufferBetter
```
tfenne: There's an in build assumption here that the transcript always iterates exons in genomic order (as opposed to say transcript order).  I wonder if it wouldn't be better to not have an `iterator` method, but instead have two methods that provide access in genomic and transcript order and make it explicit, e.g. `transcript.exonsInGenomeOrder` and `transcript.exonsInTranscriptOrder` or similar.

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+        exons.next()
+      }
+    }
+
+    overlap
+  }
+}
+
+object EstimateRnaSeqInsertSize {
+
```
tfenne: Bingo!

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+    new FailsVendorReadQualityFilter,
+    new AlignedFilter(true),
+    new SecondaryOrSupplementaryFilter,
+    new MappingQualityFilter(minimumMappingQuality),
+    new DuplicatesFilter(includeDuplicates=includeDuplicates),
+    new FirstOfPairOnlyFilter,
+    new DifferentReferenceIndexFilter
+  )
+
+  trait SingleEndSamRecordFilter extends SamRecordFilter {
```
tfenne: If you wanted to make this more compact I think you could do something like this:
```scala
object SamRecordFilters {
  private trait SeFilter extends SamRecordFilter { override def filterOut(r1: SAMRecord, r2: SAMRecord) = filterOut(r1) || filterOut(r2) } 
  def filter(f: SAMRecord => Boolean): SamRecordFilter = new SeFilter { 
    override def filterOut(r: SAMRecord): Boolean = f(r)
  }

  def ReadPairedFilter = filter(!_.getReadPairedFlat)
  def MateMappedFilter = filter(r => !r.getReadPairedFlag || r.getMateUnmappedFlag)
  ...
}
```

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+  private class FirstOfPairOnlyFilter extends SingleEndSamRecordFilter {
+    override def filterOut(record: SAMRecord): Boolean = !record.getFirstOfPairFlag
+  }
+
+  private class DifferentReferenceIndexFilter extends SingleEndSamRecordFilter {
+    override def filterOut(record: SAMRecord): Boolean = record.getReferenceIndex != record.getMateReferenceIndex
+  }
+}
+
+case class InsertSizeMetric(pairOrientation: PairOrientation,
```
tfenne: `pair_orientation`?

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+                            read_pairs: Long = 0,
+                            mean: Double = 0,
+                            median: Double = 0,
+                            min: Long = 0,
+                            max: Long = 0,
+                            standard_deviation: Double = 0,
```
tfenne: could sd and mad go next to mean and median respectively please

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -0,0 +1,52 @@
...
+object GeneAnnotations {
+
+  /** A gene with zero or more transcripts, that all are on the same transcription strand. */
+  case class Gene(contig: String, start: Int, end: Int, negativeStrand: Boolean, name: String, transcripts: Seq[Transcript])
+    extends Interval(contig, start, end, negativeStrand, name) with Iterable[Transcript] {
+    def iterator: Iterator[Transcript] = transcripts.toIterator
+  }
+
+  /** A transcript associated with a given gene (which stores the transcription strand).  Contains zero or more exons. */
+  case class Transcript(name: String, txStart: Int, txEnd: Int, cdsStart: Int, cdsEnd: Int, exons: Seq[Exon]) extends Iterable[Exon] {
```
tfenne: I'd vote for calling `txStart` and `txEnd` just `start` and `end`.

tfenne: Also, see earlier comment about providing two exon iterators to make it clear what the ordering is.

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -0,0 +1,52 @@
...
+    if (exons.length > 1) {
+      val exonsOverlap = exons.sliding(2).map { e => (e.head, e.last) }.exists { case (e1, e2) => CoordMath.overlaps(e1.start, e1.end, e2.start, e2.end) }
+      require(!exonsOverlap, s"exons overlap for transcript: $name")
+    }
+    def iterator: Iterator[Exon] = exons.toIterator
+  }
+
+  /** Defines an exonic sequence within a transcript. */
+  case class Exon(start: Int, end: Int) {
```
tfenne: I sort of wonder whether Transcript and Exon should be inner classes of gene since they lack things like a reference to Gene and the contig on which they are mapped?

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,154 @@
...
+  /** Creates a new refflat source from a sequence of lines. */
+  def apply(lines: Seq[String], dict: Option[SAMSequenceDictionary]): RefFlatSource = new RefFlatSource(lines.iterator, dict)
+
+  /** Creates a new fasrefflattq source from an iterator of lines. */
+  def apply(lines: Iterator[String], dict: Option[SAMSequenceDictionary]): RefFlatSource = new RefFlatSource(lines, dict)
+
+  /** Creates a new refflat source from an input stream. */
+  def apply(stream: InputStream, dict: Option[SAMSequenceDictionary]): RefFlatSource = new RefFlatSource(Source.fromInputStream(stream).getLines(), dict, Some(stream))
+
+  /** Creates a new refflat source from an input stream. */
```
tfenne: Comment copy paste error.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,154 @@
...
+  def apply(source: Source, dict: Option[SAMSequenceDictionary]): RefFlatSource = new RefFlatSource(source.getLines(), dict, Some(source))
+
+  /** Creates a new refflat source from a File. */
+  def apply(file: File, dict: Option[SAMSequenceDictionary]): RefFlatSource = apply(path=file.toPath, dict=dict)
```
tfenne: So many applies!  I might kill this one as we should always try to use `Path` vs. `File`.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,154 @@
...
+
+  /** Creates a new refflat source from a Path. */
+  def apply(path: PathToFastq, dict: Option[SAMSequenceDictionary]): RefFlatSource = apply(Io.toInputStream(path), dict)
```
tfenne: Not a fastq.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,154 @@
...
+
+  private val Header: String = Seq(
```
tfenne: The UCSC column names are all like `geneName` vs. `GENE_NAME`.  Is this for Picard compatibility?

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    // Each line is a gene and transcript.  We want to group all transcripts from the same gene.
+    new DelimitedDataParser(lines=_lines, delimiter='\t').flatMap { row =>
+      val geneName       = row[String]("GENE_NAME")
+      val transcriptName = row[String]("TRANSCRIPT_NAME")
+      val contig         = row[String]("CHROMOSOME")
+      val strand         = row[String]("STRAND")
+      val exonCount      = row[Int]("EXON_COUNT")
+      val exonStarts     = row[String]("EXON_STARTS").split(",").filter(_.nonEmpty).map(_.toInt)
```
tfenne: If only `row[Array[Int]]` worked!

tfenne: Scala has a more efficient `split(Char)` you could use here.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,154 @@
...
+        None
+      }
+      else {
+        require(exonStarts.length == exonCount, s"Number of EXON_STARTS does not equal EXON_COUNT for $geneName/$transcriptName")
+        require(exonEnds.length == exonCount, s"Number of EXON_ENDS does not equal EXON_COUNT for $geneName/$transcriptName")
+
+        // Convert from 0-based half-open to 1-based inclusive intervals.
+        val transcript = Transcript(
+          name     = transcriptName,
+          txStart  = row[String]("TX_START").toInt + 1,
```
tfenne: row[Int]?

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,154 @@
...
+          txEnd    = row[String]("TX_END").toInt,
+          cdsStart = row[String]("CDS_START").toInt + 1,
+          cdsEnd   = row[String]("CDS_END").toInt,
+          exons    = exonStarts.zip(exonEnds).map { case (s, e) => Exon(start=s.toInt + 1, end=e.toInt) }
```
tfenne: Some days I just really love scala.  This is pretty sweet.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,154 @@
...
+
+        Some(gene)
+      }
+    }
+      .toSeq
+      .groupBy(_.name)
+      .map { case (name, _genes) =>
+        val contig         = _genes.head.contig
+        val negativeStrand = _genes.head.negativeStrand
+        require(_genes.forall(_.contig == contig), s"Chromosomes differ for gene $name")
```
tfenne: This is going to hurt.  It's not unusual for gene databases at UCSC to have mappings to multiple chromosomes for a small fraction of genes.  If you throw an exception when you see this, you're going to have a bad time trying to clean up your refFlat data.

I'd suggest that instead, you either filter out those genes, or you only provide a single set of mappings, or you have a flag/def that tells you whether there are multiple contig mappings and return everything, so that clients can decide what to do about it.

src/test/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSizeTest.scala
```diff
@@ -0,0 +1,192 @@
...
+import com.fulcrumgenomics.util.{Io, Metric}
+import com.fulcrumgenomics.testing.{SamRecordSetBuilder, UnitSpec}
+import com.fulcrumgenomics.testing.SamRecordSetBuilder._
+import dagr.commons.io.PathUtil
+import htsjdk.samtools.SamPairUtil.PairOrientation
+
+class EstimateRnaSeqInsertSizeTest extends UnitSpec {
+
+  private val RefFlatFile = {
+    val lines = Seq(
```
tfenne: Could you at least make sure that in your test data here that the transcript IDs `NM_*` are unique please?

src/test/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSizeTest.scala
```diff
@@ -0,0 +1,192 @@
...
+      // a run-of-the-mill gene, with one transcript and one exon
+      Seq("ACKR4-3", "NM_178445", "chr3", "+", "133801670", "133804175", "133801931", "133802984", "1", "133801670", "133804175"),
+      // two genes that overlap
```
tfenne: FWIW it's more common for this to be on opposite strands like:
```
gene 1: --->        --->        --->
gene 2:                  <----       <--   <-----
```

src/test/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSizeTest.scala
```diff
@@ -0,0 +1,192 @@
...
+    {
+      new EstimateRnaSeqInsertSize(input=bam, refFlat=RefFlatFile, minimumOverlap=3/200.0).execute()
+      val metrics = Metric.read[InsertSizeMetric](path=out)
+      metrics.length shouldBe PairOrientation.values().length
+      metrics.zip(EmptyMetrics).foreach { case (actual, expected) =>
+        actual shouldBe expected
+      }
+    }
+  }
+}
```
tfenne: I would feel better if there were a few direct test cases for the calculation of the insert size given a set of genes and a SAMRecord.  I think what you have now probably tests sufficiently well, but is going to be really painful to maintain should we decide to change the algorithm at all.

src/test/scala/com/fulcrumgenomics/util/RefFlatSourceTest.scala
```diff
@@ -0,0 +1,129 @@
...
+
+import java.nio.file.Paths
+
+import com.fulcrumgenomics.testing.UnitSpec
+import com.fulcrumgenomics.util.GeneAnnotations.Exon
+import htsjdk.samtools.{SAMSequenceDictionary, SAMSequenceRecord}
+
+class RefFlatSourceTest extends UnitSpec {
+
+  // Adapted from http://hgdownload.cse.ucsc.edu/goldenPath/hg18/database/refFlat.txt.gz
```
tfenne: Really, the HG18 one?

src/test/scala/com/fulcrumgenomics/util/RefFlatSourceTest.scala
```diff
@@ -0,0 +1,129 @@
...
+  private val RefFlatFile = Paths.get("src/test/resources/com/fulcrumgenomics/util/refFlat.txt.gz")
+
+  private val Dictionary = {
+    import scala.collection.JavaConversions.seqAsJavaList
+    new SAMSequenceDictionary(Seq(new SAMSequenceRecord("chr1", 249250621)))
```
tfenne: You could just use the java `Collections.singletonList()` here.

src/test/scala/com/fulcrumgenomics/util/RefFlatSourceTest.scala
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  "RefFlatSource" should "read valid refFlat from various kinds of input resources" in {
+    val rf = RefFlatFile
+    val lines = Io.readLines(rf)
+
+
```
tfenne: image

src/test/scala/com/fulcrumgenomics/util/RefFlatSourceTest.scala
```diff
@@ -0,0 +1,129 @@
...
+      transcript.cdsEnd shouldBe 141655128
+      transcript.size shouldBe 5
+
+      //verify the first exon in the first transcript
+      val exon = transcript.head
+      exon shouldBe Exon(141638944, 141639396)
+    })
+  }
+
+  it should "fail if the chromosomes differ across transcripts" in {
```
tfenne: Harping on my earlier point, this is not uncommon.  Running the following on the hg38 refFlat file from UCSC:
```bash
gzcat refFlat.txt.gz | cut -f1,3,4 | sort -u | cut -f1 | sort | uniq -c | sort -rn | \
    awk '{if ($1 > 1) dupes +=1; total+=1;} END {print dupes, total, dupes/total}'
```
yeilds:
`1491 27238 0.0547397`

I.e. there are 1491 genes that have more than one `contig,strand` in the file.  Even if you exclude mappings to `alt` and `random` contigs it's still:
`442 27221 0.0162375`

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,267 @@
...
+  Io.assertReadable(input)
+  Io.assertReadable(refFlat)
+  prefix.foreach(Io.assertCanWriteFile(_))
+
+  override def execute(): Unit = {
+    val progress            = new ProgressLogger(logger, verb = "read", unit = 5e6.toInt)
+    val in                  = SamReaderFactory.make.open(input)
+    val refFlatSource       = RefFlatSource(refFlat, Some(in.getFileHeader.getSequenceDictionary))
+    val geneOverlapDetector = new OverlapDetector[Gene](0, 0)
+    val counters            = PairOrientation.values().toList.map { pairOrientation => (pairOrientation, new NumericCounter[Long]()) }.toMap
```
nh13: `map` was showing up before, weird, perhaps I imported a scala conversion later and now it works?

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
nh13: And you'd be right.

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
nh13: Here's your 💵

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
nh13: Fixed in `FastqSource` ;)

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
nh13: Just mirroring `FastqSource` for expediency.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
nh13: You got me.

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
nh13: Actually, it gets to be a pain to create transcripts and exons and insert them correctly, even with "add" methods and the like.  Take a look at how painful the Picard code is for creating the same thing.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
nh13: Yes, Picard uses the latter, which sucks, but it breaks if it doesn't have them that way :/

nh13: I think I can have two recognized headers.

src/test/scala/com/fulcrumgenomics/util/RefFlatSourceTest.scala
nh13: 👍

src/test/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSizeTest.scala
nh13: I totally agree about the tests, I'll try and fix them at a later date as I did some refactoring that should make it easier.

nh13:
@tfenne I made a bunch of changes as you suggested.  The two big things are:
1. I did not make `Transcript` and `Exon` inner classes of `Gene` since it complicates how to create a `Gene`/`Transcript`/`Exon`.  Just look at Picard's treatment.
2. I did not re-write the tests for `EstimateRnaSeqInsertSize` but I did re-factor `EstimateRnaSeqInsertSize` to make it a lot easier for the future.

Before merging this PR, I would like to clean up the `EstimateRnaSeqInsertSize` object by adding documentation and making it a bit cleaner.  After that, I am going to re-factor the tests.  Just putting the changes up here for now.

src/main/scala/com/fulcrumgenomics/util/NumericCounter.scala
```diff
@@ -35,6 +35,13 @@ object NumericCounter {
     ts.foreach(counter.count)
     counter
   }
+
+  /** Generates a counter that has counted all the numerics provided. */
+  def from[T](ts: TraversableOnce[(T, Long)])(implicit numeric: Numeric[T]) : NumericCounter[T] = {
```
tfenne: Let's leave this here for now, but I think ideally this could be replaced/removed it we implement CanBuildFrom.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,210 @@
...
+      }
+    }
+      .toSeq
+      .groupBy(_.name)
+      .map { case (name, _genes) =>
+        val contig         = _genes.head.contig
+        val negativeStrand = _genes.head.negativeStrand
+
+        val transcripts = _genes.flatMap { gene =>
+          require(gene.length() == 1)
```
tfenne: Is this necessary?  It confused the crap out of me, because I initially thought your were asserting that the gene had a single transcript at this point, and that was an odd use of `length()`.  After looking around I see that you're just asserting your two `-1`s above haven't been modified.  Maybe if you want to do that you could assert `start == -1` instead since I think the intent is clearer?

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -0,0 +1,210 @@
...
+          if (gene.contig != contig) {
+            val transcript = gene.head
+            numDifferentChromsomes += 1
```
tfenne: Technically this is `numOnDifferentChromosomes` right?

src/main/scala/com/fulcrumgenomics/bam/EstimateRnaSeqInsertSize.scala
```diff
@@ -0,0 +1,315 @@
...
+
+      metric.copy(mean=filteredMean, standard_deviation=filteredCounter.stddev(m=filteredMean))
+    }
+
+    // Write the metrics
+    val metricPath = PathUtil.pathTo(prefix.getOrElse(PathUtil.removeExtension(input)) + RnaSeqInsertSizeMetricExtension)
+    Metric.write(metrics=metrics, path=metricPath)
+
+    // Write the histogram
+    val histogramPath   = PathUtil.pathTo(prefix.getOrElse(PathUtil.removeExtension(input)) + RnaSeqInsertSizeMetricHistogramExtension)
```
nh13: I mentioned its file extension ;), but I'll add more.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
nh13: Actually, I wanted `gene.length()  == 1`, since at this point we should have one transcript per gene, and now we are aggregating all transcripts for a given gene.

tfenne:
Then don’t you want to check `gene.transcripts.size` instead?

nh13:
Ugh, I forgot `Gene` was an `Iterable[Transcript]`.

PR #158, opened 2017-01-17
FindTechnicalReads: Changed tagging to output all sequences to which reads match.

tfenne:
Added support for providing a name for each read.


nh13:
@tfenne do you want to keep working on this PR?

tfenne:
@nh13 I'm closing this one but will leave the branch around in case I need to come back to it.

PR #159, opened 2017-01-19
New DuplexConsensusCaller and major refactoring of existing single-UMI calling code

tfenne:
@nh13 There a _lot_ going on in this pull request.  There's probably a lot of code that's marked as new or changed in a big way, that is mostly re-organized.

`VanillaUmiConsensusCaller` got broken into `UmiConsensusCaller` (a base trait), `VanillaUmiConsensusCaller` (the single-umi caller) and `ConsensusCallingIterator` that coordinates the flow of records when doing more than just calling a single consensus/pair.

I've also remove the option to emit one half of a pair's consensus if the other fails for some reason.  It's now always all-or-nothing for a template.  Likewise I made some opinionated choices on the duplex caller:
- It always requires one pair of AB and one of BA (i.e. it won't emit single-strand supported reads)
- it doesn't offer a minReads or minConsensusQual parameter, those are to come in a later filtering tool.

I feel like the `VanillaUmiConsensusCaller` tests could do with a little tidying, and probably breaking out the tests for `ConsensusCallingIterator` separately, but I'm going to leave that for another day.


src/test/scala/com/fulcrumgenomics/testing/UnitSpec.scala
```diff
@@ -49,6 +55,15 @@ trait UnitSpec extends FlatSpec with Matchers {
     val in = SamReaderFactory.make().open(bam.toFile)
     yieldAndThen(in.iterator().toIndexedSeq) { in.safelyClose() }
   }
+
+  /** Generates a command line parser for a class to check that the argument annotations are valid. */
+  protected def checkClpAnnotations[T <: FgBioTool](implicit ct: ClassTag[T], tt: TypeTag[T]): Unit = {
```
tfenne: @nh13 I got tired of [me] screwing up CLP stuff and our tests not catching it.  This probably won't catch everything, but it will at least catch cases where we forget the `clp` annotation or forget `@arg` on params that don't have defaults, or re-use an argument name or flag.

Once this is merged I'll make a sweep and add the test to each our CLP tests.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -291,14 +162,14 @@ class VanillaUmiConsensusCaller
     *
     * The same number of base sequences and quality sequences should be given.
     * */
-  private[umi] def consensusCall(reads: Seq[SourceRead]): Option[ConsensusRead] = {
+  private[umi] def consensusCall(reads: Seq[SourceRead]): Option[VanillaConsensusRead] = {
     // check to see if we have enough reads.
     if (reads.size < this.options.minReads) {
```
nh13: Already checked on 141, is this because we want to use this method elsewhere?

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -344,163 +215,32 @@ class VanillaUmiConsensusCaller
...
-        rec.setReadPairedFlag(true)
-        rec.setSecondOfPairFlag(true)
-        rec.setMateUnmappedFlag(true)
-    }
-    rec.setReadBases(read.bases)
-    rec.setBaseQualities(read.quals)
-    rec.setAttribute(SAMTag.RG.name(), readGroupId)
-    rec.setAttribute(options.tag, records.head.getStringAttribute(options.tag))
-
+  override protected def createSamRecord(read: VanillaConsensusRead, readType: ReadType): SAMRecord = {
```
nh13: Why is this method needed?

src/test/scala/com/fulcrumgenomics/testing/UnitSpec.scala
nh13: 👍

nh13: If you're really ambitious, you could instead just search for all CLPs with the CLP annotation and check those directly.  Then we don't need to add a test for each new CLP for this.

src/test/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReadsTest.scala
```diff
@@ -0,0 +1,81 @@
...
+      new CallDuplexConsensusReads(input=in, output=out).execute()
+    }
+  }
+
+  it should "throw an exception if either error rate is set too low" in {
+      val in  = makeTempFile("in.", ".bam")
+      val out = makeTempFile("out.", ".bam")
+    an[Exception] should be thrownBy { new CallDuplexConsensusReads(input=in, output=out, errorRatePreUmi=0.toByte).execute() }
+    an[Exception] should be thrownBy { new CallDuplexConsensusReads(input=in, output=out, errorRatePostUmi=0.toByte).execute() }
+  }
```
nh13: are these three tests overkill?

src/test/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReadsTest.scala
```diff
@@ -0,0 +1,81 @@
...
+    val out = makeTempFile("duplex.", ".bam")
+    new CallDuplexConsensusReads(input=in, output=out, readGroupId="ZZ").execute()
+    val reader = SamReaderFactory.make().open(out)
+    val recs = reader.toSeq
+
+    reader.getFileHeader.getReadGroups should have size 1
+    reader.getFileHeader.getReadGroups.head.getId shouldBe "ZZ"
+    recs should have size 2
+  }
+}
```
nh13: How about a test for when the MI tag is set incorrectly, or maybe that's below.

src/test/scala/com/fulcrumgenomics/umi/DuplexConsensusCallerTest.scala
```diff
@@ -0,0 +1,265 @@
...
+    builder.addFrag(start=1, attrs=Map(MI -> "f/abc")).foreach { r => c.sourceMoleculeId(r) shouldBe "f"   }
+    builder.addFrag(start=1, attrs=Map(MI -> "f///A")).foreach { r => c.sourceMoleculeId(r) shouldBe "f//" }
+  }
+
+  it should "throw an exception if the MI tag doesn't have a /suffix" in {
+    an[Exception] shouldBe thrownBy {
+      val builder = new SamRecordSetBuilder()
+      builder.addFrag(start=1, attrs=Map(MI -> "foo")).foreach { r => c.sourceMoleculeId(r) }
+    }
+  }
```
nh13: Add a test that it partitions into two groups: AB and BA reads.

src/test/scala/com/fulcrumgenomics/umi/DuplexConsensusCallerTest.scala
```diff
@@ -0,0 +1,265 @@
...
+
+  it should "throw an exception if the MI tag is null" in {
+    an[Exception] shouldBe thrownBy {
+      val builder = new SamRecordSetBuilder()
+      builder.addFrag(start=1).foreach { r => c.sourceMoleculeId(r) }
+    }
+  }
+
+  "DuplexConsensusCaller.consensusReadsFromSamRecords" should "not create records from fragments" in {
```
nh13: it puts the fragments in the reject basket.

src/test/scala/com/fulcrumgenomics/umi/DuplexConsensusCallerTest.scala
```diff
@@ -0,0 +1,265 @@
...
+    val builder = new SamRecordSetBuilder(readLength=10)
+    builder.addFrag(start=1, attrs=Map(MI -> "foo/A")).foreach(r => r.setReadString("AAAAAAAAAA"))
+    builder.addFrag(start=1, attrs=Map(MI -> "foo/B")).foreach(r => r.setReadString("AAAAAAAAAA"))
+    val recs = c.consensusReadsFromSamRecords(builder.toSeq)
+    recs shouldBe empty
```
nh13: could also use `recs shouldBe 'empty`; I am not sure what's the difference, but the latter is what I have been using.

src/test/scala/com/fulcrumgenomics/umi/DuplexConsensusCallerTest.scala
```diff
@@ -0,0 +1,265 @@
...
+      import ConsensusTags.PerBase._
+      r1.getSignedShortArrayAttribute(AbRawReadCount)  shouldBe Array[Byte](3,3,2,2,3,3,3,3,3,3)
+      r1.getSignedShortArrayAttribute(AbRawReadErrors) shouldBe Array[Byte](0,0,0,0,1,0,0,0,0,0)
+      r1.getSignedShortArrayAttribute(BaRawReadCount)  shouldBe Array[Byte](2,2,2,2,2,2,2,2,2,2)
+      r1.getSignedShortArrayAttribute(BaRawReadErrors) shouldBe Array[Byte](0,0,0,0,0,0,0,0,0,0)
+      r2.getSignedShortArrayAttribute(AbRawReadCount)  shouldBe Array[Byte](3,3,3,3,3,3,3,3,3,3)
+      r2.getSignedShortArrayAttribute(AbRawReadErrors) shouldBe Array[Byte](0,0,0,0,1,0,0,0,0,0)
+      r2.getSignedShortArrayAttribute(BaRawReadCount)  shouldBe Array[Byte](2,2,2,2,2,2,2,2,2,2)
+      r2.getSignedShortArrayAttribute(BaRawReadErrors) shouldBe Array[Byte](0,0,0,0,0,0,0,0,0,0)
+    }
```
nh13: Mother of God these test must have been painful to write and get right.

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -41,4 +41,7 @@ trait FgBioTool {
 
   /** Generates a new validation exception with the given message. */
   def invalid(message: String) = throw new ValidationException(message)
+
+  /** Generates a validation exception if the test value is false. */
+  def validate(test: Boolean, message: => String) = if (!test) throw new ValidationException(message)
```
tfenne: I'm not sure I see the point.  `invalid` always evaluates it's parameter, where as `validate` only does if the `test` fails.

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -0,0 +1,115 @@
...
+    |from the two single strand consensus reads.
+    |
+    |Because of the nature of duplex sequencing, this tool does not support fragment reads - if found in the
+    |input they are ignored.  Similarly read pairs for which consensus reads cannot be generated for one or
+    |other read (R1 or R2) are omitted from the output.
+    |
+    |Consensus reads have a number of additional optional tags set in the resulting BAM file.  The tag names follow
+    |a pattern where the first letter (a, b or c) denotes that the tag applies to the first single strand consensus (a),
+    |second single-strand consensus (b) or the final duplex consensus (c).  The second letter is intended to capture
+    |the meaning of the tag (e.g. d=depth, m=min depth, e=errors/error-rate) and is upper case for values that are
```
tfenne: It's error rate at the read level, and error count at the base level.

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -0,0 +1,115 @@
...
+  group = ClpGroups.Umi)
+class CallDuplexConsensusReads
+(@arg(flag="i", doc="The input SAM or BAM file.") val input: PathToBam,
+ @arg(flag="o", doc="Output SAM or BAM file to write consensus reads.") val output: PathToBam,
+ @arg(flag="p", doc="The Prefix all consensus read names") val readNamePrefix: Option[String] = None,
+ @arg(flag="R", doc="The new read group ID for all the consensus reads.") val readGroupId: String = "A",
+ @arg(flag="1", doc="The Phred-scaled error rate for an error prior to the UMIs being integrated.") val errorRatePreUmi: PhredScore = DefaultErrorRatePreUmi,
+ @arg(flag="2", doc="The Phred-scaled error rate for an error post the UMIs have been integrated.") val errorRatePostUmi: PhredScore = DefaultErrorRatePostUmi,
+ @arg(flag="m", doc="Ignore bases in raw reads that have Q below this value.") val minInputBaseQuality: PhredScore = DefaultMinInputBaseQuality,
+ @arg(flag="S", doc="The sort order of the output, if None then the same as the input.") val sortOrder: Option[SortOrder] = Some(SortOrder.queryname)
```
tfenne: The thing is that since we unmap the reads, the next logical step is to proceed into mapping.  And for that I think you really want it in queryname, so I'd like to leave that as the default so one can't forget it.  I think queryname is important here a) because I'm not sure MergeBamAlignment will like query-grouped but not sorted data (even though it could work) and b) it's probably not good to send reads through alignment that are approximately coordinate sorted.

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -0,0 +1,115 @@
...
+    // The output file is unmapped, so for now let's clear out the sequence dictionary & PGs
+    val outHeader = UmiConsensusCaller.outputHeader(in.getFileHeader, readGroupId, sortOrder)
+    val out = new SAMFileWriterFactory().makeWriter(outHeader, sortOrder.forall(_ == in.getFileHeader.getSortOrder), output.toFile, null)
+
+    val caller = new DuplexConsensusCaller(
+      readNamePrefix = readNamePrefix.getOrElse(UmiConsensusCaller.makePrefixFromSamHeader(in.getFileHeader)),
+      readGroupId    = readGroupId,
+      minInputBaseQuality = minInputBaseQuality,
+      errorRatePreUmi = errorRatePreUmi,
+      errorRatePostUmi = errorRatePostUmi
```
tfenne: image

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -0,0 +1,90 @@
...
+/**
+  * An iterator that consumes from an incoming queue of SAMRecords and generates consensus
+  * read SAMRecords using the supplied caller.
+  *
+  * @param iterator an iterator of SAMRecords
+  * @param caller   the consensus caller to use to call consensus reads
+  * @param progress an optional progress logger to which to log progress in input reads
+  */
+class ConsensusCallingIterator(iterator: Iterator[SAMRecord],
+                               val caller: UmiConsensusCaller[_],
```
tfenne: I _think_ that since the iterator never sees any of the `ConsensusRead` objects and never has to write the type anywhere, it doesn't care.  I.e. I read this as `I the iterator need a consensus caller, you might be constrained on it's type parameter, but personally I don't care`.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -0,0 +1,90 @@
...
+  def next(): SAMRecord = {
+    if (!this.hasNext()) throw new NoSuchElementException("Calling next() when hasNext() is false.")
+    this.outputQueue.dequeue()
+  }
+
+  /**
+    * Consumes the next group of records from the input iterator, based on molecule id
+    * and returns then as a Seq.
+    */
+  private def nextGroupOfRecords: Seq[SAMRecord] = {
```
tfenne: RIght you are.  I had previously been thinking that the convention was to put empty parens on procedures (i.e. functions who's entire goal is the side-effect, and the return type is `Unit`), however doing a bit of reading it does appear to be recommended for any method in scala that has side effects.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -0,0 +1,90 @@
...
+    if (iterator.isEmpty) Nil
```
tfenne: Thanks for catching this.  I _really_ wish there was a way in scala to prevent non-val constructor parameters getting val'd.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -0,0 +1,90 @@
...
+    else {
+      val idToMatch = this.caller.sourceMoleculeId(this.input.head)
+      val buffer = ListBuffer[SAMRecord]()
+      while (this.input.hasNext && this.caller.sourceMoleculeId(this.input.head) == idToMatch) {
+        buffer += this.input.next()
+      }
```
tfenne: Yes (though I had to change to using a betterBufferedIterator for a working `takeWhile`.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -0,0 +1,90 @@
...
+  private def advance(): Boolean = {
+    // get the records to create the consensus read
+    val inputs  = nextGroupOfRecords
+    val outputs = this.caller.consensusReadsFromSamRecords(inputs)
+    this.outputQueue ++= outputs
+
+    // Log progress on the _input_ reads and then return/recurse
+    for (p <- progress; r <- inputs) p.record(r)
+    if (outputs.nonEmpty) true
+    else if (this.input.hasNext) advance()
```
tfenne: Yup.  The compiler will yell if it is not.  It's still tail recursive because the current stack has no use once we go down the if - it's returning the value from `advance()` and not evaluating anything else.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -0,0 +1,277 @@
...
+    /** The ID of the molecule that generated the consensus read. */
+    def id: String
+    /** The bases of the consensus read. */
+    def bases: Array[Byte]
+    /** The quals of the consensus read. */
+    def quals: Array[Byte]
+    /** Gets the length of the consensus read. */
+    def length: Int = bases.length
+    /** Returns the consensus read a String - mostly useful for testing. */
+    def baseString = new String(bases)
```
tfenne: Done.  I've renamed the trait to `SimpleRead` because I felt like `Read` was too generic, and it's is incredibly simple 😛

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -0,0 +1,277 @@
...
+  def outputHeader(in: SAMFileHeader, readGroupId: String, sortOrder: Option[SortOrder] = None): SAMFileHeader = {
+    val oldRgs = in.getReadGroups.toSeq
+    def collapse(f: SAMReadGroupRecord => String): String = oldRgs.map(f).filter(_ != null).toSet.toList match {
+      case Nil      => null
+      case x :: Nil => x
+      case xs       => xs.mkString(",")
+    }
+
+    val rg = new SAMReadGroupRecord(readGroupId)
+    rg.setDescription(s"Consensus reads generated from ${oldRgs.size} input read groups.")
```
tfenne: Done.  I also simplified `collapse` to:
```scala
    def collapse(f: SAMReadGroupRecord => String): String = oldRgs.map(f).filter(_ != null).distinct match {
      case Nil => null
      case xs  => xs.mkString(",")
    }
```
Since Seq(one).mkString(",") just returns the one thing, and not using `toSet` will retain the ordering.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -0,0 +1,277 @@
...
+
+    val outHeader = new SAMFileHeader
+    outHeader.addReadGroup(rg)
+    outHeader.setSortOrder(sortOrder.getOrElse(SortOrder.unsorted))
+    outHeader.setGroupOrder(GroupOrder.query)
+    outHeader
+  }
+}
+
+
```
tfenne: I _always_ leave two blank lines between top level classes/traits/objects in the same file as I find the visual break useful!

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -0,0 +1,277 @@
...
+/**
+  * A trait that can be mixed in by any consensus caller that works at the read level,
+  * mapping incoming SAMRecords into consensus SAMRecords.
+  *
+  * @tparam C Internally, the type of lightweight conesnsus read that is used prior to
+  *           rebuilding SAMRecords.
```
tfenne: I'm still getting used to the whole `[[]]` for references in scaladoc.  Will do.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -0,0 +1,277 @@
...
+  /**
+    * Returns the number of raw reads filtered out due to there being insufficient reads present
+    * to build the necessary set of consensus reads.
+    */
+  def readsFilteredInsufficientSupport: Long = this._totalFiltered - this._filteredForMinorityAlignment
+
+  /** Returns the number of raw reads filtered out because their alignment disagreed with the majority alignment of
+    * all raw reads for the same source molecule
+    */
+  def readsFilteredMinorityAlignment: Long = _filteredForMinorityAlignment
```
tfenne: It's whatever the callers send to the `reject` method.  For duplexes I've been lazy and not registered the fragments as rejects because the doc says they are ignored.  For the vanilla caller, I believe it's everything that is unused.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -0,0 +1,277 @@
...
+  /** Records that the supplied records were rejected, and not used to build a consensus read. */
+  protected def rejectRecords(recs: Traversable[SAMRecord]) : Unit = {
+    this._totalFiltered += recs.size
+  }
+
+  /** A RG.ID to apply to all generated reads. */
+  protected def readGroupId: String
+
+  /** A prefix to use on all read names.  If None then a suitable prefix will be synthesized. */
+  protected val readNamePrefix: String
```
tfenne: It's an `Option[String]` on the command line programs, but a `String` here, because I wanted to do away with the need to feed the _input_ `SAMFileHeader` into the UMI caller.  There's a utility method in the companion object to make a prefix from the input header, so that's still available.  

I would only want to make it an option here if it's truly optional, and when absent we don't use a prefix at all.  Is that what you're suggesting?  That would make valid BAM files that would be nearly impossible to merge with other consensus BAM files later.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -0,0 +1,277 @@
...
+
+  /**
+    * Needs to be implemented to to return a value from a SAMRecord that represents the unit of
+    * grouping, e.g. the MI tag for vanilla UMI data and the MI tag minus the /?? suffix for
+    * duplex data.
+    * @param rec a SAMRecord
+    * @return an identified for the source molecule
+    */
+  protected[umi] def sourceMoleculeId(rec: SAMRecord): String
```
tfenne: Protected so sub-classes can see it to implement it.  Package so it can be invoked from tests.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -0,0 +1,277 @@
...
+  protected def subGroupRecords(records: Seq[SAMRecord]): (Seq[SAMRecord], Seq[SAMRecord],Seq[SAMRecord]) = {
+    val fragments    = records.filter { rec => !rec.getReadPairedFlag }
+    val firstOfPair  = records.filter { rec => rec.getReadPairedFlag && rec.getFirstOfPairFlag }
+    val secondOfPair = records.filter { rec => rec.getReadPairedFlag && rec.getSecondOfPairFlag }
+    (fragments, firstOfPair, secondOfPair)
+  }
+
+
+  /**
+    * Takes in a seq of SAMRecords and filters them such that the returned seq only contains
```
tfenne: If a seq is empty, is it a seq of anything ;)

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -0,0 +1,277 @@
...
+    val keepers = groups.values.find(_.size == max).getOrElse(unreachable("At least one group must have the max size!"))
+    groups.values.filterNot(_ eq keepers).foreach(rejectRecords)
+    this._filteredForMinorityAlignment += (recs.size - keepers.size)
+    keepers
+  }
+
+  /** Creates a `SAMRecord` from the called consensus base and qualities. */
+  protected def createSamRecord(read: C, readType: ReadType): SAMRecord = {
+    val rec = new SAMRecord(null)
+    rec.setReadName(this.readNamePrefix + ":" + read.id)
```
tfenne: image

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+  */
+class DuplexConsensusCaller(override val readNamePrefix: String,
+                            override val readGroupId: String = "A",
+                            val minInputBaseQuality: PhredScore = DuplexConsensusCaller.MinInputBaseQuality,
+                            val errorRatePreUmi: PhredScore = DuplexConsensusCaller.ErrorRatePreUmi,
+                            val errorRatePostUmi: PhredScore = DuplexConsensusCaller.ErrorRatePostUmi
+                           ) extends UmiConsensusCaller[DuplexConsensusRead] {
+
+  private val ssCaller =
+    new VanillaUmiConsensusCaller(readNamePrefix="x", options=new VanillaUmiConsensusCallerOptions(
```
tfenne: Let's chat about this.  I think it's necessary to have a read name prefix, to ensure that the read names in the output BAM have a reasonable chance of being unique.  Otherwise if you generate two BAMs from two libraries, and decide you want to merge them for variant calling, all hell will break loose when you have multiple templates with the same name.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+    * Takes in all the reads for a source molecule and, if possible, generates one or more
+    * output consensus reads as SAM records.
+    *
+    * @param recs the full set of source SAMRecords for a source molecule
+    * @return a seq of consensus SAM records, may be empty
+    */
+  override protected def consensusSamRecordsFromSamRecords(recs: Seq[SAMRecord]): Seq[SAMRecord] = {
+    // Group the reads by /A vs. /B and then sort so that the larger group is first, and then by suffix
+    val groups = recs.filter(_.getReadPairedFlag)
+      .groupBy(_.getStringAttribute(ConsensusTags.MolecularId))
```
tfenne: Yes, the `sourceMoleculeId` checks that the `MI` tag is present and that it contains a `/`.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+      .toSeq.sortWith {
+      case ((lhsMi, lhsRecs), (rhsMi, rhsRecs)) =>
+        if (lhsRecs.size > rhsRecs.size) true
+        else if (rhsRecs.size > lhsRecs.size) false
+        else lhsMi < rhsMi
+    }.map(_._2)
```
tfenne: I went back and forth on this.  At least in `GroupReadsByUmi` the /A and /B are largely artbirary.  Largely I just wanted some predictability for the tests.  I'll change it so that it just sorts by the id's for now.  That does mean that in filtering we have to be careful that the `a?` tags may have the lower values, but I think that's ok.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
tfenne: Eh, I felt a little guilty so I went back and updated the duplex caller to `reject` frags.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+
+    groups match {
```
tfenne: I tend not to like really long and complex expressions getting matched on.  My brain finds it easier to come back to if I give a name to the result of the big expression before matching.  Maybe that's just me.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+        val baR2Consensus = ssCaller.consensusFromSamRecords(filteredBaR2s)
+
+        // Call the duplex reads
+        val duplexR1 = duplexConsensus(abR1Consensus, baR2Consensus)
+        val duplexR2 = duplexConsensus(abR2Consensus, baR1Consensus)
+
+        // Convert to SAMRecords and return
+        (duplexR1, duplexR2) match {
+          case (Some(r1), Some(r2)) =>
+            Seq(createSamRecord(r1, FirstOfPair), createSamRecord(r2, SecondOfPair))
```
tfenne: The not-most-common alignment ones are handled by the `filterToMostCommonAlignment` method.  I've added tracking of the frags.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+          case _                    =>
+            rejectRecords(recs)
+            Nil
+        }
+      case gs =>
+        unreachable("SAMRecords supplied with more than two distinct MI values.")
```
tfenne: That's hard to do without making it more complicate, e.g. `gs if gs.size > 2` and since it's truly an exception case I'd rather not evaluate that every time when it should almost never happen.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+    * If either of the incoming reads are undefined, the duplex read will be undefined.
+    *
+    * @param ab the single-strand consensus from one strand
+    * @param ba the single-strand consensus from the other strand
+    * @return a duplex consensus if one can be built
+    */
+  private[umi] def duplexConsensus(ab: Option[VanillaConsensusRead], ba: Option[VanillaConsensusRead]): Option[DuplexConsensusRead] = {
+    (ab, ba) match {
+      case (Some(a), Some(b)) =>
+        val len = min(a.length, b.length)
```
tfenne: Yeah.  It's complicated to describe though ;)  Basically what you get back is the longest length at which you had at least one AB R1 and one BA R2 (for read 1) and vice versa.  I feel like trying to explain that might cause more confusion than help.  I'll see what I can do though.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+    * Creates a SAMRecord with a ton of additional tags annotating the duplex read.
+    */
+  override protected def createSamRecord(read: DuplexConsensusRead, readType: ReadType): SAMRecord = {
+    val r = super.createSamRecord(read, readType)
+    val ab = read.abConsensus
+    val ba = read.baConsensus
+
+    // Calculate the total depths and errors across both SS consensus reads
+    val totalDepths = read.abConsensus.depths.zip(read.baConsensus.depths).map(x => x._1 + x._2)
+    val totalErrors = new Array[Short](totalDepths.length)
```
tfenne: I guess it doesn't really.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+    forloop (from=0, until=totalErrors.length) { i =>
+      var e = 0
+      if (read.bases(i) == ab.bases(i)) e += ab.errors(i) else e += ab.depths(i) - ab.errors(i)
```
tfenne: You're not wrong, and I did think about this.  The choice is between making things quite a bit more complicated (the vanilla consensus read would now need to carry around counts of all four bases at each position instead of depths and errors, so that they could be used in duplex calling, not because they are needed for vanilla calling) vs. a minor inaccuracy in an edge case that I think it not terribly important.

I say "not terribly important" because the only way this can arise is if the two single-strand consensus reads disagree at a position _and_ there's > 2 different bases observed in one or both of them.  If the two disagree the consensus call is already going to be << Q45 and the annotated error rate is already going to be high.  

So really it's a question of whether this is worth fixing to be 100% accurate, when I think it's functionally ok.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,255 @@
...
+
+    { import ConsensusTags.PerBase._
+      // TODO: should we also set the `RawReadCount` and `RawReadErrors` tags? Totally redundant, but maybe nice for consistency?
+      r.setAttribute(AbRawReadCount, read.abConsensus.depths)
+      r.setAttribute(BaRawReadCount, read.baConsensus.depths)
+      r.setAttribute(AbRawReadErrors, read.abConsensus.errors)
+      r.setAttribute(BaRawReadErrors, read.baConsensus.errors)
+    }
+
+    r
```
tfenne: 😝 Are you going to make me rename `ab` and `ba` next?  (I change r -> rec for you).

src/main/scala/com/fulcrumgenomics/umi/ConsensusTags.scala
```diff
@@ -0,0 +1,79 @@
...
+    val RawReadErrors   = "ce" // consensus errors
+
+    // Duplex versions of the above tags for the two single strand consensus reads
+    val AbRawReadCount  = "ad"
+    val BaRawReadCount  = "bd"
+    val AbRawReadErrors = "ae"
+    val BaRawReadErrors = "be"
+  }
+
+  object PerRead {
```
tfenne: I added a sentence to the doc of the CLPs to this effect.  Not sure if there's anywhere else it's useful.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -28,6 +28,7 @@ import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.bam.Bams
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.umi.VanillaUmiConsensusCallerOptions._
+import com.fulcrumgenomics.util.NumericTypes.PhredScore
```
tfenne: One of the CLP params is of type PhredScore.  We had at some point introduced extra type defs for PhredScore in the consensus caller objects, which I've now eliminated, making this import necessary.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -125,141 +75,74 @@ case class SourceRead(bases: Array[Byte], quals: Array[Byte]) {
...
-
+  protected[umi] def consensusFromSamRecords(records: Seq[SAMRecord]): Option[VanillaConsensusRead] = {
     if (records.size < this.options.minReads) {
-      this._filteredForMinReads += records.size
       None
     }
     else {
-      val filteredRecords = VanillaUmiConsensusCaller.filterToMostCommonAlignment(records)
-      this._filteredForMinorityAlignment += (records.size - filteredRecords.size)
+      val filteredRecords = filterToMostCommonAlignment(records)
```
tfenne: It's handled internally.

tfenne: I've added a NOTE in the doc on the filter method to make it more obvious.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: It's a little belt-and-suspenders, but it is necessary.  The order of operations is:
1. Check that we have enough reads on 141
2. Filter to the most common alignment
3. Map reads to source reads and call this method
4. Check that we have enough reads left after filtering.

I'll move the check up to immediately after filtering and (so long as it doesn't break any tests) remove it from here.

tfenne: Hrm, I've added the check higher up (mostly so we don't create SourceReads if we don't need to).  But the original author (you I think?) decided there should be a test for this method that passed in insufficient reads and ensured a None is returned.  Leaving it as is for now, though we could convert it to not use Option and eliminate/refactor that test.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
tfenne: Because each caller fills in it's own set of extra tags.

src/test/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReadsTest.scala
tfenne: I'm shooting for 100% coverage ;)  I don't think they hurt.

src/test/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReadsTest.scala
tfenne: Most of the testing is in the DuplexConsensusCallerTest, the tests up here are really just testing that the CLP works conditioned on all the other tests passing.

src/test/scala/com/fulcrumgenomics/umi/DuplexConsensusCallerTest.scala
tfenne: I seem to get yelled at by the compiler when I try that, hence fishing around and finding this.

src/test/scala/com/fulcrumgenomics/umi/DuplexConsensusCallerTest.scala
tfenne: Yup.  The fact that the R strand stuff gets reversed/RC's is a bit PITA for test writing.

tfenne:
@nh13 back to you.  The filtering tool is now complete including copious tests, and I think all your earlier review points are addressed.  Can you take another pass through please?

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -0,0 +1,114 @@
...
+class CallDuplexConsensusReads
+(@arg(flag="i", doc="The input SAM or BAM file.") val input: PathToBam,
+ @arg(flag="o", doc="Output SAM or BAM file to write consensus reads.") val output: PathToBam,
+ @arg(flag="p", doc="The prefix all consensus read names") val readNamePrefix: Option[String] = None,
+ @arg(flag="R", doc="The new read group ID for all the consensus reads.") val readGroupId: String = "A",
+ @arg(flag="1", doc="The Phred-scaled error rate for an error prior to the UMIs being integrated.") val errorRatePreUmi: PhredScore = DefaultErrorRatePreUmi,
+ @arg(flag="2", doc="The Phred-scaled error rate for an error post the UMIs have been integrated.") val errorRatePostUmi: PhredScore = DefaultErrorRatePostUmi,
+ @arg(flag="m", doc="Ignore bases in raw reads that have Q below this value.") val minInputBaseQuality: PhredScore = DefaultMinInputBaseQuality,
+ @arg(flag="S", doc="The sort order of the output, if None then the same as the input.") val sortOrder: Option[SortOrder] = Some(SortOrder.queryname)
+) extends FgBioTool with LazyLogging {
```
nh13: How about adding the `outputPerBaseTags` option similar to the vanilla caller?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+            if      (aBase == NoCall || bBase == NoCall) (NoCall, PhredScore.MinValue)
+            else if (aBase == bBase) (aBase, (aQual + bQual).toByte)
+            else if (aQual > bQual)  (aBase, (aQual - bQual).toByte)
+            else if (bQual > aQual)  (bBase, (bQual - aQual).toByte)
+            else (NoCall, PhredScore.MinValue)
+          }
+
+          bases(i)  = base
+          quals(i)  = qual
+          errors(i) = sourceReads.count(s => s.bases(i) != NoCall && s.bases(i) != base).toShort
```
nh13: I don't think the `toShort` will cap it to 32767 but instead allow for an overflow and become negative.  Can you add a guard here?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+    { import ConsensusTags.PerRead._
+      rec.setAttribute(RawReadCount,       totalDepths.max)
+      rec.setAttribute(MinRawReadCount,    totalDepths.min)
+      rec.setAttribute(RawReadErrorRate,   read.errors.sum / totalDepths.sum.toFloat)
+      rec.setAttribute(AbRawReadCount,     read.abConsensus.depths.max.toInt)
+      rec.setAttribute(BaRawReadCount,     read.baConsensus.depths.max.toInt)
+      rec.setAttribute(AbMinRawReadCount,  read.abConsensus.depths.min.toInt)
+      rec.setAttribute(BaMinRawReadCount,  read.baConsensus.depths.min.toInt)
+      rec.setAttribute(AbRawReadErrorRate, read.abConsensus.errors.sum / read.abConsensus.depths.sum.toFloat)
+      rec.setAttribute(BaRawReadErrorRate, read.baConsensus.errors.sum / read.baConsensus.depths.sum.toFloat)
```
nh13: I don't see how the sum will guard against an overflow here.  Can you add some guards against that and cap it?

nh13: Here and elsewhere.

PR #164, opened 2017-01-31
Converting ErrorRateByReadPosition to allow:

tfenne:
1. Restriction to a set of intervals
  2. Exclusion of known variant sites
  3. Include substitution type specific error rates


src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
-import htsjdk.samtools.util.SequenceUtil
-import htsjdk.samtools.util.SequenceUtil.{isValidBase => validBase}
-import htsjdk.samtools.{CigarOperator, SAMRecord, SamReaderFactory}
+import htsjdk.samtools.filter.{DuplicateReadFilter, FailsVendorReadQualityFilter, SamRecordFilter, SecondaryOrSupplementaryFilter}
+import htsjdk.samtools.reference.ReferenceSequenceFileWalker
+import htsjdk.samtools.util.{IntervalList, SamLocusIterator, SequenceUtil}
+import htsjdk.samtools.{SAMRecord, SAMSequenceDictionary, SamReader, SamReaderFactory}
+import htsjdk.variant.vcf.VCFFileReader
+
+import scala.collection.JavaConversions.asScalaIterator
```
nh13: should we use our own implicits ;)

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
-  * @param bases_total the total number of bases observed at this position.
-  */
-case class ErrorRateByReadPositionMetric
-( read_number: Int,
-  position: Int,
-  error_rate: Double,
-  bases_total: Long
-) extends Metric
+    locusIterator.filterNot(l => variantMask.isVariant(l.getSequenceIndex, l.getPosition)).foreach { locus =>
+      val refBase = SequenceUtil.upperCase(refWalker.get(locus.getSequenceIndex).getBases()(locus.getPosition - 1))
```
nh13: This is so ugly it makes me want a wrapper around the refWalker.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
+      if (SequenceUtil.isValidBase(refBase)) {
```
nh13: Any harm in importing either all of SequenceUtil or the few methods you use?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
+          (ref, read) match {
+            case ('A', 'A') => counter.a_ref_obs  += 1
+            case ('A', 'C') => counter.a_to_c_obs += 1
+            case ('A', 'G') => counter.a_to_g_obs += 1
+            case ('A', 'T') => counter.a_to_t_obs += 1
+            case ('C', 'A') => counter.c_to_a_obs += 1
+            case ('C', 'C') => counter.c_ref_obs  += 1
+            case ('C', 'G') => counter.a_to_g_obs += 1
+            case ('C', 'T') => counter.c_to_t_obs += 1
+          }
```
nh13: How hard would it be to add support for insertions as errors?  I think the locus iterator has support for gathering indels, but I don't know how much effort it would be to add.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
-    updateErrors(rec=rec, readIndex=readIndex)
-  }
+    in.safelyClose()
 
-  protected def maybeReverseComplement(bases: Array[Byte], rc: Boolean): Array[Byte] = {
-    if (rc) SequenceUtil.reverseComplement(bases)
-    bases
-  }
+    // Ensure that all the maps have values up to the max
+    for ((counter, readNum) <- counters.filter(_.nonEmpty).zipWithIndex; cycle <- 1 to counter.keys.max)
```
nh13: I like this, I did not know you could do this, I like this.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
+      counter.getOrElseUpdate(cycle, new ObsCounter(readNum, cycle))
```
nh13: but could I have some brackets, pretty please?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
-        if (op == CigarOperator.INSERTION) {
-          var i = 0
-          while (i < elem.getLength) {
-            val qual = quals(readOffset)
-            increment(readIndex = readIndex, readOffset = readOffset, qual=qual)
-            readOffset += readOffsetIncrement
-            i += 1
-          }
+  /** Generates a variant mask object. */
+  private def buildVariantMask(variants: Option[PathToVcf], intervals: Option[IntervalList], dict: SAMSequenceDictionary): VariantMask = {
```
nh13: Could this be an apply method in `VariantMask`?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
+  def toMetric: ErrorRateByReadPositionMetric = {
+    val totalA = a_ref_obs + a_to_c_obs + a_to_g_obs + a_to_t_obs
+    val totalC = c_ref_obs + c_to_a_obs + c_to_g_obs + c_to_t_obs
+    val total  = totalA + totalC
+
+    new ErrorRateByReadPositionMetric(
+      read_number = readNumber,
+      position    = position,
+      bases_total = total,
+      error_rate  = if (total == 0) 0 else (total - a_ref_obs - c_ref_obs) / total.toDouble,
```
nh13: We do this so often, and sometimes I forget, perhaps we should put this in an fgbiodef.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
-        val error_bases = errCounter.countOf(offset)
-        val error_total = totalCounter.countOf(offset)
-        val error_rate  = if (0 == error_total) 0 else error_bases / error_total.toDouble
-        new ErrorRateByReadPositionMetric(
-          read_number = idx,
-          position    = offset + 1,
-          error_rate  = error_rate,
-          bases_total = error_total
-        )
+          (ref, read) match {
```
nh13: I'd rather have this in a method in ObsCounter.

src/main/scala/com/fulcrumgenomics/util/Rscript.scala
```diff
@@ -0,0 +1,91 @@
...
+
+import scala.io.Source
+import scala.util.{Success, Try}
+
+/**
+  * Object that enables running of R scripts via the Rscript command line too.
+  */
+object Rscript extends LazyLogging {
+  /** The name of the Rscript executable. */
+  private val Executable = "Rscript"
```
nh13: We can update this when we have a need, but if Rscript is in a specific location configured by a pipeline, this needs to be configurable.

src/main/scala/com/fulcrumgenomics/util/Rscript.scala
```diff
@@ -0,0 +1,91 @@
...
+    val path = Io.makeTempFile("script.", ".R")
+    path.toFile.deleteOnExit()
+    val out = Io.toWriter(path)
+
+    Source.fromInputStream(in).getLines().foreach { l => out.write(l); out.newLine() }
+    in.close()
+    out.close()
+    path
+  }
+}
```
nh13: Lots of red codecov here.

src/main/scala/com/fulcrumgenomics/vcf/VariantMask.scala
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+  * Simple mask that loads variants one reference sequence at a time and creates a compact
+  * representation allowing for rapid querying of whether or not positions are overlapped
+  * by one or more variants.
+  *
+  * @param variants a coordinate sorted iterator of variants
+  * @param dict the sequence dictionary for the reference
+  */
+class VariantMask(variants: Iterator[VariantContext], val dict: SAMSequenceDictionary) {
```
nh13: Why not just use an OverlapDetector?  Is it really that many intervals?

nh13: Also, this class could be really useful not only for variants but Intervals in general, if it is really needed over an overlap detector.

src/main/scala/com/fulcrumgenomics/vcf/VariantMask.scala
```diff
@@ -0,0 +1,97 @@
...
+    * Causes the variant mask to advance to the chosen reference sequence and load up the mask.
+    */
+  private def advanceTo(refIndex: Int): Unit = {
+    require(refIndex > this.currentIndex, "Requested to navigate to an earlier contig. Is your VCF sorted?")
+    require(refIndex >= 0 && refIndex < dict.size(), s"Requested navigation to contig #${refIndex} which is invalid.")
+
+    val ref  = dict.getSequence(refIndex)
+    val bits = new mutable.BitSet(ref.getSequenceLength)
+    iterator.dropWhile(v => dict.getSequenceIndex(v.getContig) < refIndex)
+      .takeWhile(v => dict.getSequenceIndex(v.getContig) == refIndex)
```
nh13: Is the `getSequenceIndex` fast?  You could get the contig name for `refIndex` instead and pay for a string comparison.

src/test/scala/com/fulcrumgenomics/bam/ErrorRateByReadPositionTest.scala
```diff
@@ -25,229 +25,170 @@
...
-      builder.addFrag(start = 100, strand=strand).foreach { rec => rec.setAttribute("MD", "100"); rec.setReadString("A" * 100) }
-      builder.addFrag(start = 100, cigar=cigar, strand=strand).foreach { rec => rec.setAttribute("MD", md); rec.setReadString(readString) }
-      val input = builder.toTempFile()
-      val (out, pre) = outputAndPrefix
-      new ErrorRateByReadPosition(input = input, output = Some(pre), ref=maybeRef, includeInsertions=includeInsertions).execute()
-      val metrics = Metric.read[ErrorRateByReadPositionMetric](out)
-      metrics.length shouldBe 100
-      metrics
-    }
+  it should "work on a file with all unmapped reads" in {
```
nh13: unmapped fragments, since this is where frag support is tested

src/test/scala/com/fulcrumgenomics/bam/ErrorRateByReadPositionTest.scala
```diff
@@ -25,229 +25,170 @@
...
+    val metrics = new ErrorRateByReadPosition(input=builder.toTempFile(), output=Some(pre), ref=ref, variants=Some(vcf)).computeMetrics
+    metrics.find(_.position == 1).get.error_rate shouldBe 1.0
+    metrics.find(_.position == 11).get.error_rate shouldBe 0.0
+    metrics.filter(m => m.position != 1 && m.position != 11).foreach { m =>
+      m.bases_total shouldBe 4
+      m.error_rate  shouldBe 0.0
     }
   }
 
-  it should "ignore soft-clipped bases at the end of the read" in {
```
nh13: I assume these are gone since we are assuming SamLocusIterator does this for us?

src/main/scala/com/fulcrumgenomics/util/Rscript.scala
nh13: Could we install R manually? http://jtleek.com/protocols/travis_bioc_devel/

src/test/scala/com/fulcrumgenomics/bam/ErrorRateByReadPositionTest.scala
```diff
@@ -25,229 +25,170 @@
...
-  def outputAndPrefix: (Path, Path) = {
-    val out = makeTempFile("ErrorRateByReadPositionTest", ErrorRateByReadPositionMetric.FileExtension)
-    val pre = PathUtil.removeExtension(PathUtil.removeExtension(out))
+class ErrorRateByReadPositionTest extends UnitSpec {
+  /////////////////////////////////////////////////////////////////////////////
+  // Make a reference and a set of variants for use in the tests
+  /////////////////////////////////////////////////////////////////////////////
+  private val ref = {
+    val builder = new ReferenceSetBuilder()
+    builder.add("chr0").add("RYSWKMBDHV", 50).add("N", 500) // Sooo ambiguous
```
nh13: just so

src/test/scala/com/fulcrumgenomics/bam/ErrorRateByReadPositionTest.scala
```diff
@@ -25,229 +25,170 @@
...
-      }
-      else if (metric.position <= 20) {
-        metric shouldBe new ErrorRateByReadPositionMetric(0, metric.position, 0.5d, 2l)
-      }
-      else {
-        metric shouldBe new ErrorRateByReadPositionMetric(0, metric.position, 0.0d, 2l)
-      }
+    if (Rscript.Available) {
+      val plot = PathUtil.pathTo(pre + ErrorRateByReadPositionMetric.PlotExtension)
+      Files.exists(plot) shouldBe true
```
nh13: For our travis build, I'd like us to generate the plots.  Is that feasible?

src/main/resources/com/fulcrumgenomics/bam/ErrorRateByReadPosition.R
```diff
@@ -0,0 +1,59 @@
+#################################################################################
```
tfenne: We don't execute the script directly, we invoke Rscript and pass it, so it wouldn't do anything.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
tfenne: Once that PR is merged, yes.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
-    |contain a row per logical read and per position in that read.  The 'read_number' will be 0 for fragments, 1 for the
-    |first end of a pair, and 2 for the second end of a pair. Both the error rate and number of bases observed will be
-    |provided.  Soft-clipped bases, and unmapped, secondary, supplementary, and vendor failing reads are ignored.
-    |Optionally, duplicates are ignored.
+    |The following are reads / bases are excluded from the analysis:
+    |  - Unmapped reads
+    |  - Reads marked as failing vendor quality
+    |  - Reads marked as duplicates (unless --include-duplicates is specified)
+    |  - Secondary and supplemental records
+    |  - Soft-clipped bases in records
```
tfenne: Of course.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
+  """
+  )
 class ErrorRateByReadPosition
 ( @arg(flag="i", doc="Input BAM file.") val input: PathToBam,
-  @arg(flag="o", doc="Output metrics prefix. If not given, will use the input BAM basename as a prefix.")
-  val output: Option[PathPrefix] = None,
-  @arg(flag="r", doc="Reference fasta. If not given, will use the MD tag.") val ref: Option[PathToFasta] = None,
+  @arg(flag="o", doc="Output metrics prefix. If not given, will use the input BAM basename.") val output: Option[PathPrefix] = None,
+  @arg(flag="r", doc="Reference sequence fasta file.") val ref: PathToFasta,
+  @arg(flag="v", doc="Optional file of variant sites to ignore.") val variants: Option[PathToVcf] = None,
```
tfenne: We could, but that would be markedly more complicated for, IMHO, pretty much no benefit.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -25,266 +25,234 @@
...
-        (rec.getMappingQuality < minMappingQuality)
-    }
-    .foreach { rec =>
-      counter.accept(rec)
-      progress.record(rec)
-    }
+  /** Computes the metrics from the BAM file. */
+  private[bam] def computeMetrics: Seq[ErrorRateByReadPositionMetric] = {
+        val progress = new ProgressLogger(logger, verb="Processed", noun="loci", unit=50000)
+    val in = SamReaderFactory.make.open(input)
```
tfenne: I guess.  Silly CRAM.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
tfenne: refWalker is so simple, would probably be easier to just write our own in Scala that provides some nicer APIs.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
tfenne: I tend to prefer referencing like this, as I find easier (when reading later) to determine where the method is coming from.  Particularly with the mess of utility classes in HTSJDK, I can never figure out if something is in SequenceUtil or StringUtil or SamUtil or SamPairUtil etc.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
tfenne: I'd like to leave that for another day.  It's not super hard to grab the reads that have insertions, but why count insertions and not deletions?  It also opens up questions like, do you count each inserted base as an error, or just once for the whole insertion?  Etc.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
tfenne: I don't think VariantMask should deal with `Option[PathToVcf]` or `Option[IntervalList]`...

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
tfenne: Agreed.  Some kind of `x.safeDivide(y)` that returns 0 if both are 0, else the result of the division.  Need to think about how to do that nicely though.

src/main/scala/com/fulcrumgenomics/util/Rscript.scala
tfenne: I agree that would be good to do, but it's also a bunch of work and might make our builds less robust.  I'm not sure what to do about that.

src/main/scala/com/fulcrumgenomics/vcf/VariantMask.scala
tfenne: It depends on usage.  If you're looking at a small set of intervals with truth, maybe not.  If you're looking at WGS data with dbSNP, several million entries?

tfenne: We can always refactor later to make it a "GenomeMask" trait that just needs implementations to supply a stream of positions or ranges on a chromosome.

src/main/scala/com/fulcrumgenomics/vcf/VariantMask.scala
tfenne: It is a hash lookup.  Unfortunately I don't think I can do as you suggest.  If the iterator doesn't contain any variants on, say, `chr20` and I ask to advance to `chr20` I need to know to stop either when I find a variant on `chr20`, or I've gone past `chr20`.

src/main/scala/com/fulcrumgenomics/util/Rscript.scala
tfenne: I think that's awkward.  That would imply that either every fgbio tool that uses R would need to take a parameter for where to find `Rscript`, or we'd have to have some global way of doing that.  Otherwise there's no way for a pipeline to tell the tool where to find Rscript.

nh13: Yeah, a Defaults or a Config class is a bit unwieldy for a tool, good point.

src/main/scala/com/fulcrumgenomics/util/Rscript.scala
nh13: Picard must run R in its tests that run on travis-ci.

nh13:
👍

PR #165, opened 2017-02-01
Two new implicits to replace all use of JavaConversions in fgbio non-test code

tfenne:


src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -46,6 +46,17 @@ class FgBioDef extends CommonsDef {
     def bufferBetter = new BetterBufferedIterator(scala.collection.JavaConversions.asScalaIterator(iterator))
   }
 
+  /** Implicit that wraps a Java iterator into a Scala iterator. */
+  implicit class JavaIteratorAdapter[A](private val underlying: java.util.Iterator[A]) extends Iterator[A] {
```
nh13: Could there be an implicit class that takes in a `java.util.Iterable` in its constructor and creates an `Iterator`?  Then we can replace `iterableAsScalaIterable` with `JavaIterableAsIterator` or the like.  This would mean that the `toIterator` is not right in your face, but I think the thing we want is iterable->iterator.

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -43,7 +43,36 @@ class FgBioDef extends CommonsDef {
...
+
+  /** Implicit that wraps a scala iterator as a Java iterator. */
+  implicit def javaIteratorAsScalaIterator[A](iterator: java.util.Iterator[A]): Iterator[A] = iterator match {
+    case iter: ScalaIteratorAdapter[A] => iter.underlying
+    case iter: java.util.Iterator[A]   => new JavaIteratorAdapter[A](iter)
+  }
+
+  /** Implicit that wraps a scala iterator as a Java iterator. */
+  implicit def scalaIteratorAsJavaIterator[A](iterator: Iterator[A]): java.util.Iterator[A] = iterator match {
+    case iter: JavaIteratorAdapter[A] => iter.underlying
```
nh13: This is clever, here and above, nice.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -190,7 +189,7 @@ class EstimatePoolingFractions
     iterator.setIncludeNonPfReads(false)
     iterator.setMappingQualityScoreCutoff(this.minMappingQuality)
     iterator.setQualityScoreCutoff(this.minBaseQuality)
-    iterator.toIterator
+    javaIteratorAsScalaIterator(iterator)
   }
```
nh13: It's interesting to see that we construct SamLocusIterators in a similar fashion across a lot of code.  It's starting to make me think it needs a proper treatment in the `Bam` object, but that's for another day.

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -93,6 +92,7 @@ class FilterBam
     }
     logger.info("Kept " + new DecimalFormat("#,##0").format(kept) + " records.")
     out.close()
+    iterator.safelyClose()
     CloserUtil.close(iterator)
```
nh13: Can remove this line now.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -250,7 +249,7 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
     * @param rec the record to be clipped
     * @param numberOfBasesToClip the number of additional bases to be clipped
     */
-  def clip3PrimeEndOfRead(rec: SAMRecord, numberOfBasesToClip: Int) = {
+  def clip3PrimeEndOfRead(rec: SAMRecord, numberOfBasesToClip: Int): Unit = {
```
nh13: Public methods _should_ have a return type ;), thanks!

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -183,8 +182,7 @@ class AssessPhasing
 
     val intervalListForContig = intervalList.map { oldList =>
       val newList = new IntervalList(oldList.getHeader)
-      import scala.collection.JavaConversions._
-      newList.addall(oldList.getIntervals.filter { _.getContig == contig })
+      newList.addall(oldList.getIntervals.filter { _.getContig == contig }.toList.asJava)
```
nh13: `toJavaList`?  I know you were debating adding this earlier.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -271,14 +268,11 @@ class AssessPhasing
                                        contig: String,
                                        contigLength: Int,
                                        intervalList: Option[IntervalList] = None): Iterator[VariantContext] = {
-    import scala.collection.JavaConversions.asScalaBuffer
-    val sampleName = reader.getFileHeader.getSampleNamesInOrder.head
+    val sampleName = reader.getFileHeader.getSampleNamesInOrder.iterator().next()
```
nh13: This is ugly.

src/main/scala/com/fulcrumgenomics/vcf/HapCutToVcf.scala
```diff
@@ -128,14 +127,14 @@ class HapCutToVcf
     // get the header lines in the input header that we wish to skip/replace with our own definitions
     val headerLinesToSkip = HeaderLines.formatHeaderKeys(hapCutType).flatMap(key => Option(inputHeader.getFormatHeaderLine(key)))
     val headerLines: util.Set[VCFHeaderLine] = new util.HashSet[VCFHeaderLine](
-      inputHeader.getMetaDataInSortedOrder.filterNot(headerLinesToSkip.contains).toSet
+      inputHeader.getMetaDataInSortedOrder.filterNot(headerLinesToSkip.contains).toSet.asJava
```
nh13: toJavaSet?

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -43,7 +43,36 @@ class FgBioDef extends CommonsDef {
...
+    override def next(): A = underlying.next()
+  }
+
+  /** Class that wraps a Scala iterator into a Java iterator. */
+  private final class ScalaIteratorAdapter[A](private[FgBioDef] val underlying: Iterator[A]) extends java.util.Iterator[A] {
+    override def hasNext: Boolean = underlying.hasNext
+    override def next(): A = underlying.next()
+  }
+
+  /** Implicit that wraps a scala iterator as a Java iterator. */
```
tfenne: Doh, copy-paste error.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: That's not a bad idea.  It's also annoying that SamLocusIterator implements `Iterable` and `Iterator`, making the implicits ambiguous and this explicit call necessary.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
tfenne: I know, but there's really not a great alternative.  Could call `.get(0)` in this instance since it's a list, not a set.  But converting to scala just to call `.head` seems just as bad.

tfenne:
@nh13 I made a couple of changes:
1. Added an implicit class that wraps an `Iterator` and provides efficient `toJavaList` and `toJavaSet` methods.
2. Reworked the iterator wrappers to extend from a new trait `DualIterator` that extents both scala and Java iterators.

The latter I did because the APIs are identical, and it just means that we can treat one as the other without constantly wrapping/unwrapping things, which is how I think it should be.

On your point about moving towards explicit implicits (ahem), I'm of two minds.  I have absolutely no issue at all with the implicits to go back and forth between a Java and a Scala iterator, as they are essentially the same thing, and the conversion is very lightweight, and unlikely to mislead.  The java Iterable -> scala Iterator is slightly more questionable.  Removing it would just mean having to call `thing.iterator().scalaMethods` instead of `thing.scalaMethods`.  I could go either way on that.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
nh13: That's a good point, we could have our own SamLocusIterator case class that extends only Iterator.

PR #166, opened 2017-02-02
Change how per-base errors are counted when calling consensuses.

nh13:
When the final consensus call is a no-call (N), we instead count the
number of errors relative to the most likely consensus call, whereas we
previously used the total depth.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
@@ -199,7 +199,7 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
 
         // Generate the values for depth and count of errors
         val depth  = builder.contributions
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -187,13 +187,13 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
         }
 
         // Call the consensus and do any additional filtering
-        val (base, qual) = {
+        val (rawBase, rawQual) = builder.call()
+        val (base, qual)       = {
```
tfenne: How about:
```scala
val (base, qual) = 
    if (builder.contributions < this.options.minReads)       (NoCall, NotEnoughReadsQual)
    else if (rawQual < this.options.minConsensusBaseQuality) (NoCall, TooLowQualityQual)
    else (rawBase, rawQual)
```

PR #167, opened 2017-02-04
Adding esoteric tools for basecalling

nh13:


src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+    |The sample sheet should be an Illumina Experiment Manager sample sheet. The tool writes two files to the output
+    |directory: a barcode parameter file and a library parameter file.
+    |
+    |The barcode parameter file is used by Picard's ExtractIlluminaBarcodes and CollectIlluminaBasecallingMetrics to
+    |determine how to match sample barcodes to each read.  The parameter file will be written to the output directory
+    |with name "barcode_params.lane-<lane>.txt".
+    |
+    |The library parameter file is used by Picard's IlluminaBasecallsToSam to demultiplex samples and name each sample's
+    |output BAM file.  The parameter file will be written to the output directory with name
+    |"library_params.lane-<lane>.txt".  The path to each sample's BAM file will be specified in the library parameter
```
tfenne: "samples's BAM" probably needs turning into a phrase like "the output BAM file path for each Sample" to indicate that you're writing paths to a file to tell IlluminaBasecallsToSam where it should write the BAMs.  Ugh.

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+    |file.  Each BAM file will have path "<output>/<sample-name>.<library-id>.bam" if the library identifier is given or
```
tfenne: I'd suggest maybe using `sample-name.barcode-sequence` as a) that way you should always have both and b) it'll always be unique.

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+    |"<output>/<sample-name>.bam" if library identifer is not specified in the sample sheet.
+  """)
+class ExtractBasecallingParams
+(
+  @arg(flag="i", doc="The sample sheet.") val sampleSheet: FilePath,
```
tfenne: I think either this should be `input` or the flag should be `s`.

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+    |determine how to match sample barcodes to each read.  The parameter file will be written to the output directory
+    |with name "barcode_params.lane-<lane>.txt".
+    |
+    |The library parameter file is used by Picard's IlluminaBasecallsToSam to demultiplex samples and name each sample's
+    |output BAM file.  The parameter file will be written to the output directory with name
+    |"library_params.lane-<lane>.txt".  The path to each sample's BAM file will be specified in the library parameter
+    |file.  Each BAM file will have path "<output>/<sample-name>.<library-id>.bam" if the library identifier is given or
+    |"<output>/<sample-name>.bam" if library identifer is not specified in the sample sheet.
+  """)
+class ExtractBasecallingParams
```
tfenne: What do you think about either `ExtractIlluminaBasecallingParams` or `ExtractBasecallingParamsForPicard`?

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+(
+  @arg(flag="i", doc="The sample sheet.") val sampleSheet: FilePath,
+  @arg(flag="o", doc="The output folder to where lane-specific data should be written.") val output: DirPath,
```
tfenne: How about "The output folder where per-lane parameter files should be written."

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+  @arg(flag="l", doc="The lane (1-based).") val lane: Int
```
tfenne: I'm confused why this is single-valued.  Wouldn't I want to run this once to generate files for each lane?

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
tfenne: I think probably there should be a second parameter, something like `bamOutput: Option[DirPath]` that defaults to `output` at runtime, but allows users to specify a different directory to output BAMs to.

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+  /** Extracts sample and library information from an sample sheet for a given lane.  The resulting files are used
+    * to run Picard's ExtractIlluminaBarcodes, CollectIlluminaBasecallingMetrics, and IlluminaBasecallsToSam. See
+    * [[ExtractBasecallingParams]] for more information.
+    *
+    * @param sampleSheet the Illumina Experiment Manager sample sheet.
+    * @param lane the lane number.
+    * @param output the output directory where the barcode, library param, and BAM files should be written.
+    */
+  def from(sampleSheet: SampleSheet, lane: Int, output: DirPath): BasecallingParams = {
+    require(0 < lane)
```
tfenne: Can you please add a message for the requirement.

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+    val libraryParamLines = new ListBuffer[String]()
+    val samples           = sampleSheet.filter(_.lane.forall(_ == lane)).toSeq
+    val barcodeFile       = barcodeFileFrom(output=output, lane=lane)
+    val libraryParamsFile = libraryParamsFileFrom(output=output, lane=lane)
+
+    // Dual vs single indexed
+    val dualIndexed = samples.exists(_.i5IndexBases.nonEmpty)
+
+    // Check for project/description on the sample
+    val includeDescription    = samples.exists(s => s.project.nonEmpty || s.description.nonEmpty)
```
tfenne: I don't love this.  I've seen a lot of labs re-use the "project" field for things that have no place in a BAM file (instead of creating their own fields).  E.g. one lab puts the email address of the person who made the libraries in that field!!

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+      barcodeLines += Seq("barcode_sequence_1", "barcode_sequence_2", "barcode_name", "library_name").mkString("\t")
+      libraryParamLines += (Seq("BARCODE_1", "BARCODE_2") ++ standardLibraryParams).mkString("\t")
+    }
+    else {
+      barcodeLines += Seq("barcode_sequence_1", "barcode_name", "library_name").mkString("\t")
+      libraryParamLines += (Seq("BARCODE_1") ++ standardLibraryParams).mkString("\t")
+    }
+
+    // Samples
+    val bams = samples.map { sample =>
```
tfenne: I feel like it is bad form to use `map` when also doing side-effecting things, and not just transforming the collection.  Any objections to just declaring a `val bams = ListBuffer[Path]` and using `foreach`?

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+        case (Some(barcode1Bases), Some(barcode2Bases)) => Seq(barcode1Bases, barcode2Bases)
+        case (Some(barcode1Bases), None) => Seq(barcode1Bases)
+        case _ => unreachable(s"No i7 index bases found for sample: ${sample.sampleName}")
+      }
+
+      // Get the description
+      val description = if (includeDescription) Seq(Seq(sample.project, sample.description).flatten.mkString(" - ")) else Seq.empty
+
+      // Add the barcode line
+      barcodeLines += (barcodes ++ Seq(barcodes.mkString, libraryId)).mkString("\t")
```
tfenne: It would be a bit weird, but will this create an invalid file if one or more samples has both I7 and I5 indices, and another sample has only an I7 index in the sample sheet?

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+    // Samples
+    val bams = samples.map { sample =>
+      // Get the library id
+      val libraryId = sample.libraryId.getOrElse(unreachable(s"No library found for sample: ${sample.sampleName}"))
+
+      // Get the barcodes
+      val barcodes = (sample.i7IndexBases, sample.i5IndexBases) match {
+        case (Some(barcode1Bases), Some(barcode2Bases)) => Seq(barcode1Bases, barcode2Bases)
+        case (Some(barcode1Bases), None) => Seq(barcode1Bases)
+        case _ => unreachable(s"No i7 index bases found for sample: ${sample.sampleName}")
```
tfenne: Is the genuinely unreachable?  Or is it possible given a poorly setup sample sheet?

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+      // Add the barcode line
+      barcodeLines += (barcodes ++ Seq(barcodes.mkString, libraryId)).mkString("\t")
+
+      // Add the library param line
+      val bam            = bamFileFrom(output=output, sample=sample, lane=lane)
+      val sampleAlias    = sample.sampleName
+      libraryParamLines += (barcodes ++ Seq(bam, sampleAlias, libraryId) ++ description).mkString("\t")
+      bam
+    }
+    if (bams.toSet.size != bams.size) unreachable("Outputs are not unique when writing the library params file.")
```
tfenne: Why not just check up front if the combinations you need (sample + library) are unique, and if not, emit the list of ones that are going to cause problems?

tfenne: Also, again, is this truly unreachable (i.e. something previously ensures that this won't happen) or is this more correctly a `require(bams.toSet.size == bams.size)`?

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
```diff
@@ -0,0 +1,191 @@
...
+
+  /** The lane-specific file containing information necessary to run Picard's ExtractIlluminaBarcodes and
+    * CollectIlluminaBasecallingMetrics.
+    * @param output the base output directory.
+    * @param lane the lane number.
+    * @return the path to the barcode file, named "barcode_data.lane-<lane>.txt".
+    */
+  def barcodeFileFrom(output: DirPath, lane: Int): FilePath = {
+    require(0 < lane)
+    output.resolve(s"barcode_params.lane-$lane.txt")
```
tfenne: FWIW when I do this I tend to call these files `barcode-params.$lane.txt`.  Illumina tends to use `L00$lane`.  Not sure what's best, but at a minimum it would make me happier if the names used only one of `-` and `_`!

src/main/scala/com/fulcrumgenomics/basecalling/ExtractRunInfo.scala
```diff
@@ -0,0 +1,124 @@
...
+import com.fulcrumgenomics.util.{Io, ReadStructure, SampleBarcode, Template}
+import htsjdk.samtools.util.Iso8601Date
+import dagr.sopt.{arg, clp}
+
+@clp(group=ClpGroups.Basecalling, description=
+  """
+    |Extracts information about an Illumina sequencing run from the RunInfo.xml.
+    |
+    |The output file will contain a header line and a single line containing the following columns:
+    |1. Run Barcode: the unique identifier for the sequencing run and flowcell, stored as "<instrument-name>_<flowcell-barcode>".
```
tfenne: It _might_ be nicer to output this as:
```
run_barcode=foo.123
flowcell_barcode=foo
...
```

As most languages (and even shells) have a way to parse something like that back into a map/dictionary/properties object.  Thoughts?

src/main/scala/com/fulcrumgenomics/basecalling/ExtractRunInfo.scala
```diff
@@ -0,0 +1,124 @@
...
+    |2. Flowcell Barcode: the flowcell barcode.
+    |3. Run Date: the date of the sequencing run.
+    |4. Read Structure: the description of the logical structure of cycles within the sequencing run, including which cycles
+    |   correspond to sample barcodes, molecular barcodes, template bases, and bases that should be skipped.
+    |5. Number of Lanes: the number of lanes in the flowcell.
+  """)
+class ExtractRunInfo
```
tfenne: ExtractIlluminaRunInfo?

src/main/scala/com/fulcrumgenomics/basecalling/ExtractRunInfo.scala
```diff
@@ -0,0 +1,124 @@
...
+(
+  @arg(flag="i", doc="The input RunInfo.xml typically found in the run folder.") val input: FilePath,
+  @arg(flag="o", doc="The output file.") val output: FilePath,
+  @arg(flag="d", doc="The delimiter to use") val delim: String = "\t"
```
tfenne: Surprised to see this.  Figured you'd be using a Metrics class to format the output given the description above.

src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParams.scala
nh13: This currently supports sample sheets that are single or dual indexed

src/main/scala/com/fulcrumgenomics/basecalling/ExtractRunInfo.scala
nh13: Not all workflow language support this unfortunately.  Lets go with the current use-case for now, and then we can change it if needed.  I implemented most of the extraction code in its own object so that it could be an in-jvm dagr task for most of our purposes.

src/main/scala/com/fulcrumgenomics/basecalling/ExtractRunInfo.scala
nh13: I need a CSV file for output, and for dagr tasks, I can just have an in-jvm task to get and store the info myself.

PR #170, opened 2017-02-10
Adding a detailed text output file to review consensus variants with …

tfenne:
…information about each variant and consensus read.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
@@ -96,7 +96,7 @@ object ReviewConsensusVariants {
     */
   private[umi] def toInsertString(r: SAMRecord): String = {
     if (!r.getReadPairedFlag || r.getReadUnmappedFlag || r.getMateUnmappedFlag ||
-      r.getReferenceIndex.toInt != r.getMateReferenceIndex.toInt ||
+      r.getReferenceIndex != r.getMateReferenceIndex ||
       SamPairUtil.getPairOrientation(r) != PairOrientation.FR) {
       "NA"
     }
@@ -236,7 +236,6 @@ class ReviewConsensusVariants
 
       // Setup ready to iterate through the consensus reads
       val consensusCounts  = BaseCounts(consensusPileup.getRecordAndOffsets.asScala)
-
       val rawByMiAndReadNum = groupedPileup.getRecordAndOffsets.asScala.groupBy(r => toMi(r.getRecord) + readNumberSuffix(r.getRecord))
 
       consensusPileup.getRecordAndOffsets
@@ -245,7 +244,7 @@ class ReviewConsensusVariants
         .toSeq.sortBy(r => toMi(r.getRecord) + readNumberSuffix(r.getRecord))
         .foreach { c =>
           val mi = toMi(c.getRecord)
-          val consensusReadName = c.getRecord.getReadName + ( if (c.getRecord.getReadPairedFlag && c.getRecord.getSecondOfPairFlag) "/2" else "/1" )
+          val consensusReadName = c.getRecord.getReadName + readNumberSuffix(c.getRecord)
           val rawCounts = BaseCounts(rawByMiAndReadNum(mi + readNumberSuffix(c.getRecord)))
 
           val m =  ConsensusVariantReviewInfo(
@@ -271,6 +270,7 @@ class ReviewConsensusVariants
           )
 
           writer.write(m)
+          writer.flush()
         }
     }
 
--- a/src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
@@ -25,22 +25,99 @@
 package com.fulcrumgenomics.umi
 
 import java.nio.file.Path
+import java.util.Collections
 
 import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.bam.BaseCounts
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
-import com.fulcrumgenomics.util.Io
+import com.fulcrumgenomics.umi.ReviewConsensusVariants._
+import com.fulcrumgenomics.util.{Io, Metric}
 import dagr.commons.util.LazyLogging
+import dagr.sopt
 import dagr.sopt.{arg, clp}
+import htsjdk.samtools.SamPairUtil.PairOrientation
+import htsjdk.samtools._
 import htsjdk.samtools.reference.{ReferenceSequenceFile, ReferenceSequenceFileFactory}
 import htsjdk.samtools.util.IOUtil.{VCF_EXTENSIONS => VcfExtensions}
-import htsjdk.samtools.util.{IntervalList, SequenceUtil}
-import htsjdk.samtools.{QueryInterval, SAMFileWriterFactory, SAMRecord, SamReaderFactory}
+import htsjdk.samtools.util._
+import htsjdk.variant.variantcontext.VariantContext
 import htsjdk.variant.vcf.VCFFileReader
 
+import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 
-@clp(group=ClpGroups.Umi, description=
+object ReviewConsensusVariants {
+
+  /**
+    * Metric class that is used to output detailed information on each consensus read that carries a variant allele
+    * @param chrom the chromosome on which the variant exists
+    * @param pos the position of the variant
+    * @param ref the reference allele at the position
+    * @param genotype the genotype of the sample in question
+    * @param filters the set of filters applied to the variant in the VCF
+    * @param A the count of A observations at the variant locus
+    * @param C the count of C observations at the variant locus
+    * @param G the count of G observations at the variant locus
+    * @param T the count of T observations at the variant locus
+    * @param N the count of N observations at the variant locus
+    * @param consensus_read the consensus read name for which the following fields contain values
+    * @param consensus_insert a description of the insert that generated the consensus read
+    * @param consensus_call the base call from the consensus read
+    * @param consensus_qual the quality score from the consensus read
+    * @param a the number of As in raw-reads contributing to the consensus base call at the variant site
+    * @param c the number of Cs in raw-reads contributing to the consensus base call at the variant site
+    * @param g the number of Gs in raw-reads contributing to the consensus base call at the variant site
+    * @param t the number of Ts in raw-reads contributing to the consensus base call at the variant site
+    * @param n the number of Ns in raw-reads contributing to the consensus base call at the variant site
+    */
+  case class ConsensusVariantReviewInfo
+  ( // First set is all the same for each variant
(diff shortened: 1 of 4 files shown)


tfenne:
@nh13 There's a bunch going on here, including:
1. The ability to write metrics in a stream (yay!)
2. The ability to filter the set of variants for review by MAF when using a VCF as input
3. Generation of the detailed review file

I think this could be cleaned up a bunch, and improved, once we have our own pileup generation code, but that will come later.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -52,82 +129,197 @@ import scala.collection.mutable.ListBuffer
...
 
     if (VcfExtensions.exists(ext => path.getFileName.toString.endsWith(ext))) {
       val in = new VCFFileReader(input.toFile, false)
-      in.filter(_.isSNP).foreach(v => buffer += LocAndRef(v.getContig, v.getStart, v.getReference.getBases()(0).toChar.toUpper))
+      val sample = this.sample.orElse {
+        if (in.getFileHeader.getSampleNamesInOrder.size() == 1) Some(in.getFileHeader.getSampleNamesInOrder.get(0))
+        else None
+      }
+
+      in.filter(_.isSNP).filter(v => mafFromGenotype(v, sample).forall(_ <= this.maf)).foreach { v => buffer += Variant(
```
nh13: I am not sure if MNP or complex variants are going to bite us here.  Is `isSNP` just for single allele variants?

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -52,82 +129,197 @@ import scala.collection.mutable.ListBuffer
...
 
     buffer.toList
   }
 
+  /** Attempts to determine the minor allele fraction from a VCF genotype object. */
+  private def mafFromGenotype(ctx: VariantContext, sample: Option[String]): Option[Double] = sample match {
+    case None    => None
+    case Some(s) =>
+      val gt = ctx.getGenotype(s)
+      Option(gt.getExtendedAttribute("AF")) match {
```
nh13: I see lots of "AF" hard-coded in various tools (ex. `AlleleFractionField` in MakeMixtureVcf).  Perhaps its time to pull those out?

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -52,82 +129,197 @@ import scala.collection.mutable.ListBuffer
...
+        case Some(af) => Some(af.asInstanceOf[String].toDouble)
+        case None =>
+          val ad = gt.getAD
+          if (ad == null || ad.isEmpty) None
+          else Some (1 - ad(0) / ad.sum.toDouble)
```
nh13: Looks correct, but codecov says its red.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -92,22 +107,18 @@ object Metric {
...
     */
-  def write[T <: Metric](metrics: Seq[T], path: Path)(implicit tt: ru.TypeTag[T]): Unit = {
-    val writer = Io.toWriter(path)
-    write(metrics, writer)
-    writer.close()
-  }
+  def write[T <: Metric](metrics: Seq[T], path: Path)(implicit tt: ru.TypeTag[T]): Unit = write(metrics, Io.toWriter(path))
+
+  /** Returns a MetricWriter that can be used to stream metrics out to a file. */
+  def writer[T <: Metric](path: Path)(implicit tt: ru.TypeTag[T]): MetricWriter[T] = new MetricWriter[T](Io.toWriter(path))
```
nh13: I like these changes.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -25,22 +25,99 @@
...
+    * @param consensus_qual the quality score from the consensus read
+    * @param a the number of As in raw-reads contributing to the consensus base call at the variant site
+    * @param c the number of Cs in raw-reads contributing to the consensus base call at the variant site
+    * @param g the number of Gs in raw-reads contributing to the consensus base call at the variant site
+    * @param t the number of Ts in raw-reads contributing to the consensus base call at the variant site
+    * @param n the number of Ns in raw-reads contributing to the consensus base call at the variant site
+    */
+  case class ConsensusVariantReviewInfo
+  ( // First set is all the same for each variant
+    chrom: String, pos: Int, ref: String, genotype: String, filters: String, A: Int, C: Int, G: Int, T: Int, N: Int,
```
tfenne: It would be nice, but since it also has to play nice with how Metric works, I'm afraid of losing a lot of time trying to find a solution that "just works".

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -25,22 +25,99 @@
...
+  /** Generate a /1 or /2 suffix based on whether a read is first or second of pair or unpaired. */
+  private[umi] def readNumberSuffix(r : SAMRecord): String = if (r.getReadPairedFlag && r.getSecondOfPairFlag) "/2" else "/1"
+
+  /**
+    * Generates a string of the format 'chr1:500-650 | F1R2' for the insert represented by a read.
+    * Only generates strings for FR pairs, all other reads get "NA".
+    */
+  private[umi] def toInsertString(r: SAMRecord): String = {
+    if (!r.getReadPairedFlag || r.getReadUnmappedFlag || r.getMateUnmappedFlag ||
+      r.getReferenceIndex.toInt != r.getMateReferenceIndex.toInt ||
```
tfenne: I was/am concerned that since both methods return java.lang.Integer it might compare without unboxing, but I just tested in the interpreter and it works fine without the `.toInt`

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -25,22 +25,99 @@
...
+      SamPairUtil.getPairOrientation(r) != PairOrientation.FR) {
+      "NA"
+    }
+    else {
+      val outer   = if (r.getReadNegativeStrandFlag) r.getAlignmentEnd else r.getAlignmentStart
+      val isize   = r.getInferredInsertSize
+      val Seq(start, end) = Seq(outer, outer + isize + (if (isize < 0) 1 else -1)).sorted
```
tfenne: This is something else I think we might want to have centralized somewhere - get the start and end of the insert for a paired record with mate on the same chrom, since it's marginally tricky.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -25,22 +25,99 @@
...
+        case (false, true)  => "F2R1"
+        case (false, false) => "F1R2"
+      }
+
+      s"${r.getReferenceName}:${start}-${end} | ${pairing}"
+    }
+  }
+}
+
+@sopt.clp(group=ClpGroups.Umi, description=
```
tfenne: I'm done fighting it.  It's valid, and until Intellij stops losing it's mind it's just not worth fighting.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -52,82 +129,197 @@ import scala.collection.mutable.ListBuffer
...
   Io.assertReadable(Seq(input, consensusBam, groupedBam, ref))
   Io.assertCanWriteFile(output)
   private val refFile = ReferenceSequenceFileFactory.getReferenceSequenceFile(ref)
   private val dict = refFile.getSequenceDictionary
 
-  private[umi] case class LocAndRef(chrom: String, start: Int, refBase: Char) {
+  /** Simple case class to hold the relevant information from a variant context for easy access. */
+  private[umi] case class Variant(chrom: String, start: Int, refBase: Char, genotype: Option[String], filters: Option[String]) {
     def toQueryInterval: QueryInterval = new QueryInterval(dict.getSequenceIndex(chrom), start, start)
+    def toInterval: Interval = new Interval(chrom, start, start)
```
tfenne: The second line of the usage does state "SNPs only".

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -52,82 +129,197 @@ import scala.collection.mutable.ListBuffer
...
+    * Generates the detailed consensus review file.
+    *
+    * @param consensusBam the original consensus bam (not the review consensus bam)
+    * @param groupedBam the grouped bam (the review version)
+    * @param variants the set of variants to report on
+    * @param output the file to write
+    */
+  private def generateDetailsFile(consensusBam: PathToBam, groupedBam: PathToBam, variants: Seq[Variant], output: FilePath): Unit = {
+    val consensusReader = SamReaderFactory.make().open(consensusBam)
+    val groupedReader   = SamReaderFactory.make().open(groupedBam)
```
tfenne: I guess?  CRAM still seems pretty broken to me and I'm not super keen on passing more stuff around to support CRAM until it actually works.

Perhaps what we should do is centralize all generation of readers/writers into methods in the `Bams` object so it's easy to track them down?

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -52,82 +129,197 @@ import scala.collection.mutable.ListBuffer
...
+            N = consensusCounts.n,
+            consensus_read = consensusReadName,
+            consensus_insert = toInsertString(c.getRecord),
+            consensus_call = c.getReadBase.toString.toUpperCase,
+            consensus_qual = c.getBaseQuality,
+            a = rawCounts.a,
+            c = rawCounts.c,
+            g = rawCounts.g,
+            t = rawCounts.t,
+            n = rawCounts.n
```
tfenne: I don't really see the point ... that's actually _more_ code since this is the only use of the constructor.  So I'd have to have a second apply() method that does this, but still the main constructor/apply to keep Metric happy.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
tfenne: `isSNP` is for single-nucleotide variants, though it can include ones with > 1 alt allele.  However, we're looking at a single sample's genotype which I'm making the assumption is diploid.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
tfenne: I'm honestly not sure what the right thing to do here is.  It's _not_ a field that is standardized in VCF, so some tools use `AF` others use other fields.  `AlleleFractionField` in `MakeMixtureVcf` represents the choice in that tool to _write_ a field called `AF`, but note that it supports the user's choice of input field.  Freakin VCF.

PR #171, opened 2017-02-13
Use sbt 2.11.8

nh13:


tfenne:
Can you just re-write the commit message to say `scala 2.11.8` instead of `sbt 2.11.8` for posterity, then merge?

PR #172, opened 2017-02-16
Add the instrument name to ExtractIlluminaRunInfo.

nh13:


src/main/scala/com/fulcrumgenomics/basecalling/ExtractIlluminaRunInfo.scala
```diff
@@ -38,10 +38,11 @@ import dagr.sopt.{arg, clp}
     |The output file will contain a header line and a single line containing the following columns:
     |1. Run Barcode: the unique identifier for the sequencing run and flowcell, stored as "<instrument-name>_<flowcell-barcode>".
     |2. Flowcell Barcode: the flowcell barcode.
-    |3. Run Date: the date of the sequencing run.
-    |4. Read Structure: the description of the logical structure of cycles within the sequencing run, including which cycles
+    |3. Instrument Name: the instrument name.
```
tfenne: I'm just realizing that it would be nice in the usage to have the actual column names (e.g. `instrument_name`) vs. how you have it now.

nh13:
@tfenne this one became more involved as I added and deprecated a few methods in `Metric`.  I also changed the `ExtractIlluminaRunInfo` tool to emit a metric.  Unfortunately, even using a secondary `String` constructor for `ReadStructure` I can't use a `ReadStructure` in a metric.  I'll think about adding support in Dagr.

src/main/scala/com/fulcrumgenomics/basecalling/ExtractIlluminaRunInfo.scala
```diff
@@ -27,74 +27,67 @@ package com.fulcrumgenomics.basecalling
...
-  flowcellBarcode: String,
-  runDate: Iso8601Date,
-  readStructure: ReadStructure,
-  numLanes: Int
-)
+( run_barcode: String,
+  flowcell_barcode: String,
+  instrument_name: String,
+  run_date: Iso8601Date,
+  read_structure: String, // NB: would like this to be ReadStructure
```
tfenne: Does it not work?  Will Metric not toString it and use the `apply(String)` to re-hydrate it?

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -103,19 +103,45 @@ object Metric {
...
+
+  /** Writes one or more metrics to the given path.  The first line will be a header with the field names.  Each subsequent
+    * line is a single metric.
+    */
+  def write[T <: Metric](path: Path, metric: T*)(implicit tt: ru.TypeTag[T]): Unit = write(Io.toWriter(path), metric:_*)
+
+  /** Writes a metric to the given writer.  The first line will be a header with the field names.  Each subsequent
+    * line is a single metric.
+    */
+  def write[T <: Metric](writer: Writer, metrics: TraversableOnce[T])(implicit tt: ru.TypeTag[T]): Unit = write(writer, metrics.toSeq:_*)
```
tfenne: Something doesn't feel awesome about accepting a `TraversableOnce` and then immediately calling `toSeq` on it.  I.e. if I provided an `Iterator[MyMetric]` that produced a ton of metrics, I might get surprised by this.

nh13: Good point.  I'll fix this but note that I am not fixing up metric reading, which also uses `Seq` and doesn't have a `MetricReader`.

nh13:
@tfenne one last look?

src/main/scala/com/fulcrumgenomics/basecalling/ExtractIlluminaRunInfo.scala
```diff
@@ -27,74 +27,67 @@ package com.fulcrumgenomics.basecalling
...
-  flowcellBarcode: String,
-  runDate: Iso8601Date,
-  readStructure: ReadStructure,
-  numLanes: Int
-)
+( run_barcode: String,
+  flowcell_barcode: String,
+  instrument_name: String,
+  run_date: Iso8601Date,
+  read_structure: ReadStructure, // NB: would like this to be ReadStructure
```
tfenne: Comment can go now.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -31,13 +31,16 @@ import scala.reflect.ClassTag
 // would be good to have support for * and + to indicate zero or more and one or more bases.
 object ReadStructure {
   /** Creates a read structure from a string */
-  def apply(readStructure: String): ReadStructure = {
+  def apply(readStructure: String): ReadStructure = new ReadStructure(segs=segments(readStructure=readStructure))
+
+  /** Creates a sequence of read segments from a string. */
+  private[ReadStructure] def segments(readStructure: String): Seq[ReadSegment] = {
```
tfenne: Isn't `private[ReadStructure]` the same thing as `private` here?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -94,6 +97,9 @@ object ReadStructure {
   */
 class ReadStructure(segs: Seq[ReadSegment], resetOffsets: Boolean = false) extends immutable.Seq[ReadSegment] {
   import ReadStructure.SubRead
+
+  def this(readStructure: String) = this(ReadStructure.segments(readStructure=readStructure))
```
tfenne: Can you just add a one-line comment here please?

PR #173, opened 2017-02-20
Improve control of sorting in `Bams` sorting methods

nh13:
I am indifferent to exposing the "reads in memory" option in `ClipOverlappingReads` but I do like the change to `Bams` to make the `maxInMemory` parameter `Optional` as well as adding in a param `tagsToRemove` in case we have very many very large per-record tags.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
@@ -112,6 +112,10 @@ object Template {
   * Utility methods for working with BAMs.
   */
 object Bams extends LazyLogging {
+
+  /** The default maximum # of records to keep and sort in memory. */
+  val MaxInMemory: Int = 1e6.toInt
+
   /**
     * Builds a sorting collection for sorting SAMRecords.
     *
@@ -123,7 +127,7 @@ object Bams extends LazyLogging {
     */
   def sortingCollection(order: SortOrder,
                         header: SAMFileHeader,
-                        maxInMemory: Int = 1e6.toInt,
+                        maxInMemory: Int = MaxInMemory,
                         tmpDir: Option[DirPath] = None): SortingCollection[SAMRecord] = {
     SortingCollection.newInstance(
       classOf[SAMRecord],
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -112,24 +112,28 @@ object Template {
...
     * @param header the header to use to encode records during sorting
-    * @param maxInMemory the maximum number of records to keep and sort in memory
+    * @param maxInMemory optionally the maximum number of records to keep and sort in memory
     * @param tmpDir an optional temp directory to use for temporary sorting files
     * @return a SortingCollection
     */
   def sortingCollection(order: SortOrder,
                         header: SAMFileHeader,
-                        maxInMemory: Int = 1e6.toInt,
+                        maxInMemory: Option[Int] = None,
```
tfenne: I don't like this change.  What does it mean to set it to `None` exactly?  That there will be none kept in memory?  Since you immediately substitute in the default value, what is the point of this change other than obfuscating that there is a default?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -140,21 +144,25 @@ object Bams extends LazyLogging {
...
     * @param tmpDir an optional temp directory to use for temporary sorting files if needed
+    * @param tagsToRemove an optional list of SAM tags to remove after reading a record.
     * @return an Iterator with reads from the same query grouped together
     */
   def queryGroupedIterator(in: SamReader,
-                           maxInMemory: Int = 1e6.toInt,
-                           tmpDir: Option[DirPath] = None): BetterBufferedIterator[SAMRecord] = {
+                           maxInMemory: Option[Int] = None,
+                           tmpDir: Option[DirPath] = None,
+                           tagsToRemove: Seq[String] = Seq.empty): BetterBufferedIterator[SAMRecord] = {
```
tfenne: This is super-specific, and feels like a response to a particular problem. Instead of doing this, how about creating versions of these methods that take in `SAMFileHeader` and `Iterator[SAMRecord]` (that the current versions could delegate to), thus allowing you to `filter`, `map`, etc. on the iterators to your heart's content before you pass them in rather than building in one specific kind of transform here?

src/main/scala/com/fulcrumgenomics/bam/ClipOverlappingReads.scala
```diff
@@ -59,21 +59,25 @@ class ClipOverlappingReads
   @arg(flag="o", doc="Output SAM or BAM file.") val output: PathToBam,
   @arg(flag="s", doc="Soft clip reads instead of hard clipping.") val softClip: Boolean = false,
   @arg(flag="r", doc="Reference sequence fasta file.") val ref: PathToFasta,
-  @arg(flag="a", doc="Automatically clip extended attributes that are the same length as bases.") val autoClipAttributes: Boolean = false
+  @arg(flag="a", doc="Automatically clip extended attributes that are the same length as bases.") val autoClipAttributes: Boolean = false,
+  @arg(flag="m", doc="The maximum # of reads in memory for any internal sorter") val maxReadsInMemory: Option[Int] = None,
+  @arg(flag="t", doc="SAM tags to remove from the input") val tagsToRemove: Seq[String] = Seq.empty
```
tfenne: I really don't like having this here.  I feel like it has nothing to do with clipping overlapping reads, and is a first step down a picard-like path of adding odd-ball options to tools to workaround other problem (e.g. memory usage).  If the tags are really a substantial part of the memory, why not just pipe the data through a tool to strip out tags first, instead of building that in here (and potentially elsewhere)?

nh13:
Talke in-person, we agreed on:
- a tool to strip tags
- adding an `Iterator/SamFileHeader` interface to `Bams`.

tfenne:
I restarted your builds that had failed for a silly reason (temporary failure to download `r-core`, but now their failing for real because there are two methods in `Bams` with the same names and default arguments 😞 .

nh13:
@tfenne I think I will keep this tool simple for now as I totally agree that the expanded tool you propose is worthwhile, I just don't have a need for it at the moment.  If and when we get around to writing it, we can deprecate and remove this tool, and standardize on that one.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -145,16 +149,34 @@ object Bams extends LazyLogging {
...
+    * @param maxInMemory the maximum number of records to keep and sort in memory if sorting is needed
+    * @param tmpDir an optional temp directory to use for temporary sorting files if needed
+    * @return an Iterator with reads from the same query grouped together
+    */
+  def queryGroupedIterator(iterator: Iterator[SAMRecord],
+                           header: SAMFileHeader,
+                           maxInMemory: Int,
+                           tmpDir: Option[DirPath]): BetterBufferedIterator[SAMRecord] = {
+    if (header.getSortOrder == SortOrder.queryname || header.getGroupOrder == GroupOrder.query) {
+      iterator.bufferBetter
```
nh13: I opted to have the same name rather than using something like the [Magnet pattern](http://spray.io/blog/2012-12-13-the-magnet-pattern/).

src/main/scala/com/fulcrumgenomics/bam/RemoveSamTags.scala
```diff
@@ -0,0 +1,60 @@
...
+
+import com.fulcrumgenomics.FgBioDef.{PathToBam, javaIterableToIterator}
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.util.Io
+import dagr.commons.CommonsDef.SafelyClosable
+import dagr.sopt.{arg, clp}
+import htsjdk.samtools.{SAMFileWriter, SAMFileWriterFactory, SamReader, SamReaderFactory}
+
+@clp(
+  description = "Removes SAM tags from a SAM or BAM file.  If no tags to remove are given, the original file is produced.",
```
tfenne: Can you just validate that running this without a `--tags-to-remove` works, or just kill the second sentence?  I kinda recall that it'll force you to supply at least one value by default if there's no default value.

PR #174, opened 2017-02-21
Improved error/warning logging for EndRepairArtifactFilter.

tfenne:


src/main/scala/com/fulcrumgenomics/vcf/filtration/EndRepairArtifactLikelihoodFilter.scala
```diff
@@ -135,8 +135,14 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
     * skewed towards the artifact at very low MAFs, and towards mutations as the MAF increases towards 0.5.
     */
   private[filtration] def priors(pileup: Pileup[PileupEntry], maf: Double) : (Double, Double) = {
-    require(maf > 0 && maf <= 1, "maf must be greater than 0 and no more than 1.")
-    val priorMutation = min(pow(maf * 2, 2), 0.9999)
+    require(maf >= 0 && maf <= 1, s"Maf must be between 0 and 1. Observed maf ${maf} at ${pileup.refName}:${pileup.pos}.")
+
+    val m = if (maf != 0) maf else {
```
tfenne: This should only happen when the filtering is run with different base/mapq cutoffs than the caller is run.  Unfortunately some callers do not make it clear what cutoffs and filters they use, so it's a little hard to avoid.  In the case that the pileup doesn't actually contain any alt allele it estimates it as 1/depth.

src/main/scala/com/fulcrumgenomics/vcf/filtration/EndRepairArtifactLikelihoodFilter.scala
```diff
@@ -135,8 +135,14 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+      logger.warning(s"No alt allele observations found with selected cutoffs at: ${pileup.refName}:${pileup.pos}")
+      1.0 / pileup.depth
```
nh13: `1 / pileup.depth.toDouble` ;)

PR #176, opened 2017-02-22
Support for cross-building scala 2.11 and 2.12.

tfenne:
I slightly more involved PR since I had to update some plugin versions and also fix the scala-xml dependency.

I believe this will fail until the `dagr` PR for cross-building gets merged and a 2.12 snapshot published.


nh13:
👍

PR #179, opened 2017-02-24
Adding a tool to demultiplex FASTQs.

nh13:


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+          .getOrElse(illegal(s"Sample with id '${sample.sampleId}' did not have a sample barcode"))
+          .concatenatedBarcode
+      val libraryId: String = sample.libraryId.getOrElse(illegal(s"Sample with id '${sample.sampleId}' did not have a sample barcode"))
+      SampleBarcodeMetric(barcodeName=sample.sampleName, libraryName=libraryId, barcode=barcode)
+    }
+    sampleMetrics.toSeq ++ Seq(SampleBarcodeMetric(UnmatchedSampleId, UnmatchedSampleId, noMatchBarcode))
+  }
+
+  /** Increments the metrics for the given sample ordinal, which matched with the given number of mismatches.  The
+    * sample ordinal less than zero indicates the "unmatched" sample. */
```
nh13: less than zero not true anymore -> remove

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+  def expectedNumberOfReads: Int = readStructures.length
+
+  /** True if we the read structure implies paired end reads will be produced, false otherwise. */
+  val pairedEnd: Boolean = readStructures.count(_.template.nonEmpty) match {
+    case 1 => false
+    case 2 => true
+    case n => throw new IllegalArgumentException(s"Found $n read structures with template bases but expected 1 or 2.")
+  }
+
+  /** The concatenation of all read structures read structure. */
```
nh13: remove extra `read structure`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+  require(readStructures.nonEmpty, "No read structures were given")
+  require(samples.map(_.sampleOrdinal).toList == samples.map(_.sampleOrdinal).sorted, "Samples not provided in ascending order by sample ordinal")
+  require(samples.flatMap(_.sampleBarcode).map(_.concatenatedBarcode).sorted.distinct.length == samples.length, "Unique sample barcodes required for all samples")
+  require(headers.length == samples.length + 1, s"Not enough SAMFileHeaders; found ${headers.length} but should have ${samples.length+1}.")
+  require(headers.forall(_.getReadGroups.size == 1), "All headers must have one read group. Ordinals are: " + headers.zipWithIndex.filter(_._1.getReadGroups.size != 1).map(_._2).mkString(", "))
+
+  /** The number of reads that are expected to be given to the [[demultiplex()]]. */
+  def expectedNumberOfReads: Int = readStructures.length
+
+  /** True if we the read structure implies paired end reads will be produced, false otherwise. */
```
nh13: change to: `True if the read structure implies paired end reads will be produced, false otherwise`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+                                 val maxNoCalls: Int = 2) {
+  import FastqDemultiplexer._
+
+  require(readStructures.nonEmpty, "No read structures were given")
+  require(samples.map(_.sampleOrdinal).toList == samples.map(_.sampleOrdinal).sorted, "Samples not provided in ascending order by sample ordinal")
+  require(samples.flatMap(_.sampleBarcode).map(_.concatenatedBarcode).sorted.distinct.length == samples.length, "Unique sample barcodes required for all samples")
+  require(headers.length == samples.length + 1, s"Not enough SAMFileHeaders; found ${headers.length} but should have ${samples.length+1}.")
+  require(headers.forall(_.getReadGroups.size == 1), "All headers must have one read group. Ordinals are: " + headers.zipWithIndex.filter(_._1.getReadGroups.size != 1).map(_._2).mkString(", "))
+
+  /** The number of reads that are expected to be given to the [[demultiplex()]]. */
```
nh13: add `method` to the end of the sentence.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+      |columns
+      |  - Sample Identifier:  Sample_ID
+      |  - Sample Name:        Sample_Name
+      |  - Library Identifier: Library_ID
+      |  - Sample Project:     Sample_Project
+      |  - Description:        Description
+      |  - Sample Barcode:     Index
+      |The expected sample barcode should be placed in the 'Index' column.  If the sample barcode is present across
+      |multiple reads (ex. dual-index, or inline in both reads of a pair), then the expected barcode bases from each read
+      |should be concatenated and placed in teh 'Index' column.  The concatenation should be in the same order as the
```
nh13: teh typo

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+  /** Get the molecular barcode bases across all bases. */
+  private def molecularBarcode(allBases: String): String = {
+    ReadStructure.structureRead(bases=allBases, segments=molecularBarcodeSegments)
+      .map(_.bases)
+      .filter(_.nonEmpty)
+      .mkString("-")
+  }
+
+    /** Gets the sample ordinal given the bases from all cycles and the number of mismatches between the bases and matched
+      * sample barcode.  If no match is found, the ordinal for the oridinal for the 'unmatched" sample is given. */
```
nh13: indentation.

nh13: the ordinal for the oridinal for the 'unmatched" -> the "unmatched" ordinal

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+    * structures.
+    *
+    * The sample barcoded bases from each read are extracted and concatenated in the same order as the given reads. They
+    * are matched against the sample barcode bases for each sample.  The
+    * */
+  def demultiplex(reads: FastqRecord*): DemuxRecord = {
+    require(reads.nonEmpty, "No reads given for demultiplexing.")
+    require(reads.length == expectedNumberOfReads, s"Expected '$expectedNumberOfReads' number of reads but found '${reads.length}'.")
+
+    // Get the sample ordinal
```
nh13: _sample index_

src/test/scala/com/fulcrumgenomics/fastq/DemuxFastqsTest.scala
```diff
@@ -0,0 +1,336 @@
...
+import com.fulcrumgenomics.fastq.DemuxFastqs.UnmatchedSampleId
+import com.fulcrumgenomics.fastq.FastqDemultiplexer.DemuxRecord
+import org.scalatest.OptionValues
+import com.fulcrumgenomics.util.miseq.{Sample, SampleBarcode, SampleSheet}
+import dagr.commons.io.PathUtil
+import htsjdk.samtools.SAMUtils
+
+import scala.collection.mutable.ListBuffer
+
+class DemuxFastqsTest extends UnitSpec with OptionValues with ErrorLogLevel {
```
nh13: @tfenne I really like both the `OptionValues` and `ErrorLogLevel` traits.  The former allows us to replace the `get` method with the `value` method for `Option`s and scalatest report a meaningful error when the `Option` is `None`.  The latter allows us to turn off logging to the console when running end-to-end `tool.execute()` tests.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+  * @param minMismatchDelta the minimum difference between number of mismatches in the best and second best barcodes for
+  *                         a barcode to be considered a match.
+  * @param maxNoCalls the maximum number of no calls in the sample barcode bases allowed for matching.
+  */
+private class FastqDemultiplexer(val headers: Seq[SAMFileHeader],
+                                 val samples: Seq[Sample],
+                                 val readStructures: Seq[ReadStructure],
+                                 val umiTag: String = "RX",
+                                 val qualityFormat: FastqQualityFormat = FastqQualityFormat.Standard,
+                                 val minQ: Int = 0,
```
tfenne: I feel like enforcement of the min/max quality doesn't belong here.  This is really validation that the fastqs aren't borked, and surely that should reside with the thing reading the fastqs?  Perhaps we should add options to FastqSource to translate qualities and enforce the quality range, so that it's re-useable elsewhere?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+  require(headers.forall(_.getReadGroups.size == 1), "All headers must have one read group. Ordinals are: " + headers.zipWithIndex.filter(_._1.getReadGroups.size != 1).map(_._2).mkString(", "))
+
+  /** The number of reads that are expected to be given to the [[demultiplex()]]. */
+  def expectedNumberOfReads: Int = readStructures.length
+
+  /** True if we the read structure implies paired end reads will be produced, false otherwise. */
+  val pairedEnd: Boolean = readStructures.count(_.template.nonEmpty) match {
+    case 1 => false
+    case 2 => true
+    case n => throw new IllegalArgumentException(s"Found $n read structures with template bases but expected 1 or 2.")
```
tfenne: Move this check up into DemuxFastqs?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+  }
+
+  /** The concatenation of all read structures read structure. */
+  private val fullReadStructure: ReadStructure = new ReadStructure(segs=readStructures.flatten, resetOffsets=true)
```
tfenne: I see where this is going, but I think there's a problem.  I routinely see data generated on MiSeqs where the fastqs were generated with the option to trim adapters.  The result is that the template reads are not fixed length, and thus you can't just concatenate the read structures and templates like this.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+
+  /** The quality format converter. */
+  private val solexaQualityConverter: SolexaQualityConverter = SolexaQualityConverter.getSingleton
```
tfenne: The `SolexaQualityConverter` is pretty icky for a bunch of reasons.  I wonder if it would make sense to port it to scala (it's only about a dozen lines of actual code) and also offer `String => String` versions of the methods?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+
+  /** The sub-reads for the sample barcodes across all read structures. */
+  private val sampleBarcodeSegments = fullReadStructure.sampleBarcode
+
+  /** Get the sample barcode bases across all bases. */
+  private def sampleBarcodeBases(allBases: String): String = {
+    ReadStructure.structureRead(bases=allBases, segments=sampleBarcodeSegments).map(_.bases).mkString
```
tfenne: I think I see why there are both, but it would be nice to note why you have this and also the one below.  Could even just be a single method that takes in the delimiter so you can use it both ways?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+    val numNoCalls      = observedBarcode.count(base => SequenceUtil.isNoCall(base))
+
+    // Get the best and second best sample barcode matches.
+    val (bestSampleIndex, bestMismatches, secondBestMismatches) = if (numNoCalls <= maxNoCalls) {
+      samples.map { sample =>
+        val expectedBarcode = sample.sampleBarcode
+          .getOrElse(unreachable(s"Sample with id '${sample.sampleId}' did not have a sample barcode"))
+          .barcodeBytes
+        val numMismatches = countMismatches(observedBarcode, expectedBarcode)
+        (sample.sampleOrdinal-1, numMismatches)
```
tfenne: Why `sampleOrdinal - 1`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+      }
+      .toList.sortBy(_._2).take(2) match {
+        case Nil                              => (-1,           Int.MaxValue, Int.MaxValue)
+        case List(bestTuple)                  => (bestTuple._1, bestTuple._2, Int.MaxValue)
+        case List(bestTuple, secondBestTuple) => (bestTuple._1, bestTuple._2, secondBestTuple._2)
+      }
+    }
+    else {
+      (this.samples.size, Int.MaxValue, Int.MaxValue)
```
tfenne: It would be nice to extract `this.samples.size` into a class level val called `UnmatchedSampleOrdinal` and use that everywhere to make it clear what's going on.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+    require(reads.length == expectedNumberOfReads, s"Expected '$expectedNumberOfReads' number of reads but found '${reads.length}'.")
+
+    // Get the sample ordinal
+    val allBases                     = reads.map(_.bases).mkString
+    val (sampleIndex, numMismatches) = matchSampleBarcode(allBases)
+
+    // Create the SAMRecords.
+    val header            = headers(sampleIndex)
+    val sampleId          = header.getReadGroups.get(0).getId
+    val molecularBarcode  = this.molecularBarcode(allBases)
```
tfenne: prefixing method names with `this` is generally frowned upon

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+    val records           = reads.zip(readStructures).flatMap { case (read, readStructure) =>
+      val templateSegments = readStructure.template
+      if (templateSegments.isEmpty) {
+        None
+      }
+      else {
+        val subReads: Seq[SubRead] = ReadStructure.structureReadWithQualities(bases=read.bases, qualities=read.quals, segments=templateSegments)
+        val bases                  = subReads.map(_.bases).mkString
+        val quals                  = subReads.flatMap(_.quals).mkString
+        require(bases.length == quals.length, s"Bases and qualities have differing lengths forread: ${read.header}")
```
tfenne: forread -> for read

tfenne:
Thinking about my point 4 a bit more, I think you could literally create a single case class:
`case class SampleInfo(sample: Sample: writer: SAMFileWriter, header: SAMFileHeader, rg: SAMReadGroupRecord)`, make a `Seq` of those, and pass the `Seq` around to things that need to see all the samples, and a `SampleInfo` anywhere you need just one, then all the lookups could be avoided without too much surgery.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+    val unmatchedWriter = {
+      val outputBam: FilePath   = output.resolve(unmatched)
+      val header: SAMFileHeader = sampleToHeader(new Sample(writers.length, UnmatchedSampleId, UnmatchedSampleId, Some(UnmatchedSampleId), None, None))
+      toWriter(header, outputBam)
+    }
+    writers ++ List(unmatchedWriter)
+  }
+
+  /** Gets the quality format of the FASTQs. */
+  private def determineQualityFormat(fastqs: Seq[PathToFastq], qualityFormat: Option[FastqQualityFormat] = None, logger: Option[Logger] = None): FastqQualityFormat = {
```
nh13: It's actually the expected quality format.  If it's `None`, it guesses the best format, and if it is `Some(format)` then it verifies the quality format.  This is the same behavior as Picard's `FastqToSam`, which I am trying to keep a little bit of.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+      |the order of the reads' FASTQs and read structures given to this tool.
+      |
+      |The read structures will be used to extract the observed sample barcode, template bses, and molecular identifiers
+      |from each read.  The observed sample barcode will be matched to the sample barcodes extracted from the bases in
+      |the sample sheet and associated read structures.
+    """,
+  group=ClpGroups.Fastq
+)
+class DemuxFastqs
+(@arg(flag="f", doc="One or more input fastq files each corresponding to a sub-read (ex. index read, read one, read two, fragment).") val fastqs:  Seq[PathToFastq],
```
nh13: If you have multiple FASTQs, then yes, you will need to run this tool once per group of FASTQs, or at least combine them prior to running this tool.  As a by-product, the former allows for simple parallelization in a pipeline!

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+    * @param numMismatches the # of mismatches if it matched a sample with a sample barcode, -1 otherwise (the 'unmatched'
+    *                      sample).
+    * @param records the records, one for each read that have template bases.
+    */
+  case class DemuxRecord(sampleIndex: Int, numMismatches: Int, records: Seq[SAMRecord])
+
+  /** Counts the nucleotide mismatches between two strings of the same length.  Ignores no calls in expectedBases.
+    * Observed base qualities less than the minimum base quality are counted as mismatches if not a no call.
+    */
+  private[fastq] def countMismatches(observedBases: Array[Byte], expectedBases: Array[Byte]): Int = {
```
nh13: The optimization in `matchSampleBarcode` is not to even check against _any_ sample barcodes if there are too many no calls in the observed bases.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,614 @@
...
+    var idx = 0
+    var count = 0
+    while (idx < observedBases.length) {
```
nh13: I like _while_ every once in a while.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
nh13: Another day.

nh13:
@tfenne I think I addressed all your concerns, and made a note where I did not, which were few and far between.  I think `Sample` and `SampleSheet` could be further improved, but I running out of steam to do that here.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,573 @@
...
+    require(reads.nonEmpty, "No reads given for demultiplexing.")
+    require(reads.length == expectedNumberOfReads, s"Expected '$expectedNumberOfReads' number of reads but found '${reads.length}'.")
+
+    // Get the sample
+    val allBases                    = reads.map(_.bases)
+    val (sampleInfo, numMismatches) = matchSampleBarcode(allBases)
+
+    // Create the SAMRecords.
+    val header            = sampleInfo.header
+    val sampleId          = sampleInfo.rg.getId
```
tfenne: This should probably be `rgId`.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,573 @@
...
+
+        // Build the [[SAMRecord]]
+        val record  = new SAMRecord(header)
+        record.setReadName(read.name)
+        record.setReadString(bases)
+        record.setReadUnmappedFlag(true)
+        if (pairedEnd) {
+          record.setReadPairedFlag(true)
+          record.setMateUnmappedFlag(true)
+          if (readNumber == 1) {
```
tfenne: This could be:
```scala
readNumber match {
  case 1 => record.setFirstOfPairFlag(true)
  case 2 => record.setSecondOfPairFlag(true) 
  case n => throw new IllegalStateException(s"Invalid read number '$n' for read: ${read.header}")
}
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,573 @@
...
+            record.setFirstOfPairFlag(true)
+          }
+          else {
+            require(readNumber == 2, s"Expected read number 2 but found '$readNumber' for read: ${read.header}")
+            record.setSecondOfPairFlag(true)
+          }
+        }
+        record.setAttribute("BC", sampleBarcode)
+        record.setAttribute(ReservedTagConstants.READ_GROUP_ID, sampleId)
+        if (molecularBarcode.nonEmpty) record.setAttribute("RX", molecularBarcode)
```
tfenne: I believe this should be `this.umiTag` not `"RX"`.

src/main/scala/com/fulcrumgenomics/illumina/Sample.scala
```diff
@@ -42,22 +46,30 @@ package com.fulcrumgenomics.util.miseq
...
   extendedAttributes.keysIterator.foreach { key => require(key == key.toUpperCase) }
 
+  import Sample.SampleBarcodeDelimiter
+
+  def sampleBarcodes: Seq[Option[String]] = Seq(i7IndexBases, i5IndexBases)
+
+  lazy val sampleBarcodeBytes: Array[Byte]  = sampleBarcodes.flatten.filter(_.nonEmpty).flatMap(_.getBytes).toArray
+  lazy val sampleBarcodeString: String      = sampleBarcodes.flatten.filter(_.nonEmpty).mkString(SampleBarcodeDelimiter)
+  override def equals(o: Any): Boolean = o.isInstanceOf[Sample] && this.sampleBarcodeString == o.asInstanceOf[Sample].sampleBarcodeString
+  override def hashCode: Int           = this.sampleBarcodeString.hashCode
```
tfenne: I'm not sure about this.  This essentially means that if you have a sample sheet that doesn't have sample barcodes defined, then the hashCode will be `"".hashCode` which seems wrong.  Is the goal here to just have something faster than the standard case class `hashCode`?  Maybe you could `+ sampleOrdinal`?

src/main/scala/com/fulcrumgenomics/illumina/Sample.scala
```diff
@@ -42,22 +46,30 @@ package com.fulcrumgenomics.util.miseq
...
+                  extendedAttributes: Map[String, String] = Map.empty) {
   extendedAttributes.keysIterator.foreach { key => require(key == key.toUpperCase) }
 
+  import Sample.SampleBarcodeDelimiter
+
+  def sampleBarcodes: Seq[Option[String]] = Seq(i7IndexBases, i5IndexBases)
+
+  lazy val sampleBarcodeBytes: Array[Byte]  = sampleBarcodes.flatten.filter(_.nonEmpty).flatMap(_.getBytes).toArray
+  lazy val sampleBarcodeString: String      = sampleBarcodes.flatten.filter(_.nonEmpty).mkString(SampleBarcodeDelimiter)
+  override def equals(o: Any): Boolean = o.isInstanceOf[Sample] && this.sampleBarcodeString == o.asInstanceOf[Sample].sampleBarcodeString
```
tfenne: Why not leave the default case class implementation of equals?  This seems over-simplified.

src/main/scala/com/fulcrumgenomics/illumina/Sample.scala
```diff
@@ -42,22 +46,30 @@ package com.fulcrumgenomics.util.miseq
...
+  override def hashCode: Int           = this.sampleBarcodeString.hashCode
+  override def toString: String        = this.sampleBarcodeString
```
tfenne: Same comment here.  This seems very specific to your use in DemuxFastqs.  Perhaps the toString could be at least `sampleName/sampleBarcodeString`?

src/test/scala/com/fulcrumgenomics/illumina/SampleTest.scala
```diff
@@ -54,6 +54,22 @@ class SampleTest extends UnitSpec with OptionValues{
   }
 
   it should "support throw an exception if extended attribute keys are not all uppercase" in {
-    an[Exception] should be thrownBy new Sample(0, "ID", "NAME", extendedAttributes=Seq(("foo", "1"), ("bar", "2  ")).toMap)
+    an[Exception] should be thrownBy new Sample(0, "ID", "NAME", "LIBRARY", extendedAttributes=Seq(("foo", "1"), ("bar", "2  ")).toMap)
+  }
+
+  "Sample.hashCode" should "be the hash code of the concatenated barcode sequences" in {
```
tfenne: I think it would be better to write these tests to test the desired behaviour (e.g. hashCodes don't result in collisions for different samples) rather than the specific implementation.

src/test/scala/com/fulcrumgenomics/fastq/DemuxFastqsTest.scala
```diff
@@ -0,0 +1,524 @@
...
+      |151,,,,,,,,,
+      |,,,,,,,,,
+      |[Settings],,,,,,,,,
+      |ReverseComplement,0,,,,,,,,
+      |Adapter,AGATCGGAAGAGCACACGTCTGAACTCCAGTCA,,,,,,,,
+      |AdapterRead2,AGATCGGAAGAGCGTCGTGTAGGGAAAGAGTGT,,,,,,,,
+      |,,,,,,,,,
+      |[Data],,,,,,,,,
+      |Sample_ID,Sample_Name,Sample_Plate,Sample_Well,index,index2,Sample_Barcode,Sample_Project,Description
+      |20000101-EXPID-1,Sample_Name_1,,,,,AAAAAAAAGATTACAGA,Sample_Project_1,Description_1
```
tfenne: Since you use these elsewhere it might be nice to extract the barcode sequences into four vals and use `${barcode1}` syntax to insert them here.  Then you can refer to them in tests without hard-coding them in multiple places.

src/test/scala/com/fulcrumgenomics/fastq/DemuxFastqsTest.scala
```diff
@@ -0,0 +1,524 @@
...
+    demuxRecord.records.length shouldBe 2
+    val r1 = demuxRecord.records.headOption.value
+    val r2 = demuxRecord.records.lastOption.value
+
+    r1.getReadName shouldBe "pair"
+    r1.getReadString shouldBe "A"*100
+    r1.getBaseQualityString shouldBe "I"*100
+    r1.getReadPairedFlag shouldBe true
+    r1.getFirstOfPairFlag shouldBe true
+    r1.getReadUnmappedFlag shouldBe true
```
tfenne: to be on the safe side you could also assert that `getSecondOfPairFlag` is `false`.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -0,0 +1,573 @@
...
+  /** True if the read structure implies paired end reads will be produced, false otherwise. */
+  val pairedEnd: Boolean = readStructures.count(_.template.nonEmpty) == 2
+
+  /** The quality format converter. */
+  private val solexaQualityConverter: SolexaQualityConverter = SolexaQualityConverter.getSingleton
+
+  /** Converts the segments across all reads to a string, with each read delimited by the given delimiter. */
+  private def toBarcode(allBases: Seq[String], delimiter: String, allSegments: Seq[Seq[ReadSegment]]): String = {
+    allBases.zip(allSegments).map { case (bases, segments) =>
+      ReadStructure.structureRead(bases=bases, segments=segments).map(_.bases).filter(_.nonEmpty).mkString
```
nh13: The barcodes are delimited by read, not segment.

src/main/scala/com/fulcrumgenomics/illumina/Sample.scala
nh13: Copy and past error from the previous `SampleBarcode` class.  Thanks for catching this.

src/main/scala/com/fulcrumgenomics/illumina/Sample.scala
nh13: Ditto about the copy-paste error.

PR #180, opened 2017-02-28
Prototype of semi-automatic sizing for sorting collections.

tfenne:
@nh13 Very rough prototype here of the auto-sizing sorting collection stuff if you want to take a peek.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/basecalling/SortSam.scala
+++ b/src/main/scala/com/fulcrumgenomics/basecalling/SortSam.scala
@@ -1,0 +1,65 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/basecalling/SortSam.scala
```diff
@@ -0,0 +1,65 @@
...
+import dagr.commons.util.LazyLogging
+import dagr.sopt.{arg, clp}
+import htsjdk.samtools.SAMFileHeader.SortOrder
+import htsjdk.samtools.{BAMRecordCodec, SAMFileWriterFactory, SamReaderFactory}
+
+@clp(group=ClpGroups.SamOrBam, description=
+  """
+    |Sorts a SAM or BAM file.
+  """)
+class SortSam
```
tfenne: I'll probably kill this class, it's mostly just here to enable some interactive testing.

src/main/scala/com/fulcrumgenomics/util/Sorter.scala
```diff
@@ -0,0 +1,126 @@
...
+  def +=(a: A): Unit = (this.sizer, this.collection) match {
+    case (None, Some(c)) =>
+      c.add(a)
+    case (Some(s), None) =>
+      s.buffer += a
+      this.codec.setOutputStream(s.stream)
+      this.codec.encode(a)
+
+      if (s.buffer.size >= this.sampleSize) {
+        System.gc()
```
tfenne: This is also here just for testing ... so that I can debug with a breakpoint here and see what the heap looks like immediately after GC'ing.

PR #181, opened 2017-02-28
Changes to consensus calling default parameters.  Includes:

tfenne:
1. Removing the default for min-reads in both CallMolecularConsensusReads AND FilterConsensusReads
2. Deprecating and setting back to Q2, the min-consensus-quality parameter in CallMolecularConsensusReads
3. Removing the default for min-consensus-quality in FilterConsensusReads


tfenne:
@mjafin Let's continue the discussion about default parameters here if you don't mind.  We discussed this internally and decided that the right thing to do was to _remove_ some defaults since there isn't really a value that makes sense for all applications, and instead provide some guidance on how to set reasonable values.

So ... now `min-reads` is a required parameter without a default in `CallMolecularConsensusReads`.  In addition we've deprecated `min-consensus-quality` in `CallMolecularConsensusReads` as we are trying to move to a model where all _output_ filtering is done in `FilterConsensusReads` instead of `CallMolecularConsensusReads`.  `min-reads` is an exception to that rule solely because it can significantly change the number of reads output and the computational cost of re-mapping and filtering.

I'm not sure if you have looked at `FilterConsensusReads` yet, but it offers an expanded set of filters that can be used to remove lower-quality consensus bases and reads.  

Let me know if you have any feedback.

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -69,6 +69,13 @@ import htsjdk.samtools._
     |calls each end of a pair independently, and does not jointly call bases that overlap within a pair.  Insertion or
     |deletion errors in the reads are not considered in the consensus model.
     |
+    |Particular attention should be paid to setting the --min-reads parameter as this can have a dramatic effect on
+    |both results and runtime.  For libraries with low duplication rates (e.g. 100-300X exomes libraries) in which it
+    |is desirable to retain singleton reads while making consensus reads from sets of duplicates, --min-reads=1 is
+    |appropriate.  For libraries with high duplication rates where it is desirable to only produce consensus reads
+    |supported by 2+ reads to allow error correction, --min-reads=2 or higher is appropriate.  After generation,
+    |consensus reads can be further filtered using the FilterConsensusReads tool.
```
nh13: Also mention that the main reason to have `--min-reads` is performance, namely to upfront shrink the data size and not have a big BAM.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -96,8 +96,8 @@ class FilterConsensusReads
   val maxReadErrorRate: Seq[Double] = Seq(0.025),
   @arg(flag="e", minElements=1, maxElements=3, doc="The maximum error rate for a single consensus base.")
   val maxBaseErrorRate: Seq[Double] = Seq(0.1),
-  @arg(flag="N", minElements=1, maxElements=3, doc="Mask (make 'N') consensus bases with quality less than this threshold.")
-  val minBaseQuality: PhredScore = DefaultMinConsensusBaseQuality,
+  @arg(flag="N", doc="Mask (make 'N') consensus bases with quality less than this threshold.")
+  val minBaseQuality: PhredScore,
```
nh13: Why not set it to some reasonable low value (ex. `2.toByte`) so that the user can run the tool end-to-end.  There are enough filter options that have defaults.

tfenne: @nh13 I'm actually tempted to go the other way and also remove the default for `--min-reads` in the filtering tool.  All the other filters that have defaults are not very depth dependent.  The error rates and %Ns should work equally well for consensus reads with 1 raw read as for ones with 50.  But `min-reads` and `min-base-quality` both are very dependent on the tag family size distribution.

mjafin:
I like these changes, thumbs up.

PR #183, opened 2017-03-01
Made the output from Rscript a little quieter.

tfenne:
@nh13 A small change to push stderr from Rscript to the log at `info` and stdout at `debug` since R tends to be a bit noisy.  Also changed the R script for error rate by read position so that it supresses warnings (errors will still be emitted) and to use `library` in place of `require` since it does more or less the same thing but is quieter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,16 @@ cache:
   directories:
   - $HOME/.ivy2/cache
   - $HOME/.sbt/boot/
+addons:
+  apt:
+    sources:
+      - r-packages-precise
+    packages:
+      - r-core
+      - r-cran-ggplot2      
 before_install:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -12,10 +12,16 @@ cache:
   directories:
   - $HOME/.ivy2/cache
   - $HOME/.sbt/boot/
```
nh13: Do you want to cache r-packages?
```
cache:
  packages: true
```

nh13:
👍

.travis.yml
tfenne: I'm not sure that works if you don't have `language: r` but I've added a directory cache for where R is installing it's packages.

PR #184, opened 2017-03-02
Change FilterSomaticVcf to support/filter out reads with invalid insert sizes.

tfenne:


src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -246,4 +248,34 @@ object Bams extends LazyLogging {
...
     }
   }
+
+  /** Returns true if the read is mapped in an FR pair, false otherwise. */
+  def isFrPair(rec: SAMRecord): Boolean =
+    rec.getReadPairedFlag &&
+      !rec.getReadUnmappedFlag &&
+      !rec.getMateUnmappedFlag &&
+      rec.getReferenceIndex == rec.getMateReferenceIndex &&
+      SamPairUtil.getPairOrientation(rec) == PairOrientation.FR
```
nh13: I am of the opinion of the statement isn't a one-liner there should be brackets.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -246,4 +248,34 @@ object Bams extends LazyLogging {
...
+
+  /**
+    * Calculates the coordinates of the insert represented by this record and returns them
+    * as a pair of 1-based closed ended coordinated.
```
nh13: coordinated -> coordinates

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -246,4 +248,34 @@ object Bams extends LazyLogging {
...
+    * @return the start and end position of the insert as a tuple, always with start <= end
+    */
+  def insertCoordinates(rec: SAMRecord): (Int, Int) = {
+    require(rec.getReadPairedFlag, s"Read ${rec.getReadName} is not paired, cannot calculate insert coordinates.")
+    require(!rec.getReadUnmappedFlag, s"Read ${rec.getReadName} must be mapped to calculate insert coordinates.")
+    require(!rec.getMateUnmappedFlag, s"Read ${rec.getReadName} must have mapped mate to calculate insert coordinates.")
+    require(rec.getReferenceIndex == rec.getMateReferenceIndex, s"Read ${rec.getReadName} must be mapped to same chrom as it's mate.")
+
+    val isize      = rec.getInferredInsertSize
+    val thisEnd    = if (rec.getReadNegativeStrandFlag) rec.getAlignmentEnd else rec.getAlignmentStart
```
nh13: `thisEnd` was confusing since I thought it meant the `end position of this records` not what I think it means which is `the 5' coordinate of this record (end)`.

src/main/scala/com/fulcrumgenomics/vcf/filtration/EndRepairArtifactLikelihoodFilter.scala
```diff
@@ -228,22 +228,16 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
     if (rec.getReadPairedFlag && !rec.getReadUnmappedFlag && !rec.getMateUnmappedFlag &&
       rec.getReferenceIndex == rec.getMateReferenceIndex && rec.getInferredInsertSize != 0 &&
       SamPairUtil.getPairOrientation(rec) == PairOrientation.FR) {
-
+      
       val isize      = rec.getInferredInsertSize
       val thisEnd    = if (rec.getReadNegativeStrandFlag) rec.getAlignmentEnd else rec.getAlignmentStart
       val adjustment = if (rec.getInferredInsertSize < 0) 1 else -1
       val otherEnd   = thisEnd + rec.getInferredInsertSize + adjustment
+      val (start, end) = (min(thisEnd, otherEnd), max(thisEnd,otherEnd))
```
nh13: adjust the alignment of `=` sir

nh13: Should you use your shiny new method `insertCoordinates` here?

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
```diff
@@ -138,6 +138,34 @@ class FilterSomaticVcf
...
+              |incompatible insert size yielding insert coordinates of ${pileup.refName}:$start-$end which do not overlap the variant.
+            """.stripMargin.trim.lines.mkString(" "))
+          false
+        }
+        else {
+          true
+        }
+      }
+      else {
+        true
```
nh13: I was expecting this to be false, since we potentially can have the same problem with other pair orientations.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
tfenne: I'll give you your brackets, but I actually prefer the opposite.  I think it makes it clear that it's all one statement if there are no brackets, vs. with the brackets you have to look more closely to see if it's one statement or many.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
tfenne: I've changed it to `firstEnd` and `secondEnd`, I couldn't think of anything better.

src/main/scala/com/fulcrumgenomics/vcf/filtration/EndRepairArtifactLikelihoodFilter.scala
tfenne: I wanted to, but it's important here to know which of the two ends of the insert are which, which I kinda lose with the shiny new method.

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
tfenne: If it were false we'd lose all non-FR reads, which might not be desirable.  I have a hard time reasoning about what the expectations are for non-FR reads, since if the insert size if 5' to 5', then all other pair orientations will by definition have bases outside of the region defined by the insert size.

tfenne:
Back to you @nh13

nh13:
@tfenne back to you.

PR #185, opened 2017-03-02
Fix to sort/unique intervals first in ErrorRateByReadPosition.

tfenne:
@nh13 This started out as a trivial PR to add `uniqued(false)` to the intervals in `ErrorRateByReadPosition`, but grew a bit.  It turned out the root of the problem was that `ByIntervalListVariantContextIterator` was not defensive in ensuring the `IntervalList` it was given was sorted/uniqued.  I added that, and then ran into a bug in HTSJDK where calling `uniqued(false)` on intervals without names throws an exception!  Hence the addition of `false, "foo"` to all the intervals in the tests for `ByIntervalListVariantContextIterator`.

I've submitted a PR to htsjdk, but don't want to wait on a release there.


nh13:
This is probably not a big deal for interval lists, since sorting and unique I presume is fast and doesn't take too much memory.  If you agree, then 👍

PR #189, opened 2017-03-07
Fix so that FilterSomaticVcf doesn't NPE if the input VCF lacks a sequence dictionary.

tfenne:
The htsjdk VCF code, in it's ultimate wisdom will return `null` from `VCFHeader.getSequenceDictionary()` if there are no contig lines, and then turn around and throw a `NullPointerException` at you if you call `VCFHeader.setSequenceDictionary(null)`.  So it's not safe to just do `newHeader.setSequenceDictionary(oldHeader.getSequenceDictionary())`.  Sigh.


tfenne:
@nh13 Good call on the `foreach` vs. `match`.  I agree VCFs without dictionaries are nasty, but so many tools produce them I think we should handle them for now and then _maybe_ at a later date centralize our code for opening a VCF and `warn` or `error` if opening a VCF without contig lines.

PR #190, opened 2017-03-13
Added the ability to limit the maximum number of reads going into a consensus.

tfenne:
@nh13 This PR introduces a parameter to allow you to limit the consensus caller to using `n` input reads.  My goal in doing this is to be able to evaluate consensus reads generated at different levels of input more easily.  I.e. what is my reduction in error in going from tag family size 2 to 3, to 4 to 10, etc.  This is hampered by the fact that in a real library the distribution of tag family sizes is often pretty broad and using `min=3` yields consensus reads that have 3+ family members, and often most of the data is at n > 3.

I'm open to putting the downsampling in at a different point, but I figured where I put it a) makes testing simpler (can be done with `SourceRead`s vs. `SAMRecord`s) and b) is after the indel-based filtering, so has the chance to downsample from usable reads vs. reads that may make it in, therefore increasing potential yield.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
@@ -108,7 +108,11 @@ class CallMolecularConsensusReads
    """)
  val minConsensusBaseQuality: PhredScore = 2.toByte,
  @arg(flag="M", doc="The minimum number of reads to produce a consensus base.") val minReads: Int,
- @arg(doc="The maximum number of reads to use when building a consensus.") val maxReads: Option[Int] = None,
+ @arg(doc="""
+            |The maximum number of reads to use when building a consensus. If more than this many reads are
+            |present in a tag family, the family is randomly downsampled to exactly max-reads reads.
+          """)
+ val maxReads: Option[Int] = None,
  @arg(flag="B", doc="If true produce tags on consensus reads that contain per-base information.") val outputPerBaseTags: Boolean = DefaultProducePerBaseTags,
  @arg(flag="S", doc="The sort order of the output, if None then the same as the input.") val sortOrder: Option[SortOrder] = Some(SortOrder.queryname),
  @arg(flag="D", doc="Turn on debug logging.") val debug: Boolean = false
--- a/src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
@@ -108,6 +108,7 @@ class CallMolecularConsensusReads
    """)
  val minConsensusBaseQuality: PhredScore = 2.toByte,
  @arg(flag="M", doc="The minimum number of reads to produce a consensus base.") val minReads: Int,
+ @arg(doc="The maximum number of reads to use when building a consensus.") val maxReads: Option[Int] = None,
  @arg(flag="B", doc="If true produce tags on consensus reads that contain per-base information.") val outputPerBaseTags: Boolean = DefaultProducePerBaseTags,
  @arg(flag="S", doc="The sort order of the output, if None then the same as the input.") val sortOrder: Option[SortOrder] = Some(SortOrder.queryname),
  @arg(flag="D", doc="Turn on debug logging.") val debug: Boolean = false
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -108,6 +108,7 @@ class CallMolecularConsensusReads
    """)
  val minConsensusBaseQuality: PhredScore = 2.toByte,
  @arg(flag="M", doc="The minimum number of reads to produce a consensus base.") val minReads: Int,
+ @arg(doc="The maximum number of reads to use when building a consensus.") val maxReads: Option[Int] = None,
```
nh13: One thought I have is instead of downsampling what if you just want only families with no and no less than the given # of reads?  I just wonder if you downsample families with `20+` to `n=3` if you aren't biasing the error rate at the end of the day to those that amplified well, which could have their own properties (i.e. amplified well because they have few errors).

Regardless, I think that the usage needs to state that the reads are downsampled if more than this # are found.

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -139,6 +140,7 @@ class CallMolecularConsensusReads
       minInputBaseQuality          = minInputBaseQuality,
       minConsensusBaseQuality      = minConsensusBaseQuality,
       minReads                     = minReads,
+      maxReads                     = maxReads.getOrElse(Int.MaxValue),
```
nh13: `Int.MaxValue` -> `DefaultMaxReads`

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -100,13 +103,15 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
                                 val rejects: Option[SAMFileWriter] = None
                                ) extends UmiConsensusCaller[VanillaConsensusRead] with LazyLogging {
 
+  private val NotEnoughReadsQual: PhredScore = 0.toByte // Score output when masking to N due to insufficient input reads
```
nh13: Why zero and not two?

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -100,13 +103,15 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
...
+  private val TooLowQualityQual: PhredScore = 2.toByte  // Score output when masking to N due to too low consensus quality
   private val DnaBasesUpperCase: Array[Byte] = Array('A', 'C', 'G', 'T').map(_.toByte)
   private val LogThree = LogProbability.toLogProbability(3.0)
+
```
nh13: found one

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -167,8 +172,11 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
       None
     }
     else {
+      // First limit to max reads if necessary
+      val capped  = if (reads.size < this.options.maxReads) reads else this.random.shuffle(reads).take(this.options.maxReads)
```
nh13: `<=`

src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -153,6 +153,22 @@ class VanillaUmiConsensusCallerTest extends UnitSpec with OptionValues {
...
+    for (max <- Seq(3, 1000); n <- Range.inclusive(1, 10)) {
+      val srcs      = Seq.tabulate(n)(_ => r)
+      val caller    = cc(cco(minReads=1, maxReads=max))
+      val consensus = caller.consensusCall(srcs)
+
+      consensus match {
+        case None    => fail("Consensus should have been generated")
+        case Some(c) =>
+          if (n < max) c.depths.forall(_ <= n)   shouldBe true
+          else         c.depths.forall(_ <= max) shouldBe true
```
nh13: I think this can be written as:
```
 consensus match {
         case None               => fail("Consensus should have been generated")
         case Some(c) if n < max => c.depths.forall(_ <= n)   shouldBe true
         case Some(c)            => c.depths.forall(_ <= max) shouldBe true
}
```
Also, you may want `n <= max` for clarity.

PR #191, opened 2017-03-15
Yet another useful method moved to Bams.

nh13:
I want this really useful method elsewhere!

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
@@ -31,10 +31,10 @@ import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
 import htsjdk.samtools.SamPairUtil.PairOrientation
 import htsjdk.samtools._
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker
-import htsjdk.samtools.util.{SequenceUtil, SortingCollection}
+import htsjdk.samtools.util.{CoordMath, SequenceUtil, SortingCollection}
 
 import scala.collection.mutable.ListBuffer
-import scala.math.{min, max}
+import scala.math.{max, min}
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -279,4 +279,28 @@ object Bams extends LazyLogging {
...
+    * @param genomicPosition the genomic position of interest (NOT the position within the read)
+    */
+  def positionFromOtherEndOfTemplate(rec: SAMRecord, genomicPosition: Int): Option[Int] = {
+    if (isFrPair(rec) && rec.getInferredInsertSize != 0) {
+      val isize        = rec.getInferredInsertSize
+      val thisEnd      = if (rec.getReadNegativeStrandFlag) rec.getAlignmentEnd else rec.getAlignmentStart
+      val adjustment   = if (rec.getInferredInsertSize < 0) 1 else -1
+      val otherEnd     = thisEnd + rec.getInferredInsertSize + adjustment
+      val (start, end) = (min(thisEnd, otherEnd), max(thisEnd,otherEnd))
+      require(genomicPosition >= start && genomicPosition <= end, s"genomicPosition is outside of template for $rec")
```
tfenne: If we're moving this out into a utility method (it was in a filter class that was protected by some other checks in FilterSomaticVcf) I wonder if we should convert this from a `require()` into an `if (...) None else Some()`?  Otherwise we have to build that some check in everywhere?  OTOH, I kinda like that it's a requirement here.

nh13: I agree that returning `None` where the insert size indicates that the genomic position is _outside_ the insert should be handled elsewhere.  I'd like the keep the `require` here until and have callers check (and perhaps warn) like in `FilterSomaticVcf.filterPileup`.

PR #197, opened 2017-03-30
Implementation of NeedlemanWunch aligner.

tfenne:


src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -0,0 +1,187 @@
...
+    // The interior of the matrix
+    forloop(from=1, until=query.length+1) { i =>
+      forloop(from=1, until=target.length+1) { j =>
+        val diagonalScore = scoring(i-1, j-1) + scoringFunction(query(i-1), target(j-1))
+        val upScore       = scoring(i-1, j  ) + gapExtend + (if (trace(i-1, j  ) == Up) 0 else gapOpen)
+        val leftScore     = scoring(i,   j-1) + gapExtend + (if (trace(i,   j-1) == Left) 0 else gapOpen)
+
+        val maxScore = max(max(diagonalScore, upScore), leftScore)
+        scoring(i,j) = maxScore
+        trace(i,j)   = if (maxScore == diagonalScore) Diagonal else if (maxScore == leftScore) Left else Up
```
nh13: I think it may be useful to detail that when we have two equally likely choices, we prefer mismatches, deletions, and insertions, in that order.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -0,0 +1,187 @@
...
+    * @param query the query sequence
+    * @param target the target sequence
+    * @param scoring the scoring matrix
+    * @param trace the trace back matrix
+    * @return an [[Alignment]] object representing the alignment
+    */
+  protected def generateAlignment(query: Array[Byte], target: Array[Byte], scoring: Matrix[Int], trace: Matrix[Direction]): Alignment = {
+    var i = query.length
+    var j = target.length
+    var currOperator: CigarOperator = null
```
nh13: I wish we could extend the enum to a `NoOp` or `None`.  Alas.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -0,0 +1,187 @@
...
+
+      if (extending) {
+        currLength += 1
+        if (i == 0 && j == 0) stack.push(CigarElem(currOperator, currLength))
+      }
+      else {
+        if (currLength > 0) stack.push(CigarElem(currOperator, currLength))
+        currOperator = op
+        currLength   = 1
+        if (i == 0 && j == 0) stack.push(CigarElem(currOperator, currLength))
```
nh13: Can move this to after the `if/else` and remove the same call within the `if/else`:
```
if (i == 0 && j == 0) stack.push(CigarElem(currOperator, currLength))
```

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -0,0 +1,78 @@
...
+  * @param query the query sequence
+  * @param target the target sequence
+  * @param queryStart the 1-based position in the query sequence where the alignment begins
+  * @param targetStart the 1-based position in the target sequence where the alignment begins
+  * @param cigar a [[Cigar]] object describing the alignment of the two sequences
+  * @param score the alignment score
+  */
+case class Alignment(query: Array[Byte],
+                     target: Array[Byte],
+                     queryStart: Int,
```
nh13: The more and more I look at this, the less I am in love with `queryStart/targetStart` without having `queryEnd/targetEnd`.  I assume you include this here so it can be extensible.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -0,0 +1,187 @@
...
+    var currOperator: CigarOperator = null
+    var currLength: Int = 0
+    val stack = new mutable.Stack[CigarElem]()
+
+    while (i != 0 || j != 0) {
+      val op = trace(i,j) match {
+        case Done     =>
+          i = 0
+          j = 0
+          null
```
nh13: Remove this case, as I don't think we will ever hit it, since `Done` is only stored in `trace(0, 0)`.  If you intend this to be used for other types of alignment (maybe that's why it's `protected`) ignore this comment, since then yeah, this is needed.

src/test/scala/com/fulcrumgenomics/alignment/MatrixTest.scala
```diff
@@ -0,0 +1,81 @@
...
+    val m = Matrix[Int](4,5)
+    an[Exception] shouldBe thrownBy { m(-1, -1) }
+    an[Exception] shouldBe thrownBy { m( 1,  7) }
+    an[Exception] shouldBe thrownBy { m( 6,  1) }
+    an[Exception] shouldBe thrownBy { m( 5,  5) }
+
+    an[Exception] shouldBe thrownBy { m(-1, -1) = 1}
+    an[Exception] shouldBe thrownBy { m( 1,  7) = 1}
+    an[Exception] shouldBe thrownBy { m( 6,  1) = 1}
+    an[Exception] shouldBe thrownBy { m( 5,  5) = 1}
```
nh13: Test one negative index and one positive index, I bet you can make it not throw an exception.

src/test/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAlignerTest.scala
```diff
@@ -0,0 +1,238 @@
...
+  "NeedlemanWunschAligner.align" should "align two identical sequences with all matches" in {
+    val result = NeedlemanWunschAligner(1, -1, -3, -1).align(s("ACGTAACC"), s("ACGTAACC"))
+    assertValidGlobalAlignment(result)
+    result.cigar.toString() shouldBe "8="
+    result.score shouldBe 8
+  }
+
+  it should "align two sequences with a single mismatch in them" in {
+    val q = s("AACCGGTT")
+    val t = s("AACCGtTT")
```
nh13: Are different cases counted as mismatches?

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -0,0 +1,78 @@
...
+  /** Returns the canonical Cigar string. */
+  override def toString(): String = elems.mkString
+}
+
+
+/**
+  * A general class to describe the alignment between two sequences or partial ranges thereof
+  * @param query the query sequence
+  * @param target the target sequence
+  * @param queryStart the 1-based position in the query sequence where the alignment begins
```
tfenne: I will add queryEnd/targetEnd methods, since yes they are computable through the cigar.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: Yes, I wanted to have `Alignment` be the same if/when we also implement local alignment, and didn't want to have different result classes for local and global.  For global the starts are always 1 and the ends should always be the last base of each sequence.  I'll add the `end` methods though so that at least we have both.

src/main/scala/com/fulcrumgenomics/alignment/Matrix.scala
```diff
@@ -0,0 +1,101 @@
...
+trait Matrix[A] {
+  /** Access a cell in the matrix. */
+  def apply(i: Int, j: Int): A
+
+  /** Updates a cell in the matrix. */
+  def update(i: Int, j: Int, value: A): Unit
+
+  /**
+    * Returns the dimensions of the matrix.  E.g. a 2x2 matrix will return (2,2), although the maximum
+    * valid offsets are (1,1).
```
tfenne: Meaning in a 2x2 matrix, the dimensions are (2,2), but the last cell is accessed at `(1,1)` because it's all zero-based like arrays.  I'll try again.

src/main/scala/com/fulcrumgenomics/alignment/Matrix.scala
```diff
@@ -0,0 +1,101 @@
...
+    val (x,y) = dimensions
+    val buffer = new StringBuilder
+    forloop (from=0, until=x) { i =>
+      if (i > 0) buffer.append("\n")
+
+      forloop (from=0, until=y) { j =>
+        if (j > 0) buffer.append("\t")
+        buffer.append(String.valueOf(this(i,j)))
+      }
+    }
```
tfenne: Good call.  I was in the mode of "must make everything efficient", but since this will only ever be used for debugging, that's not important.

src/main/scala/com/fulcrumgenomics/alignment/Matrix.scala
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /** Updates a cell in the matrix. */
+  override def update(i: Int, j: Int, value: A): Unit = {
+    if (i >= x || j >= y) throw new NoSuchElementException(s"Illegal matrix indices: (${i}, ${j})")
+    data(i*y + j) = value
+  }
+
+  /** Returns the dimensions of the matrix. */
+  override def dimensions: (Int, Int) = (x, y)
```
tfenne: It's not needed, but it is helpful.  E.g. it will cause a compile error if it does not match the method name in the trait (e.g. because the trait removed that method).

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -0,0 +1,187 @@
...
+  /** Creates a NW aligner with fixed match and mismatch scores. */
+  def apply(matchScore: Int, mismatchScore: Int, gapOpen: Int, gapExtend: Int): NeedlemanWunschAligner = {
+    new NeedlemanWunschAligner(scoringFunction=simpleScoringFunction(matchScore, mismatchScore), gapOpen, gapExtend)
+  }
+
+  /** Directions within the trace back matrix. */
+  type Direction = Byte
+  val Left: Direction     = 1.toByte
+  val Up  : Direction     = 2.toByte
+  val Diagonal: Direction = 3.toByte
```
tfenne: Or, you know, I could make it `Leftward`, `Upwards` and `Diagonal` to get them as close as possible to the same length :p

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -0,0 +1,187 @@
...
+  val Done: Direction     = 4.toByte
+}
+
+/**
+  * Implementation of the Needleman-Wunsch algorithm for global alignment of two sequences with support for an
+  * affine gap penalty.
```
tfenne: Yup, thanks.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Yeah.  I thought about making this `Option[CigarOperator]` but I want to keep things in here as lean as possible and adding `Option` to handle that first time through the loop seemed not the best choice.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Commented.  To my mind Mismatches first is important to left-align indels.  It's a toss up then between deletions and insertions, so I just picked one.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Nope, I thought mostly this was necessary at one point, but with how I ended up it's clearly not.

src/test/scala/com/fulcrumgenomics/alignment/MatrixTest.scala
tfenne: I cannot (because I followed your earlier comment to test for negative indices ;))

src/test/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAlignerTest.scala
tfenne: Yes, they are currently.  The `s()` method in the test upper-cases everything and strips out the `-`s that I use to align the sequences for readability.  So the aligner actually sees: `AACCGGTT` and `AACCGTTT`, but I lower-case the T here so it's more visible where the mismatch is.

tfenne:
@nh13 I think I made all your suggested changes (except shortening `Diagonal`).  Want to take another look or shall I merge it?

nh13:
@tfenne merge it.

PR #198, opened 2017-03-30
DemuxFastqs improvements

nh13:
Three main features have been added:
1. The output metrics file name is no longer required.
2. A CSV file can be given with sample metadata in lieu of a sample sheet.  Some users requested this feature as they are not experts on Illumina sample sheets.
3. Gzipped FASTQs can now also be written.  Users requested this feature, even though Picard's `SamToFastq` can be used on the outputs of this tool.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -25,14 +25,16 @@
 
 package com.fulcrumgenomics.fastq
 
+import java.io.{Closeable, Serializable}
+
 import com.fulcrumgenomics.FgBioDef.unreachable
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
-import com.fulcrumgenomics.fastq.FastqDemultiplexer.DemuxRecord
+import com.fulcrumgenomics.fastq.FastqDemultiplexer.{DemuxRecord, DemuxResult}
 import com.fulcrumgenomics.illumina.{Sample, SampleSheet}
 import com.fulcrumgenomics.umi.ConsensusTags
 import com.fulcrumgenomics.util.ReadStructure.SubRead
 import com.fulcrumgenomics.util.{ReadStructure, SampleBarcodeMetric, SampleBarcode => _, _}
-import dagr.commons.CommonsDef.{DirPath, FilePath, PathToBam, PathToFastq}
+import dagr.commons.CommonsDef.{DirPath, FilePath, PathPrefix, PathToBam, PathToFastq, yieldAndThen}
 import dagr.commons.io.PathUtil
 import dagr.commons.util.{LazyLogging, Logger}
 import dagr.sopt.{arg, clp}
@@ -44,24 +46,28 @@ import htsjdk.samtools.{SAMRecord, _}
 import scala.concurrent.forkjoin.ForkJoinPool
 
 object DemuxFastqs {
+
+  /** The name of the metrics file if none is given by the user. */
+  val DefaultDemuxMetricsFileName = "demux_barcode_metrics.txt"
+
   /** The name of the sample for unmatched reads. */
   val UnmatchedSampleId: String = "unmatched"
 
   /** The maximum # of records in RAM per writer. */
-  private val MaxRecordsInRamPerSamFileWriter: Int = 1e6.toInt
+  private[fastq] val MaxRecordsInRamPerSamFileWriter: Int = 1e6.toInt
 
   /** Decides whether or not to use asynchronous IO for the writers. This has a big performance benefit at the cost of
     * some RAM. RAM may balloon if there is a need to sort the output. */
-  private val UseAsyncIo: Boolean  = true
+  private[fastq] val UseAsyncIo: Boolean  = true
 
   /** The number of records to batch when demultiplexing in parallel. */
   private val DemuxBatchRecordsSize = 1e5.toInt
 
   /** Creates the sample output BAM path for the given sample. */
-  private[fastq] def sampleOutputBam(output: DirPath, sample: Sample): PathToBam = {
+  private[fastq] def outputPrefixFrom(output: DirPath, sample: Sample): PathPrefix = {
     val sampleBarcode = sample.sampleBarcodeString
     require(sampleBarcode.nonEmpty, s"Sample barcode missing for sample: ${sample.sampleName}")
-    output.resolve(PathUtil.sanitizeFileName(s"${sample.sampleId}-${sample.sampleName}-$sampleBarcode.bam"))
+    output.resolve(PathUtil.sanitizeFileName(s"${sample.sampleId}-${sample.sampleName}-$sampleBarcode"))
   }
 
   /** Gets the quality format of the FASTQs. */
@@ -121,6 +127,10 @@ object DemuxFastqs {
     }
   }
 
+  private[fastq] def toSampleOutputPrefix(sample: Sample, isUnmatched: Boolean, output: DirPath, unmatched: String): PathPrefix = {
+    if (isUnmatched) PathUtil.removeExtension(output.resolve(unmatched)) else outputPrefixFrom(output, sample)
+  }
+
   /** Creates a demultiplexing iterator that performs demultiplexing in parallel.
     *
     * @param sources the FASTQ sources, one per read.
@@ -130,7 +140,7 @@ object DemuxFastqs {
   def demultiplexingIterator(sources: Seq[Iterator[FastqRecord]],
                              demultiplexer: FastqDemultiplexer,
                              threads: Int,
-                             batchSize: Int = DemuxBatchRecordsSize): Iterator[DemuxRecord] = {
+                             batchSize: Int = DemuxBatchRecordsSize): Iterator[DemuxResult] = {
 
     require(demultiplexer.expectedNumberOfReads == sources.length,
       s"The demultiplexer expects ${demultiplexer.expectedNumberOfReads} reads but ${sources.length} FASTQ sources given.")
@@ -164,11 +174,18 @@ object DemuxFastqs {
       |the FASTQs, to assign each read to a sample.  Reads that do not match any sample within the given error tolerance
       |will be placed in the 'unmatched' file.
       |
-      |The output directory will contain one BAM file per sample in the sample sheet, plus a BAM for reads that could
-      |not be assigned to a sample given the criteria.  The output file names will be the concatenation of sample id, sample
-      |name, and sample barcode bases (expected not observed), delimited by "-".  A metrics file will also be output
-      |providing analagous information to the metric desribed here:
+      |The output directory will contain one BAM file per sample in the sample sheet or metadata CSV file, plus a BAM for
+      |reads that could not be assigned to a sample given the criteria.  The output file names will be the concatenation
+      |of sample id, sample name, and sample barcode bases (expected not observed), delimited by "-".  A metrics file
+      |will also be output providing analogous information to the metric described here:
       |https://broadinstitute.github.io/picard/picard-metric-definitions.html#SampleBarcodeMetric
+      |
+      |Alternatively, gzipped FASTQs can be written using the "--output-fastqs=true" option instead of BAMs.  For paired
+      |end data, the output will have the suffix "_R1.fastq.gz" and "_R2.fastq.gz" for read one and read two respectively.
+      |The sample barcode and molecular barcodes (concatenated) will be appended to the read name and delimited by a
+      |colon.
+      |
+      |The output base qualities will be standardized to Sanger/SAM format.
       |
       |FASTQs and associated read structures for each read should be given:
       |- a single fragment read should have one FASTQ and one read structure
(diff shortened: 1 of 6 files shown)


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -312,12 +363,37 @@ val qualityFormat: Option[FastqQualityFormat] = None,
     // Close the writer; NB: the inputs close automatically
     sampleInfos.foreach(_.writer.close())
 
+    // Optionally create the FASTQs
+    if (outputFastqs) {
```
tfenne: Ok, this feels _really_ weird to me.  I figured the idea here was to add an option to output _either_ BAMs or FASTQs.  I find it hard to believe that anyone would really want both - i.e. if I prefer to demux to fastq files, then I'm probably going to ignore the BAM files and vice versa.  Shouldn't the options be exclusive?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -312,12 +363,37 @@ val qualityFormat: Option[FastqQualityFormat] = None,
...
+      logger.info("Writing FASTQs")
+      val fastqProgress = new ProgressLogger(this.logger, unit=1e6.toInt)
```
tfenne: If things stay this way you probably want to give different noun/verbs here so that it's clear what the progress is for.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -312,12 +363,37 @@ val qualityFormat: Option[FastqQualityFormat] = None,
...
+      sampleInfos.foreach { info =>
+        val outputFastq    = PathUtil.replaceExtension(path=info.outputBam, ext=".fastq.gz")
+        val bufferedWriter = IOUtil.openFileForBufferedWriting(outputFastq.toFile)
+        val fastqWriter    = FastqWriter(bufferedWriter)
+        val bamReader      = SamReaderFactory.makeDefault().open(info.outputBam)
+        import com.fulcrumgenomics.FgBioDef.javaIterableToIterator
+        bamReader.foreach { bamRecord =>
+          val fastqRecord = FastqRecord(
```
tfenne: This appears to write all the reads to a single fastq file, without their read number suffixes.  Shouldn't this either output to a pair of fastq files, or supply `readNumber` below?

src/main/scala/com/fulcrumgenomics/util/DelimitedDataParser.scala
```diff
@@ -65,6 +65,12 @@ class Row private[util] (private val headerIndices: Map[String,Int], private val
 
   /** Gets a value from the specified column. If the value is null or empty returns None. */
   def get[A](column: String)(implicit tag: ru.TypeTag[A]): Option[A] = get(headerIndices(column))
+
+  /** Gets a value from the specified column. If the no column with the given name exists, or the value is null or empty
+    * returns None. */
+  def getOrNone[A](column: String)(implicit tag: ru.TypeTag[A]): Option[A] = if (headerIndices.keys.exists(_ == column)) {
```
tfenne: A few thoughts:
1) This needs a clearer name since `get()` already returns `Option`. How about instead of a separate method, having `allowMissingColumn: Boolean = false` on `get()`?
2) We could also just make this a parameter to delimited data parser, so it's a mode for the parser?
3) I think you could implement the current version as `headerIndices.get(column).map(get[A](_))` and use the fact that map.get returns an `Option`.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -164,11 +168,13 @@ object DemuxFastqs {
...
-      |The output directory will contain one BAM file per sample in the sample sheet, plus a BAM for reads that could
-      |not be assigned to a sample given the criteria.  The output file names will be the concatenation of sample id, sample
-      |name, and sample barcode bases (expected not observed), delimited by "-".  A metrics file will also be output
-      |providing analagous information to the metric desribed here:
+      |The output directory will contain one BAM file per sample in the sample sheet or metadata CSV file, plus a BAM for
+      |reads that could not be assigned to a sample given the criteria.  The output file names will be the concatenation
+      |of sample id, sample name, and sample barcode bases (expected not observed), delimited by "-".  A metrics file
+      |will also be output providing analogous information to the metric described here:
       |https://broadinstitute.github.io/picard/picard-metric-definitions.html#SampleBarcodeMetric
+      |In addition to BAM files, gzipped FASTQs can be written using the "--output-fastqs=true" option.
```
nh13: Now generates either FASTQs or BAM, but not both.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -193,31 +207,43 @@ object DemuxFastqs {
...
+      |  --inputs r1.fq i1.fq i2.fq r2.fq --read-structures 8B92T 8B 8B 100T \
       |    --sample-sheet SampleSheet.csv --metrics metrics.txt --output output_folder
     """,
   group=ClpGroups.Fastq
 )
 class DemuxFastqs
 (@arg(flag="i", doc="One or more input fastq files each corresponding to a sub-read (ex. index read, read one, read two, fragment).") val inputs: Seq[PathToFastq],
  @arg(flag="o", doc="The output directory in which to place sample BAMs.") val output: DirPath,
- @arg(flag="s", doc="The sample sheet.") val sampleSheet: FilePath,
+ @arg(flag="s", doc="The sample sheet if the metadata CSV file is not given.") val sampleSheet: Option[FilePath] = None,
```
nh13: I thought of parsing it both ways but I want to inform the user if either file is malformed along with a useful error message (for the sample sheet at least).

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -238,31 +264,56 @@ val qualityFormat: Option[FastqQualityFormat] = None,
...
+  private[fastq] val metricsPath = metrics.getOrElse(output.resolve(DefaultDemuxMetricsFileName))
+
   validate(inputs.length == readStructures.length, "The same number of read structures should be given as FASTQs.")
   validate(readStructures.flatMap(_.sampleBarcode).nonEmpty, s"No sample barcodes found in read structures: " + readStructures.map(_.toString).mkString(", "))
 
-  /** True if the read structure implies paired end reads will be produced, false otherwise. */
-  val pairedEnd: Boolean = readStructures.count(_.template.nonEmpty) match {
-    case 1 => false
-    case 2 => true
+  readStructures.count(_.template.nonEmpty) match {
```
nh13: readStructures is a `Seq[ReadStructure]`, while `_.template` returns a list of all template segments (`Seq[ReadSegment]`) in that read structure, so we count how many read structures have at least one read segment that is a template.  I think it handles the case you're think about.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -238,31 +264,56 @@ val qualityFormat: Option[FastqQualityFormat] = None,
...
+    case 1 | 2 => Unit
     case n => invalid(s"Found $n read structures with template bases but expected 1 or 2.")
   }
 
   Io.assertReadable(inputs)
   Io.assertReadable(sampleSheet)
   Io.assertWritableDirectory(output)
 
+  /** Create a sample sheet from either the input sample sheet path or the metadata CSV. */
```
nh13: 💥

nh13:
@tfenne I think I have addressed all your comments.  By changing the behavior to be either BAM or FASTQ but not both, I was able to some more refactoring, hence the number of changes.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -44,24 +46,28 @@ import htsjdk.samtools.{SAMRecord, _}
...
     * some RAM. RAM may balloon if there is a need to sort the output. */
-  private val UseAsyncIo: Boolean  = true
+  private[fastq] val UseAsyncIo: Boolean  = true
 
   /** The number of records to batch when demultiplexing in parallel. */
   private val DemuxBatchRecordsSize = 1e5.toInt
 
   /** Creates the sample output BAM path for the given sample. */
-  private[fastq] def sampleOutputBam(output: DirPath, sample: Sample): PathToBam = {
+  private[fastq] def outputPrefixFrom(output: DirPath, sample: Sample): PathToBam = {
```
tfenne: I know they're only typedefs, but could the return type get changed to `PathPrefix` as well please?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -238,33 +264,52 @@ val qualityFormat: Option[FastqQualityFormat] = None,
...
   Io.assertReadable(inputs)
-  Io.assertReadable(sampleSheet)
+  Io.assertReadable(metadata)
   Io.assertWritableDirectory(output)
 
+  /** The quality format converter. */
+  private val solexaQualityConverter: SolexaQualityConverter = SolexaQualityConverter.getSingleton
+
+  /** Create a sample sheet from either the input sample sheet path or the metadata CSV. */
+  private val sampleSheet: SampleSheet = {
```
tfenne: So nice!

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -238,33 +264,52 @@ val qualityFormat: Option[FastqQualityFormat] = None,
...
     // Get the FASTQ quality encoding format
     val qualityFormat = this.qualityFormat.getOrElse(determineQualityFormat(inputs, this.qualityFormat, Some(this.logger)))
 
     // Read in the sample sheet and create the sample information
-    val samplesFromSampleSheet = SampleSheet(this.sampleSheet).map(s => withCustomSampleBarcode(s, columnForSampleBarcode))
+    val samplesFromSampleSheet = sampleSheet.map(s => withCustomSampleBarcode(s, columnForSampleBarcode))
     val samples                = samplesFromSampleSheet.toSeq :+ unmatchedSample(samplesFromSampleSheet.size, this.readStructures)
-    val sampleInfos            = samples.map(toSampleInfo)
+    val sampleInfos            = samples.map(sample => SampleInfo(sample, sample.sampleName == UnmatchedSampleId))
+    val sampleToWriter         = sampleInfos.map { info =>  (info.sample, toWriter(info)) }.toMap
```
tfenne: The two are equivelant, but I kinda prefer this syntax for map creation, how about you:
`sampleInfos.map { info =>  info.sample -> toWriter(info) }.toMap`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -302,60 +346,138 @@ val qualityFormat: Option[FastqQualityFormat] = None,
...
     iterator.foreach { demuxRecord =>
       demuxRecord.sampleInfo.metric.increment(numMismatches=demuxRecord.numMismatches)
-      val writer = demuxRecord.sampleInfo.writer
+      val writer = sampleToWriter(demuxRecord.sampleInfo.sample)
       demuxRecord.records.foreach { rec =>
-        writer.addAlignment(rec)
-        progress.record(rec)
+
+        val readQualsBytes = rec.quals.getBytes
+        convertQuality(readQualsBytes, qualityFormat, rec.name)
```
tfenne: Maybe we should note in the usage that even if you are outputting fastq, your quality scores will be standardized to sanger/SAM format?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -302,60 +346,138 @@ val qualityFormat: Option[FastqQualityFormat] = None,
...
+    record.setReadName(rec.name)
+    record.setReadString(rec.bases)
+    record.setBaseQualities(rec.quals.getBytes)
+    record.setReadUnmappedFlag(true)
+    if (rec.pairedEnd) {
+      record.setReadPairedFlag(true)
+      record.setMateUnmappedFlag(true)
+      if (rec.readNumber == 1) record.setFirstOfPairFlag(true) else record.setSecondOfPairFlag(true)
+    }
+    record.setAttribute("BC", rec.sampleBarcode.orNull)
```
tfenne: For this and the molecularBarcode below why not use the usual?:
`rec.sampleBarcode.foreach(bc => record.setAttribute("BC", bc))`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -302,60 +346,138 @@ val qualityFormat: Option[FastqQualityFormat] = None,
...
 
-/** A class to store information about a sample. */
-private[fastq] case class SampleInfo(sample: Sample, writer: SAMFileWriter, isUnmatched: Boolean = false) {
-  def header: SAMFileHeader = writer.getFileHeader
-  def rg: SAMReadGroupRecord = header.getReadGroups.get(0)
-  def rgId: String = rg.getId
+/** A writer that writes [[DemuxRecord]]s as [[FastqRecord]]s. */
+private class FastqRecordWriter(prefix: PathPrefix, val pairedEnd: Boolean) extends DemuxWriter {
+  private val writers: IndexedSeq[FastqWriter] = {
+    def f(ext: String): PathToFastq = PathUtil.pathTo(prefix + ext)
```
tfenne: I think this could be a little cleaner as:
```scala
  (if (pairedEnd) Seq("_R1.fastq.gz", "_R2.fastq.gz") else Seq(".fastq.gz"))
    .map(s => FastqWriter(Io.toWriter(PathUtil.pathTo(prefix + ext))))
    .toIndexedSeq
```

At the very least it could use our `Io.toWriter` instead of `IOUtil.openFileForBufferedWriting`

nh13: Actually, `IOUtil. openFileForBufferedWriting` supports gzip output, while `Io.toWriter` does not, so I can add `Io.toWriter` to fgbio's `Io`.

nh13: My bad, it's overriden in `toOutputStream`

PR #199, opened 2017-03-31
Tool to correct UMIs when a fixed set of known UMI sequences are used.

tfenne:


src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -0,0 +1,229 @@
...
+import CorrectUmis._
+
+object CorrectUmis {
+  /**
+    * Holds information about the match of a UMI to an expected UMI.
+    * @param matched true if the match is acceptable, false otherwise
+    * @param umi the fixed UMI sequence that was the closest match
+    * @param mismatches the number of mismatches between the UMI and the reported best matching fixed UMI
+    *  */
+  private[umi] case class UmiMatch(matched: Boolean, umi: String, mismatches: Int)
```
nh13: `type Umi = String`?

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -0,0 +1,229 @@
...
+    |
+    |All of the fixed UMIs must be of he same length, and all UMIs in the BAM file must also have the same
+    |length.  Multiple UMIs that are concatenated with hyphens (e.g. AACCAGT-AGGTAGA) are split apart,
+    |corrected individually and then re-assembled.  A read is accepted only if all the UMIs can be corrected.
+    |
+    |Correction is controlled by two parameters:
+    |  1. --max-mismatches controls how many mismatches (no-calls are counted as mismatches) are tolerated
+    |         between a UMI as read and a fixed UMI.
+    |  2. --min-distance controls how many more mismatches the next best hit must have
+    |
```
nh13: I think it's worth saying that these parameters apply to matching each UMI individually, which matters in the case of having multiple UMIs (ex. `AAA-GGG`).

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -0,0 +1,229 @@
...
+    */
+  case class UmiCorrectionMetrics(umi: String,
+                                  var total_matches: Long = 0,
+                                  var perfect_matches: Long = 0,
+                                  var one_mismatch_matches: Long = 0,
+                                  var two_mismatch_matches: Long = 0,
+                                  var other_matches: Long = 0,
+                                  var fraction_of_matches: Double = 0,
+                                  var representation: Double = 0
+                                 ) extends Metric
```
nh13: In the case of two or more UMIs per template, I'd like to see a metric per UMI as well as an overall metric for the template/read

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -0,0 +1,229 @@
...
+  override def execute(): Unit = {
+    // Construct the full set of UMI sequences to match again
+    val (umiSequences, umiLength) = {
+      val set = mutable.HashSet[String](umis:_*)
+      umiFiles.foreach(Io.readLines(_).map(_.trim).filter(_.nonEmpty).foreach(set.add))
+      validate(set.nonEmpty, s"At least one UMI sequence must be provided; none found in files ${umiFiles.mkString(", ")}")
+
+      val lengths = set.map(_.length)
+      validate(lengths.size == 1, s"UMIs of multiple lengths found. Lengths: ${lengths.mkString(", ")}")
+      (set.toArray, lengths.head)
```
nh13: It's up to you, but it may be worthwhile double-checking that the UMIs are sufficiently dissimilar relative to `minDistance`.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -0,0 +1,229 @@
...
+    |  2. --min-distance controls how many more mismatches the next best hit must have
+    |
+    |For example, with two fixed UMIs AAAAA and CCCCC and max-mismatches=3 and min-distance=2 the
+    |following would happen:
+    |  - AAAAA would match to AAAAA
+    |  - AAGTG would match to AAAAA with three mismatches because CCCCCC has six mismatches and 6 >= 3 + 2
+    |  - AACCA would be rejected because it is 2 mismatches to AAAAA and 3 to CCCCCC and 3 <= 2 + 2
+    |
+    |The set of fixed UMIs may be specified on the command line using --umis umi1 umi2 ... or via one or
+    |more files of UMIs with a single sequence per line using --umi-files umis.txt more_umis.txt
```
nh13: It's also worth saying that the expected UMIs will be used to match against each UMI read in a template (ex. when you have hyphens).

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -0,0 +1,229 @@
...
+class CorrectUmis
+( @arg(flag="i", doc="Input SAM or BAM file.")  val input: PathToBam,
+  @arg(flag="o", doc="Output SAM or BAM file.") val output: PathToBam,
+  @arg(flag="r", doc="Reject BAM file to save unassigned reads.") val rejects: Option[PathToBam] = None,
+  @arg(flag="M", doc="Metrics file to write.") val metrics: Option[FilePath] = None,
+  @arg(flag="m", doc="Maximum number of mismatches between a UMI and an expected UMI.") val maxMismatches: Int,
+  @arg(flag="d", doc="Minimum distance (in mismatches) to next best UMI.") val minDistance: Int,
+  @arg(flag="u", doc="Expected UMI sequences.", minElements=0) val umis: Seq[String] = Seq.empty,
+  @arg(flag="U", doc="File of UMI sequences, one per line.", minElements=0) val umiFiles: Seq[FilePath] = Seq.empty,
+  @arg(flag="t", doc="Tag in which UMIs are stored.") val umiTag: String = ConsensusTags.UmiBases
```
nh13: Given hts-specs#119 I wonder if the "corrected" umi tag should `MI` by its definition and then the "raw" umi would be in the `RX`.  This would require two command line options.  I am not in favor, but just putting it out there.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    in.safelyClose()
+    out.close()
+    rejectOut.foreach(_.close())
+
+    // Finalize the metrics
+    val sortedMetrics        = umiMetrics.values.toSeq.sortBy(_.umi)
+    val totalWithUnmatched   = sortedMetrics.map(_.total_matches).sum.toDouble
+    val meanWithoutUnmatched = sortedMetrics.filter(_.umi != unmatchedUmi).map(_.total_matches).sum / (sortedMetrics.size - 1d)
```
nh13: Could divide by zero if there is only one UMI.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -0,0 +1,229 @@
...
+      if (missingUmisRecords > 0) logger.error(s"# ${mismatchedRecords} were missing UMI attributes in the BAM file!")
+      if (wrongLengthRecords > 0) logger.error(s"# ${wrongLengthRecords} had unexpected UMIs of differing lengths in the BAM file!")
+      logger.error("###################################################################")
+    }
+  }
+
+  /** Given a UMI sequence and a set of fixed UMIs, report the best match. */
+  private[umi] def findBestMatch(bases: String, umis: Array[String]): UmiMatch = {
+    val mismatchLimit = this.maxMismatches + minDistance
+    val mismatches    = umis.map(umi => Sequences.countMismatches(bases, umi))
```
nh13: looking at the `countMismatches` code, it treats a no-call as a mismatch.  Perhaps we should not count it as a mismatch if the no-call is in the expected UMI sequence?

nh13: If you find that this method is slow, you could add parameter to `.countMismatches` to have a maximum # of mismatches before it aborts, as well as changing the `val min = mismatches.min` to return the smallest two values (ex. `val min = mismatches.min(2)` or something like that.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -36,10 +36,16 @@ object Sequences {
...
 
-    s1.zip(s2).count(pair => pair._1.toUpper != pair._2.toUpper)
+    var count = 0
+    forloop (from=0, until=s1.length) { i =>
+      val a = Character.toUpperCase(s1.charAt(i))
+      val b = Character.toUpperCase(s2.charAt(i))
+      if (a != b) count += 1
+    }
+
+    count
```
nh13: Was this slow and that's why it changed?

src/test/scala/com/fulcrumgenomics/testing/UnitSpec.scala
```diff
@@ -71,6 +73,25 @@ trait UnitSpec extends FlatSpec with Matchers {
     ParsingUtil.findClpAnnotation(cl).getOrElse(fail(s"${name} is missing the clp annotation."))
     new CommandLineProgramParser(cl)
   }
+
+  /**
+    * Executes the provided tool and returns the tools logging output as a list of String.
+    */
+  protected def executeFgbioTool(tool: FgBioTool): Seq[String] = {
```
nh13: I don't think this is thread safe unfortunately.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -0,0 +1,229 @@
...
+( @arg(flag="i", doc="Input SAM or BAM file.")  val input: PathToBam,
+  @arg(flag="o", doc="Output SAM or BAM file.") val output: PathToBam,
+  @arg(flag="r", doc="Reject BAM file to save unassigned reads.") val rejects: Option[PathToBam] = None,
+  @arg(flag="M", doc="Metrics file to write.") val metrics: Option[FilePath] = None,
+  @arg(flag="m", doc="Maximum number of mismatches between a UMI and an expected UMI.") val maxMismatches: Int,
+  @arg(flag="d", doc="Minimum distance (in mismatches) to next best UMI.") val minDistance: Int,
+  @arg(flag="u", doc="Expected UMI sequences.", minElements=0) val umis: Seq[String] = Seq.empty,
+  @arg(flag="U", doc="File of UMI sequences, one per line.", minElements=0) val umiFiles: Seq[FilePath] = Seq.empty,
+  @arg(flag="t", doc="Tag in which UMIs are stored.") val umiTag: String = ConsensusTags.UmiBases
+) extends FgBioTool with LazyLogging {
```
nh13: I think this tool will not complain about two reads with the following UMIs:
1. `AAACCC`
2. `AAA-CCC`
though we should treat them differently with respect to metrics collection (r1, r2, template)

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
tfenne: I'm not sure that adds anything.  The Umis are pretty much the only strings being flung around, and I think if they were type-def'd I'd spend more time thinking about "wait, are they Strings or byte[]s underneath?"

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
tfenne: Definitely not `MI`, but yes;  I'm just assuming I'll have to come back and make another pass once that PR gets finalized.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
tfenne: It absolutely will complain!  It requires that a) all the fixed UMIs be the same length, and then b) all the individual UMIs it examines be the same length as the fixed UMIs.  So you can't have a 6bp UMI and a pair of 3bp UMIs in the same BAM without getting a fairly strong warning/error.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
tfenne: I kinda feel like the UMIs are what they are by the time you're running this tool.  It's true that if they are too close together they may not be able to use the full range of matching distances, but I think the time to be thinking about that is during design.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
tfenne: It's a bit confusing, but `sortedMetrics` is the set of all UMIs + 1 entry for unmatched.  So min size is 2, and that's why the `-1` is there.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
tfenne: It seems to work nicely right now for a small-ish (10s) numbers of UMIs.  I think it's probably fine to leave it as is and tune it more if we find it slow on larger sets of UMIs later.

I always tie myself in knots with thinking about capping the number of mismatches, and I figure since UMIs are generally short (say 4-20bp) there's probably not a ton to be gained by aborting early on the counting mismatches.  It _might_ be worth trying to abort early if two sequences are found with < minMismatches, but that's a bit more complicated.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
tfenne: I have to admit, I didn't test, but I can't imagine it was very efficient creating a `Seq[(Char,Char)]` for each comparison, and also using `Char.toUpper` which implicitly generates a `RichChar` object per char.

src/test/scala/com/fulcrumgenomics/testing/UnitSpec.scala
tfenne: Crap, you're right.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
nh13: Can you log an issue and link it to hts-specs#119?

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
nh13: 👍

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
nh13: I agree the time to think about it is during design, but I don't think we can always rely on the design being good since we cannot always perform the design ourselves.  Thinking about it more, I want like to see a double check:
```
set.toSeq
    .combinations(2)
    .map { vec => (vec(0), vec(1) }  // for readability
    .map { case (first, second) => 
        val numMismatches = Sequences.countMismatches(first, second)
        if (numMismatches < minDistance) {
            log.warn(s"Umis are within min-distance $minDistance: $first and $second have $numMismatches distance")
        }
    }
```

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
nh13: 👍

nh13:
LGTM, I really think that you want to know how many UMIs matched each end of a read pair in addition to the full template.  I've assigned an issue to myself since I do want to get to it at some point: https://github.com/fulcrumgenomics/fgbio/issues/207

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
tfenne: I don't agree, but it's easier to implement than to argue more, so I did it.  I did something slightly different though to avoid reporting the same pair as (A,B) and (B,A):

```scala
  def findUmiPairsWithinDistance(umis: Seq[String], distance: Int): Seq[(String,String,Int)] = {
    umis.tails.flatMap {
      case x +: ys => ys.map(y => (x, y, Sequences.countMismatches(x, y))).filter(_._3 <= distance)
      case _      => Seq.empty
    }.toList
  }
```

nh13: Thank-you, fgbio doesn't want its parents to fight.

PR #200, opened 2017-04-03
Various improvements to duplex consensus calling

tfenne:


tfenne:
@nh13 This is definitely a work in progress (at least until I end debugging my latest dataset), but would you like to take a look and see what you think of the direction things are taking?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -196,17 +197,21 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+            else                     (aBase, PhredScore.MinValue)
           }
 
+          // Then mask it if appropriate
+          val (base, qual) = if (aBase == NoCall || bBase == NoCall || rawQual == PhredScore.MinValue) (NoCall, NoCallQual) else (rawBase, rawQual)
+
           bases(i)  = base
           quals(i)  = qual
-          errors(i) = min(sourceReads.count(s => s.bases(i) != NoCall && s.bases(i) != base), Short.MaxValue).toShort
+          errors(i) = min(sourceReads.count(s => s.length > i && s.bases(i) != NoCall && rawBase != NoCall && s.bases(i) != rawBase), Short.MaxValue).toShort
```
nh13: The conditional is a lengthy one and took me a while to agree with.  Any chance you can move it out and comment well?

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -160,25 +162,47 @@ trait UmiConsensusCaller[C <: SimpleRead] {
...
         quals(i) = NoCallQual
       }
     }
 
-    SourceRead(sourceMoleculeId(rec), bases, quals)
+    // Find the last non-N base of sufficient quality in the record, starting from either the
+    // end of the read, or the end of the insert, whichever is shorter!
+    val len = {
+      var index = if (Bams.isFrPair(rec)) min(abs(rec.getInferredInsertSize), bases.length) - 1 else bases.length - 1
+      while (index >= 0 && (bases(index) == NoCall || quals(index) < minBaseQuality)) index -= 1
```
nh13: I think `quals(index) < minBaseQuality` is redundant since you masked them above.

nh13: Thinking about this more, why move the filtering and trimming to the filtering tool?  Or do you want to filter the single strand consensus reads _prior_ to duplex calling?

tfenne: Good catch on the `quals(index)` check. I reorganized this function a couple of times and had meant to remove that.

The filtering that's done here is on the input/`SourceRead`s, so it has to be done up front.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
tfenne: I've changed that line to:
```scala
          errors(i) = min(sourceReads.count(s => s.length > i && isError(s.bases(i), rawBase)), Short.MaxValue).toShort
```
With the new function looking like:
```scala
  /** Function that returns true if the pair of bases are both valid/called bases and do not match each other. */
  @inline private def isError(lhs: Byte, rhs: Byte): Boolean = lhs != NoCall && rhs != NoCall && lhs != rhs
```

tfenne:
@nh13 I think I'm ready to go with this - I don't yet feel the need to make more changes, and I think these are positive changes.  I've addressed your review comments.  Do you want to take one more time with a more critical eye before I merge?

PR #205, opened 2017-04-06
Expose reading in an Illumina RunInfo.xml file.

nh13:


src/main/scala/com/fulcrumgenomics/basecalling/RunInfo.scala
```diff
@@ -0,0 +1,88 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.basecalling
```
tfenne: Now that it's not tied to creating a file for basecalling, it should probably live in the `illumina` package.

src/main/scala/com/fulcrumgenomics/basecalling/RunInfo.scala
```diff
@@ -0,0 +1,88 @@
...
+import htsjdk.samtools.util.Iso8601Date
+
+/** Stores the result of parsing the run info file.
+  *
+  * @param run_barcode the unique identifier for the sequencing run and flowcell, stored as
+  *                   "<instrument-name>_<flowcell-barcode>".
+  * @param flowcell_barcode the flowcell barcode.
+  * @param instrument_name the instrument name.
+  * @param run_date the date of the sequencing run.
+  * @param read_structure the description of the logical structure of cycles within the sequencing run, including which cycles
```
tfenne: We can't actually tell molecular barcodes from a run info though right?  So in practice this will only contain template and sample barcode operators?

src/main/scala/com/fulcrumgenomics/basecalling/RunInfo.scala
```diff
@@ -0,0 +1,88 @@
...
+      run_date         = formatDate(runDate),
+      read_structure   = readStructure,
+      num_lanes        = numLanes
+    )
+  }
+
+  /** Formats the given date string.  The date string should be six or eight characters long with formats YYMMDD or
+    * YYYYMMDD respectively.
+    */
+  private def formatDate(date: String): Iso8601Date = {
```
tfenne: Technically this should be `parseDate` not `formatDate` since the latter implies going from `Date` -> `String`.

src/test/scala/com/fulcrumgenomics/basecalling/ExtractIlluminaRunInfoTest.scala
```diff
@@ -26,77 +26,11 @@
...
-  it should "handle a complicated read structure" in {
-    val info = RunInfo(runInfo(date="170204", readStructure=ReadStructure("8B10T8B10T10T8B")))
-    info.run_barcode             shouldBe "BCDEFGHIJ_NS123456"
-    info.flowcell_barcode        shouldBe "NS123456"
-    info.instrument_name         shouldBe "BCDEFGHIJ"
-    info.run_date                shouldBe  new Iso8601Date("2017-02-04")
-    info.read_structure.toString shouldBe "8B10T8B10T10T8B"
-    info.num_lanes               shouldBe 4
-  }
+  import RunInfoTest._
```
tfenne: I don't love this.  Mostly because it would be easy to e.g. extend the RunInfoTest later and then only realize after doing all the work that this test got broken.  At the very least can you put a big comment in RunInfoTest that this test depends on some of the stuff there?

PR #211, opened 2017-04-20
Added utility methods to parse cigars from String and display pairwise alignments in a simple text format.

tfenne:


tfenne:
Nice little bite-size PR for you here @nh13

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -82,4 +120,55 @@ case class Alignment(query: Array[Byte],
...
   /** One based closed coordinate of the end of the alignment on the query sequence. */
   def targetEnd: Int = targetStart + cigar.lengthOnTarget - 1
+
+  /**
+    * Generates a padded text representation of the alignment for visualization. The returned
+    * sequence will consist of three lines as follows (minus the labels on the left):
+    *
+    * query : ACGTGAACTGACT-ACTGTATGCG
+    * align : |||||  |||||| ||||||||.|
+    * target: ACGTG--CTGACTGACTGTATGGG
```
nh13: UCSC's BLAT output will use `.` instead of `-` for a missing base in either target/query:
```
00000051 agaaatttgggatgtacatcttccagaaattttttt..gaaatcatctga 00000098
>>>>>>>> ||||||||||||||||||||||||||||||||||||  |||||||||||| >>>>>>>>
69999771 agaaatttgggatgtacatcttccagaaatttttttttgaaatcatctga
```
I also like that it gives the start and end coordinates before and after the alignment respectively.  This would be useful for local alignments.  I am sure folks are very familiar with that text format, so perhaps it is good to try to emulate.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -82,4 +120,55 @@ case class Alignment(query: Array[Byte],
...
+    *
+    * @param matchChar the character to use in the alignment line when the bases match
+    * @param mismatchChar the character to use in the alignment line when the bases mismatch
+    * @param gapChar the character to use in the alignment line when one of the sequences is gapped
+    * @return Three lines representing the alignment
+    */
+  def paddedString(matchChar: Char = '|', mismatchChar: Char = '.', gapChar: Char = ' '): Seq[String] = {
```
nh13: You could add a param `seqGapChar: Char = '.') here if you wanted.  Or just make a `case class` that is optionally passed to format the padded string.  Up to you.

nh13: Also, the return may be nicer if it was a case class with `query`, `alignment`, and `target`, with an `unapply` method to give the three lines.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -82,4 +120,55 @@ case class Alignment(query: Array[Byte],
...
+    val buffers = Seq(new StringBuilder, new StringBuilder, new StringBuilder)
+    val Seq(queryBuffer, alignBuffer, targetBuffer) = buffers
+
+    var qOffset = queryStart - 1
+    var tOffset = targetStart - 1
+    cigar.foreach {
+      case CigarElem(CigarOperator.M | CigarOperator.EQ | CigarOperator.X, len) =>
```
nh13: I ❤️  scala so much for this.  It's so easy to understand.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -82,4 +120,55 @@ case class Alignment(query: Array[Byte],
...
+    * @param gapChar the character to use in the alignment line when one of the sequences is gapped
+    * @return Three lines representing the alignment
+    */
+  def paddedString(matchChar: Char = '|', mismatchChar: Char = '.', gapChar: Char = ' '): Seq[String] = {
+    val buffers = Seq(new StringBuilder, new StringBuilder, new StringBuilder)
+    val Seq(queryBuffer, alignBuffer, targetBuffer) = buffers
+
+    var qOffset = queryStart - 1
+    var tOffset = targetStart - 1
+    cigar.foreach {
```
nh13: Part of me wishes that if you could write this as `cigar.foldLeft` and not have to use `vars`, but then you'd have to carry forward five variables, which would be pretty ugly (unless it was part of a case class :P).  I find myself doing this a lot too, namely using `var`s and a `foreach` instead of `foldLeft`.  Sigh.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -40,6 +42,34 @@ case class CigarElem(operator: CigarOperator, length: Int) {
   override def toString: String = length + operator.toString
 }
 
+/** Companion object for Cigar that offers alternative constructors. */
+object Cigar {
+  private val ZeroCharAsInt = '0'.toInt
+
+  /** Constructs a Cigar object from a Cigar string. */
+  def apply(cigar: String): Cigar = {
+    require(cigar != null && cigar.length > 0, "Can't parse a null or empty cigar string.")
```
nh13: Or you could let it blow up in the cigar constructor for an empty string.  Probably nice to have it here so the error message is more specific.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -40,6 +42,34 @@ case class CigarElem(operator: CigarOperator, length: Int) {
...
+    val elements = new ArrayBuffer[CigarElem]
+    val chars    = cigar.toCharArray
+
+    var i = 0
+    while (i < chars.length) {
```
nh13: `forloop`?

nh13: Ah, I see, you need to update `i` below too.

src/test/scala/com/fulcrumgenomics/alignment/AlignmentTest.scala
```diff
@@ -0,0 +1,118 @@
...
+      +||  |||..|||| || |||
+      +AACCGGGTAAACCCGG-TTT
+       """.stripMargin('+').trim.lines.toSeq
+
+    val alignment = Alignment(expected.head.replace("-", ""), expected.last.replace("-", ""), 1, 1, Cigar("2M2D9M1D2M1I3M"), 1)
+    alignment.paddedString() shouldBe expected
+  }
+
+  Seq("5S10M", "10M5H", "5M50N5M", "50P10M").foreach { cigar =>
+    it should s"throw an exception if with unsupported operator contained in cigar $cigar" in {
```
nh13: `if with` -> `if an... is`

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: I'm happy to parameterize it so it's possible to use whatever character you like.  I'm pretty used to seeing `-` though in multiple alignments.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: I did think about that.  My thinking was that a `Seq[String]` would work better if we ever wanted to do something more sensible for longer alignments where we want to wrap them, instead of creating one big long line for each.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: Agreed, this doesn't feel ideal, but the `fold` solution seems even worse to me.

PR #212, opened 2017-04-20
Add some explicit assumption checks in DuplexConsensusCaller.

nh13:
1. That AB-R1s and BA-R2s should be on the same strand, and the
same for AB-R2s and BA-R1s.
2. That UmiConsensusCaller.filterToMostCommonAlignment should only
receive reads from the same strand.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
@@ -139,6 +139,29 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
         val (_, abR1s, abR2s) = subGroupRecords(ab)
         val (_, baR1s, baR2s) = subGroupRecords(ba)
 
+        // Get all the alignments to one end of the source molecule
+        val singleStrand1 = abR1s ++ baR2s
+        val singleStrand2 = abR2s ++ baR1s
+
+        // The orientation of AB and BA reads should be:
+        // AB R1: +  AB R2: -
+        // BA R1: -  BA R2: +
+        // or vice versa (AB-R1:-, AB-R2:+, AB-R1:-, AB-R2: +
+        // Therefore, AB-R1s and BA-R2s should be on the same strand, and the same for AB-R2s and BA-R1s
+        // Check for this explicitly here.
+        if (singleStrand1.nonEmpty) {
+          val ss1Flag = singleStrand1.head.getReadNegativeStrandFlag
+          val ss1MI   = singleStrand1.head.getStringAttribute(ConsensusTags.MolecularId)
+          require(singleStrand1.forall(_.getReadNegativeStrandFlag == ss1Flag),
+            s"Not all AB-R1s and BA-R2s were on the same strand for molecule with id: $ss1MI")
+        }
+        if (singleStrand2.nonEmpty) {
+          val ss2Flag = singleStrand2.head.getReadNegativeStrandFlag
+          val ss2MI   = singleStrand2.head.getStringAttribute(ConsensusTags.MolecularId)
+          require(singleStrand2.forall(_.getReadNegativeStrandFlag == ss2Flag),
+            s"Not all AB-R2s and BA-R1s were on the same strand for molecule with id: $ss2MI")
+        }
+
         // Filter by common indel pattern with AB and BA together
         val filteredXs = filterToMostCommonAlignment(abR1s ++ baR2s)
         val filteredYs = filterToMostCommonAlignment(abR2s ++ baR1s)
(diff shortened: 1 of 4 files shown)


src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -383,6 +383,20 @@ class VanillaUmiConsensusCallerTest extends UnitSpec with OptionValues {
     consensus.getFloatAttribute(ConsensusTags.PerRead.RawReadErrorRate) shouldBe 0.toFloat
   }
 
+  it should "calculate the # of errors relative to the most likely consensus call, even when the final call is an N" in {
```
nh13: Moved this test to the section it belonged to.  It wasn't testing `filterToMostCommonAlignment`

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -139,6 +139,14 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
         val (_, abR1s, abR2s) = subGroupRecords(ab)
         val (_, baR1s, baR2s) = subGroupRecords(ba)
 
+        // The orientation of AB and BA reads should be:
+        // AB R1: +  AB R2: -
```
tfenne: Is this true?  I ask because I honestly can't recall whether `GroupReadsByUmi` constrains things so that AB are always +/- and BA are always -/+, or just that they're the opposite of each other.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -139,6 +139,14 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+        // BA R1: -  BA R2: +
+        // Therefore, AB-R1s and BA-R2s should be on the same strand, and the same for AB-R2s and BA-R1s
+        // Check for this explicitly here.
+        require((abR1s ++ baR2s).map(_.getReadNegativeStrandFlag).distinct.size <= 1, "Not all AB-R1s and BA-R2s were on the same strand")
```
tfenne: Would be nice to assign the result of `abR1s ++ baR2s` to a `val` since we do the exact same concatenation right below, and don't need to do it twice.

tfenne: Also would be nice to print out something here that would allow tracking back to the BAM records when there is a failure.  E.g. including the `MI` in the `require()` message.

tfenne: Also creating a `Seq` of booleans and then calling `distinct` isn't terribly efficient.  If you already have a val for this, then you could do this instead:
`require(xs.forall(_.getReadNegativeStrand == xs.head.getReadNegativeStrand), "fooey")`

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -253,6 +253,7 @@ trait UmiConsensusCaller[C <: SimpleRead] {
     * NOTE: filtered out reads are sent to the [[rejectRecords()]] method and do not need further handling
     */
   protected[umi] def filterToMostCommonAlignment(recs: Seq[SAMRecord]): Seq[SAMRecord] = {
+    require(recs.map(_.getReadNegativeStrandFlag).distinct.size <= 1, "Not all records were on the same strand.")
```
tfenne: Why do this in both places?  This one method is invoked by both duplex and vanilla consensus callers, so will cover both.

src/test/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReadsTest.scala
```diff
@@ -58,14 +59,24 @@ class CallDuplexConsensusReadsTest extends UnitSpec {
     checkClpAnnotations[CallDuplexConsensusReads]
   }
 
+  it should "run fail if AB-R1s are not on the same strand ads BA-R2s" in {
+    val builder = new SamRecordSetBuilder(readLength=10)
+    builder.addPair(name="ab1", start1=100, start2=100, attrs=Map(MI -> "1/A")).foreach { _.setReadString("AAAAAAAAAA") }
```
tfenne: It would be nice to change `start2` just so that it's a little more clear.  The test is valid, but having R1 and R2 starts be the same makes it harder to see that.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
nh13: In case we change this method somehow and the constraint goes away or is modified.  I want it _explicit_ in the duplex caller.

PR #214, opened 2017-04-21
New CollectDuplexSeqMetrics tool for QC'ing duplex seq data

tfenne:


src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,501 @@
...
+    |  - Definition of "duplex" and how minAbReads and minBaReads relate to it
+    |  - 1-2 sentence summary for each the various metrics files produced
+    |
+    |For plots to be generated R must be installed and the ggplot2 package installed with suggested
+    |dependencies. Successfully executing the following in R will ensure a working installation:
+    |    install.packages("ggplot2", repos="http://cran.us.r-project.org", dependencies=TRUE)
+  """)
+class CollectDuplexSeqMetrics
+( @arg(flag="i", doc="Input BAM file generated by GroupReadByUmi.") val input: PathToBam,
+  @arg(flag="o", doc="Prefix of output files to write.") val output: PathPrefix,
```
nh13: Can this be `Option[PathPrefix] = None` and default to use the BAM file after removing the .bam extension?  Obviously it wont work if the input is stdin.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,501 @@
...
+  @arg(flag="l", doc="Optional set of intervals over which to restrict analysis.") val intervals: Option[PathToIntervals] = None,
+  @arg(flag="d", doc="Description of dataset used to label plots. Defaults to sample/library.") val description: Option[String] = None,
+  @arg(flag="a", doc="Minimum AB reads to call a tag family a 'duplex'.") val minAbReads: Int = 1,
+  @arg(flag="b", doc="Minimum BA reads to call a tag family a 'duplex'.") val minBaReads: Int = 1
+) extends FgBioTool with LazyLogging {
+  import CollectDuplexSeqMetrics._
```
nh13: Is there a requirement on the sort order of the input?

tfenne: Yes, but it can't be expressed as a sort order.  It requires things to be in the order that they are produced in by GroupReadsByUmi.  Hence the doc for input = `Input BAM file generated by GroupReadByUmi`.  That order is super useful, but it's not a `SortOrder` - it's sorted by the lower 5' coordinate of the pair, and then by MI.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
tfenne: Maybe?

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,501 @@
...
+    |  - Definition of CS, SS and DS families and how they relate
+    |  - Definition of "duplex" and how minAbReads and minBaReads relate to it
+    |  - 1-2 sentence summary for each the various metrics files produced
+    |
+    |For plots to be generated R must be installed and the ggplot2 package installed with suggested
+    |dependencies. Successfully executing the following in R will ensure a working installation:
+    |    install.packages("ggplot2", repos="http://cran.us.r-project.org", dependencies=TRUE)
+  """)
+class CollectDuplexSeqMetrics
+( @arg(flag="i", doc="Input BAM file generated by GroupReadByUmi.") val input: PathToBam,
```
nh13: You could explicitly state the sort order in case someone has sorted it to `coordinate` and needs to get it back in the correct order ;p

tfenne:
@nh13 Tests are in.  I think this is finally ready for review!

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+                               ss_families: Long,
+                               ds_families: Long,
+                               ds_duplexes: Long,
+                               ds_fraction_duplexes: Double,
+                               ds_fraction_duplexes_ideal: Double) extends Metric
+
+  /**
+    * Metrics describing the set of observed UMI sequences and the frequency of their observations.  The UMI
+    * sequences reported may have been corrected using information within a double-stranded tag family.  For
+    * example if a tag family is comprised of three read pairs with UMIs ACGT-TGGT ACGT-TGGT ACGT-TGGG that
```
nh13: that -> then

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+    * a consensus UMI of will be generated ACGT-TGGT will be generated, and 3 raw observations counted for each
```
nh13: 3 -> three

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+    |```
+  """)
+class CollectDuplexSeqMetrics
+( @arg(flag='i', doc="Input BAM file generated by GroupReadByUmi.") val input: PathToBam,
+  @arg(flag='o', doc="Prefix of output files to write.") val output: PathPrefix,
+  @arg(flag='l', doc="Optional set of intervals over which to restrict analysis.") val intervals: Option[PathToIntervals] = None,
+  @arg(flag='d', doc="Description of dataset used to label plots. Defaults to sample/library.") val description: Option[String] = None,
+  @arg(flag='a', doc="Minimum AB reads to call a tag family a 'duplex'.") val minAbReads: Int = 1,
+  @arg(flag='b', doc="Minimum BA reads to call a tag family a 'duplex'.") val minBaReads: Int = 1,
+  val generatePlots: Boolean = true // not a CLP arg - here to allow disabling of plots to speed up testing
```
nh13: private val?

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+      case None       => _filteredIterator
+      case Some(path) =>
+        val ilist    = IntervalList.fromFile(path.toFile).uniqued(false)
+        val detector = new OverlapDetector[Interval](0,0)
+        detector.addAll(ilist.getIntervals, ilist.getIntervals)
+        _filteredIterator.filter { rec =>
+          val (start, end) = if (rec.refIndex == rec.mateRefIndex) Bams.insertCoordinates(rec) else (rec.start, rec.end)
+          detector.overlapsAny(new Interval(rec.refName, start, end))
+        }
+    }
```
nh13: Have we now done this three times?  Should we move this somewhere?

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+      val libraries = in.readGroups.map(_.getLibrary).distinct
+      (samples, libraries) match {
+        case (Seq(sample), Seq(library)) => s"$sample / $library"
+        case _                           => input.getFileName.toString
+      }
+    }
+
+    // Do a bunch of metrics collection
+    collect(iterator)
+    in.close()
```
nh13: safely please

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+
+    // Write the output files
+    write(description)
+  }
+
+  /** Consumes from the iterator and collects information internal from which to generate metrics. */
```
nh13: internal from which?

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+    val umi2s = new ArrayBuffer[String]
+
+    ab.iterator.map(r => r[String](RX).split('-')).foreach { case Array(u1, u2) => umi1s += u1; umi2s += u2 }
+    ba.iterator.map(r => r[String](RX).split('-')).foreach { case Array(u1, u2) => umi1s += u2; umi2s += u1 }
+
+    Seq(umi1s, umi2s).foreach { umis =>
+      val consensus = callConsensus(umis)
+      if (consensus.contains('N')) {
+        val x = 0
+      }
```
nh13: What does this do?

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+    val ssTotal = metrics.map(_.ss_count).sum.toDouble
+    val dsTotal = metrics.map(_.ds_count).sum.toDouble
+
+    var csInverseCumulativeFraction = 0.0
+    var ssInverseCumulativeFraction = 0.0
+    var dsInverseCumulativeFraction = 0.0
+    metrics.foreach { m =>
+      m.cs_fraction = m.cs_count / csTotal
+      m.ss_fraction = m.ss_count / ssTotal
+      m.ds_fraction = m.ds_count / dsTotal
```
nh13: Protect against divide by zero

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+
+    // Set the fractions
+    val total = metrics.map(_.count).sum.toDouble
+    metrics.foreach(m => m.fraction = m.count / total)
+
+    // Set the cumulative fractions - there's probably a smarter way to do this!
+    metrics.foreach { m =>
+      val countGtOrEq = metrics.iterator.filter(n => n.ab_size >= m.ab_size && n.ba_size >= m.ba_size).map(_.count).sum
+      m.fraction_gt_or_eq_size = countGtOrEq / total
+    }
```
nh13: Nah, this is readable, but this is less likely, but `total` could be zero.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+
+      val dsFamilies       = duplexCounter.map { case (Pair(a,b), count) => count }.sum
+      val countOfDuplexes  = duplexCounter.map {
+        case (Pair(a,b), count) if a >= this.minAbReads && b >= this.minBaReads => count
+        case _                                                                  => 0
+      }.sum
+
+      val countOfDuplexesIdeal = duplexCounter.map { case (pair, count) => count * pDuplexIdeal(pair.ab + pair.ba) }.sum
+
+
```
nh13: nl

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -0,0 +1,540 @@
...
+      new DuplexYieldMetric(
+        fraction                   = fraction,
+        read_pairs                 = startStopCounter.totalMass.toLong,
+        cs_families                = startStopCounter.total,
+        ss_families                = duplexCounter.map { case (Pair(a,b), count) => if (b>0) count*2 else count }.sum,
+        ds_families                = dsFamilies,
+        ds_duplexes                = countOfDuplexes,
+        ds_fraction_duplexes       = countOfDuplexes / dsFamilies.toDouble,
+        ds_fraction_duplexes_ideal = countOfDuplexesIdeal / dsFamilies.toDouble
```
nh13: `dsFamilies ` could be zero

src/test/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetricsTest.scala
```diff
@@ -0,0 +1,299 @@
...
+      families       = Paths.get(prefix.toString + CollectDuplexSeqMetrics.FamilySizeMetricsExt),
+      duplexFamilies = Paths.get(prefix.toString + CollectDuplexSeqMetrics.DuplexFamilySizeMetricsExt),
+      umis           = Paths.get(prefix.toString + CollectDuplexSeqMetrics.UmiMetricsExt),
+      yields         = Paths.get(prefix.toString + CollectDuplexSeqMetrics.YieldMetricsExt),
+      plots          = Paths.get(prefix.toString + CollectDuplexSeqMetrics.PlotsExt)
+    )
+  }
+
+  "CollectDuplexSeqMetrics" should "have acceptable CLP annotations" in {
+    checkClpAnnotations[CollectDuplexSeqMetrics]
```
nh13: Cool, I should use this.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
tfenne: Oops, I think I had that in there so I could catch and set a breakpoint, I'll pull it back out.

src/main/resources/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.R
```diff
@@ -0,0 +1,118 @@
...
+
+#################################################################################
+# R script to generate QC plots for duplex sequencing data QC's with
+# CollectDuplexSeqMetrics tools.  All arguments are positional and required:
+#   1. The family size file with data on families by start/stop, ss and ds
+#   2. The duplex family size file
+#   3. The duplex yield metrics file
+#   4. The UMI metrics file
+#   5. An output PDF to write
+#   6+ One or more strings to use in plot titles to ID the sample/dataset
```
tfenne: The plus was intentional to indicate arguments 6 and on.  I've added a period after the plus to make it more clear it's not just a mistake.

src/main/resources/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.R
```diff
@@ -0,0 +1,118 @@
...
+# Plot #3a - Duplex Family Size Distribution
+ggplot(duplexFamilyData) +
+  aes(x=ab_size, y=ba_size, z=count) +
+  stat_summary_2d(fun=sum, binwidth=1) +
+  scale_fill_gradient2(low=fgblue, mid="white", high=fggreen, space="Lab",
+                       midpoint=log10(max(duplexFamilyData$count)), guide = "colourbar", trans="log2") +
+  labs(x="AB Reads", y="BA Reads", title=paste("Duplex Tag Family Size Distribution for", sampleInfo)) +
+  theme(plot.title = element_text(hjust = 0.5))
+
+# Plot #3b - Duplex Family Size Distribution for families with ab > 1 and ba > 1
```
tfenne: Heh.  I called them 3a and 3b since they're the same plot just with slightly different ranges of data displayed.

src/main/resources/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.R
```diff
@@ -0,0 +1,118 @@
...
+                       midpoint=log10(max(duplexFamilyData$count)), guide = "colourbar", trans="log2") +
+  labs(x="AB Reads", y="BA Reads", title=paste("Duplex Tag Family Size Distribution for ", sampleInfo, " ; Only Families with AB>0 & BA>0", sep="")) +
+  theme(plot.title = element_text(hjust = 0.5))
+
+# Plot #4 - Duplex Yield
+ggplot(yieldData) +
+  aes(x=read_pairs) +
+  geom_area(aes(y=ds_families, fill="DS Families")) +
+  geom_area(aes(y=ds_families * ds_fraction_duplexes_ideal, fill="Duplexes - Ideal")) +
+  geom_area(aes(y=ds_duplexes, fill="Duplexes - Actual")) +
```
tfenne: That's why it's _after_.  Since these are distinct `geom_areas` they are added to the plot in the order they are specified here.  There are always more families than ideal duplexes, and more ideal duplexes than actual duplexes, so doing it this way they stack nicely and are all visibible.  Re-arranging would mean that those with more would hide those with less.

src/main/resources/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.R
```diff
@@ -0,0 +1,118 @@
...
+# Plot #5 UMI distribution (for UMIs that don't contain any no-calls)
+ggplot(subset(umiData, !grepl("N", umiData$umi, fixed=T))) +
+  aes(x=raw_observations, y=unique_observations) +
+  geom_point(color=fggreen) +
+  labs(x="Observations of UMI in Raw Reads", y="Unique Observations (Tag Families w/UMI)", title=paste("UMI Representation in", sampleInfo)) +
+  theme(plot.title = element_text(hjust = 0.5), legend.title=element_blank())
+
+# Plot #6 - Distribution of reads within tag families
+ggplot(familyData) + aes(x=family_size) +
+  geom_line(aes(y=cs_count * family_size, color="By Position")) +
```
tfenne: Yeah, I'll change it.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
tfenne: I feel like each time we do it something is slightly different - e.g. the set of filters ... using the read vs. insert cooridnates etc.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
tfenne: The totals can only be zero if there are no observations at all, in which case `map` and `metrics` will both be empty so we never get into this `foreach`

PR #215, opened 2017-04-24
Tool to convert from Fastq to unmapped BAM

tfenne:


src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -24,88 +24,123 @@
...
+  val AnyLengthChar: Char = '+'
+
+  /** Can be subtracted from any digit character to get it's integer value. */
+  private val DigitOffset = '0'.toInt
+
+  /** Contains the bases and optionally base qualities that correspond to the given read segment. */
+  sealed trait SubRead {
+    def bases: String
+    def segment: ReadSegment
+    def kind: SegmentType = this.segment.kind
```
tfenne: Now that `SegmentType` is a class, I think I would actually like to get rid of `segment` in `SubRead` and replace it with a direct reference to the `SegmentType`.  Any objections?

tfenne:
@nh13 A few thoughts on the `ReadStructure` stuff in particular:

1. I turned `SubRead` into a trait with two subclasses, one that has quals and one that doesn't.  I felt like 99% of the time we're going to want to break down reads and quals, and having quals as an `Option[String]` is annoying and inefficient when doing so.
2. `ReadSegment` used to have four sub-types, one for each kind of segment.  I think that was my suggestion originally, and now I'm saying I think that was wrong.  Their behaviors are identical, and having four sub-types led two things I think were sub-optimal:
    1. Because we had types for segments, we didn't bother having a class for SegmentType so the only way to pass that information around was either as a `Char` or by passing around the whole `ReadSegment`
    2. Lots of extra code to do things like copy a segment, because you had to pattern match by type to call `.copy()` on it
3. I've introduced `SegmentType` which is basically an `enum` implemented with a `sealed trait` and case _objects_, and collapsed `ReadSegment` into a single case class.  I think this makes using this stuff much tidier.

I'm sorely tempted to remove public assess to the object method `extract(bases: String, quals: String, segments: Seq[ReadSegment]): Seq[SubReadWithQuals]` that allows extraction of a sub-set of segments.  I think this came about as a reaction to how awkward it was to segregate sub-reads by type after doing the extraction on the whole read.  But now you can write things like:

```scala
val subs = rs.extract(bases, quals)
val umi = subs.filter(_.kind == MolecularBarcode).mkString("-")
val templates = subs.filter(_.kind == Template)
require(1 to 2 contains templates.length, "One 1 or 2 template reads allowed")
templates.foreach { t => write(t) }
```

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
nh13: My only concern would be that we would want some information from `ReadSegment` (i.e. `length` or `offset`) that we would lose by getting rid of it...

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -24,88 +24,136 @@
...
+  }
+  case class SubReadWithoutQuals(bases: String, segment: ReadSegment) extends SubRead
+  case class SubReadWithQuals(bases: String, quals: String, segment: ReadSegment) extends SubRead
```
nh13: I see you too were getting tired of the `Option[String]` for quals.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -24,88 +24,136 @@
...
+    * @param bases the sequenced bases
+    * @param segments the set of segments to extract
+    * @return a sequence of [[SubRead]] objects, one for each segment
+    */
+  def extract(bases: String, segments: Seq[ReadSegment]): Seq[SubReadWithoutQuals] = segments.map(_.extract(bases))
+
+  /**
+    * Creates sub-reads for the set of read segments provided.
+    * @param bases the sequenced bases
+    * @param quals the quality scores for the bases (must be same length as bases)
```
nh13: I think you could get away with the qualities being longer than bases fwiw.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -24,88 +24,136 @@
...
     **/
-  def structureRead(bases: String, segments: Seq[ReadSegment], strict: Boolean = true): Seq[SubRead] = {
-    // TODO: TF: strict == false allows us to skip segments (in the filter)
-    segments.filter(s => strict || s.offset < bases.length).map {
-      segment =>
-        val (start, end) = segment.range(bases.length, strict=strict)
-        SubRead(bases=bases.substring(start, end), segment=ReadSegment(segment, end - start))
-    }
+  @deprecated(message="Use ReadStructure.extract() instead.")
+  def structureRead(bases: String, segments: Seq[ReadSegment]): Seq[SubReadWithoutQuals] = {
```
nh13: Since both the parameters and return type have changed, I am in favor of removing the previous method altogether!

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -24,88 +24,136 @@
...
+    * return [[com.fulcrumgenomics.util.ReadStructure.SubRead]]s for which we have bases in `bases`, otherwise throw an exception.
     **/
-  def structureReadWithQualities(bases: String, qualities: String, segments: Seq[ReadSegment], strict: Boolean = true): Seq[SubRead] = {
-    assert(bases.length == qualities.length)
-    segments.filter(s => strict || s.offset < bases.length).map {
-      segment =>
-        val (start, end) = segment.range(Math.min(bases.length, qualities.length), strict=strict)
-        SubRead(bases=bases.substring(start, end), quals=Some(qualities.substring(start,end)), segment=ReadSegment(segment, end - start))
-    }
+  @deprecated(message="Use ReadStructure.extract() instead.")
```
nh13: Ditto to killing this.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -114,141 +162,125 @@ object ReadStructure {
...
-    case seg: SampleBarcode    => SampleBarcode(offset, length)
-    case seg: MolecularBarcode => MolecularBarcode(offset, length)
-    case seg: Skip             => Skip(offset, length)
+  /** Returns the [[SegmentType]] for the given code/letter. */
+  def apply(code: Char): SegmentType = code match {
+    case 'T' => Template
+    case 'B' => SampleBarcode
+    case 'M' => MolecularBarcode
+    case 'S' => Skip
+    case _   => throw new IllegalArgumentException(s"Invalid read segment type: $code")
```
nh13: Perhaps we just store an immutable map internally?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
```diff
@@ -114,141 +162,125 @@ object ReadStructure {
...
+
+  /** Creates a new ReadSegment with undefined length (i.e. length=0 or more). */
+  def apply(offset: Int, c: Char): ReadSegment = new ReadSegment(offset, None, kind=SegmentType(c))
+
+  /** Constructs a ReadSegment with a definite length using Char c to find the segment type. */
+  def apply(offset: Int, length: Int, c: Char): ReadSegment = new ReadSegment(offset=offset, length=Some(length), kind=SegmentType(c))
+}
+
+/** Encapsulates all the information about a segment within a read structure. */
+case class ReadSegment (offset: Int, length: Option[Int], kind: SegmentType) {
```
nh13: Add a quick comment about what it means to have `length=None`.

nh13: What about a method:
```
def getDefiniteLength: Int = this.length.get
```
which will throw an exception when it is `None`.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -234,7 +226,7 @@ object ExtractUmisFromBam {
     // get the bases associated with each segment
     val readStructureBases = readStructure.structureReadWithQualities(bases, qualities, strict = false)
     // get the molecular index segments
-    val molecularIndexBases = readStructureBases.collect { case SubRead(b: String, _, m: MolecularBarcode) => b }
+    val molecularIndexBases = readStructureBases.filter(_.segment.kind == SegmentType.MolecularBarcode).map(_.bases)
```
nh13: `_.segment.kind` to `_.kind`

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -246,17 +238,15 @@ object ExtractUmisFromBam {
...
     }
+
     // keep only template bases and qualities in the output read
-    val basesAndQualities = readStructureBases.flatMap { r =>
-      r.segment match {
-        case m: Template => Some((r.bases, r.quals.getOrElse(unreachable())))
-        case _ => None
-      }
-    }
+    val basesAndQualities = readStructureBases.filter(_.segment.kind == SegmentType.Template)
```
nh13: ditto: `_.kind`

src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
```diff
@@ -29,10 +29,10 @@ import htsjdk.samtools.util.AbstractProgressLogger
 /**
   * A subclass of HTSJDK's progress logger that uses fgbio's logging system.
   */
-class ProgressLogger(val logger: Logger,
-                     noun: String = "record",
-                     verb: String = "processed",
-                     unit: Int = 1000 * 1000) extends AbstractProgressLogger(noun, verb, unit) {
+case class ProgressLogger(logger: Logger,
```
nh13: Why remove the `val` when it's going to make it a `val` anyhow?

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
tfenne: Woohoo!  Done!

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
tfenne: So long as this is only in one place I'm happy with the match statement.

src/main/scala/com/fulcrumgenomics/util/ReadStructure.scala
tfenne: Good call.  I've changed the `hasDefiniteLength` to `hasFixedLength`.  I originally thought the former was a good idea since it mimic'd what scala collections use, but that's really for distinguishing finite vs. infinite not exact vs. variable, so I think `fixed` is better.  I've added `fixedLength` and used it where appropriate.

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
```diff
@@ -38,6 +38,11 @@ class ExtractUmisFromBamTest extends UnitSpec with OptionValues {
 
   def newBam = makeTempFile("extract_umis_from_bam_test.", ".bam")
 
+  /** An implicit conversion to read stucture to make test writing easier below by allowing the use of Strings
+    * directly instead of having to write ReadStructure("75T") everywhere.
+    */
+  private implicit def stringToReadStructure(rs: String): ReadStructure = ReadStructure(rs)
```
tfenne: I'm really pleased I thought of this instead of rewriting all the lines in the test that construct `ExtractReadsFromUmi` to have e.g. `readStructures=Seq(ReadStructure("75T"))` instead of `readStructures=Seq("75T")`.  This makes me thing that we should actually make _more_ use of private implicits in our tests to make things easier to read.

tfenne:
@nh13 I've made the final set of changes we talked about.  I'll hold off merging until you're ready, but you might want to take a look at the last commit.

PR #217, opened 2017-04-26
Ensure MI is in increasing order

nh13:


nh13:
@tfenne curious about what you think.

nh13:
@tfenne https://www.youtube.com/watch?v=8lYBVi3ALWA

nh13:
@tfenne I updated based on your suggestion.  Lets just hope that the first few groups are OK, since this solution _doesn't_ work if the first 100 reads are grouped (i.e. `seq 1 100 | sort` vs `seq 1 100 | sort -n`)

tfenne:
👍  I'm not sure I understand your comment about 1-100 though.

PR #219, opened 2017-04-30
Wrappers around HTSJDK's SAM/BAM reading/writing classes and better disk-backed sorting.

tfenne:
@nh13 This is still something of a work in progress.  Specifically it needs a) tests, b) to go through the entire rest of the codebase and replace usages.  But before I did all of that I wanted for you to take a look and see if you liked where I'm going and have any feedback.

The main ideas here are:
1. Provide a cleaner scala API for reading/writing BAM files and for constructing the reader/writer objects
2. Provide some centralization so it's easy for us to change how we do these things across the codebase should we find the need
3. Provide a trait-based mechanism for us to add methods and functionality to SAMRecord that works for both SAMRecord and BAMRecord without duplication.

It's not a ton of code, but it's quite a change from how we're doing things now.  If you have a little time I'd suggest starting with `SamRecord.scala` and `SamRecordFactory.scala`, and then going from there.


tfenne:
@nh13 This isn't quite done yet - I still need to write some more tests if nothing else.  But I would really appreciate you taking a look at this point and giving some feedback on the API.

Particularly I think it would be helpful if you could look at:

1. `SamRecord` and friends in that file
2. `SamSource` and `SamWriter`
3. `Sorter` and `SamOrder` and friends
4. A handful of example usages in the rest of the code base

Feel free to go as deep as you'd like, or to do a higher level review and await a final PR for a full review.  One question: do you think it's worth it to add scala-doc to all the property accessors in the new `SamRecord`?   I don't mind writing them, but I feel like it's going to make the class harder to read.

src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
```diff
@@ -0,0 +1,221 @@
...
+  /** The subsort string to be placed *after* the sort order in the SS tag. */
+  def subSort: Option[String]
+
+  /** Function to generate the sort key for sorting records into this ordering. */
+  def sortkey: (SamRecord) => A
+}
+
+object SamOrder {
+  /** The set of all possible values. */
+  val values: Seq[SamOrder] = Seq(Coordinate, Queryname, Random, RandomQuery, TemplateCoordinate, Unsorted, Unknown)
```
nh13: I recall some PR in dagr to get all the classes that extend a trait reflectively :P

src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+    * Returns the SamOrder for the given name. Throws an exception if the name doesn't match a known ordering.
+    * Performs case insensitive matching so that we can use capitalized class names, but have backwards
+    * compatibility to SAM spec sort orders that are all lower-case.
+    * */
+  def apply(name: String): SamOrder = values.find(_.productPrefix.equalsIgnoreCase(name)).getOrElse {
+    throw new NoSuchElementException("No SamOrder: " + name)
+  }
```
nh13: Are you building an Enum?

src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
```diff
@@ -0,0 +1,221 @@
...
+
+  /** Ordering object for coordinate order per the SAM spec. */
+  case object Coordinate extends SamOrder {
+    override type A = CoordinateKey
+    override val sortOrder:  SortOrder      = SortOrder.coordinate
+    override val groupOrder: GroupOrder     = GroupOrder.none
+    override val subSort:    Option[String] = None
+    override val sortkey: SamRecord => A = rec => {
+      val ref = rec.refIndex
+      val idx = if (ref < 0) Int.MaxValue else ref
```
nh13: Add a note that makes it so records without coordinates are sorted to the end.

src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
```diff
@@ -0,0 +1,221 @@
...
+  case object Queryname extends SamOrder {
+    override type A = QuerynameKey
+    override val sortOrder:  SortOrder      = SortOrder.queryname
+    override val groupOrder: GroupOrder     = GroupOrder.none
+    override val subSort:    Option[String] = None
+    override val sortkey: (SamRecord => A)  = r => QuerynameKey(r.name, r.asSam.getFlags)
+  }
+
+  /** Ordered key object for Queryname order. */
+  case class QuerynameKey(name: String, flags: Int) extends Ordered[QuerynameKey] {
```
nh13: Can `flags` be a `Short`?

src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
```diff
@@ -0,0 +1,221 @@
...
+  case object Random extends SamOrder {
+    override type A = RandomKey
+    private  val hasher = new Murmur3(42)
+    override val sortOrder:  SortOrder      = SortOrder.unsorted
+    override val groupOrder: GroupOrder     = GroupOrder.none
+    override val subSort:    Option[String] = None
+    override val sortkey: (SamRecord => A)  = rec => RandomKey(hasher.hashUnencodedChars(rec.id + rec.basesString), rec.asSam.getFlags)
+  }
+
+  /** Ordering object for generating a random order with queryname grouping.. */
```
nh13: extra period

src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
tfenne: It _could_ but it gets quite messy because Java only really supports unsigned `Int` and `Long`, and not `Short`.  So I could store it as a `Short`, but then have to use methods to convert it to an unsigned int at every use.  Not clear that would be a win.

tfenne:
@nh13 I've addressed a bunch of your comments, and also added some new tests to cover the new functionality.  I'm sure there could be more tests, but I think coverage is reasonable, especially since so much of the code is tested via other tests that use `SamBuilder` etc.

Want to take a final look?  I'd love to get this merged and be able to rebase other PRs onto it.

nh13:
@tfenne ha, reviewed before you asked.

PR #231, opened 2017-05-16
Add a tool to SplitBamByReadGroup

nh13:


src/main/scala/com/fulcrumgenomics/bam/SplitBamByReadGroup.scala
```diff
@@ -0,0 +1,113 @@
...
+import com.fulcrumgenomics.commons.CommonsDef.{PathPrefix, PathToBam, SafelyClosable, javaIterableToIterator, javaIteratorAsScalaIterator}
+import com.fulcrumgenomics.commons.io.{Io, PathUtil}
+import com.fulcrumgenomics.commons.util.{LazyLogging, SimpleCounter}
+import com.fulcrumgenomics.sopt.{arg, clp}
+import com.fulcrumgenomics.util.ProgressLogger
+import htsjdk.samtools._
+
+@clp(group = ClpGroups.SamOrBam, description=
+  """
+    |Splits a BAM into multiple BAMs per-read group (or library).
```
tfenne: ..BAMs, one per read group....

src/main/scala/com/fulcrumgenomics/bam/SplitBamByReadGroup.scala
```diff
@@ -0,0 +1,113 @@
...
+    |
+    |The resulting BAM will be named `<output-prefix>.<read-group-id>.bam`, or `<output-prefix>.<library-name>.bam` when
```
tfenne: BAM -> BAMs

src/main/scala/com/fulcrumgenomics/bam/SplitBamByReadGroup.scala
```diff
@@ -0,0 +1,113 @@
...
+    |splitting by the library.  All reads without a read group, or without a library when splitting by library, will be
+    |written to `<output-prefix>.unknown.bam`.
+  """)
+class SplitBamByReadGroup
+( @arg(flag='i', doc="Input SAM or BAM file.") val input: PathToBam,
+  @arg(flag='o', doc="Output prefix for all SAM or BAM files (ex. output/sample-name).") val output: PathPrefix,
+  @arg(flag='l', doc="Split by library instead of read group") val byLibrary: Boolean = false,
```
tfenne: Could this be a string of enum argument?  With options "ReadGroup" and "Library".  That way, if we want to add another split type later (e.g. split by read number), it won't break compatibility.

src/main/scala/com/fulcrumgenomics/bam/SplitBamByReadGroup.scala
```diff
@@ -0,0 +1,113 @@
...
+  @arg(flag='u', doc="The name to use for the unknown file") val unknown: String = "unknown"
+) extends FgBioTool with LazyLogging {
+
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+
+  override def execute(): Unit = {
+    val in       = SamReaderFactory.make().open(input)
+    val progress = ProgressLogger(logger)
+    val unknownBamAndWriter = {
```
tfenne: It would be nice if the unknown writer could be lazily constructed so that you don't end up with an empty unknown bam on each run.  Or alternatively delete the BAM if no records are output to it?

src/main/scala/com/fulcrumgenomics/bam/SplitBamByReadGroup.scala
```diff
@@ -0,0 +1,113 @@
...
+      val unknownBam: PathToBam = toOutput(unknown)
+      val unknownWriter: SAMFileWriter = new SAMFileWriterFactory().makeWriter(in.getFileHeader, true, toOutput(unknown).toFile, null)
+      WriterInfo(name=unknown, bam=unknownBam, writer=unknownWriter)
+    }
+    val writers  = createWriters(header=in.getFileHeader, unknownBamAndWriter=unknownBamAndWriter)
```
tfenne: Why not just have `createWriters` also create the unknown too?

tfenne: I see .... you're just using it as the default in the map.  I think in that case it would be clearer if you just changed this line to:
`createWriters(header=in.getFileHeader).withDefaultValue(unknownBamAndWriter)`

src/main/scala/com/fulcrumgenomics/bam/SplitBamByReadGroup.scala
```diff
@@ -0,0 +1,113 @@
...
+    unknownBamAndWriter.close()
+    in.safelyClose()
+
+    counter.toSeq sortBy { _._1.name } foreach { case (info, count) =>
+      logger.info(s"Wrote $count records to ${info.bam.getFileName}")
+    }
+  }
+
+  /** Stores the path to the BAM and the associated writer */
+  private case class WriterInfo(name: String, bam: PathToBam, writer: SAMFileWriter) {
```
tfenne: Since it implements it you could `extends Closeable`

src/main/scala/com/fulcrumgenomics/bam/SplitBamByReadGroup.scala
```diff
@@ -0,0 +1,113 @@
...
+        rg -> WriterInfo(name=rg.getId, bam=bam, writer=writer)
+      }
+    }
+
+    toWriter.toMap.withDefaultValue(unknownBamAndWriter)
+  }
+
+  private[bam] def toOutput(name: String): PathToBam = {
+    val outputDir = output.getParent
+    val prefix    = PathUtil.basename(output)
```
tfenne: I think you want just `output.getFileName` here.  `PathUtil.basename` will also remove any extension.  I.e. if I say I want my prefix to be `/tmp/x.by_lib` it will strip off the `by_lib` which I wouldn't expect.

src/test/scala/com/fulcrumgenomics/bam/SplitBamByReadGroupTest.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  private def newOutput = Files.createTempDirectory("split_bam_by_read_group_test").resolve("prefix")
+
+  // HACK!
+  private val defaultRgId: String = new SAMRecordSetBuilder().getHeader.getReadGroups.map(_.getId).toSeq.head
+  private val otherRgId: String = "2"
+
+  // Make a set of records to run multiple tests on
+  private val builder = new SamRecordSetBuilder(readGroupId=Some(otherRgId))
+  builder.addPair("rg_default_1", 0, 100, 300).foreach(_.setAttribute("RG", defaultRgId))
```
tfenne: Can you use `attrs=Map("RG" -> defaultRgId)` instead please?

tfenne: Also, it'll make moving this to the new builder much easier if you name the parameters please.  I.e. 
```scala
builder.addPair("rg_default_1", 0, 100, 300).foreach(_.setAttribute("RG", defaultRgId))
->
builder.addPair(name="rg_default_1", contig=0, start1=100, start2=300, attrs=Map("RG" -> defaultRgId))
```

src/main/scala/com/fulcrumgenomics/bam/SplitBamByReadGroup.scala
```diff
@@ -0,0 +1,113 @@
...
+
+@clp(group = ClpGroups.SamOrBam, description=
+  """
+    |Splits a BAM into multiple BAMs per-read group (or library).
+    |
+    |The resulting BAM will be named `<output-prefix>.<read-group-id>.bam`, or `<output-prefix>.<library-name>.bam` when
+    |splitting by the library.  All reads without a read group, or without a library when splitting by library, will be
+    |written to `<output-prefix>.unknown.bam`.
+  """)
+class SplitBamByReadGroup
```
tfenne: One last comment - can we just call it `SplitBam` since it splits by RG or Library?

nh13:
@tfenne could you take one look, in particular to the new `ClpEnum` and `ClpEnumEntry` traits in `FgBioDef` that use the `enumeratum` library, and then the usage of that in the `SplitBam` tool?

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -25,8 +25,23 @@
...
+
+  /** Extends this trait in your enumeration object to enable the case objects to be created on the command line.
+    * You should implement the [[Enum#values]] method in your object, for example:
+    * `def values: IndexedSeq[T] = findValues`.
+    * */
+  trait ClpEnum[T<:ClpEnumEntry] extends Enum[T] {
+    // NB: we cannot put `def values: IndexedSeq[T] = findValues` here since ClpEnum is not a class!
+
+    /** The string constructor method that enables us to create the case object instance by name. */
+    def apply(name: String): T = withName(name)
```
tfenne: It sorta sucks that we have to do this here.  I would have thought that Enumeratum would supply an `apply()` method :/

nh13:
The `values` method is required to make it all work.  Unfortunately it has to be defined in the enum class and not the base trait due to reflection magic under the covers to my best understanding.

PR #233, opened 2017-05-18
Clip a fixed amount on the 5' end of reads.

nh13:
@tfenne I was loathe to create yet-another-tool and chain these together because:
1. I would have to add a output `sort order` to one of the two tools so that we don't sort unnecessarily (`coordinate -> query -> coordinate` twice).
2. The actual amount of code is so little, I would hate to duplicate it.
The option I _could_ do is to create another tool called `ClipReads` and have `ClipOverlappingReads` extend it but fix `numStartClip=0`.  That way we have the previous tool and a new tool, though the documentation (usage and arg annotations) would be duplicated.  Therefore I chose not to go down this route.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
@@ -1,0 +1,144 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.bam
+
+import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.bam.SamRecordClipper.ClippingMode
+import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamSource, SamWriter}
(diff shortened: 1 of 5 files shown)


src/main/scala/com/fulcrumgenomics/bam/ClipOverlappingReads.scala
```diff
@@ -117,6 +121,12 @@ class ClipOverlappingReads
     else {
       val (f,r) = if (r1.getReadNegativeStrandFlag) (r2, r1) else (r1, r2)
 
+      // CLip the 5' end of both reads if desired
+      if (this.numStartClip > 0) {
```
tfenne: Note that adding this here means it will not clip any reads that are not F/R pairs on the same chromosome.  That seems ok if this is only used to feed the consensus pipeline, but it's a general tool so should probably clip all reads.

nh13:
Added a few more tests and created the `ClipBam` tool.  Rebased onto master (not _that_ bad).  I am merging when tests pass.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -0,0 +1,145 @@
...
+import com.fulcrumgenomics.sopt.{arg, clp}
+import com.fulcrumgenomics.util.{Io, ProgressLogger}
+import htsjdk.samtools.SAMFileHeader.SortOrder
+import htsjdk.samtools.SamPairUtil.PairOrientation
+import htsjdk.samtools._
+import htsjdk.samtools.reference.ReferenceSequenceFileWalker
+
+@clp(group = ClpGroups.SamOrBam, description=
+  """
+    |Clips reads from the same template.Ensures that at least N bases are clipped from any end of the read (i.e.
```
tfenne: Needs a space after the period.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -0,0 +1,145 @@
...
+    |R1 5' end, R1 3' end, R2 5' end, and R2 3' end).  Optionally lips reads from the same template to eliminate overlap
```
tfenne: lips -> clips

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -0,0 +1,145 @@
...
+    |resorted into coordinate order, any existing `NM`, `UQ` and `MD` tags are repaired, and the output is
+    |written in coordinate order.
+  """)
+class ClipBam
+( @arg(flag='i', doc="Input SAM or BAM file of aligned reads in coordinate order.") val input: PathToBam,
+  @arg(flag='o', doc="Output SAM or BAM file.") val output: PathToBam,
+  @arg(flag='r', doc="Reference sequence fasta file.") val ref: PathToFasta,
+  @arg(flag='s', doc="Soft clip reads instead of hard clipping.") val softClip: Boolean = false,
+  @arg(flag='a', doc="Automatically clip extended attributes that are the same length as bases.") val autoClipAttributes: Boolean = false,
+  @arg(          doc="Require at least this number of bases to be clipped on R1 5'") val readOneFivePrime: Int  = 0,
```
tfenne: `on R1 5'` -> `on the 5' end of R1.`

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -0,0 +1,145 @@
...
+      // we could just trim read bases in a loop until the overlap is eliminated!
+      while (f.end >= r.start && f.mapped && r.mapped) {
+        val lengthToClip = f.end - r.start + 1
+        val firstHalf    = lengthToClip / 2
+        val secondHalf   = lengthToClip - firstHalf // safe guard against rounding on odd lengths
+        this.clipper.clip3PrimeEndOfAlignment(r1, firstHalf)
+        this.clipper.clip3PrimeEndOfAlignment(r2, secondHalf)
+      }
+    }
+
```
tfenne: Bonus newline!

Issue #222, opened 2017-05-04
HapCUT2 format has changed slightly


pjedge:
Sorry for the change in format, but it this change will really improve interpretability of HapCUT2 output.

The value in HapCUT2 output corresponding to the 'SE' tag is now represented by a floating point phred-scaled quality score (closer to 0 means switch is likely, closer to 100 means switch is unlikely).
Similarly, the value corresponding to 'ME' is represented by a phred-scaled quality score (closer to 0 means mismatch is likely, closer to 100 means mismatch is unlikely)

If that were all, I'd just commit a change to the description string in https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/vcf/HapCutToVcf.scala with the updated information. However there is also a change that if a score is not explicitly computed (based on the flags used) it is represented as a period '.' rather than just being given the max score. I'm not familiar with scala at all so if someone else could add code to handle this case that'd be fantastic.

nh13:
@pjedge thanks for keeping us in the loop on this, and I'll take a look at updating it.

nh13:
@pjedge do you have some example output you could send me.  I wouldn't mind having a test case or two from a real example dataset.

pjedge:
Hey Nils. Sorry for the delay. Sent you 3 example files to your gmail.

nh13:
@pjedge can you take a look and potentially try the tool on your data?

PR #235, opened 2017-05-19
Updated HapCutToVcf to support updated HapCut2 outputs.

nh13:
The "pruned" column, and likelihoods (no switch, mismatch) columns can have missing values (".").

The likelihoods are no phred-scaled.

Fixes https://github.com/fulcrumgenomics/fgbio/issues/222

@pjedge would you like to review the definitions of SE and NE to see if they match up?


nh13:
@pjedge does this work for you in your hands?  Can you take a quick look to make sure I am defining the outputs correctly?

nh13:
@pjedge bump!

pjedge:
hey Nils, sorry for the delay.

A couple things:
(1) In the "skip_prune" test case I notice a mismatch between the number of format fields (5) and the number of corresponding values fields for the individual (4) for the phased positions. Looks like SE is being left off?
(2) The annotation at the top of the VCF for the PR field is confusing -- the field refers to column 9 of the HapCUT2 format I believe? But technically HapCUT2 prunes using a threshold on the value of NE field and PR field is unrelated. PR field refers to status (unpruned or pruned) using a discrete pruning strategy used by other haplotyping tools but not used by default in HapCUT2 (prune if there are equal number of alleles consistent and inconsistent with the phase).
(3) I'm not knowledgeable on this, is it standard to use the first genomic position to define a phase set? If so, it seems technically you need to look at both chromosome and PS values together to know what's a block right (since you may have blocks on different chromosomes starting at the same position)? It's common sense that they're not phased together but I'm just not sure on standards here.

Other than these things, it looks good as far as I can tell

nh13:
@pjedge 
(1) this occurs when " if a score is not explicitly computed (based on the flags used) it is represented as a period '.' rather than just being given the max score".  When it is present, it is outputted (I confirmed manually).
(2) how about `"1 if HapCut2 pruned this variant when using --discrete_pruning, 0 otherwise"`
(3) see page 10 and the "PS" tag here: https://samtools.github.io/hts-specs/VCFv4.3.pdf.  By definition, they apply only to genotypes on the same chromosome (does it make sense to phase them on different chromosomes if there are no translocations?).  It further says: "The recommended convention is to use the
position of the first variant in the set as the PS identifier (although this is not required)."

If you could give me a 👍  on (2) I will merge.

pjedge:
Right, for (1) I meant shouldn't you either print out that '.' in the
individual's column explicitly, or exclude the SE tag in the tag column?
Right now there's just an SE tag in the tag column that corresponds to
nothing at all, not even a '.' value. It's a mismatch in the number of
fields.
(2) seems fine.

nh13:
You will be surprised since values for format fields can be empty!  This is true for the last value in the format fields strangely.

nh13:
@pjedge here's the relevant line in HTSJDK: https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/variantcontext/writer/Options.java#L40

pjedge:
interesting. thanks for informing me

PR #236, opened 2017-05-23
Bug fix: validation of sample metadata

nh13:


src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -149,6 +150,16 @@ object SampleSheet {
   }
 
   protected[illumina] val SplitRegex = ",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)"
+
+  /** Validate that the sample ids are unique, and that the combination of sample name and library id are unique. */
+  private def validateSamples(samples: Seq[Sample]): Unit = {
+    // Validate some properties for a sample sheet
+    require(samples.map(_.sampleId).toSet.size == samples.length,
+      "Sample identifiers were not unique: " + samples.groupBy(_.sampleId).filter(_._2.length > 2).keys.mkString(", "))
+    require(samples.map(s => s"${s.sampleName} ${s.sampleName}").toSet.size == samples.length,
```
tfenne: I don't understand why this is `s.sampleName` twice - I think that's probably a mistake?  Should the second value on this line and on 159 be `s.libraryId`?

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -168,12 +179,11 @@ object SampleSheet {
...
 class SampleSheet(samples: Seq[Sample]) extends Iterable[Sample] {
 
-  // Validate some properties for a sample sheet
-  require(samples.map(_.sampleId).toSet.size == samples.length,
-    "Sample identifiers were not unique: " + samples.groupBy(_.sampleId).filter(_._2.length > 2).keys.mkString(", "))
-  require(samples.map(s => s"${s.sampleId} ${s.libraryId}").toSet.size == samples.length,
-    "Sample name and library identifier combinations were not unique: " +
-      samples.groupBy(s => s"${s.sampleId} ${s.libraryId}").filter(_._2.length > 2).keys.mkString(", "))
+  // Validate samples
+  samples.flatMap(_.lane).toSet[Int].toList.foreach { lane =>
```
tfenne: I'd suggest changing this to `.map(_.lane.getOrElse(1))` so you don't need the extra check on 186.

src/test/scala/com/fulcrumgenomics/illumina/SampleSheetTest.scala
```diff
@@ -84,6 +84,33 @@ class SampleSheetTest extends FlatSpec with Matchers with OptionValues {
...
+
+  it should "not throw an exception if sample ids are unique and if the combination of sample name and library id are unique" in {
+    val sample = Sample(sampleOrdinal=0, sampleId="IDA", sampleName="NAMEA", libraryId="LIBRARY")
+    new SampleSheet(Seq(sample, sample.copy(sampleId="IDB", sampleName="NAMEB"))).size shouldBe 2
+
+    // within lane
+    val sampleForLane = Sample(sampleOrdinal=0, sampleId="IDA", sampleName="NAMEA", libraryId="LIBRARY", lane=Some(1))
+    new SampleSheet(Seq(sampleForLane, sampleForLane.copy(sampleId="IDB", sampleName="NAMEB"))).size shouldBe 2
+  }
+
```
tfenne: It would be nice to see one test with ~2 samples that have different id/name/library values, but are then specified on multiple lanes (i.e. are replicated on lanes 1 and 2) and that this doesn't throw an exception.

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
nh13: It wont work since the samples themselves will still have `None` for `lane`

nh13:
@tfenne one last look or am I am good to merge?

PR #247, opened 2017-06-14
Fix for GroupReadsByUmi with strategy=paired with same/similar UMIs on each read.

tfenne:
@nh13 Can you take a look at this please?  Given the way everything is structured in `GroupReadsByUmi` I was left with two choices:

1. Do some pretty large scale refactoring to have the `UmiAssigner` trait and all it's sub-classes take in something more than just the UMI (e.g. a small case class with UMI and position information for R1 and R2).  This would be a pretty substantive change, would make testing much more awkward and would make single-UMI processing less efficient (since those case classes would have to be made and passed around).
2. For `paired` embed some extra information into the "UMI" to ensure we don't match similar UMIs from different ends of the molecules.

I chose to go with (2) because it's a much more limited change and therefore easier to introduce and test.  I don't think it's ideal, but it does work.

Originally I wanted to modify the UMIs as per the following example:

```
AAA-TTT R1=100(+)  R2=300(-) => 100P:AAA-300M:TTT
TTT-AAA R1=300(-)  R2=100(+) => 300M:TTT-100P:AAA
```

But was concerned that this may occasionally introduce different length prefixes onto the UMIs, which would cause functions that count mismatches and expect two strings of equal length to fail.  I could generate said strings (e.g. `100P`) and then pad them out to the same length, but instead chose to just prefix a fixed string (of length `edits+1`) instead, e.g.:

```
AAA-TTT R1=100(+)  R2=300(-) => aa:AAA-bb:TTT
TTT-AAA R1=300(-)  R2=100(+) => bb:TTT-aa:AAA
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -231,11 +231,11 @@ object GroupReadsByUmi {
     * @param maxMismatches the maximum number of mismatches between UMIs
     */
   class PairedUmiAssigner(maxMismatches: Int) extends AdjacencyUmiAssigner(maxMismatches) {
-    /** String that is prefixed onto the UMI from the genomically earlier/lower read. */
-    val lowerReadUmiPrefix: String = ("a" * (maxMismatches+1)) + ":"
+    /** String that is prefixed onto the UMI from the read with that maps to a lower coordinate in the genome.. */
+    private[umi] val lowerReadUmiPrefix: String = ("a" * (maxMismatches+1)) + ":"
 
-    /** String that is prefixed onto the UMI from the genomically later/higher read. */
-    val higherReadUmiPrefix: String = ("b" * (maxMismatches+1)) + ":"
+    /** String that is prefixed onto the UMI from the read with that maps to a higher coordinate in the genome.. */
+    private[umi] val higherReadUmiPrefix: String = ("b" * (maxMismatches+1)) + ":"
 
     /** Takes a UMI of the form "A-B" and returns "B-A". */
     def reverse(umi: Umi): Umi = umi.indexOf('-') match {
@@ -455,11 +455,10 @@ class GroupReadsByUmi
     * sub-grouping into UMI groups by original molecule.
     */
   def assignUmiGroups(pairs: Seq[ReadPair]): Unit = {
-    val umis = pairs.map { case (r1, r2) => umiForRead(r1, r2) }
+    val umis    = pairs.map { case (r1, r2) => umiForRead(r1, r2) }
     val rawToId = this.assigner.assign(umis)
 
-    pairs.foreach { case (r1, r2) =>
-      val umi = umiForRead(r1, r2)
+    pairs.iterator.zip(umis.iterator).foreach { case ((r1, r2), umi) =>
       val id  = rawToId(umi)
       r1(this.assignTag) = id
       r2(this.assignTag) = id
@@ -474,9 +473,9 @@ class GroupReadsByUmi
     * or later read on the genome.  This is necessary to ensure that, when the two paired UMIs are the same
     * or highly similar, that the A vs. B groups are constructed correctly.
     */
-  private def umiForRead(r1: SamRecord, r2: SamRecord): String = r1.get[String](this.rawTag) match {
+  private def umiForRead(r1: SamRecord, r2: SamRecord): Umi = r1.get[String](this.rawTag) match {
     case None | Some("") => fail(s"Record '$r1' was missing the raw UMI tag '${this.rawTag}'")
-    case Some(chs) =>
+    case Some(chs)       =>
       val umi = chs.toUpperCase
 
       this.assigner match {
@@ -484,12 +483,12 @@ class GroupReadsByUmi
           require(r1.refIndex == r2.refIndex, s"Mates on different references not supported: ${r1.name}")
           val pos1 = if (r1.positiveStrand) r1.unclippedStart else r1.unclippedEnd
           val pos2 = if (r2.positiveStrand) r2.unclippedStart else r2.unclippedEnd
-
+          val r1Lower = pos1 < pos2 || pos1 == pos2 && r1.positiveStrand
           val umis = umi.split('-')
           require(umis.length == 2, s"Paired strategy used but umi did not contain 2 segments: $umi")
 
-          if (pos1 < pos2) s"${paired.lowerReadUmiPrefix}:${umis(0)}-${paired.higherReadUmiPrefix}:${umis(1)}"
-          else             s"${paired.higherReadUmiPrefix}:${umis(0)}-${paired.lowerReadUmiPrefix}:${umis(1)}"
+          if (r1Lower) paired.lowerReadUmiPrefix  + ":" + umis(0) + "-" + paired.higherReadUmiPrefix + ":" + umis(1)
+          else         paired.higherReadUmiPrefix + ":" + umis(0) + "-" + paired.lowerReadUmiPrefix  + ":" + umis(1)
         case _ =>
           umi
       }
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -39,6 +39,7 @@ import com.fulcrumgenomics.util.Sequences.countMismatches
 import com.fulcrumgenomics.util._
 import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
 import htsjdk.samtools._
+import htsjdk.samtools.util.StringUtil
 
 import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -446,18 +455,43 @@ class GroupReadsByUmi
...
-    case Some(umi) => umi.toUpperCase
+  /**
+    * Retrieves the UMI to use for a read pair.  In the case of single-umi strategies this is just
+    * the upper-case UMI sequence.
+    *
+    * For Paired the UMI is extended to encode which "side" of the template the read came from - the earlier
+    * or later read on the genome.  This is necessary to ensure that, when the two paired UMIs are the same
+    * or highly similar, that the A vs. B groups are constructed correctly.
+    */
+  private def umiForRead(r1: SamRecord, r2: SamRecord): String = r1.get[String](this.rawTag) match {
```
nh13: Return type should be `Umi`

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -230,6 +231,12 @@ object GroupReadsByUmi {
     * @param maxMismatches the maximum number of mismatches between UMIs
     */
   class PairedUmiAssigner(maxMismatches: Int) extends AdjacencyUmiAssigner(maxMismatches) {
+    /** String that is prefixed onto the UMI from the genomically earlier/lower read. */
+    val lowerReadUmiPrefix: String = ("a" * (maxMismatches+1)) + ":"
```
nh13: I think these can be private

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -230,6 +231,12 @@ object GroupReadsByUmi {
...
+
+    /** String that is prefixed onto the UMI from the genomically later/higher read. */
```
nh13: genomically?  How about "the read with larger genomic position/coordinate"

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -396,13 +403,15 @@ class GroupReadsByUmi
...
-    val iterator = sorter.iterator.grouped(2).map(i => (i(0), i(1))).buffered // consume in pairs
+    val iterator = sorter.iterator.grouped(2).map { case Seq(x,y) => if (x.firstOfPair) (x, y) else (y, x) }.buffered // consume in pairs
     val tagFamilySizeCounter = new NumericCounter[Int]()
 
     while (iterator.hasNext) {
       // Take the next set of pairs by position and assign UMIs
       val pairs = takeNextGroup(iterator)
-      pairs.foreach { case(r1, r2) => assert(r1.name == r2.name, s"Reads out of order @ ${r1.id} + ${r2.id}") }
+      pairs.foreach { case(r1, r2) =>
+        assert(r1.name == r2.name && r1.firstOfPair && r2.secondOfPair, s"Reads out of order @ ${r1.id} + ${r2.id}")
```
nh13: Can you add a test to hit this assertion?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -446,18 +455,43 @@ class GroupReadsByUmi
...
   def assignUmiGroups(pairs: Seq[ReadPair]): Unit = {
-    val rawUmis = pairs.map(_._1).map(getRawUmi)
-    val rawToId = this.assigner.assign(rawUmis)
-
-    pairs.foreach(pair => Seq(pair._1, pair._2).foreach(rec => {
-      val raw = getRawUmi(rec)
-      val id  = rawToId(raw)
-      rec(this.assignTag) = id
-    }))
+    val umis = pairs.map { case (r1, r2) => umiForRead(r1, r2) }
```
nh13: align equals since it was aligned before.

nh13: Also `umiForRead` is called twice, here (458) and below (462).  It doesn't look to expensive, but you could just return a tuple of `(r1, r2, umi)` or `((r1, r2), umi)` and re-use that for 459 and 461.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -446,18 +455,43 @@ class GroupReadsByUmi
...
+    *
+    * For Paired the UMI is extended to encode which "side" of the template the read came from - the earlier
+    * or later read on the genome.  This is necessary to ensure that, when the two paired UMIs are the same
+    * or highly similar, that the A vs. B groups are constructed correctly.
+    */
+  private def umiForRead(r1: SamRecord, r2: SamRecord): String = r1.get[String](this.rawTag) match {
+    case None | Some("") => fail(s"Record '$r1' was missing the raw UMI tag '${this.rawTag}'")
+    case Some(chs) =>
+      val umi = chs.toUpperCase
+
```
nh13: align `=>`

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -446,18 +455,43 @@ class GroupReadsByUmi
...
+          val pos1 = if (r1.positiveStrand) r1.unclippedStart else r1.unclippedEnd
+          val pos2 = if (r2.positiveStrand) r2.unclippedStart else r2.unclippedEnd
+
+          val umis = umi.split('-')
+          require(umis.length == 2, s"Paired strategy used but umi did not contain 2 segments: $umi")
+
+          if (pos1 < pos2) s"${paired.lowerReadUmiPrefix}:${umis(0)}-${paired.higherReadUmiPrefix}:${umis(1)}"
+          else             s"${paired.higherReadUmiPrefix}:${umis(0)}-${paired.lowerReadUmiPrefix}:${umis(1)}"
+        case _ =>
+          umi
```
nh13: `umi.toUpperCase`?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -446,18 +455,43 @@ class GroupReadsByUmi
...
+    * or highly similar, that the A vs. B groups are constructed correctly.
+    */
+  private def umiForRead(r1: SamRecord, r2: SamRecord): String = r1.get[String](this.rawTag) match {
+    case None | Some("") => fail(s"Record '$r1' was missing the raw UMI tag '${this.rawTag}'")
+    case Some(chs) =>
+      val umi = chs.toUpperCase
+
+      this.assigner match {
+        case paired: PairedUmiAssigner =>
+          require(r1.refIndex == r2.refIndex, s"Mates on different references not supported: ${r1.name}")
```
nh13: Hmmn, this could bite us later too.  Since we only care about which is earlier and which is later, why can't you just use the sequence dictionary to decide?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -446,18 +455,43 @@ class GroupReadsByUmi
...
+          val pos1 = if (r1.positiveStrand) r1.unclippedStart else r1.unclippedEnd
+          val pos2 = if (r2.positiveStrand) r2.unclippedStart else r2.unclippedEnd
```
nh13: See my comment about calling this method twice, since I think getting the unclipped positions isn't the fastest (cigar extraction or alignment blocks?)

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: Cal also call these left/right so they have the same variable name length ;)

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -446,18 +455,43 @@ class GroupReadsByUmi
...
+
+          val umis = umi.split('-')
+          require(umis.length == 2, s"Paired strategy used but umi did not contain 2 segments: $umi")
+
+          if (pos1 < pos2) s"${paired.lowerReadUmiPrefix}:${umis(0)}-${paired.higherReadUmiPrefix}:${umis(1)}"
+          else             s"${paired.higherReadUmiPrefix}:${umis(0)}-${paired.lowerReadUmiPrefix}:${umis(1)}"
```
nh13: Since these are all strings, string concatenation will be faster: https://medium.com/@dkomanov/scala-string-interpolation-performance-21dc85e83afd

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: `umi` is already upper-cased a few lines above.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I actually can't.  It's there to catch any potential bug/regression, but between the filtering and sorting that's done and then the grouping on line 406 this assertion _should_ never fail.

PR #253, opened 2017-06-19
Sample Sheet: fix a case where the error message may not be displayed…

nh13:
… correctly.


nh13:
@tfenne a quick fix.

nh13:
@tfenne one more thought, we then check that the combination of `Sample_Name` and `Library_ID` are unique (within a lane if specified).  If the `Library_ID` is not given, we default it to the `Sample_Name`, and so this then implies that `Sample_Name` must be unique.  What if we instead defaulted `Library_ID` to `Sample_ID`, which _should_ be unique across samples?

nh13:
Starting 0.2.0 release notes here: https://github.com/fulcrumgenomics/fgbio/releases

I looked at `fgbio` and other repositories, and found only a single instance that changes, which worries me a bit, since should it not have broken more things?

PR #254, opened 2017-06-19
Metric: adding apply methods and making it iterable.

nh13:


src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -176,6 +178,22 @@ trait Metric extends Product {
...
+  /** Gets the value of the field by name. */
+  def apply(name: String): Option[String] = {
+    this.names.indexOf(name) match {
+      case -1   => None
+      case idx  => Some(this.values(idx))
+    }
+  }
+
+  /** Gets the value for the ith field. */
+  def apply(i: Int): Option[String] = {
```
tfenne: Could this be private please?  I think it's fine to do internally, but I don't think it should be part of the public API as the _only_ safe and reliable way to get the index is via scanning the names, and this might encourage the idea that the indices are durable and safe to use.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -176,6 +178,22 @@ trait Metric extends Product {
   /** Get the values of the arguments in the order they were defined. */
   def values: Seq[String] = productIterator.map(formatValues).toSeq
 
+  /** Gets the value of the field by name. */
+  def apply(name: String): Option[String] = {
```
tfenne: If possible I'd prefer to keep the semantics of `Map` here; i.e. `apply` returns `String` (in this case) and throws an exception for an invalid key, and `get` returns `Option[String]`.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -176,6 +178,22 @@ trait Metric extends Product {
...
+    this.names.indexOf(name) match {
+      case -1   => None
+      case idx  => Some(this.values(idx))
+    }
+  }
+
+  /** Gets the value for the ith field. */
+  def apply(i: Int): Option[String] = {
+    if (i < 0 || this.length <= i) None else Some(this.values(i))
```
tfenne: Are you ok with `this.value(i)` being very inefficient?  It looks like for each access it will a) reflectively access all fields, b) format all fields to Strings, c) create a Seq with all those values, d) return the `i`th value from the Seq.  For accessing small collections of metrics that's probably not a big deal, but if you're iterating through thousands of rows with lots of fields each, that might suck.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: I am going to remove it until needed.

Issue #257, opened 2017-06-22
AB-R1s, BA-R2s on same strand check in CallDuplexConsensusReads


chapmanb:
Tim and Nils;
After your help working through the memory issues I'm running into an apparent input data related issue on trying to test this for duplex input reads. After running through GroupReadsByUmi with `-s paired`, I'm getting a complaint about the orientation of the reads using the latest development version:
```
[2017/06/22 13:39:04 | FgBioMain | Info] CallDuplexConsensusReads failed. Elapsed time: 33.38 minutes.
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed: Not all AB-R1s and BA-R2s were on the same strand for molecule with id: 2520614/A
 at scala.Predef$.require(Predef.scala:277)
 at com.fulcrumgenomics.umi.DuplexConsensusCaller.consensusSamRecordsFromSamRecords(DuplexConsensusCaller.scala:166)
 at com.fulcrumgenomics.umi.UmiConsensusCaller.consensusReadsFromSamRecords(UmiConsensusCaller.scala:244)
 at com.fulcrumgenomics.umi.UmiConsensusCaller.consensusReadsFromSamRecords$(UmiConsensusCaller.scala:242)
 at com.fulcrumgenomics.umi.DuplexConsensusCaller.consensusReadsFromSamRecords(DuplexConsensusCaller.scala:96)
```
I'm not totally sure what this indicates about our data/prep process and if there is anything we can to mitigate or filter problematic reads. Do you have any tips/tricks to diagnose or work around the issue? Thanks much.

nh13:
@chapmanb can you try with the latest (released) fgbio.  I am wondering if it is the same issue as here: https://github.com/fulcrumgenomics/fgbio/issues/246#issuecomment-307903215, namely a bug that `GroupReadsByUmi` incorrectly assigned the reads when the UMIs are similar, which happens much more often with short (ex. 3bp) UMIs.

chapmanb:
Nils;
Thanks for following up on this. The 0.2.0 release does resolve most of the issues but I'm still running into cases failing with the same problem. I was able to put together a minimal example that demonstrates the issue:
```
wget https://s3.amazonaws.com/chapmanb/testcases/fgbio-grouped-ab.bam
fgbio -Xms500m -Xmx1g CallDuplexConsensusReads -i fgbio-grouped-ab.bam -o /dev/null
```
[...]

PR #259, opened 2017-06-26
Filter out possible collisions between independent source molecules

nh13:
Fixes https://github.com/fulcrumgenomics/fgbio/issues/257.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
@@ -28,6 +28,7 @@ import java.lang.Math.min
 
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.bam.api.SamRecord
+import com.fulcrumgenomics.commons.util.LazyLogging
 import com.fulcrumgenomics.umi.DuplexConsensusCaller._
 import com.fulcrumgenomics.umi.UmiConsensusCaller.ReadType.{ReadType, _}
 import com.fulcrumgenomics.umi.UmiConsensusCaller.{SimpleRead, SourceRead}
@@ -48,6 +49,7 @@ object DuplexConsensusCaller {
   val FilterAorBOnly    = "Only Seeing A or B Reads"
   val FilterFragments   = "Being Fragment/Non-Paired Reads"
   val FilterSsConsensus = "Only Generating One Strand Consensus"
+  val FilterCollision   = "Potential collision between independent duplex molecules"
 
   /**
     * Stores information about a consensus read.  Bases, arrays, and the two single
@@ -99,7 +101,7 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
                             val trim: Boolean = false,
                             val errorRatePreUmi: PhredScore     = DuplexConsensusCaller.ErrorRatePreUmi,
                             val errorRatePostUmi: PhredScore    = DuplexConsensusCaller.ErrorRatePostUmi
-                           ) extends UmiConsensusCaller[DuplexConsensusRead] {
+                           ) extends UmiConsensusCaller[DuplexConsensusRead] with LazyLogging {
 
   private val ssCaller = new VanillaUmiConsensusCaller(readNamePrefix="x", options=new VanillaUmiConsensusCallerOptions(
       errorRatePreUmi         = this.errorRatePreUmi,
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -161,15 +162,17 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
         // Check for this explicitly here.
         if (singleStrand1.nonEmpty) {
           val ss1Flag = singleStrand1.head.negativeStrand
-          val ss1MI   = singleStrand1.head.apply[String](ConsensusTags.MolecularId)
-          require(singleStrand1.forall(_.negativeStrand == ss1Flag),
-            s"Not all AB-R1s and BA-R2s were on the same strand for molecule with id: $ss1MI")
+          if (singleStrand1.exists(_.negativeStrand != ss1Flag)) {
+            rejectRecords(ab ++ ba, FilterCollision)
```
tfenne: Do you think it would be reasonable to add a `logger.debug()` here with the `MI` of the offending group?  I think that when this error is seen it might be nice to be able to see the group with collisions and confirm that it's not some other issue.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -161,15 +162,17 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+            return Nil
+          }
         }
         if (singleStrand2.nonEmpty) {
           val ss2Flag = singleStrand2.head.negativeStrand
-          val ss2MI   = singleStrand2.head.apply[String](ConsensusTags.MolecularId)
-          require(singleStrand2.forall(_.negativeStrand == ss2Flag),
-            s"Not all AB-R2s and BA-R1s were on the same strand for molecule with id: $ss2MI")
+          if (singleStrand2.exists(_.negativeStrand != ss2Flag)) {
+            rejectRecords(ab ++ ba, FilterCollision)
```
tfenne: Doing this here and above causes a line to be logged at the end with the count, right?

src/test/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReadsTest.scala
```diff
@@ -58,14 +58,20 @@ class CallDuplexConsensusReadsTest extends UnitSpec {
     checkClpAnnotations[CallDuplexConsensusReads]
   }
 
-  it should "run fail if AB-R1s are not on the same strand ads BA-R2s" in {
+  it should "not generate a consensus if AB-R1s are not on the same strand ads BA-R2s" in {
     val builder = new SamBuilder(readLength=10)
     builder.addPair(name="ab1", start1=100, start2=200, attrs=Map(MI -> "1/A"), bases1="AAAAAAAAAA", bases2="AAAAAAAAAA")
```
tfenne: It would be nice to be explicit about the strands here even though it's mirroring the defaults.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
nh13: It does:
```
[2017/06/26 15:40:25 | FgBioMain | Info] Executing CallDuplexConsensusReads from fgbio version 0.2.1-SNAPSHOT as nhomer@Nils on JRE 1.8.0_66-b17 without snappy
[2017/06/26 15:40:26 | CallDuplexConsensusReads | Info] Total Raw Reads Considered: 10.
[2017/06/26 15:40:26 | CallDuplexConsensusReads | Info] Raw Reads Filtered Due to Potential collision between independent duplex molecules: 10 (1.0000).
[2017/06/26 15:40:26 | CallDuplexConsensusReads | Info] Raw Reads Filtered Due to Being Fragment/Non-Paired Reads: 0 (0.0000).
[2017/06/26 15:40:26 | CallDuplexConsensusReads | Info] Consensus reads emitted: 0.
[2017/06/26 15:40:26 | FgBioMain | Info] CallDuplexConsensusReads completed. Elapsed time: 0.05 minutes.
```

src/test/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReadsTest.scala
nh13: Done

PR #265, opened 2017-07-07
Bug fix: ClipBam alwasy clipped overlapping reads.

nh13:
@tfenne this is a fairly embarrassing bug.  I am wondering if I/we should do a release next week.  If so, I'd like to keep the version #s across commons, sopt, fgbio, and dagr the same for ease, what do you think, or is that too much?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
@@ -72,6 +72,11 @@ class ClipBam
   Io.assertReadable(ref)
   Io.assertCanWriteFile(output)
 
+  if (Seq(readOneFivePrime, readOneThreePrime, readTwoFivePrime, readTwoThreePrime).forall(_ == 0) && !overlappingReads) {
+    validate(overlappingReads || Seq(readOneFivePrime, readOneThreePrime, readTwoFivePrime, readTwoThreePrime).exists(_ != 0),
+      "At least one clipping option is required")
+  }
+
   private val clipper = new SamRecordClipper(mode=if (softClip) ClippingMode.Soft else ClippingMode.Hard, autoClipAttributes=autoClipAttributes)
 
   override def execute(): Unit = {
@@ -123,7 +128,7 @@ class ClipBam
     this.clipper.clip5PrimeEndOfRead(r2, readTwoFivePrime)
     this.clipper.clip3PrimeEndOfRead(r2, readTwoThreePrime)
 
-    if (r1.unmapped || r2.unmapped || r1.refIndex != r2.refIndex || r1.pairOrientation != PairOrientation.FR) {
+    if (!overlappingReads || r1.unmapped || r2.unmapped || r1.refIndex != r2.refIndex || r1.pairOrientation != PairOrientation.FR) {
       // Do nothing
     }
     else {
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -123,7 +128,7 @@ class ClipBam
     this.clipper.clip5PrimeEndOfRead(r2, readTwoFivePrime)
     this.clipper.clip3PrimeEndOfRead(r2, readTwoThreePrime)
 
-    if (r1.unmapped || r2.unmapped || r1.refIndex != r2.refIndex || r1.pairOrientation != PairOrientation.FR) {
+    if (!overlappingReads || r1.unmapped || r2.unmapped || r1.refIndex != r2.refIndex || r1.pairOrientation != PairOrientation.FR) {
```
tfenne: I find this a little confusing.  How about `if (overlappingReads && r1.mapped && r2.mapped $$ r1.refIndex == r2.refIndex && r1.pairOrientation == FR) { ... }`.  I'm assuming there used to be something more than `// do nothing` in the first branch, but now there's not there's no reason to even have it.

Also, the ship has probably sailed, but `overlappingReads` is confusing because I interpreted it initially as "reads are overlapping" not "clip overlapping reads".

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -72,6 +72,11 @@ class ClipBam
   Io.assertReadable(ref)
   Io.assertCanWriteFile(output)
 
+  if (Seq(readOneFivePrime, readOneThreePrime, readTwoFivePrime, readTwoThreePrime).forall(_ == 0) && !overlappingReads) {
```
tfenne: You don't need the if here ... `validate` is like `require`.

nh13:
@tfenne I also refactored out the part of the code that determines _how_ overlapping reads are clipped, since I would like to extend this tool in another project and have a custom method for determining how overlapping reads are clipped.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -123,22 +127,27 @@ class ClipBam
...
-      while (f.end >= r.start && f.mapped && r.mapped) {
-        val lengthToClip = f.end - r.start + 1
-        val firstHalf    = lengthToClip / 2
-        val secondHalf   = lengthToClip - firstHalf // safe guard against rounding on odd lengths
-        this.clipper.clip3PrimeEndOfAlignment(r1, firstHalf)
-        this.clipper.clip3PrimeEndOfAlignment(r2, secondHalf)
-      }
+  /** Clips overlapping reads, where both ends of the read pair are mapped to the same chromosome, and in FR orientation. */
+  protected def clipOverlappingReads(f: SamRecord, r: SamRecord): Unit = {
+
```
tfenne: Bonus newline.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
nh13: I can fix the option issue and keep the old one there but deprecate it.

PR #268, opened 2017-08-09
Added more filters to FilterBam.

tfenne:


src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -51,6 +52,11 @@ import htsjdk.samtools.util.IntervalList
      |marked as secondary alignments, reads marked as duplicates, and if a set of Intervals are provided,
      |reads that do not overlap any of the intervals.
      |
+     |The `--remove-unpaired-reads`, which removes reads that are not part of a mapped pair will remove
```
nh13: Can you change `--remove-unpaired-reads` to `--remove-unmapped-pairs`?  I thought it was named to remove fragment reads (does it)?  Perhaps you want to have an option `--keep-only-mapped-pairs` instead.

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -51,6 +52,11 @@ import htsjdk.samtools.util.IntervalList
...
+     |unmapped reads when set to true even if `--remove-unmapped-reads` is false.  Similarly if
+     |`--min-insert-size` or `--min-mapped-bases` is specified, unmapped reads will also be removed
+     |even if `--remove-unmapped-reads` is false
```
nh13: period

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -62,7 +68,11 @@ class FilterBam
   @arg(flag='D', doc="If true remove all reads that are marked as duplicates.")   val removeDuplicates: Boolean = true,
   @arg(flag='U', doc="Remove all unmapped reads.")                                val removeUnmappedReads: Boolean = true,
   @arg(flag='M', doc="Remove all mapped reads with MAPQ lower than this number.") val minMapQ: Int = 1,
-  @arg(flag='S', doc="Remove all reads marked as secondary alignments.")          val removeSecondaryAlignments: Boolean = true
+  @arg(flag='P', doc="Remove all reads that are not part of a mapped pair.")      val removeUnpairedReads: Boolean = false,
+  @arg(flag='S', doc="Remove all reads marked as secondary alignments.")          val removeSecondaryAlignments: Boolean = true,
+  @arg(          doc="Remove all reads with insert size < this value.")           val minInsertSize: Option[Int] = None,
+  @arg(          doc="Remove all reads with insert size > this value.")           val maxInsertSize: Option[Int] = None,
+  @arg(flag='m', doc="Remove reads with fewer than this many mapped bases.")      val minMappedBases: Option[Int] = None
 ) extends FgBioTool with LazyLogging {
```
nh13: Can you add a warning when `--remove-unmapped-reads=false` but one of the other options overrides it?

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -62,7 +68,11 @@ class FilterBam
   @arg(flag='D', doc="If true remove all reads that are marked as duplicates.")   val removeDuplicates: Boolean = true,
   @arg(flag='U', doc="Remove all unmapped reads.")                                val removeUnmappedReads: Boolean = true,
   @arg(flag='M', doc="Remove all mapped reads with MAPQ lower than this number.") val minMapQ: Int = 1,
-  @arg(flag='S', doc="Remove all reads marked as secondary alignments.")          val removeSecondaryAlignments: Boolean = true
+  @arg(flag='P', doc="Remove all reads that are not part of a mapped pair.")      val removeUnpairedReads: Boolean = false,
+  @arg(flag='S', doc="Remove all reads marked as secondary alignments.")          val removeSecondaryAlignments: Boolean = true,
+  @arg(          doc="Remove all reads with insert size < this value.")           val minInsertSize: Option[Int] = None,
```
nh13: I would prefer inclusive, say `<=` rather than `<`.  Similarly for `maxInsertSize`.  I read it as "all reads with at least insert size of..."

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -77,8 +87,12 @@ class FilterBam
     val kept = iterator.count { rec => {
         val throwOut = (removeDuplicates && rec.duplicate) ||
           (removeUnmappedReads && rec.unmapped) ||
+          (removeUnpairedReads && (!rec.paired || rec.unmapped || rec.mateUnmapped)) ||
```
nh13: I guess it does remove fragment reads, as well as read pairs where both ends don't map.

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -62,7 +68,11 @@ class FilterBam
   @arg(flag='D', doc="If true remove all reads that are marked as duplicates.")   val removeDuplicates: Boolean = true,
   @arg(flag='U', doc="Remove all unmapped reads.")                                val removeUnmappedReads: Boolean = true,
   @arg(flag='M', doc="Remove all mapped reads with MAPQ lower than this number.") val minMapQ: Int = 1,
-  @arg(flag='S', doc="Remove all reads marked as secondary alignments.")          val removeSecondaryAlignments: Boolean = true
+  @arg(flag='P', doc="Remove all reads that are not part of a mapped pair.")      val removeUnpairedReads: Boolean = false,
+  @arg(flag='S', doc="Remove all reads marked as secondary alignments.")          val removeSecondaryAlignments: Boolean = true,
+  @arg(          doc="Remove all reads with insert size < this value.")           val minInsertSize: Option[Int] = None,
+  @arg(          doc="Remove all reads with insert size > this value.")           val maxInsertSize: Option[Int] = None,
+  @arg(flag='m', doc="Remove reads with fewer than this many mapped bases.")      val minMappedBases: Option[Int] = None
```
nh13: Again, I like inclusivity due to the way I read "min" here.

src/test/scala/com/fulcrumgenomics/bam/FilterBamTest.scala
```diff
@@ -107,4 +109,71 @@ class FilterBamTest extends UnitSpec {
...
+  }
+
+  it should "remove reads below the specified insert size" in {
+    val builder = new SamBuilder(readLength=10)
+    builder.addPair("q1", start1=100, start2=589) // isize=499
+    builder.addPair("q2", start1=100, start2=590) // isize=500
+    builder.addPair("q3", start1=100, start2=591) // isize=501
+    builder.addPair("q4", start1=100, start2=120) // isize=30
+    builder.addPair("q5", start1=200, start2=100, strand1=Minus, strand2=Plus) // isize=110
+    builder.addFrag("q6", start=500)
```
nh13: what about read pairs that map to different chromosomes?

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
tfenne: I'm struggling to come up with a good name for this that's not really long and is in the style of other options.  I like `--remove-*` because I think it's clearer that the option removes those things, rather than `--keep-*` because that implies things will be kept, which other options may invalidate.

`--remove-reads-not-part-of-mapped-pairs` seems overly wordy, but clearest.  Perhaps just expanding the documentation for that option is best?

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
tfenne: Are you concerned with the documentation or the actual handling of the options?  In my mind the options are inclusive already.  I.e. if you specify `--min-insert-size=100 --max-insert-size=600` then you will retain all reads with inserts in (100,600) inclusive.  Similarly for `minMappedBases` if you specify `--min-mapped-bases=50` then you'll retain records with >= 50 mapped bases.

It's just that I think the options are documented more accurately as removing things below the minimum or above the maximum, because e.g. if `--min-insert-size` was documented as "retain records with insert size >= this value", it would be confusing when combined with `--max-insert-size`.

tfenne:
@nh13 Updates pushed, want to take another quick look?  If the inclusive/exclusive option stuff still doesn't work for you let's chat.

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
nh13: I'll vote for anything that isn't `--remove-unpaired-reads` since I know in "Broad"-speak that "unpaired" meant a read pair where not both ends are mapped, but I know other folks use unpaired differently (unpaired means single-end reads, or where exactly one end maps).

By the way, not all options are `--remove-*`, see `minMapQ`!

How about two options:
1. `--remove-fragment-reads`
2. `--require-mapped-pairs`
?

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
nh13: @tfenne this didn't change, and there was no comment about why not.

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -79,6 +85,10 @@ class FilterBam
   Io.assertCanWriteFile(output)
   intervals.foreach(Io.assertReadable)
 
+  if (!removeUnmappedReads && (minInsertSize.isDefined || removeUnpairedReads))
+    logger.warning("--remove-unmapped-reads was set to false, but unmapped reads will be removed ",
+      "due to setting either --min-insert-size or --remove-unpaired-reads.")
```
nh13: Brackets please

nh13:
Lets chat about how we want the options here, and talk about if minimum is inclusive or exclusive.

PR #269, opened 2017-08-09
Adding a tool CollectErccMetrics.

nh13:
This will collect metrics for analyzing ERCC spike-ins in
RNA-Seq experiments for dose response but not fold-change
response.  The former requires a single mixture to be spiked
in the sample, while the latter would require a two mixtures
in separate samples.


src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+      .filterNot { r => progress.record(r); r.secondary || r.supplementary } // no secondary or supplementary
+      .filter { r =>
+        totalReads += 1
+        // consider only records mapped to an ERCC transcript
+        r.mapped && r.mapq >= minimumMappingQuality && erccData.contains(r.refName)
+      }
+      .filter { r =>
+        erccReads += 1
+        // consider only paired records where both ends map to the sam ERCC transcript
+        r.paired && r.mapped && r.mateMapped && r.refIndex == r.mateRefIndex && r.firstOfPair
```
tfenne: Is there a good reason to only allow paired data?  I don't know enough to know whether there are folks still doing single end RNASeq.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+      }
+      .foreach { r =>
+        erccFragments += 1
+        erccFragmentCounter.count(r.refName)
+      }
+    in.safelyClose()
+
+    // Transform to log2 counts to compute some simple correlation and linear regression metrics.
+    val counts         = ListBuffer[Double]()
```
tfenne: I think `ArrayBuffer` is generally more efficient than `ListBuffer`.  Sort of like `ArrayList` vs. `LinkedList`.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+    val concentrations = ListBuffer[Double]()
+    erccData.iterator.foreach { case (name, concentration) =>
+      val count = erccFragmentCounter.countOf(name)
+      if (minTranscriptCount <= count) {
```
tfenne: I think there's a missing step prior to this point.  The ERCC sequences have a pretty large range of lengths, from 273bp up to 2022bp.  Using raw counts is going to skew things given that we're comparing to molar concentration.  We should be length-normalizing the read counts prior to any kind of correlations or stats.

I think too, I would report both the count and the length-normalized count in the detail metrics.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    // Rocket science
+    val simpleRegression = new SimpleRegression()
+    counts.zip(concentrations).foreach { case (cnt, conc) => simpleRegression.addData(cnt, conc) }
+
+    val erccMetrics = ErccSummaryMetrics(
+      total_reads                = totalReads,
+      ercc_reads                 = erccReads,
+      pct_ercc_reads             = if (totalReads == 0) 0 else 100.0 * erccReads / totalReads.toDouble,
```
tfenne: In all other metrics classes to date we've used values between 0 and 1 and called them `fraction_...`.  Could we stick to that please?

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+      ercc_fragments             = erccFragments,
```
tfenne: `ercc_templates`?  or `ercc_read_pairs`?  Fragments make me think of single-end mappings.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+      total_transcripts          = erccFragmentCounter.iterator.length,
```
tfenne: Maybe we should add a `SimpleCounter.size` that reports the number of counted things?

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+
+      val summaryPath  = f(".ercc_summary_metrics.txt")
+      val detailedPath = f(".ercc_detailed_metrics.txt")
+      val plotPath     = f(".ercc_plot.pdf")
+
+      // summary metrics file
+      Metric.write[ErccSummaryMetrics](summaryPath, erccMetrics)
+
+      // detailed (per-ERCC-transcript) file
+      val metrics = in.header.getSequenceDictionary
```
tfenne: Why use the dict and filter it vs. just iterating over the `errcData`?

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+          ErccDetailedMetric(name=name, concentration=concentration, count=count)
+        }.toSeq
+      Metric.write[ErccDetailedMetric](detailedPath, metrics)
+
+      // And try plotting
+      if (metrics.isEmpty || totalReads == 0 || simpleRegression.getN <= 1) {
+        logger.warning("No metrics generated (is your BAM empty?). Plots will not be generated.")
+      }
+      else {
+        val name = PathUtil.basename(input, trimExt=true).toString
```
tfenne: Or maybe re-use `output` here?  I think elsewhere I've tended to pull the sample or library name from the header.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  /** The natural log of 2! */
+  private val lnOf2 = scala.math.log(2) // natural log of 2
+
+  /** Computes the log base 2. */
+  private def log2(x: Double): Double = scala.math.log(x) / lnOf2
+
+  /** Use this method to wrap the calculation of correlations such that if insufficient data are given, zero is returned. */
+  private def zeroIfNaNOrInsufficientData(f: => Double): Double = {
```
tfenne: You could make the metric values be `Option[Double]` and return `None` here if correlation can't be generated.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+  }
+}
+
+/**
+  * Metrics produced by `CollectErccMetrics` describing various per-transcript metrics related to the spike-in of ERCC
+  * (External RNA Controls Consortium) into an RNA-Seq experiment.  One metric per ERCC transcript will be present.
+  *
+  * @param name the name (or ID) of the ERCC transcript.
+  * @param concentration the expected concentration as input to `CollectErccMetrics`.
+  * @param count the observed count of the # of fragments.
```
tfenne: `#` -> number; fragments -> read pairs?

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+  * The correlation coefficients and linear regression are calculated based on the log2 observed fragment count versus
+  * the log2 expected concentration.
+  *
+  * @param total_reads the total # of reads considered.
+  * @param ercc_reads the total # of reads mapping to an ERCC transcript.
+  * @param pct_ercc_reads the fraction of total reads that map to an ERCC transcript.
+  * @param ercc_fragments the total # of read pairs mapping to an ERCC transcript.
+  * @param total_transcripts the total # of ERCC transcripts with at least one read observed.
+  * @param total_filtered_transcripts the total # of ERCC transcripts with at least the user-set minimum # of reads observed.
+  * @param pearsons_correlation the Pearson's correlation coefficient.
```
tfenne: I may be wrong, but it's weird to me to have "the" before "Pearson's correlation coefficient" and likewise for Spearman's.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+  * @param spearmans_correlation the Spearman's correlation coefficient.
+  * @param slope the slope of the linear regression.
+  * @param r_square the r-square of the linear regression.
+  * @param intercept the intercept of the linear regression.
```
tfenne: Could this come prior to slope please?

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
```diff
@@ -0,0 +1,253 @@
...
+  * @param total_reads the total # of reads considered.
+  * @param ercc_reads the total # of reads mapping to an ERCC transcript.
+  * @param pct_ercc_reads the fraction of total reads that map to an ERCC transcript.
+  * @param ercc_fragments the total # of read pairs mapping to an ERCC transcript.
+  * @param total_transcripts the total # of ERCC transcripts with at least one read observed.
+  * @param total_filtered_transcripts the total # of ERCC transcripts with at least the user-set minimum # of reads observed.
+  * @param pearsons_correlation the Pearson's correlation coefficient.
+  * @param spearmans_correlation the Spearman's correlation coefficient.
+  * @param slope the slope of the linear regression.
+  * @param r_square the r-square of the linear regression.
```
tfenne: r-square**d**

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
nh13: I wanted to be conservative in my approach, and only support paired end data.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
nh13: Picard is sneaking back in.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
nh13: I want to return them in the order found in the Sam header.

src/main/scala/com/fulcrumgenomics/bam/CollectErccMetrics.scala
nh13: I'll standardize this, `ErrorRateByReadPosition`, and `CollectDuplexSeqMetrics`

nh13:
@tfenne no real objections to any of the suggestions or comments.  Would you care to take one more look?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -110,18 +111,18 @@ class ErrorRateByReadPosition
...
       }
     }
   }
 
   /** Computes the metrics from the BAM file. */
-  private[bam] def computeMetrics: Seq[ErrorRateByReadPositionMetric] = {
+  private[bam] def computeMetrics(source: Option[SamSource] = None): Seq[ErrorRateByReadPositionMetric] = {
     val progress = ProgressLogger(logger, verb="Processed", noun="loci", unit=50000)
-    val in       = SamSource(input, ref=Some(ref))
+    val in       = source.getOrElse(SamSource(input, ref=Some(ref)))
```
tfenne: This `source = None` argument confused me for a second.  Why not just have the argument be of type `SamSource` with a default value of `SamSource(input, ref=Some(ref))`?

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+    |BAM should have sequence lines in the header matching the ERCC ids (ex. ERCC-00130 or ERCC-00004).
+    |
+    |The standard ERCC transcripts, including their unique IDs and concentrations, are taken from
+    |[here](https://assets.thermofisher.com/TFS-Assets/LSG/manuals/cms_095046.txt).  The second column lists the ERCC
+    |transcript identifier which should be present as a sequence line in the BAM header, while columns four and five
+    |give the concentration of the transcript for mixtures #1 and #2.
+    |
+    |The choice of mixture to use can be specified with the `--mixture-name` option (either `Mix1` or `Mix2`), or with
+    |a file containing a custom list of transcripts and concentrations using the `--custom-mixture` option as follows.
+    |The custom mixture file should be tab-delimited file containing the following columns:
```
tfenne: With or without headers?

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+ @arg(flag='c', doc="Minimum # of counts required to include an ERCC transcript.") val minTranscriptCount: Int = 3,
+ @arg(flag='M', doc="The minimum mapping quality") val minimumMappingQuality: Int = 10
+) extends FgBioTool with LazyLogging {
+
+  Io.assertReadable(input)
+  customMixture.foreach(Io.assertReadable)
+  Io.assertCanWriteFile(output)
+
+  if (mixtureName.isEmpty && customMixture.isEmpty) {
+    throw new ValidationException("Either --mixture-name or --mixture-metadata")
```
tfenne: ... `must be provided`?

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+
+  private val ScriptPath = "com/fulcrumgenomics/rnaseq/CollectErccMetrics.R"
+
+  override def execute(): Unit = {
+    val in       = SamSource(input)
+    val progress = ProgressLogger(logger, unit=2e6.toInt)
+    val erccData = customMixture match {
+      case Some(path) => DelimitedDataParser(path, '\t').map { row => row[String](0) -> row[Double](1) }.toMap
+      case None       =>
+        val stream    = getClass.getResourceAsStream(s"${CollectErccMetrics.StandardMixtureMetadataPath}")
```
tfenne: I don't think you need the `s"${}"` right?

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+    throw new ValidationException("Either --mixture-name or --mixture-metadata")
+  }
+
+  private val ScriptPath = "com/fulcrumgenomics/rnaseq/CollectErccMetrics.R"
+
+  override def execute(): Unit = {
+    val in       = SamSource(input)
+    val progress = ProgressLogger(logger, unit=2e6.toInt)
+    val erccData = customMixture match {
+      case Some(path) => DelimitedDataParser(path, '\t').map { row => row[String](0) -> row[Double](1) }.toMap
```
tfenne: Just realized that if the file does _not_ contain headers that this will eat the first data row because `DelimitedDataParser` always expects a header row.

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    // check for no ERCC data
+    if (erccData.isEmpty) {
+      val description = customMixture.map("in '" + _).getOrElse(mixtureName.map("for '" + _).get) + "'"
+      throw new IllegalArgumentException(s"No ERCC data given " + description)
+    }
+
+    // require that the SAM/BAM header contains all the ERCC transcripts
+
```
tfenne: Bingo!

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    // Rocket science
+    val simpleRegression = new SimpleRegression()
+    normalizedCounts.zip(concentrations).foreach { case (cnt, conc) => simpleRegression.addData(cnt, conc) }
+
+    val erccMetrics = ErccSummaryMetrics(
+      total_reads                = totalReads,
+      ercc_reads                 = erccReads,
+      fraction_ercc_reads        = if (totalReads == 0) 0 else 100.0 * erccReads / totalReads.toDouble,
```
tfenne: Should presumably drop the `100.0 *` here.

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+      ercc_templates             = erccReadPairs,
+      total_transcripts          = erccReadPairCounter.iterator.length,
+      total_filtered_transcripts = erccReadPairCounter.iterator.count(_._2 >= minTranscriptCount),
+      pearsons_correlation       = noneIfNaNOrInsufficientData { new PearsonsCorrelation().correlation(normalizedCounts.toArray, concentrations.toArray) },
+      spearmans_correlation      = noneIfNaNOrInsufficientData { new SpearmansCorrelation().correlation(normalizedCounts.toArray, concentrations.toArray) },
+      intercept                  = noneIfNaN { simpleRegression.getIntercept },
+      slope                      = noneIfNaN { simpleRegression.getSlope },
+      r_square                   = noneIfNaN { simpleRegression.getRSquare }
```
tfenne: r_square -> r_squared

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+    val simpleRegression = new SimpleRegression()
+    normalizedCounts.zip(concentrations).foreach { case (cnt, conc) => simpleRegression.addData(cnt, conc) }
+
+    val erccMetrics = ErccSummaryMetrics(
+      total_reads                = totalReads,
+      ercc_reads                 = erccReads,
+      fraction_ercc_reads        = if (totalReads == 0) 0 else 100.0 * erccReads / totalReads.toDouble,
+      ercc_templates             = erccReadPairs,
+      total_transcripts          = erccReadPairCounter.iterator.length,
+      total_filtered_transcripts = erccReadPairCounter.iterator.count(_._2 >= minTranscriptCount),
```
tfenne: `passing_filter_transcripts`?  I struggle with `filtered` because it's never clear to me whether it means things that were filtered _out_, or things kept after filtering.

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+      Metric.write[ErccSummaryMetrics](summaryPath, erccMetrics)
+
+      // detailed (per-ERCC-transcript) file, written in the order defined in the Sam header
+      val metrics = in.header.getSequenceDictionary
+        .getSequences
+        .map(_.getSequenceName)
+        .filter { name => erccData.contains(name) }
+        .map { name =>
+          val count           = erccReadPairCounter.countOf(name)
+          val normalizedCount = count / in.header.getSequence(name).getSequenceLength.toDouble
```
tfenne: I think it's a little odd that you normalize in two different ways.  Above it's by `maxTxLen / len` and here's it just be `len`.  They're equivalent since `maxTxLen` is a constant, but it seems strange to do it differently in two places.

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -0,0 +1,294 @@
...
+  * The correlation coefficients and linear regression are calculated based on the log2 observed read pair count normalized
+  * by ERCC transcript length versus the log2 expected concentration.
+  *
+  * @param total_reads the total number of reads considered.
+  * @param ercc_reads the total number of reads mapping to an ERCC transcript.
+  * @param fraction_ercc_reads the fraction of total reads that map to an ERCC transcript.
+  * @param ercc_templates the total number of read pairs mapping to an ERCC transcript.
+  * @param total_transcripts the total number of ERCC transcripts with at least one read observed.
+  * @param total_filtered_transcripts the total number of ERCC transcripts with at least the user-set minimum # of reads observed.
+  * @param pearsons_correlation Pearson's correlation coefficient.
```
tfenne: ... for correlation of concentration and normalized counts.

src/main/scala/com/fulcrumgenomics/rnaseq/EstimateRnaSeqInsertSize.scala
```diff
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2017 Fulcrum Genomics
+ * Copyright (c) $year Fulcrum Genomics
```
tfenne: Did IntelliJ do this to you?  It should be kept as the year the file was first created.

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
nh13: Yeah, `DelimitedDataParser` requires a header in the file (which is silly) but I am not going to address that here.

src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
nh13: Silly `DelimitedDataParser`.

Issue #274, opened 2017-08-16
SamRecord.apply should thrown an exception if the tag was not found


nh13:
@tfenne I _think_ it will do the wrong thing currently.

```
 @inline final def apply[A](name: String): A              = getAttribute(name).asInstanceOf[A]
```
if `getAttribute` returns `null`, and `A` is `Int`, then, we have `null.asInstanceOf[Int]`, which in my REPL returns `0`!

tfenne:
Well that's horrifying.  I would have expected that `null.asInstanceOf[Int]` to either return `null` itself, or fail with an NPE.  WTF is it doing returning the default value for `Int`?

PR #276, opened 2017-08-16
Added a test for SamRecord.apply[A].

tfenne:


tfenne:
@nh13 This is a test for your issue in #274.  I see what you see when typing `null.asInstanceOf[Int]` in the repl, but I think it's the repl playing games, because in the test it works as expected.  And if you breakpoint in the test and check the value of calling `rec[Int]("bx")` the return value is null.

One could argue that it should throw an exception instead of returning null, but it's at least not returning an erroneous value.

nh13:
👍  works for me, silly repl

PR #278, opened 2017-08-31
ClipBam does not validate when using the --clip-overlapping-reads-option

nh13:


nh13:
@tfenne merging once tests pass, as this is a simple, yet urgent bug fix.

PR #279, opened 2017-09-05
Tune the amount of memory used by DemuxFastqs.

nh13:
Too many reads were stored in RAM when sorting reads with a large # of output samples.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -53,8 +53,8 @@ object DemuxFastqs {
   /** The name of the sample for unmatched reads. */
   val UnmatchedSampleId: String = "unmatched"
 
-  /** The maximum # of records in RAM per writer. */
-  private[fastq] val MaxRecordsInRamPerSamFileWriter: Int = 1e6.toInt
+  /** The maximum # of records in RAM per SAM/BAM writer. */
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -410,10 +410,12 @@ private trait DemuxWriter extends Closeable {
...
 private class SamRecordWriter(output: PathToBam,
                               val header: SAMFileHeader,
-                              val umiTag: String) extends DemuxWriter {
+                              val umiTag: String,
+                              val numSamples: Int) extends DemuxWriter {
   val order = if (header.getSortOrder == SortOrder.unsorted) None else SamOrder(header)
   private val writer = SamWriter(output, header, sort=order,
-    async = DemuxFastqs.UseAsyncIo, maxRecordsInRam = DemuxFastqs.MaxRecordsInRamPerSamFileWriter)
+    async = DemuxFastqs.UseAsyncIo,
+    maxRecordsInRam = Math.max(10000,  DemuxFastqs.MaxRecordsInRamPerSamFileWriter / numSamples))
```
tfenne: I think this constant should get renamed to just `MaxRecordsInRam` or `MaxRecordsInRamWhenSortingSams`, since it's now treated as a total and divided by the number of samples.

nh13:
@tfenne also fixing up a travis-ci build issue, as I believe travis-ci moved from precise to trusty in their builds, so we need to update R.

tfenne:
👍

PR #280, opened 2017-09-08
Output the consensus UMI bases when calling consensus reads.

nh13:
For vanilla consensus reads, this is just the consensus sequence across all input reads.  The consensus is generated per-end for a read-pair.

For duplex consensus reads, we group the reads into two sets (AB R1s with BA R2s, and BA R1s with AB R2s), and call the consensus UMI on each set respectively.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
@@ -197,13 +197,18 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
         val baR2Consensus = ssCaller.consensusCall(filteredBaR2s)
 
         // Call the duplex reads
-        val duplexR1 = duplexConsensus(abR1Consensus, baR2Consensus, filteredAbR1s ++ filteredBaR2s)
-        val duplexR2 = duplexConsensus(abR2Consensus, baR1Consensus, filteredAbR2s ++ filteredBaR1s)
+        val duplexR1Sources = filteredAbR1s ++ filteredBaR2s
+        val duplexR2Sources = filteredAbR2s ++ filteredBaR1s
+        val duplexR1 = duplexConsensus(abR1Consensus, baR2Consensus, duplexR1Sources)
+        val duplexR2 = duplexConsensus(abR2Consensus, baR1Consensus, duplexR2Sources)
 
         // Convert to SamRecords and return
         (duplexR1, duplexR2) match {
           case (Some(r1), Some(r2)) =>
-            Seq(createSamRecord(r1, FirstOfPair), createSamRecord(r2, SecondOfPair))
+            Seq(
+              createSamRecord(r1, FirstOfPair, toUmiBases(duplexR1Sources)),
+              createSamRecord(r2, SecondOfPair, toUmiBases(duplexR2Sources))
+            )
           case _                    =>
             // NB: some reads may have been rejected already in filterToMostCommonAlignment, so just
             //     reject those records that survived the initial filtering.
@@ -214,6 +219,14 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
       case _ =>
         unreachable("SamRecords supplied with more than two distinct MI values.")
     }
+  }
+
+  /** Extracts the UMI bases for each source read, ignoring reads without UMI bases */
+  private def toUmiBases(reads: Seq[SourceRead]): Seq[String] = {
+    // The source read may not have the RX tag, and if so, we ignore them through
+    // the use of .flatMap and .get, with the latter returning an `Some(bases)` if the
+    // UMI bases are present, `None` otherwise.
+    reads.flatMap(_.sam).flatMap(_.get[String](ConsensusTags.UmiBases))
   }
 
   /**
@@ -271,8 +284,8 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
   /**
     * Creates a SamRecord with a ton of additional tags annotating the duplex read.
     */
-  override protected def createSamRecord(read: DuplexConsensusRead, readType: ReadType): SamRecord = {
-    val rec = super.createSamRecord(read, readType)
+  override protected def createSamRecord(read: DuplexConsensusRead, readType: ReadType, umis: Seq[String] = Seq.empty): SamRecord = {
+    val rec = super.createSamRecord(read, readType, umis)
     val ab = read.abConsensus
     val ba = read.baConsensus
 
--- a/src/main/scala/com/fulcrumgenomics/umi/SimpleConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/SimpleConsensusCaller.scala
@@ -1,0 +1,82 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.umi
+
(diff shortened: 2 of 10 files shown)


src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -203,7 +203,10 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
         // Convert to SamRecords and return
         (duplexR1, duplexR2) match {
           case (Some(r1), Some(r2)) =>
-            Seq(createSamRecord(r1, FirstOfPair), createSamRecord(r2, SecondOfPair))
+            Seq(
+              createSamRecord(r1, FirstOfPair, toUmiBases(filteredAbR1s ++ filteredBaR2s)),
```
tfenne: Looking at which `++` gets called, these calls to `filteredAbR1s ++ filteredBaRs2` do appear to construct a new collection each time and add all the elements.  Might be nice to just store the collections before lines 200-201 and reuse them here.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -216,6 +219,9 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
     }
   }
 
+  /** Extracts the UMI bases for each source read. */
+  private def toUmiBases(reads: Seq[SourceRead]): Seq[String] = reads.flatMap(_.sam).flatMap(_.get[String](ConsensusTags.UmiBases))
```
tfenne: Is the use of `.get` and `.flatMap` here to handle the case that we're presented a BAM that doesn't have `RX`?  If so, maybe just note that so we don't forget?

src/main/scala/com/fulcrumgenomics/umi/SimpleConsensusCaller.scala
```diff
@@ -0,0 +1,76 @@
...
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.umi
+
+import com.fulcrumgenomics.FgBioDef.forloop
+import com.fulcrumgenomics.util.NumericTypes.LogProbability
+
+/**
+  * A trait that can be mixed in to call sequences represented as strings of the same length.
```
tfenne: Tis a class!

src/main/scala/com/fulcrumgenomics/umi/SimpleConsensusCaller.scala
```diff
@@ -0,0 +1,76 @@
...
+  *
+  * If a non-ACGTN bases is encountered at any position in the strings, that position is not consensus called, and the
```
tfenne: Maybe in this case we should require that all the reads have a non-ACGT at the same place?  To protect ourselves against mixing of different length UMI data?  Shouldn't ever happen, but....

src/main/scala/com/fulcrumgenomics/umi/SimpleConsensusCaller.scala
```diff
@@ -0,0 +1,76 @@
...
+  * character from the first sequenced is used.
+  *
+  * @param errorRatePreLabeling The error probability prior to adding the UMI.  By default, set to almost surely zero.
+  * @param errorRatePostLabeling The error probability post to adding the UMI.  By default, set to almost surely zero
+  * @param pError The adjusted error probability.  By default set to ln(0.99), as the actual values don't matter much.
+  * @param pTruth The adjusted truth probability.  By default set to ln(0.01), as the actual values don't matter much.
+  */
+private[umi] class SimpleConsensusCaller(val errorRatePreLabeling: Byte = 90.toByte,
+                                         val errorRatePostLabeling: Byte = 90.toByte,
+                                         val pError: LogProbability = -4.60517,
```
tfenne: It might be nice to re-express these as Q scores here and transform them internally to be more consistent?

src/main/scala/com/fulcrumgenomics/umi/SimpleConsensusCaller.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  /** Calls a simple consensus sequences from a set of sequences all the same length. */
+  def callConsensus(sequences: Seq[String]): String = {
+    require(sequences.nonEmpty, "Can't call consensus on an empty set of UMIs!")
+    require(sequences.forall(_.length == sequences.head.length), "Sequences must all have the same length")
+    val buffer = new StringBuilder
+
+    forloop(from=0, until=sequences.head.length) { i =>
+      this.consensusBuilder.reset()
+      val nonDnaBase = sequences.map(sequence => sequence.charAt(i)).flatMap { char =>
```
tfenne: Given that `sequences` is a `Seq` this is going to create two more concrete `Seq`s along the way - one for the chars, and then another for the option(s).  How about something like this instead:

```scala
forloop (from=0, until=sequences.head.length) { i =>
  this.consensusBuilder.reset()
  var nonDna = 0
  for (sequence <- sequences; char <- sequence) {
    if (!this.DnaBases.contains(char)) nonDna += 1
    else this.consensusBuilder.add(char.toByte, pError=this.pError, pTruth=this.pTruth)
  }

  if (nonDna == 0) buffer.append(this.consensusBuilder.call())
  else if (nonDna == sequences.length) buffer.append(sequences.head.charAt(i))
  else throw new IllegalStateException(s"Sequences contained a mix of DNA and non-DNA characters at offset $i: $sequences")
}
```

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -125,16 +125,18 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
...
 
     // pairs
     (consensusFromSamRecords(firstOfPair), consensusFromSamRecords(secondOfPair)) match {
       case (None, Some(r2))     => rejectRecords(secondOfPair, UmiConsensusCaller.FilterOrphan)
       case (Some(r1), None)     => rejectRecords(firstOfPair,  UmiConsensusCaller.FilterOrphan)
       case (None, None)         => rejectRecords(firstOfPair ++ secondOfPair, UmiConsensusCaller.FilterOrphan)
       case (Some(r1), Some(r2)) =>
-        buffer += createSamRecord(r1, FirstOfPair)
-        buffer += createSamRecord(r2, SecondOfPair)
+        buffer += createSamRecord(r1, FirstOfPair, firstOfPair.flatMap(_.get[String](ConsensusTags.UmiBases)))
```
tfenne: Having the UMI consensus calling in `createSamRecord` means that we're doing it twice when we only need to do it once, right?  I.e. we expect/require that UMIs on R1 and R2 are the same, so we don't _really_ need to call it twice.  How would you feel about having createSamRecord take in the consensus UMI instead to avoid this?

tfenne: Maybe if we make the method more efficient we shouldn't worry about it?

src/test/scala/com/fulcrumgenomics/umi/SimpleConsensusCallerTest.scala
```diff
@@ -0,0 +1,62 @@
...
+    caller.callConsensus(Seq("A", "C", "G", "T")) shouldBe "N"
+    caller.callConsensus(Seq("A", "C", "C", "C")) shouldBe "C"
+    caller.callConsensus(Seq("C", "C", "C", "A")) shouldBe "C"
+    caller.callConsensus(Seq("GATTACA", "GATTACA", "GATTACA", "NNNNNNN")) shouldBe "GATTACA"
+  }
+
+  it should "gracefully handle non-ACGTN bases" in {
+    val caller = new SimpleConsensusCaller()
+    caller.callConsensus(Seq("GATT-ACA", "GATT-ACA", "GATT-ACA")) shouldBe "GATT-ACA"
+    caller.callConsensus(Seq("GATT-ACA", "GATTAACA", "GATTAACA")) shouldBe "GATT-ACA"
```
tfenne: I think these last three should fail, because there isn't a real logical answer to what's the consensus of `-`, `A` and `A`.

src/main/scala/com/fulcrumgenomics/umi/SimpleConsensusCaller.scala
nh13: Close, the inner loop is wrong (we want the base at index `i` for each string), but I get the idea.

nh13:
@tfenne I addressed all your comments, one look before I merge?

nh13:
I forgot to mention that this adds about ~5% to the run time on some of my local data.

src/main/scala/com/fulcrumgenomics/umi/SimpleConsensusCaller.scala
```diff
@@ -0,0 +1,83 @@
...
+  private val consensusBuilder = new ConsensusCaller(
+    errorRatePreLabeling  = this.errorRatePreLabeling,
+    errorRatePostLabeling = this.errorRatePostLabeling
+  ).builder()
+
+  private val DnaBases = Set('A', 'C', 'G', 'T', 'N', 'a', 'c', 'g', 't', 'n')
+
+  /** Calls a simple consensus sequences from a set of sequences all the same length. */
+  def callConsensus(sequences: Seq[String]): String = {
+    require(sequences.nonEmpty, "Can't call consensus on an empty set of UMIs!")
```
tfenne: UMIs -> sequences?  Hangover from where this code came from I think.

src/main/scala/com/fulcrumgenomics/umi/SimpleConsensusCaller.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    forloop (from=0, until=sequences.head.length) { i =>
+      this.consensusBuilder.reset()
+      var nonDna = 0
+      sequences.foreach { sequence =>
+        val char = sequence.charAt(i)
+        if (!this.DnaBases.contains(char)) {
+          nonDna += 1
+          // verify that all non-DNA bases are the same character
+          if (sequences.head.charAt(i) != char) {
```
tfenne: You could use `require(sequences.head.charAt(i) == char, ...)` here which is a little more compact.

nh13:
@chapman, in this PR we have added support outputting the consensus UMI in the consensus BAM (`RX` tag) in addition to the molecule identifier (`MI` tag).

PR #281, opened 2017-09-14
Count single-end reads in CollectErccMetrics.

nh13:


src/main/scala/com/fulcrumgenomics/rnaseq/CollectErccMetrics.scala
```diff
@@ -144,12 +144,13 @@ class CollectErccMetrics
       }
       .filter { r =>
         erccReads += 1
-        // consider only paired records where both ends map to the sam ERCC transcript
-        r.paired && r.mapped && r.mateMapped && r.refIndex == r.mateRefIndex && r.firstOfPair
+        // consider only single-end records, or paired records where both ends map to the sam ERCC transcript
+        !r.paired || (r.mateMapped && r.refIndex == r.mateRefIndex && r.firstOfPair)
```
tfenne: In light of this change, it feels _slightly_ odd to count unpaired reads, but not paired reads where the other end didn't map.  Not _super_ odd, but slightly. What do you think?

nh13: It feels odd that one end wouldn't map when the other one does, especially for synthetic transcripts.  I am inclined to filter those cases out and be more stringent.  I know that the tool allows data to be mixed (both single-end and paired reads) but I think that having such data combined is rare.  You either have single-end data for counting, or paired-end data for isoform detection.  My 2cents.

PR #285, opened 2017-09-29
Added subByQuery() and subByTarget() methods to Alignment.

tfenne:


src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
...
+    * @param consumes a function that returns true if the cigar element passed as a parameter consumes
+    *                 bases on the sequence on which `start` and `end` are expresssed
+    * @return a new [[Alignment]] with adjusted start, end and cigar, and with score=0
+    */
+  private def sub(start: Int, end: Int, initialStart: Int, consumes: CigarElem => Boolean): Alignment = {
+    val elems = new ArrayBuffer[CigarElem](cigar.size)
+    var (qStart, tStart, currStart) = (queryStart,targetStart,initialStart)
+
+    cigar.foreach { elem =>
+      val currEnd = if (consumes(elem)) currStart + elem.length - 1 else currStart
```
nh13: Just for our future sake, could you add a comment similar to: `// the end of the current element`?

nh13: Also, do you want to save `consumes(elem)` to a `val` since you use it multiple times below?

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
...
+    * @param start the start (on either the query or target sequence) of the desired window
+    * @param end the end (on either the query or target sequence) of the desired window
+    * @param initialStart the start position on the relevant sequence (either target OR query)
+    * @param consumes a function that returns true if the cigar element passed as a parameter consumes
+    *                 bases on the sequence on which `start` and `end` are expresssed
+    * @return a new [[Alignment]] with adjusted start, end and cigar, and with score=0
+    */
+  private def sub(start: Int, end: Int, initialStart: Int, consumes: CigarElem => Boolean): Alignment = {
+    val elems = new ArrayBuffer[CigarElem](cigar.size)
+    var (qStart, tStart, currStart) = (queryStart,targetStart,initialStart)
```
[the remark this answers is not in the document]
nh13: Also, could you add comment that `currStart` is the `start` of the current element?

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
...
+      val currEnd = if (consumes(elem)) currStart + elem.length - 1 else currStart
+
+      if (currEnd < start) {
+        // Element before the desired window, need to bump start positions
+        qStart    += elem.lengthOnQuery
+        tStart    += elem.lengthOnTarget
+        if (consumes(elem)) currStart += elem.length
+      }
+      else if (currStart > end) {
+        // Don't include, beyond the range we're interested
```
nh13: What, no cute way to break early?

nh13: But perhaps this should be the first conditional rather than the second

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
...
+      }
+      else if (currStart >= start && currEnd <= end) {
+        // Contained within the target region; ignore insertions that are at the end of the desired region
+        if (consumes(elem) || (currStart != start && currStart != end)) {
+          elems     += elem
```
nh13: who's this aligning to (extra spaces before `+=`)?

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
...
+          if (consumes(elem)) currStart += elem.length
+        }
```
nh13:
```
if (consumes(elem)) {
   // Contained within the target region
   elems += elem
   currStart += elem.length
}
else if (currStart != start && currStart != end) {
  // ignore insertions that are at the end of the desired region
  elems += elem
}
```

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
...
+      }
+      else {
+        // Element is overlaps the desired region, may enclose, be enclosed by or straddle
```
nh13: An extra `is`

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
...
+        var len = elem.length
+
+        if (currStart < start) {
+          // Element is split over the start of the desired region
+          val diff = start - currStart
+          len -= diff
+          if (elem.operator.consumesReadBases())      qStart += diff
+          if (elem.operator.consumesReferenceBases()) tStart += diff
+          currStart += diff
```
nh13: I think you could remove line 358 (`len -= diff`) and line (`currStart += diff`) since you add `len` to `currStart` on 370 anyways.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
...
+          len -= (currEnd - end)
+        }
+
+        elems += CigarElem(elem.operator, len)
+        currStart += len
+      }
+    }
+
+    copy(queryStart=qStart, targetStart=tStart, cigar=Cigar(elems), score=0)
+  }
```
nh13: +1 looks correct to me, thanks for making me do the math!

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -107,10 +113,10 @@ case class Cigar(elems: IndexedSeq[CigarElem]) extends Iterable[CigarElem] {
   def reverseIterator: Iterator[CigarElem] = elems.reverseIterator
 
   /** Returns the length of the alignment on the query sequence. */
-  def lengthOnQuery: Int = elems.filter(_.operator.consumesReadBases()).map(_.length).sum
+  def lengthOnQuery: Int = elems.foldLeft(0)((sum, elem) => sum + elem.lengthOnQuery)
```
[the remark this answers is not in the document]
tfenne: It's more that I expect this may get called quite a bit, and I figured that a fold here would be more efficient than the `.map().sum` which would presumably a) create an intermediate `Seq[Int]` and also box all the values along the way.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
 
     buffers.map(_.toString())
   }
+
+  /**
+    * Returns a subset of the Alignment representing the region defined by `start` and `end` on the query
```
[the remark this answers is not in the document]
tfenne: Done

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -278,4 +284,93 @@ case class Alignment(query: Array[Byte],
...
+    * sequence. The returned alignment will contain the entire query and target sequences, but will have
+    * adjusted `queryStart` and `targetStart` positions and an updated `cigar`.  The score is set to 0.
+    *
+    * @param start the 1-based inclusive position of the first base on the query sequence to include
+    * @param end   the 1-based inclusive position of the last base on the query sequence to include
+    * @return a new [[Alignment]] with updated coordinates and cigar
+    */
+  def subByQuery(start: Int, end: Int): Alignment = {
+    require(start >= queryStart && start <= queryEnd, "start is outside of aligned region of target sequence")
```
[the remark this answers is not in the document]
tfenne: I'm torn on the `this.queryStart` thing.  The whole scala treat "vals and defs" the same, is at odds with the java "put `this.` in front of member variables" and don't put `this.` in front of methods.  It seems a bit weird to have `this.queryStart` but not `this.queryEnd`.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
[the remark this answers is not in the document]
tfenne: Done.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: Done & Done

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: It's tough.  It could be implemented as a `while` or `forloop` but that would, I think, make it even more clunky.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: Good catch.  There was a lot of trial and error and moving code around.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: I went back and forth on this.  I think I prefer the version I have mostly for compactness.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: If I comment out those two lines I get a whole lot of test failures.  The reason being here that `len` is the length of the element to be added to the cigar, after adjusting for the fact that the current element may start before and/or end after the desired region.  So it has to get adjusted.  And then the qStart/tStart/curreStart all need to get bumped up as we're moving up the alignment.

PR #286, opened 2017-10-03
Added a global parameter for log-level.

tfenne:
Tiny PR for you @nh13 that'll hopefully take < 1 minute to review!

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
+++ b/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
@@ -30,12 +30,15 @@ import java.text.DecimalFormat
 
 import com.fulcrumgenomics.bam.api.{SamSource, SamWriter}
 import com.fulcrumgenomics.cmdline.FgBioMain.FailureException
-import com.fulcrumgenomics.commons.util.LazyLogging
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
```diff
@@ -54,7 +57,8 @@ object FgBioMain {
 class FgBioCommonArgs
 ( @arg(doc="Use asynchronous I/O where possible, e.g. for SAM and BAM files.") val asyncIo: Boolean = false,
   @arg(doc="Default GZIP compression level, BAM compression level.")           val compression: Int = 5,
-  @arg(doc="Directory to use for temporary files.")                            val tmpDir: DirPath  = Paths.get(System.getProperty("java.io.tmpdir"))
+  @arg(doc="Directory to use for temporary files.")                            val tmpDir: DirPath  = Paths.get(System.getProperty("java.io.tmpdir")),
+  @arg(doc="Minimum severity log-level to emit.")                              val logLevel: LogLevel = LogLevel.Info
```
nh13: It'd be nice if the message was the same as `dagr` ;)

tfenne: I like this message better ;)  Maybe I'll PR the message in dagr.

PR #288, opened 2017-10-05
Fixed FastqRecord.trimmedTo to use len >= length to decide whether to…

tfenne:
… trim or not


tfenne:
Single char PR!

tfenne:
Just something I saw.  I'm writing a tool that trims a lot of fastq records and went to check to see if the `.trimmedTo` as smart and saw that it tried, but was lacking.

PR #292, opened 2017-10-05
Adding some options to FilterConsensusReads and command-line groups.

nh13:
- filter out reads with low mean base quality
- filter out reads whose minimum depth is too low
- filter out duplex consensus reads where the single strand consensuses disagree


src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -88,21 +90,32 @@ import htsjdk.samtools.reference.ReferenceSequenceFileWalker
...
-  @arg(flag='E', minElements=1, maxElements=3, doc="The maximum raw-read error rate across the entire consensus read.")
+
+ // Filters that are applied per-read
+ @arg(flag='E', group="Per-read", minElements=1, maxElements=3, doc="The maximum raw-read error rate across the entire consensus read.")
   val maxReadErrorRate: Seq[Double] = Seq(0.025),
-  @arg(flag='e', minElements=1, maxElements=3, doc="The maximum error rate for a single consensus base.")
+ @arg(flag='n', group="Per-read", doc="Maximum fraction of no-calls in the read after filtering.")
+  val maxNoCallFraction: Double = 0.2,
+ @arg(flag='m', group="Per-read", minElements=0, maxElements=3, doc="The minimum number of reads supporting a consensus base at any base in the consensus read")
+  val minDepth: Seq[Int] = Seq.empty,
```
nh13: This one was hard to describe, happy to update if you have any suggestions.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -92,17 +94,28 @@ class FilterConsensusReads
   @arg(flag='o', doc="Output SAM or BAM file.") val output: PathToBam,
   @arg(flag='r', doc="Reference fasta file.") val ref: PathToFasta,
   @arg(flag='R', doc="Reverse [complement] per base tags on reverse strand reads.") val reversePerBaseTags: Boolean = false,
-  @arg(flag='M', minElements=1, maxElements=3, doc="The minimum number of reads supporting a consensus base/read.")
+
+ // Filters that are applied both per-base and per-read
+  @arg(flag='M', group="Per-read/Per-base", minElements=1, maxElements=3, doc="The minimum number of reads supporting a consensus base/read.")
```
tfenne: Either this should not be in a group, or `--min-mean-base-quality-score` should be in a group I think.  I'm not sure I'm a fan of the grouping here in general though - when we have to create 4 groups to accurately group ~12 options.

tfenne: Also, if you do keep the groups I think I would add " Filters" to the ends of the group names.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -92,17 +94,28 @@ class FilterConsensusReads
...
-  @arg(flag='E', minElements=1, maxElements=3, doc="The maximum raw-read error rate across the entire consensus read.")
+
+ // Filters that are applied per-read
+  @arg(flag='E', group="Per-read", minElements=1, maxElements=3, doc="The maximum raw-read error rate across the entire consensus read.")
   val maxReadErrorRate: Seq[Double] = Seq(0.025),
-  @arg(flag='e', minElements=1, maxElements=3, doc="The maximum error rate for a single consensus base.")
+  @arg(flag='n', group="Per-read", doc="Maximum fraction of no-calls in the read after filtering.")
+  val maxNoCallFraction: Double = 0.2,
+  @arg(flag='m', group="Per-read", minElements=0, maxElements=3, doc="The minimum number of reads supporting a consensus base at any base in the consensus read")
+  val minDepth: Seq[Int] = Seq.empty,
```
tfenne: I find this _super_ confusing.  To have two options `min-reads` and `min-depth` that do subtly but significantly different things is just going to cause errors.  I think that it will also have un-desirable results when reads have been quality-trimmed prior to consensus calling, leading to a lot of reads having lots of bases above `min-reads`, but some below `min-depth` at the end.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -226,11 +246,15 @@ class FilterConsensusReads
...
+    val minDepth = rec.get[Int](ConsensusTags.PerRead.MinRawReadCount)
     val errRate  = rec.get[Float](ConsensusTags.PerRead.RawReadErrorRate)
     if (maxDepth.isEmpty || errRate.isEmpty) fail(s"Read ${rec.name} does not appear to have consensus calling tags present.")
 
     // Only bother looking at reads where per-read criteria are met
-    if (maxDepth.exists(_ < this.ccFilters.minReads) || errRate.exists(_ > this.ccFilters.maxReadErrorRate)) {
+    if (maxDepth.exists(_ < this.ccFilters.minReads) ||
+      minDepth.exists(_ < this.ccFilters.minDepth) ||
+      errRate.exists(_ > this.ccFilters.maxReadErrorRate) ||
+      minMeanBaseQuality.exists(_ > rec.quals.map(_.toInt).sum.toDouble / rec.length)) {
```
tfenne: I think this would likely be more efficient as:

```scala
rec.quals.foldLeft(0)(_ + _).toDouble / rec.length
```

Since it'll avoid creation of an ArrayBuffer and Array[Int].

tfenne: Can you please add `this.` in front of `minMeanBaseQuality?  I see why you've done it, but it's confusing to me that the first three conditions are all `option from read </> filter value` and on the fourth one it's reversed.  Making it clear that the left hand side is an instance member would help.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -294,6 +320,8 @@ class FilterConsensusReads
       val depths2 = rec[Array[Short]](BaRawReadCount)
       val errors1 = rec[Array[Short]](AbRawReadErrors)
       val errors2 = rec[Array[Short]](BaRawReadErrors)
+      val bases1  = rec.get[String](AbConsensusBases).getOrElse("N" * rec.length)
```
tfenne: This is nasty.  Is there really no better way to handle this than manufacturing a string of Ns per each read if this is missing?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -92,17 +94,28 @@ class FilterConsensusReads
...
+  @arg(flag='e', group="Per-base", minElements=1, maxElements=3, doc="The maximum error rate for a single consensus base.")
   val maxBaseErrorRate: Seq[Double] = Seq(0.1),
-  @arg(flag='N', doc="Mask (make `N`) consensus bases with quality less than this threshold.")
+  @arg(flag='N', group="Per-base", doc="Mask (make `N`) consensus bases with quality less than this threshold.")
   val minBaseQuality: PhredScore,
-  @arg(flag='n', doc="Maximum fraction of no-calls in the read after filtering.")
-  val maxNoCallFraction: Double = 0.2
-
+  @arg(flag='s', group="Per-base", doc="Require perfect agreement between the AB and BA consensus reads when both are present (for duplex-sequencing).")
+  val singleStrandAgreement: Boolean = false
```
tfenne: This should be named so as not to be open to interpretation.  E.g. `requireSingleStrandAgreement`.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -56,11 +56,13 @@ import htsjdk.samtools.reference.ReferenceSequenceFileWalker
       |
       |  1. Reads with fewer than min-reads contributing reads are filtered out
       |  2. Reads with an average consensus error rate higher than max-read-error-rate are filtered out
-      |  3. Bases with quality scores below min-base-quality are masked to Ns
-      |  4. Bases with fewer than min-reads contributing raw reads are masked to Ns
-      |  5. Bases with a consensus error rate (defined as the fraction of contributing reads that
+      |  3. Reads with whose minimum depth less than min-depth are filtered out
+      |  4. Reads with mean base quality less than min-mean-base qulaity are filtered out (if specified)
```
tfenne: qulaity -> quality

tfenne: I know the ordering implies this, but I think we should explicitly state that this is the mean quality score of the consensus read _prior_ to any masking.

nh13:
@tfenne one more look?  I tried cleaning up the case where we are missing the single-strand consensus sequences, but it ain't any prettier.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -37,6 +37,8 @@ import com.fulcrumgenomics.util.{Io, ProgressLogger}
 import htsjdk.samtools.SAMFileHeader.SortOrder
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker
 
+import scala.collection.immutable.IndexedSeq
```
tfenne: Isn't this in PreDef?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -56,11 +58,13 @@ import htsjdk.samtools.reference.ReferenceSequenceFileWalker
       |
       |  1. Reads with fewer than min-reads contributing reads are filtered out
       |  2. Reads with an average consensus error rate higher than max-read-error-rate are filtered out
-      |  3. Bases with quality scores below min-base-quality are masked to Ns
-      |  4. Bases with fewer than min-reads contributing raw reads are masked to Ns
-      |  5. Bases with a consensus error rate (defined as the fraction of contributing reads that
+      |  3. Reads with mean base quality of the consensus read, prior to any masking, less than min-mean-base quality
```
tfenne: `min-mean-base quality` -> `min-mean-base-quality`

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -72,6 +76,9 @@ import htsjdk.samtools.reference.ReferenceSequenceFileWalker
       |FilterConsensusReads ... --min-reads 10 5 3 --max-base-error-rate 0.1
       |```
       |
+      |Furthermore, the `--require-single-strand-agreement` option (for duplex consensus reads) will only be applied
```
tfenne: Should this be in the numbered list of filters?  Something like:

```
7. For duplex reads, if --require-single-strand-agreement is provided, masks to N any bases where the base was observed in both single-strand consensus reads and the two reads did not agree.
```

tfenne: Just to build on this ... from the usage and arg documentation alone it's not clear to me how this filter works (it is in code).  I.e. I could interpret this as masking bases, or as requiring this be true of all bases in the read otherwise reject the read.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -294,6 +307,15 @@ class FilterConsensusReads
       val depths2 = rec[Array[Short]](BaRawReadCount)
       val errors1 = rec[Array[Short]](AbRawReadErrors)
       val errors2 = rec[Array[Short]](BaRawReadErrors)
+      val bases1  = rec.get[String](AbConsensusBases)
+      val bases2  = rec.get[String](BaConsensusBases)
+      val singleStrandDisagreements: Seq[Boolean] = { // pre-compute where the two single-strand consensus sequences disagree
```
tfenne: I had an idea for this!  How about a local function:

```scala
def ssAgrees: Int => Boolean = (bases1, bases2) match {
  case (Some(s1), Some(s2)) => (idx) => s1(idx) == s2(idx)
  case _                    => (idx) => true
}
```

Then below you can write:

```scala
val singleStrandDisagree = this.requireSingleStrandAgreement && !ssAgree(i)
```

PR #293, opened 2017-10-05
Adding the option to output duplex consensus reads generated from onl…

nh13:
…y a single-strand.

@tfenne Some things I want to highlight two things:
1. The option `--allow-single-strand` is off by default.
2. We do not output the single strand `SamRecord`, but instead a `SamRecord` without the per-read and per-base tags for the missing single-strand consensus.  In my own thinking, I want this behavior for consistency in downstream tools that will consume duplex `SamRecord`s.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
@@ -76,6 +76,17 @@ import com.fulcrumgenomics.util.ProgressLogger
     |
     |The per base depths and errors are both capped at 32,767. In all cases no-calls (Ns) and bases below the
     |min-input-base-quality are not counted in tag value calculations.
+    |
+    |The --min-reads option can take 1-3 values similar to `FilterConsensusReads`. For example:
+    |
+    |```
+    |CallDuplexConsensusReads ... --min-reads 10 5 3
+    |```
+    |
+    |If fewer than three values are supplied, the last value is repeated (i.e. `5 4` -> `5 4 4` and `1` -> `1 1 1`.  The
+    |first value applies to the final consensus read, the second value to one single-strand consensus, and the last
+    |value to the other single-strand consensus. It is required that if values two and three differ,
+    |the _more stringent value comes earlier_.
   """,
   group = ClpGroups.Umi)
 class CallDuplexConsensusReads
@@ -86,8 +97,9 @@ class CallDuplexConsensusReads
  @arg(flag='1', doc="The Phred-scaled error rate for an error prior to the UMIs being integrated.") val errorRatePreUmi: PhredScore = DefaultErrorRatePreUmi,
  @arg(flag='2', doc="The Phred-scaled error rate for an error post the UMIs have been integrated.") val errorRatePostUmi: PhredScore = DefaultErrorRatePostUmi,
  @arg(flag='m', doc="Ignore bases in raw reads that have Q below this value.") val minInputBaseQuality: PhredScore = DefaultMinInputBaseQuality,
- @arg(flag='t', doc="If true quality trim input reads in addition to masking low Q bases.") val trim: Boolean = false,
- @arg(flag='S', doc="The sort order of the output, if `:none:` then the same as the input.") val sortOrder: Option[SamOrder] = Some(SamOrder.Queryname)
+ @arg(flag='t', doc="If true, quality trim input reads in addition to masking low Q bases.") val trim: Boolean = false,
+ @arg(flag='S', doc="The sort order of the output, if `:none:` then the same as the input.") val sortOrder: Option[SamOrder] = Some(SamOrder.Queryname),
+ @arg(flag='M', minElements=1, maxElements=3, doc="The minimum number of input reads to a consensus read.") val minReads: Seq[Int] = Seq(1)
 ) extends FgBioTool with LazyLogging {
 
   Io.assertReadable(input)
@@ -108,7 +120,8 @@ class CallDuplexConsensusReads
       minInputBaseQuality = minInputBaseQuality,
       trim                = trim,
       errorRatePreUmi     = errorRatePreUmi,
-      errorRatePostUmi    = errorRatePostUmi
+      errorRatePostUmi    = errorRatePostUmi,
+      minReads            = minReads
     )
 
     val iterator = new ConsensusCallingIterator(in.toIterator, caller, Some(ProgressLogger(logger)))
--- a/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
@@ -46,31 +46,32 @@ object DuplexConsensusCaller {
   val NoCallQual   = PhredScore.MinValue
 
   /** Additional filter strings used when rejecting reads. */
-  val FilterAorBOnly    = "Only Seeing A or B Reads"
+  val FilterMinReads    = "Not Enough Reads (Either Total, AB, or BA)"
   val FilterFragments   = "Being Fragment/Non-Paired Reads"
   val FilterSsConsensus = "Only Generating One Strand Consensus"
   val FilterCollision   = "Potential collision between independent duplex molecules"
 
   /**
     * Stores information about a consensus read.  Bases, arrays, and the two single
-    * strand consensus reads must all be the same length.
+    * strand consensus reads (when present) must all be the same length.
     *
     * @param bases the base calls of the consensus read
     * @param quals the calculated phred-scaled quality scores of the bases
     * @param errors the total number of raw read errors vs. the final consensus sequence at each base
     * @param abConsensus the AB consensus read from which the duplex was constructed
-    * @param baConsensus the AB consensus read from which the duplex was constructed
+    * @param baConsensus the BA consensus read from which the duplex was constructed, if present. It may not be present
+    *                    if created from the AB strand reads only.
     */
   case class DuplexConsensusRead(id: String,
                                  bases: Array[Byte],
                                  quals: Array[Byte],
                                  errors: Array[Short],
                                  abConsensus: VanillaConsensusRead,
-                                 baConsensus: VanillaConsensusRead) extends SimpleRead {
+                                 baConsensus: Option[VanillaConsensusRead]) extends SimpleRead {
     require(bases.length == quals.length,  "Bases and qualities are not the same length.")
     require(bases.length == errors.length, "Bases and errors are not the same length.")
-    require(bases.length == abConsensus.length, "Bases and AB consensus are not the same length.")
-    require(bases.length == baConsensus.length, "Bases and BA consensus are not the same length.")
+    require(abConsensus.length == bases.length, "Bases and AB consensus are not the same length.")
+    require(baConsensus.forall(_.length == bases.length), "Bases and BA consensus are not the same length.")
   }
 }
 
@@ -94,14 +95,23 @@ object DuplexConsensusCaller {
   * @param trim if true, quality trim reads in addition to masking. If false just mask.
   * @param errorRatePreUmi the estimated rate of errors in the DNA prior to attaching UMIs
   * @param errorRatePostUmi the estimated rate of errors in the DNA post attaching UMIs
-  */
+  * @param minReads the minimum number of input reads to a consensus read (see [[CallDuplexConsensusReads]]).
+  * */
 class DuplexConsensusCaller(override val readNamePrefix: String,
                             override val readGroupId: String    = "A",
                             val minInputBaseQuality: PhredScore = DuplexConsensusCaller.MinInputBaseQuality,
-                            val trim: Boolean = false,
+                            val trim: Boolean                   = false,
                             val errorRatePreUmi: PhredScore     = DuplexConsensusCaller.ErrorRatePreUmi,
-                            val errorRatePostUmi: PhredScore    = DuplexConsensusCaller.ErrorRatePostUmi
+                            val errorRatePostUmi: PhredScore    = DuplexConsensusCaller.ErrorRatePostUmi,
+                            val minReads: Seq[Int]              = Seq(1)
                            ) extends UmiConsensusCaller[DuplexConsensusRead] with LazyLogging {
+
+  private val Seq(minTotalReads, minXyReads, minYxReads) = this.minReads.padTo(3, this.minReads.last)
+
+
+  // For depth thresholds it's required that ba <= ab <= cc
+  require(minXyReads <= minTotalReads, "min-reads values must be specified high to low.")
+  require(minYxReads <= minXyReads, "min-reads values must be specified high to low.")
 
   private val ssCaller = new VanillaUmiConsensusCaller(readNamePrefix="x", options=new VanillaUmiConsensusCallerOptions(
       errorRatePreUmi         = this.errorRatePreUmi,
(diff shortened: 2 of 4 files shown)


src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -108,7 +109,8 @@ class CallDuplexConsensusReads
       minInputBaseQuality = minInputBaseQuality,
       trim                = trim,
       errorRatePreUmi     = errorRatePreUmi,
-      errorRatePostUmi    = errorRatePostUmi
+      errorRatePostUmi    = errorRatePostUmi,
+      allowSingleStrand          = allowSingleStrand
```
tfenne: Bonus spaces.

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -86,8 +86,9 @@ class CallDuplexConsensusReads
  @arg(flag='1', doc="The Phred-scaled error rate for an error prior to the UMIs being integrated.") val errorRatePreUmi: PhredScore = DefaultErrorRatePreUmi,
  @arg(flag='2', doc="The Phred-scaled error rate for an error post the UMIs have been integrated.") val errorRatePostUmi: PhredScore = DefaultErrorRatePostUmi,
  @arg(flag='m', doc="Ignore bases in raw reads that have Q below this value.") val minInputBaseQuality: PhredScore = DefaultMinInputBaseQuality,
- @arg(flag='t', doc="If true quality trim input reads in addition to masking low Q bases.") val trim: Boolean = false,
- @arg(flag='S', doc="The sort order of the output, if `:none:` then the same as the input.") val sortOrder: Option[SamOrder] = Some(SamOrder.Queryname)
+ @arg(flag='t', doc="If true, quality trim input reads in addition to masking low Q bases.") val trim: Boolean = false,
+ @arg(flag='S', doc="The sort order of the output, if `:none:` then the same as the input.") val sortOrder: Option[SamOrder] = Some(SamOrder.Queryname),
+ @arg(flag='s', doc="If true, output consensus reads generated from only one strand, otherwise require both strands.") val allowSingleStrand: Boolean = false
```
tfenne: I wonder if it would be more consistent here to turn this into:

```scala
@arg(flag='M', minElements=1, maxElements=3, doc="The minimum number of input reads to a consensus read.") val minReads: Seq[Int] = Seq(1)
```

So that it more directly matches what is available in `FilterConsensusReads` and provides a bit more control?  E.g. you might be willing to take any consensus from at least 2 reads, but not single-strands from 1 read?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -94,13 +95,15 @@ object DuplexConsensusCaller {
...
+  * @param allowSingleStrand if true, output consensus reads generated from only one strand, false otherwise.
+  * */
 class DuplexConsensusCaller(override val readNamePrefix: String,
                             override val readGroupId: String    = "A",
                             val minInputBaseQuality: PhredScore = DuplexConsensusCaller.MinInputBaseQuality,
                             val trim: Boolean = false,
                             val errorRatePreUmi: PhredScore     = DuplexConsensusCaller.ErrorRatePreUmi,
-                            val errorRatePostUmi: PhredScore    = DuplexConsensusCaller.ErrorRatePostUmi
+                            val errorRatePostUmi: PhredScore    = DuplexConsensusCaller.ErrorRatePostUmi,
+                            val allowSingleStrand: Boolean      = false
```
tfenne: Same thought here about having minReads/minAbReads/minBaReads.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -137,87 +140,90 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
-          val ss2Flag = singleStrand2.head.negativeStrand
-          if (singleStrand2.exists(_.negativeStrand != ss2Flag)) {
-            val ss2MI   = singleStrand2.head.apply[String](ConsensusTags.MolecularId)
-            rejectRecords(ab ++ ba, FilterCollision)
-            logger.debug(s"Not all AB-R2s and BA-R1s were on the same strand for molecule with id: $ss2MI")
-            return Nil
-          }
-        }
+    // Partition reads into those that end with /A and those that end with /B (verify the latter
+    val (ab, ba) = pairs.partition(r => r[String](ConsensusTags.MolecularId).endsWith("/A"))
```
tfenne: I don't like how this and the subsequent lines encode direct knowledge of the expected MI structure.  The only requirement elsewhere is that the MI be of the form "common-prefix/one-of-two-suffixes".  There's even a method in this class `def sourceMoleculeId(rec: SamRecord): String` to get the molecule ID that is used elsewhere.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -137,87 +140,90 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
-          case _                    =>
-            // NB: some reads may have been rejected already in filterToMostCommonAlignment, so just
-            //     reject those records that survived the initial filtering.
-            val remainingRecs = filteredXs ++ filteredYs
-            rejectRecords(remainingRecs.flatMap(_.sam), FilterSsConsensus)
-            Nil
+    // Verify ba reads end with /B
+    require(ba.forall(r =>  r[String](ConsensusTags.MolecularId).endsWith("/B")), "Expected MIs to end with /A or /B")
+    // Verify they all have the same molecular identifiers
+    require(pairs.map(_.apply[String](ConsensusTags.MolecularId).dropRight(2)).distinct.length <= 1,
```
tfenne: This would break too if we change GroupReadsByUmi.  Please use `sourceMoleculeId(rec)`.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
tfenne: If you really need to know AB vs. BA (not just one vs. two sets) then I think we need `GroupReadsByUmi` to expose the two suffixes as public vals and re-use them here.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
tfenne: Also, this is already enforced by the iterator machinery that calls into this, I'm not sure we need to re-validate here.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -137,87 +140,90 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+      "SamRecords supplied with more than two distinct MI values.")
+
+    // If we are missing AB or BA reads, and we don't wan to produce consensus reads from a single strand
```
tfenne: wan -> want

nh13:
@tfenne I updated based on your suggestions, care to take a look?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -94,15 +95,25 @@ object DuplexConsensusCaller {
...
+                            val trim: Boolean                   = false,
                             val errorRatePreUmi: PhredScore     = DuplexConsensusCaller.ErrorRatePreUmi,
-                            val errorRatePostUmi: PhredScore    = DuplexConsensusCaller.ErrorRatePostUmi
+                            val errorRatePostUmi: PhredScore    = DuplexConsensusCaller.ErrorRatePostUmi,
+                            val minReads: Seq[Int]              = Seq(1)
                            ) extends UmiConsensusCaller[DuplexConsensusRead] with LazyLogging {
 
+  private val minTotalReads: Int = this.minReads.head
+  private val minXYReads: Int    = this.minReads.take(2).last
+  private val minYXReads: Int    = this.minReads.last
```
tfenne: I know I wrote something similar to this somewhere else.  Now though, I think it would be clearer as:

```scala
private val Seq(minTotalReads, minXyReads, minYxReads) = this.minReads.padTo(3, this.minReads.last)
```

Also can `XY` and `YX` be `Xy` and `Yx` please?  Or maybe we should just call them X and Y instead of XY and YX?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -140,84 +151,99 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
-        if (singleStrand2.nonEmpty) {
-          val ss2Flag = singleStrand2.head.negativeStrand
-          if (singleStrand2.exists(_.negativeStrand != ss2Flag)) {
-            val ss2MI   = singleStrand2.head.apply[String](ConsensusTags.MolecularId)
-            rejectRecords(ab ++ ba, FilterCollision)
-            logger.debug(s"Not all AB-R2s and BA-R1s were on the same strand for molecule with id: $ss2MI")
-            return Nil
-          }
-        }
+    if (!hasMinimumNumberOfReads(groups)) {
```
tfenne: I think this little block would be clearer like:

```scala
require(groups.length <= 2, "SamRecords supplied with more than two distinct MI values.")
val x = groups.head
val y = groups.lift(1).getOrElse(Seq.empty)

if (hasMinimumNumberOfReads(x, y)) {
  callDuplexConsensusRead(x, y)
else {
  rejectRecords(groups.flatten, FilterMinReads)
  Nil
}
```

I was worried that there wasn't going to be a check of <= 2 groups, and had to go hunting to find it buried in the `hasMinimumNumberOfReads()` function.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -140,84 +151,99 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
-            // NB: some reads may have been rejected already in filterToMostCommonAlignment, so just
-            //     reject those records that survived the initial filtering.
-            val remainingRecs = filteredXs ++ filteredYs
-            rejectRecords(remainingRecs.flatMap(_.sam), FilterSsConsensus)
-            Nil
-        }
-      case _ =>
-        unreachable("SamRecords supplied with more than two distinct MI values.")
+  /** Returns true if there enough reads according to the minReads option. */
+  private def hasMinimumNumberOfReads(groups: Seq[Seq[SamRecord]]): Boolean = {
```
tfenne: Suggested above, that this take in two Seqs instead of Seq[Seq[]]

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -140,84 +151,99 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+    val (numXY: Int, numYX: Int) = {
+      require(groups.length <= 2, "SamRecords supplied with more than two distinct MI values.")
+      val numAb = groups.headOption.map(_.count(r => r.paired && r.firstOfPair)).getOrElse(0)
+      val numBa = groups.drop(1).headOption.map(_.count(r => r.paired && r.firstOfPair)).getOrElse(0)
+      if (numAb >= numBa) (numAb, numBa) else (numBa, numAb)
+    }
+    this.minTotalReads <= numXY + numYX && this.minXYReads <= numXY && this.minYXReads <= numYX
+  }
+
+  /** Attempts to call a duplex conesnsu reads from the two sets of reads, one for each strand. */
```
tfenne: conesnsu?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -140,84 +151,99 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+    // The orientation of AB and BA reads should be:
+    // AB R1: +  AB R2: -
+    // BA R1: -  BA R2: +
+    // or vice versa (AB-R1:-, AB-R2:+, AB-R1:-, AB-R2: +
+    // Therefore, AB-R1s and BA-R2s should be on the same strand, and the same for AB-R2s and BA-R1s
+    // Check for this explicitly here.
+    if (singleStrand1.nonEmpty) {
+      val ss1Flag = singleStrand1.head.negativeStrand
+      if (singleStrand1.exists(_.negativeStrand != ss1Flag)) {
+        val ss1MI   = singleStrand1.head.apply[String](ConsensusTags.MolecularId)
```
tfenne: `ss1MI` -> `ss1Mi`; also extra spaces before `=`

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -140,84 +151,99 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+        rejectRecords(ab ++ ba, FilterCollision)
+        logger.debug(s"Not all AB-R1s and BA-R2s were on the same strand for molecule with id: $ss1MI")
+        return Nil
```
tfenne: `return` is bad: https://tpolecat.github.io/2014/05/09/return.html

tfenne: I think probably you could extract a small function `areAllSameStrand(rs: Seq[SamRecord])` and then restructure this something like:

```scala
(areAllSameStrand(singleStrand1), areAllSameStrand(singleStrand2)) match {
  case (true, true) => 
    // do all the things
  case (result1, result2) => 
    rejectRecords(ab ++ ba, FilterCollision)
    if (!result1) ....
    if (!result2) ....
    Nil
}
```

src/main/scala/com/fulcrumgenomics/umi/ConsensusTags.scala
```diff
@@ -48,6 +48,14 @@ object ConsensusTags {
   /** Post-grouping ID that is file-wide unique per source molecule. */
   val MolecularId = "MI"
 
+  /** The suffix appended to molecular identifier reads from the "top strand" of the duplex molecule. These reads have
+    * 5' coordinate for read 1 less than or equal to the 5' coordinate for read 2. */
+  val TopStrandDuplex = "/A"
```
tfenne: Everything else in here is a two-letter SAM attribute name.  It feels weird to put these constants here.  Any objection to just having them be public vals in `GroupReadsByUmi`'s companion object?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
nh13: why is my spell checker off?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
nh13: This is how it was previously written.

nh13: Fixed

dilmiperera:
Hi Tim and Nils,

I've switched to the latest version of fgbio (0.4.0) and one of the new options in this version is the option to output consensus reads that are formed from only a single strand in the CallDuplexConsensusReads method. I was wondering what parameter/argument I need to specify to get this option?

Thanks,
Dilmi

PR #295, opened 2017-10-18
Attempting to make doclet for consensus tags

nh13:
@tfenne I am really close and since you wrote the original doclet I was wondering if you could help.  I am missing the _value_ of each member variable and I am not sure how to get it.  This is all useless unless I can get the two-character tag!  Thanks for any help you can offer.


nh13:
@tfenne so close, yet so far, closing

PR #296, opened 2017-10-18
Add a SamOrder that can sort by tag.

nh13:
Also allows for records with missing values for a tag, and a transformation method to be applied to tag to generate a key on which to sort.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
@@ -124,8 +124,7 @@ object Bams extends LazyLogging {
              header: SAMFileHeader,
              maxRecordsInRam: Int = MaxInMemory,
              tmpDir: DirPath = Io.tmpDir): Sorter[SamRecord,order.A] = {
-    // FIXME: tmpDir not used
-    new Sorter(maxRecordsInRam, new SamRecordCodec(header), order.sortkey)
+    new Sorter(maxRecordsInRam, new SamRecordCodec(header), order.sortkey, tmpDir=tmpDir)
   }
 
   /** A wrapper to order objects of type [[TagType]] using the ordering given.  Used when sorting by tag where we wish
@@ -254,8 +253,7 @@ object Bams extends LazyLogging {
         }
       )
     )
-    // FIXME: tmpDir not used
-    val sort = new Sorter(maxInMemory, new SamRecordCodec(header), f)
+    val sort = new Sorter(maxInMemory, new SamRecordCodec(header), f, tmpDir=tmpDir)
     sort ++= iterator
     new SelfClosingIterator(sort.iterator, () => sort.close())
   }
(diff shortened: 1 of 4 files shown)


nh13:
@tfenne care to take another look?

A few comments before you start:
1. `transform: A => B = a => a` doesn't compile, see [this link](https://stackoverflow.com/questions/38611092/scala-missing-parameter-type-for-expanded-function)
2. `tmpDir` was not being used in the sorter in `Bams.scala`
3. `SamOrderTest` should use `sliding`, not `grouped`.

Thanks!

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -219,6 +223,65 @@ object Bams extends LazyLogging {
     }
   }
 
+  /** Returns an iterator over the records in the given reader such that the order of the records returned is determined
+    * by the value of the given SAM tag, which can optionally be transformed.
```
tfenne: Do you want to enforce any sub-ordering?  E.g. by queryname and then flags?  Or do you think that sorting by just tag is sufficient?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -219,6 +223,65 @@ object Bams extends LazyLogging {
...
+    * @return an Iterator over records sorted by the given SAM tag, optionally transformed.
+    */
+  def sortByTransformedTag[A, B](iterator: Iterator[SamRecord],
+                                 header: SAMFileHeader,
+                                 maxInMemory: Int = MaxInMemory,
+                                 tmpDir: DirPath = Io.tmpDir,
+                                 tag: String,
+                                 defaultValue: Option[A] = None,
+                                 transform: A => B)
+                                (implicit ordering: Ordering[B]): Iterator[SamRecord] = {
```
tfenne: I'm curious.  Is there a reason to prefer using this implicit vs. just having `B <: Ordered`?  I see this done elsewhere too, but I don't know why it's preferred.

tfenne: Just following on from this.... since you have the implicit here, can it be provided explicitly?  And if yes, could we collapse the two methods and do away with the transform?

I.e. could I do:

```scala
sortByTag[String](iter, header, "RX")((a,b) => a.reverse <= b.reverse)
```

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -219,6 +223,65 @@ object Bams extends LazyLogging {
     }
   }
 
+  /** Returns an iterator over the records in the given reader such that the order of the records returned is determined
```
tfenne: reader -> iterator

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
nh13: For now it is sufficient, we can add that in later (without harm) if we want it.  I don't see anywhere that sorting is stable, but if it was we could first sort by queryname/flags then by tag, but that's ugly.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
nh13: (1) `Int` doesn't implement `Ordered[Int]`.  Instead, you can require an implicit and in scala PreDef there is an implicit conversion from `Int` to `RichInt` which gives you `Ordered[Int]`.
(2) That _almost_ works, but doesn't, since `<=` relies on the type being ordered.

PR #299, opened 2017-10-30
DemuxFastqs should not append a trailing /1 or /2 to FASTQs when outp…

nh13:
…utting according to Illumina standards.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -364,7 +364,7 @@ val qualityFormat: Option[FastqQualityFormat] = None,
       maxNoCalls       = maxNoCalls
     )
 
-    val progress = new ProgressLogger(this.logger, unit=1e6.toInt)
+    val progress = ProgressLogger(this.logger, unit=1e6.toInt)
 
     // An iterator that uses the given fastq demultiplexer to convert FASTQ records from the same fragment/template to
     // DemuxRecord in parallel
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -499,12 +501,11 @@ private[fastq] class FastqRecordWriter(prefix: PathPrefix, val pairedEnd: Boolea
       bases      = rec.bases,
       quals      = rec.quals,
       comment    = None,
-      readNumber = Some(rec.readNumber)
+      readNumber = if (illuminaStandards) None else Some(rec.readNumber)
```
nh13: The real change.

src/test/scala/com/fulcrumgenomics/fastq/DemuxFastqsTest.scala
```diff
@@ -421,6 +421,72 @@ class DemuxFastqsTest extends UnitSpec with OptionValues with ErrorLogLevel {
     }
   }
 
+  Seq(true, false).foreach { illuminaStandards =>
+    // A file containing valid paired-end FASTQ records.
+    val pairedFastqPath = {
+      val fastqs = new ListBuffer[FastqRecord]()
+      fastqs += fq(name="Instrument:RunID:FlowCellID:Lane:Tile:X:1 ReadNum:FilterFlag:0:SampleNumber", bases=sampleBarcode1 + "A"*100) // matches the first sample -> first sample
```
tfenne: It's a little weird to include the name and the comment via `name=`.  If you added a read number here it would do the wrong thing and write `Instrument:RunID:FlowCellID:Lane:Tile:X:1 ReadNum:FilterFlag:0:SampleNumber/1`.

Maybe add `(..., comment: Option[String] = None)` to the `fq` method and use it?

src/test/scala/com/fulcrumgenomics/fastq/DemuxFastqsTest.scala
```diff
@@ -421,6 +421,72 @@ class DemuxFastqsTest extends UnitSpec with OptionValues with ErrorLogLevel {
     }
   }
 
+  Seq(true, false).foreach { illuminaStandards =>
+    // A file containing valid paired-end FASTQ records.
```
tfenne: What makes then paired end?

PR #300, opened 2017-10-31
Fix typo that resulted in C>G being called A>G in ErrorRateByReadPosition.

tfenne:


src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -213,7 +213,7 @@ private class ObsCounter(readNumber: Int, position: Int) {
     case ('A', 'T') => a_to_t_obs += 1
     case ('C', 'A') => c_to_a_obs += 1
     case ('C', 'C') => c_ref_obs  += 1
-    case ('C', 'G') => a_to_g_obs += 1
+    case ('C', 'G') => c_to_g_obs += 1
```
nh13: Looks like we don't have a test that covers this and a few other cases. Care to add a test that adds a different # of each and checks?

tfenne: Not today.

PR #302, opened 2017-11-09
Improved speed of reading metrics files with many rows.

tfenne:
On a test of reading 10,000 rows of `TestMetric` from an in-memory iterator of lines, this went from ~57s to 0.7s, or about a 70-fold speedup!

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/Metric.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/Metric.scala
@@ -28,7 +28,7 @@ package com.fulcrumgenomics.util
 import java.io.{Closeable, Writer}
 import java.nio.file.Path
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -68,41 +68,39 @@ object Metric {
     reflectiveBuilder.argumentLookup.ordered.map(_.name)
   }
 
-  /** Construct a metric of the given type.  The arguments given should be a sequence of tuples: the name of the field
-    * and value as a string. */
-  def build[T <: Metric](args: Seq[(String, String)])(implicit tt: ru.TypeTag[T]): T = {
```
nh13: I definitely use this method elsewhere, please do not remove.

Issue #294, opened 2017-10-13
StackOverflowError in GroupReadsByUmi in deep regions


chapmanb:
Nils and Tim;
We're running into a StackOverflowError when dealing with deep regions with a large number of reads per UMI. I've put together a reproducible example here with an input BAM and run script:

wget https://s3.amazonaws.com/chapmanb/testcases/fgbio_umi_overflow.tar.gz

The command line we're using is:
```
fgbio -Xms500m -Xmx4g -XX:+UseSerialGC --async-io=true --compression=0 GroupReadsByUmi --edits=1 --min-map-q=1 -t RX -s adjacency -i in.bam | fgbio -Xms500m -Xmx4g -XX:+UseSerialGC  --async-io=true --compression=0 CallMolecularConsensusReads --min-input-base-quality=10 --min-reads=1 --max-reads=10 --output-per-base-tags=false --sort-order=:none: -i /dev/stdin -o out.bam
```
I'd been hoping `--max-reads` would help downsample and avoid the issue but it doesn't seem to impact runtime or help resolve the underlying issue. I've also tried on a different non-failing (but slow sample) and it didn't seem to help runtimes there so I might be misusing it for trying to power through diverse deep regions.

Any suggestions or tips would be much appreciated.
```
[2017/10/13 13:49:37 | FgBioMain | Info] Executing GroupReadsByUmi from fgbio version 0.2.1-SNAPSHOT as kkrq359@gpu2 on JRE 1.8.0_102-b14 with snappy
[2017/10/13 13:49:37 | FgBioMain | Info] Executing CallMolecularConsensusReads from fgbio version 0.2.1-SNAPSHOT as kkrq359@gpu2 on JRE 1.8.0_102-b14 with snappy
[2017/10/13 13:49:37 | GroupReadsByUmi | Info] Filtering and sorting input.
[2017/10/13 13:50:03 | GroupReadsByUmi | Info] Sorted     1,000,000 record.  Elapsed time: 00:00:25s.  Time for last 1,000,000:   25s.  Last read position: chr6:157,416,972
[2017/10/13 13:50:07 | GroupReadsByUmi | Info] Assigning reads to UMIs and outputting.
[2017/10/13 14:01:41 | CallMolecularConsensusReads | Info] processed       500,000 record.  Elapsed time: 00:11:30s.  Time for last 500,000:  689s.  Last read position: chr6:157,416,666
[...]

PR #304, opened 2017-11-10
Attempting to eliminate a stack overflow in GroupReadsByUmi.

tfenne:
@nh13 This is a fix for Brad's issue in #294.  There are no new tests here because it's just an internal refactor.  There are two main changes:

1. Instead of there being a method to count the number of mismatches, the method now is just asking "does the UMI match enough" so that it can abort early when it detects too many mismatches.  This is sort of unrelated, other than improving performance overall therefore making testing these complex networks of UMIs easier.
2. Instead of having a recursive implementation (`addChildren()`) it now uses a `Stack` and an iterative approach so as not to cause overflows.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -180,13 +180,26 @@ object GroupReadsByUmi {
       })
 
       override def equals(other: scala.Any): Boolean = other.isInstanceOf[Node] && this.umi == other.asInstanceOf[Node].umi
+
+      override def toString: String = s"$umi [$count]"
     }
 
     /** Returns the count of each raw UMI that was observed. */
     protected def count(umis: Seq[Umi]): Iterator[(Umi,Long)] = SimpleCounter(umis).iterator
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -193,31 +206,25 @@ object GroupReadsByUmi {
...
-        * but as counts of the root go up they must be >= approximately twice the child count.
-        */
-      def addChildren(root: Node, others: mutable.Buffer[Node]) : Unit = {
-        val xs = others.filter(other => root.count >= 2 * other.count - 1 && differences(root.umi, other.umi) <= maxMismatches )
-        root.children ++= xs
-        others --= xs
-        root.children.foreach(r => addChildren(r, others))
-      }
+      // Make a list of counts of all UMIs in order from most to least abundant; we'll consume from this buffer
+      var remaining = count(rawUmis).map{case(umi,count) => new Node(umi, count)}.toBuffer.sortBy((n:Node) => -n.count)
```
nh13: Space after `{` and before `}` please.

PR #308, opened 2017-11-10
Updated GroupReadsByUmi to reject UMIs with Ns. Also to log counts of…

tfenne:
… rejected reads.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -410,15 +410,24 @@ class GroupReadsByUmi
     val sorter = Bams.sorter(SamOrder.TemplateCoordinate, header)
     val sortProgress = ProgressLogger(logger, verb="Sorted")
 
+    // A handful of counters for tracking reads
+    var (filteredNonPf, filteredPoorAlignment, filteredNsInUmi, kept) = (0L, 0L, 0L, 0L)
+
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -410,15 +410,24 @@ class GroupReadsByUmi
...
     // Filter and sort the input BAM file
     logger.info("Filtering and sorting input.")
     in.iterator
-      .filter(r => includeNonPfReads || r.pf)
       .filter(r => !r.secondary && !r.supplementary)
-      .filter(r => r.paired && r.mapped && r.mateMapped)
-      .filter(r => r.refIndex == r.mateRefIndex)
-      .filter(r => r.mapq >= this.minMapQ && r.get[Int](SAMTag.MQ.name()).exists(_ >= this.minMapQ))
-      .foreach(r => { sorter += r; sortProgress.record(r) })
+      .filter(r => (includeNonPfReads || r.pf)            || { filteredNonPf += 1; false })
```
nh13: What if you made the `count` method on `SimpleCounter` return the thing that was passed in, so you could do `.filter { r => counter.count(!r.secondary && !r.supplementary) }`?

tfenne: I think we do need to come up with a better way to do this, but I don't think that's it.  I'm going to merge as is, and think on this some more.

PR #309, opened 2017-11-10
Add an option to specify a minimum UMI length.

nh13:
This will filter out any UMIs less than this length, and will truncate the remaining UMIs in a cluster to the minimum UMI length in that cluster.

Will need to be fixed up after https://github.com/fulcrumgenomics/fgbio/pull/308 is merged :/

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -33,15 +33,17 @@ import com.fulcrumgenomics.bam.Bams
 import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamSource, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.commons.util.{LazyLogging, NumericCounter, SimpleCounter}
-import com.fulcrumgenomics.sopt.cmdline.ValidationException
 import com.fulcrumgenomics.sopt.{arg, clp}
+import com.fulcrumgenomics.umi.GroupReadsByUmi._
 import com.fulcrumgenomics.util.Metric.{Count, Proportion}
 import com.fulcrumgenomics.util.Sequences.countMismatches
 import com.fulcrumgenomics.util._
+import enumeratum.EnumEntry
 import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
 import htsjdk.samtools._
-import htsjdk.samtools.util.StringUtil
+import htsjdk.samtools.util.SequenceUtil
 
+import scala.collection.immutable.IndexedSeq
 import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 
@@ -255,7 +257,7 @@ object GroupReadsByUmi {
 
     /** Takes a UMI of the form "A-B" and returns "B-A". */
     def reverse(umi: Umi): Umi = umi.indexOf('-') match {
-      case -1 => throw new IllegalStateException(s"UMI ${umi} is not a paired UMI.")
+      case -1 => throw new IllegalStateException(s"UMI $umi is not a paired UMI.")
       case i  =>
         val first  = umi.substring(0, i)
         val second = umi.substring(i+1, umi.length)
@@ -331,6 +333,31 @@ case class TagFamilySizeMetric(family_size: Int,
                                var fraction: Proportion = 0,
                                var fraction_gt_or_eq_family_size: Proportion = 0) extends Metric
 
+/** The strategies implemented by [[GroupReadsByUmi]] to identify reads from the same source molecule.*/
+sealed trait Strategy extends EnumEntry {
+  def newStrategy(edits: Int): UmiAssigner
+}
+object Strategy extends FgBioEnum[Strategy] {
+  def values: IndexedSeq[Strategy] = findValues
+  /** Strategy to only reads with identical UMI sequences are grouped together. */
+  case object Identity extends Strategy {
+    def newStrategy(edits: Int = 0): UmiAssigner = {
+      require(edits == 0, "Edits should be zero when using the identity UMI assigner.")
+      new IdentityUmiAssigner
+    }
+  }
+  /** Strategy to cluster reads into groups based on mismatches between reads in clusters. */
+  case object Edit extends Strategy { def newStrategy(edits: Int): UmiAssigner = new SimpleErrorUmiAssigner(edits) }
+  /** Strategy based on the directed adjacency method described in [umi_tools](http://dx.doi.org/10.1101/051755)
+    * that allows for errors between UMIs but only when there is a count gradient.
+    */
+  case object Adjacency extends Strategy { def newStrategy(edits: Int): UmiAssigner = new AdjacencyUmiAssigner(edits) }
+  /** Strategy similar to the [[Adjacency]] strategy similar to adjacency but for methods that produce template with a
+    * pair of UMIs such that a read with A-B is related to but not identical to a read with B-A.
+    */
+  case object Paired extends Strategy { def newStrategy(edits: Int): UmiAssigner = new PairedUmiAssigner(edits)}
+}
+
 @clp(group=ClpGroups.Umi, description =
   """
     |Groups reads together that appear to have come from the same original molecule. Reads
@@ -375,6 +402,12 @@ case class TagFamilySizeMetric(family_size: Int,
     |
     |`edit`, `adjacency` and `paired` make use of the `--edits` parameter to control the matching of
     |non-identical UMIs.
+    |
+    |By default, all UMIs must be the same length. If `--min-umi-length=len` is specified then reads that have a UMI
+    |shorter than `len` will be discarded, and when comparing UMIs of different lengths, the first len bases will be
+    |compared, where `len` is the length of the shortest UMI. The UMI length is the number of [ACGT] bases in the UMI
+    |(i.e. does not count dashes and other non-ACGT characters). This option is not implemented for reads with UMI pairs
+    |(i.e. using the paired assigner).
   """
 )
 class GroupReadsByUmi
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -30,18 +30,22 @@ import java.util.concurrent.atomic.AtomicLong
 
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.bam.Bams
+import com.fulcrumgenomics.bam.SplitType.findValues
```
tfenne: This import is fishy.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -378,18 +404,17 @@ class GroupReadsByUmi
   @arg(flag='T', doc="The output tag for UMI grouping.") val assignTag: String = "MI",
   @arg(flag='m', doc="Minimum mapping quality.")         val minMapQ: Int      = 30,
   @arg(flag='n', doc="Include non-PF reads.")            val includeNonPfReads: Boolean = false,
-  @arg(flag='s', doc="The UMI assignment strategy; one of 'identity', 'edit', 'adjacency' or 'paired'.") val strategy: String,
-  @arg(flag='e', doc="The allowable number of edits between UMIs.") val edits: Int = 1
+  @arg(flag='s', doc="The UMI assignment strategy.") val strategy: Strategy,
```
tfenne: This is going to create an annoying compatibility issue, since the tool previously took all-lower-case values for strategy, and now it will only accept title-case.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -368,6 +388,12 @@ case class TagFamilySizeMetric(family_size: Int,
     |
     |`edit`, `adjacency` and `paired` make use of the `--edits` parameter to control the matching of
     |non-identical UMIs.
+    |
+    |If the `--min-umi-length` is not given, then all UMIs should have the same length.  If given, then reads that have
```
tfenne: Suggested wording:

By default, all UMIs must be the same length.  If `--min-umi-length=len` is specified then reads that have a UMI shorter than `len` will be discarded, and when comparing UMIs of different lengths, the first `len` bases will be compared, where `len` is the length of the shortest UMI.  The UMI length is the number of `[ACGT]` bases in the UMI (i.e. does not count dashes and other non-ACGT characters).  For reads with UMI pairs (i.e. using the paired assigner), the length is the combined length of the UMIs for the given read (ex. "AAAA-GGGG" has length eight).

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -324,6 +328,22 @@ case class TagFamilySizeMetric(family_size: Int,
                                var fraction: Proportion = 0,
                                var fraction_gt_or_eq_family_size: Proportion = 0) extends Metric
 
+sealed trait Strategy extends EnumEntry {
```
tfenne: scaladoc please?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -378,18 +404,17 @@ class GroupReadsByUmi
   @arg(flag='T', doc="The output tag for UMI grouping.") val assignTag: String = "MI",
   @arg(flag='m', doc="Minimum mapping quality.")         val minMapQ: Int      = 30,
   @arg(flag='n', doc="Include non-PF reads.")            val includeNonPfReads: Boolean = false,
-  @arg(flag='s', doc="The UMI assignment strategy; one of 'identity', 'edit', 'adjacency' or 'paired'.") val strategy: String,
-  @arg(flag='e', doc="The allowable number of edits between UMIs.") val edits: Int = 1
+  @arg(flag='s', doc="The UMI assignment strategy.") val strategy: Strategy,
+  @arg(flag='e', doc="The allowable number of edits between UMIs.") val edits: Int = 1,
+
+  @arg(flag='l', doc= """The minimum UMI length. If not specified then all UMIs should have the same length,
```
tfenne: s/should/must/

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -324,6 +328,22 @@ case class TagFamilySizeMetric(family_size: Int,
                                var fraction: Proportion = 0,
                                var fraction_gt_or_eq_family_size: Proportion = 0) extends Metric
 
+sealed trait Strategy extends EnumEntry {
+  def strategy(edits: Int): UmiAssigner
```
tfenne: Maybe this method should be `assigner` or `newAssigner`?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -411,6 +436,13 @@ class GroupReadsByUmi
       .filter(r => r.paired && r.mapped && r.mateMapped)
       .filter(r => r.refIndex == r.mateRefIndex)
       .filter(r => r.mapq >= this.minMapQ && r.get[Int](SAMTag.MQ.name()).exists(_ >= this.minMapQ))
+      .filter { r =>
+        this.minUmiLength.forall { l =>
```
tfenne: A few thoughts:
1) Perhaps we should add another filter here that just rejects reads that are missing UMIs?  Then everywhere else we can just use `rec[String](rawTag)`?
2) Post the other PR, this should probably count the rejected number of reads and log if > 0 were rejected
3) How about just counting non `-` instead?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -482,6 +514,30 @@ class GroupReadsByUmi
     }
   }
 
+  /** When a minimum UMI length is specified, truncates all the UMIs to the length of the shortest UMI.  For the paired
+    * assigner, truncates the first UMI and second UMI sepeartely.*/
+  private def truncateUmis(umis: Seq[Umi]): Seq[Umi] = this.minUmiLength match {
+    case None => umis
+    case Some(_minUmiLength) =>
```
tfenne: I think having `minUmiLength` and `_minUmiLength` is confusing.  How about just `len` or `minLength`?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -482,6 +514,30 @@ class GroupReadsByUmi
...
+      this.assigner match {
+        case _: PairedUmiAssigner => Seq.empty
```
tfenne: I'm not even sure we should support this for `paired`. It's so much more complicated, and even now this doesn't (I think) guarantee that the resulting pairs have len > min length.  E.g. if min length is 8 and you get `ACGTA-ACG` and `ACG-ACGGG` I think you'll reduce both to `ACG-ACG`.  Also, it's necessary that the left and right UMIs be the same length for duplex to work.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -482,6 +514,30 @@ class GroupReadsByUmi
...
+            val p = umi.split('-')
+            (p.head, p.last)
+          }
+          val minUmiLengthLeft  = pairs.map(_._1.length).min
+          val minUmiLengthRight = pairs.map(_._2.length).min
+          pairs.map { case (left, right) =>
+            left.take(minUmiLengthLeft) + "-" + right.take(minUmiLengthRight)
+          }
+        case _ =>
+          val minUmiLength = umis.map(_.length).min
```
tfenne: Ah!  A third `[_]minUmiLength`.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: Have you even tried it? It should be case insensitive!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: Yup I did the best I could.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: Yup. I was waiting on your PR being merged.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: Yeah, it doesn't belong here.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: @tfenne I am torn as well, I'll update to remove support in the pairedumiassigner, which I think will simplify things a bit.

nh13:
@tfenne care to take a look.  I made all your suggested changes, including taking out support for min-umi-length with the paired assigner.  I also added another commit to test (and prove) that specifying the value for the strategy command line option is case insensitive (by design ;)).

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Huh, TIL.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -411,7 +444,7 @@ class GroupReadsByUmi
     val sortProgress = ProgressLogger(logger, verb="Sorted")
 
     // A handful of counters for tracking reads
-    var (filteredNonPf, filteredPoorAlignment, filteredNsInUmi, kept) = (0L, 0L, 0L, 0L)
+    var (filteredNonPf, filteredPoorAlignment, filteredNsInUmi, filterUmisToOShort, kept) = (0L, 0L, 0L, 0L, 0L)
```
tfenne: filterUmisTo**O**Short -> filterUmisTooShort

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -498,6 +539,21 @@ class GroupReadsByUmi
...
+  private def truncateUmis(umis: Seq[Umi]): Seq[Umi] = this.minUmiLength match {
+    case None => umis
+    case Some(length) =>
+      this.assigner match {
+        case _: PairedUmiAssigner =>
+          throw new IllegalStateException("Cannot used the paired umi assigner when min-umi-length is defined.")
+        case _ =>
+          val minLength = umis.map(_.length).min
+          require(length <= minLength, s"Bug: UMI found that had shorter length than expected ($minLength < $length)")
+          umis.map(_.take(minLength))
```
tfenne: FYI I think _probably_ it would be more efficient here to use `_.substring(0, minLength)` as that's a direct method on String and won't go through all the collections shenanigans.

nh13:
Merging once tests pass.

bounlu:
Does the length in `--min-umi-length` include the variable part that is skipped i.e. `S`? For example, should I specify 5 or seven if my structure is `5M2S+T`?

nh13:
@bounlu No, it should only include what's contained in the raw (RX) tag.  The `2S` is discarded upstream. If you specify `5M` as your UMI when extracting upstream, then your UMI is not variable length though.  Can you please open a new issue that _references_ this PR rather than commenting here?

PR #311, opened 2017-11-12
Added a high class for calculating binomial probabilities at unlimited precision.

tfenne:
@nh13 I'd like to try and get this one last PR in before the release if you don't mind.  For some work I'm doing I needed a way to calculate cumulative binomial probabilities for situations where both R and commons math lost precision and ended up rounding to 0 or 1.  The implementations here are quite simple, though testing is slightly challenging given that I don't have a good way of calculating expected results for very large/small probabilities!

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
+++ b/src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
@@ -42,7 +42,7 @@ import scala.math.{BigDecimal, BigInt}
   * calculations as is practical.
   *
   * @param mc the MathContext to use for controlling precision and rounding. Changing from the default
-  *           `UNLIMITED` can lead to a loss of precision.
+  *           [[MathContext.UNLIMITED]] can lead to a loss of precision.
   */
 class BinomialDistribution(val mc: MathContext = MathContext.UNLIMITED) {
   // This array will get expanded with more factorials any time a higher n is queried
--- a/src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
+++ b/src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
@@ -1,0 +1,145 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.math
+
+import com.fulcrumgenomics.FgBioDef._
+
+import scala.math.{BigDecimal, BigInt}
+
+/**
+  * A high-precision implementation of the math for binomial probabilities.  This implementation is several
+  * times slower (though not orders of magnitude slower) than the implementation in Apache Commons Math, but
+  * retains precision throughout the computation where the Commons implementation underflows.
+  *
+  * One implementation choice to be aware of when using this implementation is that each instance of the
+  * class will calculate and cache factorials up to factorial(n) where n is the highest value of n
+  * supplied to any call to [[probability()]] or [[cumulativeProbability()]].  Thus to ensure reasonable
+  * performance it is recommended to create one instance of this class and reuse it for as may
+  * calculations as is practical.
+  */
+class BinomialDistribution {
+  // This array will get expanded with more factorials any time a higher n is queried
+  private var factorials = Array(BigInt(0), BigInt(1))
+
+  // Constant values for 0 and 1 in BigDecimal to avoid making them all the time
+  private val Zero = BigDecimal(0)
+  private val One  = BigDecimal(1)
+
+  /** Get the factorial of n. Retrieves the value from a cached set of factorials. If the cache
+    * does not contain factorial(n) yet, then all factorials up to factorial(n) are computed
+    * and cached.
+    *
+    * @param n the number to compute the factorial of
+    * @return the factorial value as a BigInt
+    */
+  private def factorial(n: Int): BigInt = {
+    if (n > factorials.length - 1) {
+      // Expand the array
+      val oldLength = factorials.length
+      val tmp = new Array[BigInt](n+1)
+      System.arraycopy(factorials, 0, tmp, 0, oldLength)
+      factorials = tmp
+
+      // Fill in the new slots
+      forloop (from=oldLength, until=factorials.length) { i =>
+        factorials(i) = factorials(i-1) * BigInt(i)
+      }
+    }
+
+    factorials(n)
+  }
+
+  /** Computes the binomial coefficient, i.e. the number of combinations of k items given a total of
+    * n items, often phrased "n pick k".
+    *
+    * @param n The number of items being picked from
+    * @param k The number of items being picked (must be 0 <= k <= n)
+    * @return The number of combinations of k items from n
+   */
+  def coefficient(n: Int, k: Int): BigInt = {
+    require(k <= n, s"k must be <= n. n=$n, k=$k")
+    require(k >= 0, s"Can't compute coefficient for picking fewer than 0 values. n=$n, k=$k")
+
+    if (k == 0 || k == n) 1
+    else factorial(n) / (factorial(k) * factorial(n - k))
+  }
+
+  /**
+    * Calculates the probability of exactly `k` successes in `n` trials given `p` probability.
+    * Akin to `dbinom` in R.
+    *
+    * @param k The number of successful trials
+    * @param n The number of trials
+    * @param p The probability of the success in any one trial
+    * @return The probability of exactly k successes in n trials of p probability
+    */
+  def probability(k: Int, n: Int, p: Double): BigDecimal = probability(k, n, BigDecimal(p))
+
+  /**
+    * Calculates the probability of exactly `k` successes in `n` trials given `p` probability.
+    * Akin to `dbinom` in R.
+    *
+    * @param k The number of successful trials
+    * @param n The number of trials
+    * @param p The probability of the success in any one trial
+    * @return The probability of exactly k successes in n trials of p probability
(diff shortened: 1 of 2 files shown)


tfenne:
@yfarjoun - you might be interested in this following our discussion the other day.  For values of n and k I'm currently interested in ([0,300]), this implementation is ~4-6 times slower than commons math once warmed up (factorials are computed).  But still plenty fast enough - can perform tens of thousands of cumulative probability calculations per second.

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+  * A high-precision implementation of the math for binomial probabilities.  This implementation is several
+  * times slower (though not orders of magnitude slower) than the implementation in Apache Commons Math, but
+  * retains precision throughout the computation where the Commons implementation underflows.
+  *
+  * One implementation choice to be aware of when using this implementation is that each instance of the
+  * class will calculate and cache factorials up to factorial(n) where n is the highest value of n
+  * supplied to any call to [[probability()]] or [[cumulativeProbability()]].  Thus to ensure reasonable
+  * performance it is recommended to create one instance of this class and reuse it for as may
```
yfarjoun: may->many

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+  /**
+    * Calculates the probability of exactly `k` successes in `n` trials given `p` probability.
+    * Akin to `dbinom` in R.
+    *
+    * @param k The number of successful trials
+    * @param n The number of trials
+    * @param p The probability of the success in any one trial
+    * @return The probability of exactly k successes in n trials of p probability
+    */
+  def probability(k: Int, n: Int, p: BigDecimal): BigDecimal = {
```
yfarjoun: Is there an implicit conversion from BigDecimal to double? I'm assuming that the result is desired as a double

tfenne: I don't believe there is ... though I _want_ the result as a BigDecimal, as otherwise when summing many of them there may again be precision loss.

yfarjoun: Well, if there was an implicit one, it would have been OK, since cumulativeP would get the BigDecimal it needs, but a "casual" user would get a double when they need it...anyhow, it's no big deal to call doubleValue on your result....so nm.

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+  * supplied to any call to [[probability()]] or [[cumulativeProbability()]].  Thus to ensure reasonable
+  * performance it is recommended to create one instance of this class and reuse it for as may
+  * calculations as is practical.
+  */
+class BinomialDistribution {
+  // This array will get expanded with more factorials any time a higher n is queried
+  private var factorials = Array(BigInt(0), BigInt(1))
+
+  // Constant values for 0 and 1 in BigDecimal to avoid making them all the time
+  private val Zero = BigDecimal(0)
```
yfarjoun: BigDecimal already has a static ONE and ZERO.

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+  *
+  * One implementation choice to be aware of when using this implementation is that each instance of the
+  * class will calculate and cache factorials up to factorial(n) where n is the highest value of n
+  * supplied to any call to [[probability()]] or [[cumulativeProbability()]].  Thus to ensure reasonable
+  * performance it is recommended to create one instance of this class and reuse it for as may
+  * calculations as is practical.
+  */
+class BinomialDistribution {
+  // This array will get expanded with more factorials any time a higher n is queried
+  private var factorials = Array(BigInt(0), BigInt(1))
```
yfarjoun: BigInt.ZERO and BigInt.ONE exist

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+    *
+    * @param n the number to compute the factorial of
+    * @return the factorial value as a BigInt
+    */
+  private def factorial(n: Int): BigInt = {
+    if (n > factorials.length - 1) {
+      // Expand the array
+      val oldLength = factorials.length
+      val tmp = new Array[BigInt](n+1)
+      System.arraycopy(factorials, 0, tmp, 0, oldLength)
```
yfarjoun: Is it so much better to use a Array than a List? I was a little horrified by the arraycopy here... :-)

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
tfenne: For some reason, the scala wrapper of BigDecimal doesn't expose these, and it's a different (i.e. wrapper) type than java.math.BigDecimal.

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
tfenne: Same here with the type-wrapping.  Although, scala's BigInt actually internally caches values between -1024 and 1024.

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
tfenne: _So much better_ might be a stretch, but I don't understand why seeing an arraycopy horrified you?  The array is private within the class, and all it's management is internal to one method.  This is similar to using, e.g. `IndexedSeq`, except that when I need to grow the array it's a one-time operation instead of doubling each time.

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+      forloop (from=oldLength, until=factorials.length) { i =>
+        factorials(i) = factorials(i-1) * BigInt(i)
+      }
+    }
+
+    factorials(n)
+  }
+
+  /** Computes the binomial coefficient, i.e. the number of combinations of k items given a total of
+    * n items, often phrased "n pick k".
```
nh13: I have always heard it "n choose k" (and wikipedia agrees).

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+  /**
+    * Calculates the probability of exactly `k` successes in `n` trials given `p` probability.
+    * Akin to `dbinom` in R.
+    *
+    * @param k The number of successful trials
+    * @param n The number of trials
+    * @param p The probability of the success in any one trial
+    * @return The probability of exactly k successes in n trials of p probability
+    */
+  def probability(k: Int, n: Int, p: Double): BigDecimal = probability(k, n, BigDecimal(p))
```
nh13: Is there really a need for an extra method with `p` having type `Double`?

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+  def probability(k: Int, n: Int, p: BigDecimal): BigDecimal = {
+    require(p >= Zero && p <= One, s"Probability p must be between 0 and 1. p=$p.")
+    val coeff = BigDecimal(coefficient(n, k))
+    val p1 = p.pow(k)
+    val p2 = (One-p).pow(n - k)
+    coeff * p1 * p2
+  }
+
+  /**
+    * Calculates the cumulative probability of `[0, k]` successes from n trials with `p` probability
```
nh13: n -> `n` (backticks)

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+    * of success in any individual trial.
+    * Akin to `pbinom` in R.
+    *
+    * @param k the number of successes to compute the cumulative probability up to, inclusive
+    * @param n the number of trials
+    * @param p the probability of success in any single trial
+    */
+  def cumulativeProbability(k: Int, n: Int, p: Double): BigDecimal = cumulativeProbability(k, n, BigDecimal(p))
```
nh13: Ditto, is there really a need?

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+    * Calculates the cumulative probability of `[0, k]` successes from n trials with `p` probability
```
nh13: backticks on "n"

src/test/scala/com/fulcrumgenomics/math/BinomialDistributionTest.scala
```diff
@@ -0,0 +1,77 @@
...
+    val commons = new CommonsBinomial(20, 0.05)
+    Range.inclusive(0, 20).foreach { k =>
+      binom.cumulativeProbability(k, n=20, p=0.05).toDouble shouldBe commons.cumulativeProbability(k) +- 1e-10
+    }
+  }
+
+  it should "compute values where commons math runs out of precision" in {
+    val commons = new CommonsBinomial(500, 0.01)
+    commons.cumulativeProbability(250) shouldBe 1.0
+    binom.cumulativeProbability(k=250, n=500, p=1.0) < BigDecimal(1) shouldBe true
```
nh13: Ok, what am I missing?  If the probability of success is 100%, then we should always get a success, so shouldn't `binom.cumulativeProbability(k=250, n=500, p=1.0)` be `1.0`?

nh13:
```R
> pbinom(250, 500, 1)
[1] 0
```

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
tfenne: Good question.  It looks like there are implicits to go from double to BigDecimal, so I think I can kill this.

src/test/scala/com/fulcrumgenomics/math/BinomialDistributionTest.scala
tfenne: Ugh, that was a really nasty typo on my part (it should have been `p=0.01` everywhere.  And actually that was masking a bug.  I've updated the test to a) have vals at the top for n, k and p, b) use numbers from a real test case and c) included a quick computation using commons's `logProbability` and our code for summing log probs, to show why I didn't go that way.

tfenne:
@nh13 Thanks for the review.  I've pushed up a set of changes that address your comments, and fix the bug that they led me to.

I didn't go the log route initially as I wanted to be 100% sure of the precision of the results.  I've added to one test to show that for fairly reasonable values of n, k and p that yield a real cumulative probability well within the range of `double`, using common's to get the log probabilities and summing them still loses precision such that the result rounds out to 1 (or 0).  Maybe with a bunch more work there could be an impl in log space, but the BigDecimal one is guaranteed to work.

For larger values of n and k, it is considerably slower.  So where performance is critical and one is calculating many thousands of cumulative probabilities it might be smart to first calculate with commons math, and then if the result is very close to 1 or 0, to recalculate using this method.

src/main/scala/com/fulcrumgenomics/math/BinomialDistribution.scala
```diff
@@ -38,14 +40,24 @@ import scala.math.{BigDecimal, BigInt}
   * supplied to any call to [[probability()]] or [[cumulativeProbability()]].  Thus to ensure reasonable
   * performance it is recommended to create one instance of this class and reuse it for as many
   * calculations as is practical.
+  *
+  * @param mc the MathContext to use for controlling precision and rounding. Changing from the default
+  *           `UNLIMITED` can lead to a loss of precision.
```
nh13: instead of backticks, why not `[[MathContext .UNLIMITED]]`?

PR #316, opened 2017-11-14
Add in variables for which duplex consensus tags need to be reversed …

nh13:
…and reverse complemented respectively.


nh13:
@tfenne one more look since I made some non-trivial changes to `FilterConsensusReads`.

Issue #318, opened 2017-11-15
A question about the precision of doubles and floats in Metric


nh13:
@tfenne It looks like we print decimal values down to only six digits!  If we are outputting a very small double, say the error rate in duplex sequencing, then the result may be zero, when in fact it is greater than zero.  This makes me want to roll our own metric formatting utility.  We should also not forget about `Option[Double]` in the implementation.

https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/util/Metric.scala#L199-L205
https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/util/FormatUtil.java#L62
https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/util/FormatUtil.java#L46

tfenne:
I do think that we should probably change it so that it formats doubles with five significant digits.  Perhaps something like:

```scala
val bigDoubleFmt = new DecimalFormat("0.######")
val smallDoubleFmt = new DecimalFormat("0.#####E0")

if (d >= 0.001) bigDoubleFmt.format(d) else smallDoubleFormat.format(d)
```

Note that this is possible to override in an individual `Metric` class as it stands.

tfenne:
Resolved by #322

PR #322, opened 2017-11-17
Updated Metric formatting to be more efficient and handle very small doubles

tfenne:
A fix for #318.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/Metric.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/Metric.scala
@@ -237,15 +237,15 @@ trait Metric extends Product with Iterable[(String,String)] {
 
   /** Override this method to customize how values are formatted. */
   protected def formatValue(value: Any): String = value match {
-    case null      => ""
-    case None      => ""
-    case Some(x)   => formatValue(x)
+    case null           => ""
+    case None           => ""
+    case Some(x)        => formatValue(x)
     case d: Iso8601Date => d.toString
-    case d: Date   => Metric.DateFormat.synchronized { Metric.DateFormat.format(d) }
-    case f: Float  => formatValue(f.toDouble)
+    case d: Date        => Metric.DateFormat.synchronized { Metric.DateFormat.format(d) }
+    case f: Float       => formatValue(f.toDouble)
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -225,11 +228,25 @@ trait Metric extends Product with Iterable[(String,String)] {
...
 
+  /** @deprecated use [[formatValue]] instead. */
+  @deprecated("Use formatValue instead.", since="0.5.0")
+  protected def formatValues(value: Any): String = formatValue(value)
+
   /** Override this method to customize how values are formatted. */
-  protected def formatValues(value: Any): String = formatter.format(value)
+  protected def formatValue(value: Any): String = value match {
+    case null      => ""
+    case None      => ""
```
nh13: `ReflectionUtil.SpecialEmptyOrNoneToken`?

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -225,11 +228,25 @@ trait Metric extends Product with Iterable[(String,String)] {
...
+    case Some(x)   => formatValue(x)
+    case d: Iso8601Date => d.toString
+    case d: Date   => Metric.DateFormat.synchronized { Metric.DateFormat.format(d) }
+    case f: Float  => formatValue(f.toDouble)
+    case d: Double if d < 0.00001 => Metric.SmallDoubleFormat.synchronized { Metric.SmallDoubleFormat.format(d) }
+    case d: Double                => Metric.BigDoubleFormat.synchronized { Metric.BigDoubleFormat.format(d) }
+    case other     => other.toString
```
nh13: align the `=>`?

src/main/scala/com/fulcrumgenomics/util/Metric.scala
tfenne: I don't think we want that when _formatting_.  That'd result in `:none:` getting written into the text files and I think that'd be confusing.

nh13: Sorry, right, this is `Metric`.

PR #324, opened 2017-11-17
Added support for storing the original UMI in CorrectUmis.

### LINKED ISSUE #208: CorrectUmis should store the old/uncorrected UMI in a tag once the spec says which one
Requires the spec to be a bit more settled.  See:
* https://github.com/samtools/hts-specs/issues/109
* https://github.com/samtools/hts-specs/pull/119
* https://github.com/samtools/hts-specs/pull/200

tfenne:
Another little PR for you @nh13.  This time to fix #208.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
@@ -113,6 +113,10 @@ object CorrectUmis {
     |more files of UMIs with a single sequence per line using `--umi-files umis.txt more_umis.txt`.  If there
     |are multiple UMIs per template, leading to hyphenated UMI tags, the values for the fixed UMIs should
     |be single, non-hyphenated UMIs (e.g. if a record has `RX:Z:ACGT-GGCA`, you would use `--umis ACGT GGCA`).
+    |
+    |Records which have their UMIs corrected (i.e. the UMI is not identical to one of the expected UMIs but is
+    |close enough to be corrected) will by default have their original UMI stored in the `OX` tag. This can be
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -123,7 +127,8 @@ class CorrectUmis
   @arg(flag='d', doc="Minimum distance (in mismatches) to next best UMI.") val minDistance: Int,
   @arg(flag='u', doc="Expected UMI sequences.", minElements=0) val umis: Seq[String] = Seq.empty,
   @arg(flag='U', doc="File of UMI sequences, one per line.", minElements=0) val umiFiles: Seq[FilePath] = Seq.empty,
-  @arg(flag='t', doc="Tag in which UMIs are stored.") val umiTag: String = ConsensusTags.UmiBases
+  @arg(flag='t', doc="Tag in which UMIs are stored.") val umiTag: String = ConsensusTags.UmiBases,
+  @arg(flag='x', doc="If specified, don't store original UMIs upon correction.") val dontStoreOriginalUmis: Boolean = false
```
nh13: I'd remove the "If specified" part

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -198,6 +203,11 @@ class CorrectUmis
 
             // Output the corrected read
             if (matches.forall(_.matched)) {
+              // Store the original UMI if requested, and there are mismatches
```
nh13: (1) and **if** there are mismatches
(2) if both requested and there are mismatches

Issue #327, opened 2017-11-25
DemuxFastqs's quality format is incorrect


gshiba:
Two issues:
* Auto detection of quality format is not working
* Quality encoding flag `-q` is ignored

Set up:

```
$ zcat r1.fastq.gz
@M00001:001:000000000-AAAAA:1:1101:16729:1382 1:N:0:1
TCGTGACAGGAATCAAATGAAAACACTTGGT
+
1>1>11AFF?FFGGGGGGCBDGBGCFGH11B

$ zcat r2.fastq.gz
@M00001:001:000000000-AAAAA:1:1101:16729:1382 2:N:0:1
TCGTGACAGGAATCAAATGAAAACACTTGGT
+
1>1>11AFF?FFGGGGGGCBDGBGCFGH11B

$ cat samples.csv
Sample_Barcode,Sample_ID,Sample_Name,Library_ID,Description
TCGTTCGT,foo,S1,,

$ fgbio -v
Version: 0.4.0
```

Tool will automatically `Auto-detected quality format as: Standard`, but the output quality is not the same format as input.

```
$ fgbio DemuxFastqs -i r1.fastq.gz r2.fastq.gz -o demuxed --read-structures 4B+T 4B+T --metadata samples.csv
[2017/11/24 16:29:49 | DemuxFastqs | Info] Assuming input metadata file is simple CSV file.
[2017/11/24 16:29:49 | FgBioMain | Info] Executing DemuxFastqs from fgbio version 0.4.0 as gshiba@neptune on JRE 1.8.0_121-b15 with snappy
[2017/11/24 16:29:49 | DemuxFastqs | Warning] Making ambiguous determination about fastq's quality encoding; possible encodings: Standard, Solexa, Illumina.
[2017/11/24 16:29:49 | DemuxFastqs | Info] Auto-detected quality format as: Standard
[2017/11/24 16:29:50 | FgBioMain | Info] DemuxFastqs completed. Elapsed time: 0.08 minutes.

$ ll demuxed/
demux_barcode_metrics.txt  foo-S1-TCGTTCGT.bam        unmatched.bam

$ samtools view demuxed/foo-S1-TCGTTCGT.bam
M00001:001:000000000-AAAAA:1:1101:16729:1382    77      *       0       0       *       *       0       0       GACAGGAATCAAATGAAAACACTTGGT     RRbgg`gghhhhhhdcehchdghiRRc     BC:Z:TCGT-TCGT  RG:Z:foo
[...]

PR #329, opened 2017-11-28
Nh issue 327

nh13:


nh13:
@tfenne I would welcome your opinion on whether or not we should do a minor release with this bug fix.  I am not sure we have ever done one.

nh13:
#327

nh13:
@tfenne thanks for the suggestion!  I have implemented it as you described. Care to take another look?

PR #331, opened 2017-12-01
Cleaning up deprecated command line options and other methods

nh13:
@tfenne probably the only interesting commit here is 0738c59.  I want to preserve the ability to directly build a metric from a sequence of arguments, rather than always reading from a file, so I have updated it to be "more efficient" as you had done for the `write` method a while back.

I left one deprecation since it was deprecated in `0.5.0` and so I figure we should have at least one release with the deprecation notice.


src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -80,33 +80,32 @@ object Metric {
...
+        reflectiveBuilder.argumentLookup.forField(name) match {
+          case Some(arg) =>
+            val argumentValue = ReflectionUtil.constructFromString(arg.argumentType, arg.unitType, value) match {
+              case Success(v) => v
+              case Failure(thr) => throw thr
+            }
+            arg.value_=(argumentValue)
+          case None =>
+            throw new IllegalArgumentException(s"The class '${clazz.getSimpleName}' did not have a field with name '$name'.")
+        }
```
nh13: Just a thought, but what if not all the arguments are given in a sequence of tuples, will we use the value from the previous sequence?

```
scala> class Foo(str: String, int: Int = 4)
scala> Metric.build[Foo](Seq(("str", "value1"), ("int", 2)), Seq(("str", "value2")))
res0: Seq(Foo("str", 2), Foo("value2", 2)))
```
but the result should be: `Foo("str", 2), Foo("value2", 4)))

tfenne: I agree, this seems like a problem.  For loading from a file, we'll have the same set of columns each time. But for any other use case that can't be guaranteed.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -117,33 +116,17 @@ object Metric {
     val parser = new DelimitedDataParser(lines=lines, delimiter=Delimiter, ignoreBlankLines=false, trimFields=true)
     val names  = parser.headers.toIndexedSeq
 
-    val clazz             = ReflectionUtil.typeTagToClass[T]
```
tfenne: I really don't like the changes to this method.  
Firstly, it is going to make it considerably slower as it's not generating an intermediate set of tuples for every row it reads from the file.  Secondly I think this is making the major use case (reading back a metrics file) more complicated to fit in with a use case I still don't understand the need for ("reading" metrics from a `Seq[(String,String)]`.

If the latter is the minority use case, why not implement that in terms of the existing `read` method?

src/main/scala/com/fulcrumgenomics/util/Metric.scala
nh13: Fixed

nh13:
I removed the `build` method for now, but you may see a PR with one in the future!

PR #332, opened 2017-12-02
Adding the FlagStat tool to reproduce SAMTools' flagstat command.

nh13:
@tfenne I wanted to see how hard this was to implement.  Obviously it's the testing that takes the longest.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/FlagStat.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/FlagStat.scala
@@ -1,0 +1,161 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.bam
+
+import com.fulcrumgenomics.FgBioDef.PathToBam
+import com.fulcrumgenomics.bam.api.{SamRecord, SamSource}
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/bam/FlagStat.scala
```diff
@@ -0,0 +1,161 @@
...
+  * @param read_two the number of second of pair reads/records.
+  * @param duplicate the number of duplicate reads/records.
+  * @param inter_chrom the number of mapped paired reads/records whose mate maps to a different chromosome.
+  * @param inter_chrom_mapq5 the number of mapped paired reads/records with mapping quality at least five whose mate maps to a different chromosome.
+  * @param secondary the number of secondary reads/records.
+  * @param supplementary the number of supplementary reads/records.
+  */
+case class FlagStatMetric
+(pf: Boolean,
+ var reads: Long = 0,
```
tfenne: Perhaps records?

src/main/scala/com/fulcrumgenomics/bam/FlagStat.scala
```diff
@@ -0,0 +1,161 @@
...
+ var read_two: Long = 0,
+ var duplicate: Long = 0,
+ var inter_chrom: Long = 0,
+ var inter_chrom_mapq5: Long = 0,
+ var secondary: Long = 0,
+ var supplementary: Long = 0
+) extends Metric
+
+@clp(group = ClpGroups.SamOrBam, description=
+  """|Collects metrics from the FLAG field in a SAM or BAM file.
```
tfenne: If we're going to retain this we probably want a more detailed description.

src/main/scala/com/fulcrumgenomics/bam/FlagStat.scala
```diff
@@ -0,0 +1,161 @@
...
+  """)
+class FlagStat
+( @arg(flag='i', doc="Input SAM or BAM file.") val input: PathToBam,
+  @arg(flag='o', doc="The output file in which metrics will be stored.") val output: Option[PathToBam] = None
```
tfenne: `PathToBam` -> `FilePath`.
Also maybe mention what happens if you don't provide it?

src/main/scala/com/fulcrumgenomics/bam/FlagStat.scala
```diff
@@ -0,0 +1,161 @@
...
+  }
+}
+
+object FlagStat {
+  /** Updates the metric using the given record. */
+  private[bam] def updateMetric(metric: FlagStatMetric, rec: SamRecord): Unit = {
+    require(metric.pf == rec.pf, "The metric PF value did not match the record's pf value")
+    metric.reads += 1
+    if (rec.secondary) metric.secondary += 1
+    else if (rec.supplementary) metric.supplementary += 1
```
tfenne: Why is this and the next line an `else if`.  A record can be both secondary and supplementary.  And it can definitely be secondary or supplementary and also paired.

src/main/scala/com/fulcrumgenomics/bam/FlagStat.scala
```diff
@@ -0,0 +1,161 @@
...
+    else if (rec.paired) {
+      metric.paired += 1
+      if (rec.properlyPaired && rec.mapped) metric.proper_paired +=1
+      if (rec.firstOfPair) metric.read_one += 1
+      if (rec.secondOfPair) metric.read_two += 1
+      if (rec.mapped && rec.mateUnmapped) metric.singleton += 1
+      if (rec.mapped && rec.mapped) {
```
tfenne: You have `rec.mapped` on both sides of the `&&`.

src/main/scala/com/fulcrumgenomics/bam/FlagStat.scala
```diff
@@ -0,0 +1,161 @@
...
+          if (rec.mapq >= 5) metric.inter_chrom_mapq5 += 1
+        }
+      }
+    }
+    if (rec.mapped) metric.mapped += 1
+    if (rec.duplicate) metric.duplicate += 1
+  }
+
+  /** Creates a multi-line string equivalent to the output of SAMTools flagstat. */
+  private[bam] def formatInSamtools(metrics: Map[Boolean, FlagStatMetric]): String = {
```
tfenne: Feels odd to pass a Map[Boolean].  Why not just pass two parameters?

PR #333, opened 2017-12-05
WIP: Use MSA to call duplex consensus reads.

nh13:
Three options are added to `CallDuplexConsensusReads`:

1. `--use-msa`: Use multiple-sequence-alignment when the # of records filtered due to minority alignments is too high.  
2. `--max-filter-minority-fraction`: The maximum fraction of reads filtered due to minority cigar before using multiple sequence alignment.
3. `--msa--command`: The multiple sequence alignment command to use; the input file (FASTA) will be appended.  Some example uses of the `--msa--command` option:
```
--msa-command "clustalo --outt=clustal --in"
--msa-command "mafft --thread 8 --anysymbol --clustalout --bl 62 --op 1.53 --ep 0.123 --reorder --retree 2 --treeout --maxiterate 2"
```
The idea is that when the cigars do not match up well, we can use multiple-sequence alignment to do a better job.  Since MSA is expensive to perform, only do it when a non-trivial fraction of reads are discarded.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
@@ -25,6 +25,7 @@
 package com.fulcrumgenomics.umi
 
 import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.bam.Bams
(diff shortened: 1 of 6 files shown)


nh13:
Closing, lack of progress.

PR #334, opened 2017-12-06
ClipBam optionally outputs metrics

nh13:
... and some API changes.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
@@ -30,11 +30,14 @@ import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamSource, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.commons.util.LazyLogging
 import com.fulcrumgenomics.sopt.{arg, clp}
-import com.fulcrumgenomics.util.{Io, ProgressLogger}
+import com.fulcrumgenomics.util.{Io, Metric, ProgressLogger}
+import enumeratum.EnumEntry
 import htsjdk.samtools.SAMFileHeader.SortOrder
+import htsjdk.samtools.SamPairUtil
 import htsjdk.samtools.SamPairUtil.PairOrientation
-import htsjdk.samtools._
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker
+
+import scala.collection.immutable.IndexedSeq
 
 @clp(group = ClpGroups.SamOrBam, description=
   """
@@ -59,6 +62,7 @@ import htsjdk.samtools.reference.ReferenceSequenceFileWalker
 class ClipBam
 ( @arg(flag='i', doc="Input SAM or BAM file of aligned reads in coordinate order.") val input: PathToBam,
   @arg(flag='o', doc="Output SAM or BAM file.") val output: PathToBam,
+  @arg(flag='m', doc="Optional output of clipping metrics.") val metrics: Option[FilePath] = None,
   @arg(flag='r', doc="Reference sequence fasta file.") val ref: PathToFasta,
   @arg(flag='s', doc="Soft clip reads instead of hard clipping.") val softClip: Boolean = false,
   @arg(flag='a', doc="Automatically clip extended attributes that are the same length as bases.") val autoClipAttributes: Boolean = false,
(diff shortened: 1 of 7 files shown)


src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -30,20 +30,23 @@ import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamSource, SamWriter}
...
+
 @clp(group = ClpGroups.SamOrBam, description=
   """
     |Clips reads from the same template. Ensures that at least N bases are clipped from any end of the read (i.e.
     |R1 5' end, R1 3' end, R2 5' end, and R2 3' end).  Optionally clips reads from the same template to eliminate overlap
     |between the reads.  This ensures that downstream processes, particularly variant calling, cannot double-count
     |evidence from the same template when both reads span a variant site in the same template.
     |
-    |Clipping overlapping reads is only performed on `FR` read pairs, and is implemented by clipping approximately half
+    |Clipping overlapping reads is only performed on `FR` reads, and is implemented by clipping approximately half
```
tfenne: I think this should still say "read pairs", right?  since `FR` reads doesn't quite make sense.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -112,36 +121,160 @@ class ClipBam
       out += rec
     }
 
+    this.metrics.foreach { path =>
```
tfenne: I think this might be clearer as:
```scala
(this.metrics, this.metricsMap) match {
  case (Some(path), Some(ms)) =>
    ....
  case _ => Unit
}
```

I had to stare for a minute to remember that metricMap is an `Option[Map[]]` and that's why you needed the foreach.  Also, having the second foreach use `metrics` as it's parameter name is confusing since you have an instance variable called `metrics`.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -112,36 +121,160 @@ class ClipBam
...
-
-    if (clipOverlappingReads && r1.mapped && r2.mapped && r1.refIndex == r2.refIndex && r1.pairOrientation == PairOrientation.FR) {
-      val (f,r) = if (r1.negativeStrand) (r2, r1) else (r1, r2)
-      clipOverlappingReads(f, r)
+    val numReadOneFivePrime  = this.clipper.clip5PrimeEndOfRead(r1, readOneFivePrime)
+    val numReadOneThreePrime = this.clipper.clip3PrimeEndOfRead(r1, readOneThreePrime)
+    val numReadTwoFivePrime  = this.clipper.clip5PrimeEndOfRead(r2, readTwoFivePrime)
+    val numReadTwoThreePrime = this.clipper.clip3PrimeEndOfRead(r2, readTwoThreePrime)
+
+    val (numOverlappingBasesReadOne, numOverlappingBasesReadTwo) = {
```
tfenne: I think this does the wrong thing now.  You call `clipOverlappingReads(f,r)` not `clipOverlappingReads(r1,r2)`, so I think the tuple it returns needs to be flipped half the time.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -112,36 +121,160 @@ class ClipBam
...
+      if (clipOverlappingReads && r1.mapped && r2.mapped && r1.refIndex == r2.refIndex && r1.pairOrientation == PairOrientation.FR) {
```
tfenne: Also, I think this can be replaced with just `clipOverlappingReads && rec.isFrPair` now :)

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -112,36 +121,160 @@ class ClipBam
...
+        numThreePrime       = numReadTwoThreePrime,
+        numOverlappingBases = numOverlappingBasesReadTwo
+      )
     }
   }
 
   /** Clips overlapping reads, where both ends of the read pair are mapped to the same chromosome, and in FR orientation. */
-  protected def clipOverlappingReads(f: SamRecord, r: SamRecord): Unit = {
+  protected def clipOverlappingReads(f: SamRecord, r: SamRecord): (Int, Int) = {
+    var numOverlappingBasesReadOne: Int = 0
```
tfenne: Technically these are more like `numClippedFromReadOne` and `numClippedFromReadTwo` right?  Since it can be more than the overlap in some cases.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -112,36 +121,160 @@ class ClipBam
...
-      this.clipper.clip3PrimeEndOfAlignment(f, firstHalf)
-      this.clipper.clip3PrimeEndOfAlignment(r, secondHalf)
+      numOverlappingBasesReadOne += this.clipper.clip3PrimeEndOfAlignment(f, firstHalf)
+      numOverlappingBasesReadTwo += this.clipper.clip3PrimeEndOfAlignment(r, secondHalf)
+    }
+    (numOverlappingBasesReadOne, numOverlappingBasesReadTwo)
+  }
+}
+
+sealed trait ReadType extends EnumEntry
```
tfenne: Not that we need to do this now, but I feel like we've done similar things elsewhere so the next time we need to output things by R1/R2/Frag we should try and remember this is here and elevate it.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
nh13: Search and replace error.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
nh13: fixed

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
nh13: 👍

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
nh13: This method only clips the overlap, so it is exactly the bases clipped due to overlapping reads.

Issue #319, opened 2017-11-16
ClipBam (and SamRecordClipper) don't hard clip if there is already >= n soft clipped bases


tfenne:
ClipBam does not do the right thing when hard clipping, if there is already sufficient clipping present as soft-clipping.  I was trying to use ClipBam to generate a 2x75bp BAM from a BAM with 2x151bp reads by doing:

`fgbio ClipBam -i $in -o $out -r $ref --read-one-three-prime=76 --read-two-three-prime=76`

Most of the reads were clipped correctly - a `76H` was added to the cigar at the right place and the bases and quals were cut.  However, any reads that had >= `76S` at the appropriate end were left untouched.  E.g a positive strand mapping with `100S51M` would be left with the same cigar and full length bases and quals.

I think in this case it should be "promoted" to hard clipping, with a cigar of `76H24S50M` and the bases/quals adjusted accordingly.

nh13:
@tfenne I think there needs to be some updates to `SamRecordClipper` to handle this looking at it.

nh13:
Fixed

ggydush:
I'm seeing a similar issue on Fgbio versions 0.5.0 and 0.6.1 when trying to clip 50 base pairs from 146 base pair reads (primary alignment and FR). Reads with long soft clippings end up having their cigars removed, mapping quality set to 0, and are not hard clipped (read is still 146bp).

Command:
`fgbio ClipBam -i $in -o $out -r $ref --read-one-three-prime=50 --read-two-three-prime=50`

Would be happy to supply an example bam if helpful.

nh13:
An example would be great!

ggydush:
@nh13

That makes sense - I guess I was mostly confused by the reads not being hard clipped in cases like these. Thanks for the explanation!

nh13:
@ggydush if there is a specific application where you'd prefer to have all the bases hard clipped rather than bases retained (but as an unmapped read) read let us know.  I'd prefer not to add the support to keep the tool simple, but I'd be open if you have compelling use case.  Feel free to open a new issue if so.

PR #336, opened 2017-12-07
Fixed bug with soft-clipping not being coverted to hard-clipping in S…

tfenne:
…amRecordClipper.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
@@ -114,16 +114,7 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean) 
       rec.bases = bases
       rec.quals = quals
       cleanupClippedRecord(rec)
-
-      if (mode == ClippingMode.Hard && readBasesClipped > 0 && autoClipAttributes) {
-        val oldLength = oldBases.length
-        val shortenBy = oldLength - bases.length
-        rec.attributes.foreach { 
-          case (tag, s: String)   if s.length == oldLength => rec(tag) =  s.drop(shortenBy)
-          case (tag, a: Array[_]) if a.length == oldLength => rec(tag) =  a.drop(shortenBy)
-          case _ => Unit
-        }
-      }
+      clipExtendedAttributes(rec, oldBases.length - rec.bases.length, fromStart=true)
       readBasesClipped
     }
   }
@@ -168,16 +159,7 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean) 
       rec.bases = bases.reverse
       rec.quals = quals.reverse
       cleanupClippedRecord(rec)
-
-      if (mode == ClippingMode.Hard && readBasesClipped > 0 && autoClipAttributes) {
-        val oldLength = oldBases.length
-        val newLength = bases.length
-        rec.attributes.foreach {
-            case (tag, s: String)   if s.length == oldLength => rec(tag) = s.take(newLength)
-            case (tag, a: Array[_]) if a.length == oldLength => rec(tag) = a.take(newLength)
-            case _  => Unit
-        }
-      }
+      clipExtendedAttributes(rec, oldBases.length - rec.bases.length, fromStart=false)
       readBasesClipped
     }
   }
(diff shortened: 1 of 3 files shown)


tfenne:
Fixes #319

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -341,6 +325,84 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+        case (tag, a: Array[_]) if a.length == oldLength => rec(tag) = if (fromStart) a.drop(remove) else a.take(newLength)
+        case _  => Unit
+      }
+    }
+  }
+
+  /**
+    * Ensures sufficient masking or hard clipping exists on reads that may have soft-clipping. The read will
+    * only be altered if:
+    *   1. ClippingMode is Hard of SoftWithMask
```
nh13: of -> or

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -341,6 +325,84 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+    *   2. The read is already clipped at the appropriate end
+    *   3. Soft-clipping exist within the first `length` bases of the appropriate end
+    *
+    * If all of those conditions are met, the soft-clipping will be upgraded to masking or hard clipping
+    * up until length bases are clipped or masked. E.g. if the incoming cigar is `10H10S80M` and `length` is
+    * 15, the resulting cigar will be `15H5S80M`.
+    */
+  protected def upgradeClipping(rec: SamRecord, length: Int, fromStart: Boolean): Unit = if (mode != ClippingMode.Soft) {
+    require(length > 0, "Length must be >= 0")
```
nh13: condition is `>` but message is `>=`

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -341,6 +325,84 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+    * only be altered if:
+    *   1. ClippingMode is Hard of SoftWithMask
+    *   2. The read is already clipped at the appropriate end
+    *   3. Soft-clipping exist within the first `length` bases of the appropriate end
+    *
+    * If all of those conditions are met, the soft-clipping will be upgraded to masking or hard clipping
+    * up until length bases are clipped or masked. E.g. if the incoming cigar is `10H10S80M` and `length` is
+    * 15, the resulting cigar will be `15H5S80M`.
+    */
+  protected def upgradeClipping(rec: SamRecord, length: Int, fromStart: Boolean): Unit = if (mode != ClippingMode.Soft) {
```
nh13: Can this method return `Int` for #334? Namely, zero :P

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -341,6 +325,84 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+    require(length > 0, "Length must be >= 0")
+    def iter = if (fromStart) rec.cigar.iterator else rec.cigar.reverseIterator
+    val hardClipped = iter.takeWhile(_.operator == Op.H).map(_.length).sum
+    val softClipped = iter.dropWhile(_.operator == Op.H).takeWhile(_.operator == Op.S).map(_.length).sum
+
+    if (hardClipped < length && softClipped > 0) {
```
nh13: How about just a quick comment for our future selves: `If not all the requested clipping is currently hard-clipped, and some soft-clipping exists.`

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -341,6 +325,84 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+        rec.quals = quals
+      }
+      else {
+        rec.cigar = Cigar(elems.reverse)
+        rec.bases = bases.reverse
+        rec.quals = quals.reverse
+      }
+
+      clipExtendedAttributes(rec, lengthToUpgrade, fromStart=fromStart)
+    }
```
nh13: Oh god, it took a long while to read through this and understand all the book-keeping, good work.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: I think it's odd to have it return zero just for the sake of it.  But the clipping methods that call it will return 0.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: This caused a lot of test failures in `ClipBam`.  I'm removing the requirement and changing the `if` guard above to include `&& length > 0` instead.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: Done!

PR #337, opened 2017-12-12
ClipBam can convert all clipping to hard-clipping

nh13:


src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -248,6 +248,30 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+    if (rec.unmapped || this.mode != ClippingMode.Hard) (0, 0) else {
+      val numBasesClippedStart = {
+        val clippingElems = rec.cigar.elems.takeWhile(_.operator.isClipping)
+        val numSoft       = clippingElems.filter(_.operator == Op.S).map(_.length).sum
+        if (numSoft > 0) this.clipStartOfRead(rec, clippingElems.map(_.length).sum)
+        numSoft
+      }
+      val numBasesClippedEnd = {
+        val clippingElems = rec.cigar.reverseIterator.takeWhile(_.operator.isClipping).toSeq
+        val numSoft       = clippingElems.filter(_.operator == Op.S).map(_.length).sum
```
nh13: @tfenne I would be curious about your opinion in regards to calling map/filter/sum on `IndexedSeq` relative to performance.  In your experience, is it better to convert them to iterators or streams?

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -66,6 +67,7 @@ class ClipBam
   @arg(flag='r', doc="Reference sequence fasta file.") val ref: PathToFasta,
   @arg(flag='s', doc="Soft clip reads instead of hard clipping.") val softClip: Boolean = false,
   @arg(flag='a', doc="Automatically clip extended attributes that are the same length as bases.") val autoClipAttributes: Boolean = false,
+  @arg(flag='H', doc="Convert all soft-clipping, including existing soft-clipping, to hard-clipping (requires `--soft-clip=true`).") val convertSoftClipToHardClip: Boolean = false,
```
tfenne: This is super confusing.  What this means is that if I just want to run the tool to convert existing soft-clips to hard-clips I have to specify `--soft-clip=true --convert-soft-clip-to-hard-clip=true`, which seems unnecessary.

Secondly, it's also really confusing (and implies some knowledge of the ordering in the code) that if I want to do some clipping and also convert existing clips to hard clips, I have to `--soft-clip=true ... --convert-soft-clip-to-hard-clip=true`, to get _hard_ clipping, even though my command line explicitly says soft clipping.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -102,6 +106,8 @@ class ClipBam
         case _ => Unit
       }
 
+      if (this.convertSoftClipToHardClip) template.allReads.foreach { r => this.clipper.toHardClip(r) }
```
tfenne: So I'm pretty sure this does nothing since in `SamRecordClipper` you return immediately if mode != Hard, and you require above that mode == Soft.

tfenne: This would also argue for needing a test for the changes to ClipBam.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -248,6 +248,30 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
     else clipEndOfRead(rec, numberOfBasesToClip)
   }
 
+  /** Converts all soft-clipping to hard-clipping.  Does not convert unless the clipping mode is [[ClippingMode.Hard]].
+    * Calculates any clipping required and delegates to [[clipStartOfRead()]] and [[clipEndOfRead()]].
+    * @param rec the record to be clipped
+    * @return the number of bases converted to hard-clipping from soft-clipping at the start and end of the read
+    *         respectively.
+    */
+  def toHardClip(rec: SamRecord): (Int, Int) = {
```
tfenne: I really dislike how a) this method hard-codes the clipping type given that the rest of this class clips based on `ClippingMode`, and then b) refuses to do anything unless mode == Hard.

tfenne: I would suggest calling this method something like `upgradeAllClipping`, and letting it do it's thing in either Hard or SoftWithMask mode.

tfenne: Also, you return `(Int, Int)` here, but don't use that in `ClipBam`.  Is that intentional?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -248,6 +248,30 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+    if (rec.unmapped || this.mode != ClippingMode.Hard) (0, 0) else {
+      val numBasesClippedStart = {
+        val clippingElems = rec.cigar.elems.takeWhile(_.operator.isClipping)
+        val numSoft       = clippingElems.filter(_.operator == Op.S).map(_.length).sum
+        if (numSoft > 0) this.clipStartOfRead(rec, clippingElems.map(_.length).sum)
```
tfenne: I think I would just call directly to `upgradeClipping(rec, clippingElems.map(_.length).sum, fromStart=true)`.  I think that's a bit more transparent.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
tfenne: I've started to tend towards `fold(0)(_ + _.length)` vs. `.map().sum`.  I don't have any solid empirical evidence that it's faster, but since it won't create any intermediate iterators or collections, I think it should be.  Then again, these are all 0-2 element things to being with so probably no big deal.

nh13:
@tfenne I have update this PR significantly:
1. `ClipBam` now has a `--clipping-mode` parameter, which applies the given clipping mode.  I made `--soft-clip` deprecated.
2. The  `ClipBam` `--upgrade-clipping` parameter will upgrade all clipping to the current mode prior to applying any other clipping parameter.
3. Added in tests in `CilpBam` for 2.
4. Upgrade the `ClippingMode` to be an `FgBioEnum` (yay) so we can use it as a type for a command line argument.
5. `upgradeAllClipping` in `SamRecordClipper` will upgrade all clipping to the current mode, as long as the current mode is higher/more-stringent (not sure what's the right word here).  For example, we don't upgrade from hard-clipping to soft-clipping (since we have lost the bases!).

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -58,14 +56,26 @@ import scala.collection.immutable.IndexedSeq
     |pair information can be reset across all reads for the template.  Post-clipping the reads are
     |resorted into coordinate order, any existing `NM`, `UQ` and `MD` tags are repaired, and the output is
     |written in coordinate order.
+    |
+    |Three clipping modes are supported:
+    |1. `Soft` - soft-clip the bases and qualities.
+    |2. `SoftWithMask` - soft-clip and mask the bases and qualities (make bases Ns and qualities the minimum).
+    |3. `Hard` - hard-clip the bases and qualities.
+    |
+    |The `--upgrade-clipping` parameter will convert all existing clipping in the input to given more stringent mode:
```
nh13: the given

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -58,14 +56,26 @@ import scala.collection.immutable.IndexedSeq
...
 ( @arg(flag='i', doc="Input SAM or BAM file of aligned reads in coordinate order.") val input: PathToBam,
   @arg(flag='o', doc="Output SAM or BAM file.") val output: PathToBam,
   @arg(flag='m', doc="Optional output of clipping metrics.") val metrics: Option[FilePath] = None,
   @arg(flag='r', doc="Reference sequence fasta file.") val ref: PathToFasta,
-  @arg(flag='s', doc="Soft clip reads instead of hard clipping.") val softClip: Boolean = false,
+  @deprecated("Use clipping-mode instead.", since="0.4.0")
+  @arg(flag='s', doc="Soft clip reads instead of hard clipping.", mutex=Array("clippingMode")) val softClip: Boolean = false,
+  @arg(flag='c', doc="The type of clipping to perform.", mutex=Array("softClip")) var clippingMode: Option[ClippingMode] = None,
   @arg(flag='a', doc="Automatically clip extended attributes that are the same length as bases.") val autoClipAttributes: Boolean = false,
+  @arg(flag='H', doc="Upgrade all existing clipping in the input to the given mode prior to clipping.") val upgradeClipping: Boolean = false,
```
nh13: the given mode prior to clipping -> the given clipping mode prior to applying any other clipping criteria.

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -58,14 +56,26 @@ import scala.collection.immutable.IndexedSeq
...
+    |the same prior to applying any other clipping criteria.
   """)
 class ClipBam
 ( @arg(flag='i', doc="Input SAM or BAM file of aligned reads in coordinate order.") val input: PathToBam,
   @arg(flag='o', doc="Output SAM or BAM file.") val output: PathToBam,
   @arg(flag='m', doc="Optional output of clipping metrics.") val metrics: Option[FilePath] = None,
   @arg(flag='r', doc="Reference sequence fasta file.") val ref: PathToFasta,
-  @arg(flag='s', doc="Soft clip reads instead of hard clipping.") val softClip: Boolean = false,
+  @deprecated("Use clipping-mode instead.", since="0.4.0")
+  @arg(flag='s', doc="Soft clip reads instead of hard clipping.", mutex=Array("clippingMode")) val softClip: Boolean = false,
```
tfenne: I'm sorta surprised this mutex works when softClip has a default value, but I checked it out and built and it does!

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -76,10 +86,12 @@ class ClipBam
   Io.assertReadable(ref)
   Io.assertCanWriteFile(output)
 
-  validate(clipOverlappingReads || Seq(readOneFivePrime, readOneThreePrime, readTwoFivePrime, readTwoThreePrime).exists(_ != 0),
+  this.clippingMode = this.clippingMode.orElse { if (this.softClip) Some(ClippingMode.Soft) else Some(ClippingMode.Hard) }
```
tfenne: I'd suggest making `this.clippingMode` a `val` and then doing this here instead:
`val mode = this.clippingMode.getOrElse { if (this.softClip) ClippingMode.Soft else ClippingMode.Hard }`

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -91,6 +103,9 @@ class ClipBam
 
     // Go through and clip reads and fix their mate information
     Bams.templateIterator(in).foreach { template =>
+
+      template.allReads.foreach { r => this.clipper.upgradeAllClipping(r) }
```
tfenne: Shouldn't this check `if (this.upgradeClipping)`?

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -91,6 +103,9 @@ class ClipBam
 
     // Go through and clip reads and fix their mate information
     Bams.templateIterator(in).foreach { template =>
+
```
tfenne: Extra newline

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -102,6 +117,7 @@ class ClipBam
         case _ => Unit
       }
 
+
```
tfenne: Extra newline

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
nh13: I had to build it too so that I could verify.  It makes sense to me at least, mutex doesn't ask if there _are_ values, only that the user shouldn't specify both.

PR #339, opened 2017-12-14
Adding a tool that makes reads unmapped based on their insert size.

nh13:


src/main/scala/com/fulcrumgenomics/bam/MakeUnmappedByInsertSize.scala
```diff
@@ -0,0 +1,102 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) $year Fulcrum Genomics
```
tfenne: $year -> 2017

src/main/scala/com/fulcrumgenomics/bam/MakeUnmappedByInsertSize.scala
```diff
@@ -0,0 +1,102 @@
...
+import com.fulcrumgenomics.commons.io.Io
+import com.fulcrumgenomics.commons.util.LazyLogging
+import com.fulcrumgenomics.sopt._
+import htsjdk.samtools.util.IntervalList
+
+import scala.math.abs
+
+@clp(description = 
+  """
+     |Makes reads unmapped if based on their insert size.
```
tfenne: ~~if~~

src/main/scala/com/fulcrumgenomics/bam/MakeUnmappedByInsertSize.scala
```diff
@@ -0,0 +1,102 @@
...
+     |
+     |Fragment reads will always be made unmapped. Secondary and supplementary reads whose become unmapped will instead
```
tfenne: The second sentence here doesn't scan.  If I read it as "whose primary read becomes..." then it also appears false.  The code below appears, as best I can tell, to remove all secondary and supplementary reads.

src/main/scala/com/fulcrumgenomics/bam/MakeUnmappedByInsertSize.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+
+  validate(minInsertSize.isDefined || maxInsertSize.isDefined, "Either --min-insert-size or --max-insert-size must be given.")
+
+  override def execute(): Unit = {
+    val in       = SamSource(input)
+    val out      = SamWriter(output, in.header)
+    var secondary: Long = 0
```
tfenne: Or just `var secondary = 0L`

src/main/scala/com/fulcrumgenomics/bam/MakeUnmappedByInsertSize.scala
```diff
@@ -0,0 +1,102 @@
...
+    var supplementary: Long = 0
+    val unmapped: Long = in
+      .count { rec =>
```
tfenne: Using `count` with a block that is highly side-effecting feels wrong to me.  I think it would be clearer, and actually shorter, to just use a `foreach` and add `unmapped += 1` in the once place you return `true`.

src/main/scala/com/fulcrumgenomics/bam/MakeUnmappedByInsertSize.scala
```diff
@@ -0,0 +1,102 @@
...
+        val insertSize = abs(rec.insertSize)
+        val makeUnmapped = !rec.paired ||
```
tfenne: Do you want to guard against reads where `read.unmapped || read.mateUnmapped` is true but an insert size is present?

src/main/scala/com/fulcrumgenomics/bam/MakeUnmappedByInsertSize.scala
tfenne: Looking at the code I'm guessing this is actually meant to be something like "reads whose insert sizes falls outside the desired range will be removed instead of being unmapped."

src/main/scala/com/fulcrumgenomics/bam/MakeUnmappedByInsertSize.scala
```diff
@@ -0,0 +1,102 @@
...
+     |Makes reads unmapped if based on their insert size.
+     |
+     |Fragment reads will always be made unmapped. Secondary and supplementary reads whose become unmapped will instead
+     |be removed.
+  """,
+  group = ClpGroups.SamOrBam)
+class MakeUnmappedByInsertSize
+( @arg(flag='i', doc="Input BAM file.")                                 val input: PathToBam,
+  @arg(flag='o', doc="Output BAM file.")                                val output: PathToBam,
+  @arg(flag='m', doc="Remove all reads with insert size < this value.") val minInsertSize: Option[Int] = None,
```
tfenne: Why not just use `Int` with defaults of `0` and `Int.MaxValue`?

src/main/scala/com/fulcrumgenomics/bam/MakeUnmappedByInsertSize.scala
```diff
@@ -0,0 +1,102 @@
...
+          if (rec.secondary) {
+            secondary += 1
+            false
+          }
+          else if (rec.supplementary) {
+            supplementary += 1
+            false
+          }
+          else {
+            rec.unmapped = true
```
tfenne: I think you probably want to use `SAMUtils.makeReadUnmapped(rec.asSam)` here to clean up more of the things that are invalid on an unmapped record.

nh13:
Closing as not everyone and their mother should be in here.

PR #340, opened 2017-12-14
WIP: Adding a tool to convert SamToFastq with additional bases and qualities tags

nh13:
https://github.com/samtools/hts-specs/issues/270


nh13:
Lack of progress from the hts-specs folks

PR #343, opened 2017-12-21
DemuxFastqs can write both FASTQs and BAM.

nh13:
This deprecates the --output-fastqs option.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -27,10 +27,10 @@ package com.fulcrumgenomics.fastq
 
 import java.io.Closeable
 
-import com.fulcrumgenomics.FgBioDef.unreachable
+import com.fulcrumgenomics.FgBioDef.{FgBioEnum, unreachable}
 import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
-import com.fulcrumgenomics.commons.CommonsDef.{DirPath, FilePath, PathPrefix, PathToBam, PathToFastq}
+import com.fulcrumgenomics.commons.CommonsDef.{DirPath, FilePath, PathPrefix, PathToFastq}
 import com.fulcrumgenomics.commons.io.PathUtil
 import com.fulcrumgenomics.commons.util.{LazyLogging, Logger}
 import com.fulcrumgenomics.fastq.FastqDemultiplexer.{DemuxRecord, DemuxResult}
@@ -42,7 +42,10 @@ import com.fulcrumgenomics.util.{ReadStructure, SampleBarcodeMetric, _}
 import htsjdk.samtools.SAMFileHeader.SortOrder
 import htsjdk.samtools._
 import htsjdk.samtools.util.{Iso8601Date, SequenceUtil}
+import enumeratum.EnumEntry
 
+import scala.collection.immutable.IndexedSeq
+import scala.collection.mutable.ListBuffer
 import scala.concurrent.forkjoin.ForkJoinPool
 
 object DemuxFastqs {
@@ -52,6 +55,9 @@ object DemuxFastqs {
 
   /** The name of the sample for unmatched reads. */
   val UnmatchedSampleId: String = "unmatched"
+
+  /** The default output type for [[DemuxFastqs]]. */
+  val DefaultOutputType: OutputType = OutputType.BamOnly
 
   /** The maximum # of records in RAM per SAM/BAM writer. */
   private[fastq] val MaxRecordsInRam: Int = 5e6.toInt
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -174,16 +180,20 @@ object DemuxFastqs {
...
       |
-      |Alternatively, gzipped FASTQs can be written using the `--output-fastqs=true` option instead of BAMs.  For paired
-      |end data, the output will have the suffix `_R1.fastq.gz` and `_R2.fastq.gz` for read one and read two respectively.
-      |The sample barcode and molecular barcodes (concatenated) will be appended to the read name and delimited by a
-      |colon.  If the `--illumina-standards` option is given, then the output read names and file names will follow the
+      |For gzipped FASTQ output, one or more gzipped FASTQs per sample in the sample sheet or metadata CSV file will be
+      |written to the output directory. For paired end data, the output will have the suffix `_R1.fastq.gz` and
+      |`_R2.fastq.gz` for read one and read two respectively.  The sample barcode and molecular barcodes (concatenated)
+      |will be appended to the read name and delimited by a colon.  If the `--illumina-standards` option is given, then
+      |the output read names and file names will follow the
```
nh13: this is really just reformatting text and re-line-wrapping.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -294,26 +304,32 @@ class DemuxFastqs
  @arg(doc="Platform model to insert into the group header (ex. miseq, hiseq2500, hiseqX)") val platformModel: Option[String] = None,
  @arg(doc="Comment(s) to include in the merged output file's header.", minElements = 0) val comments: List[String] = Nil,
  @arg(doc="Date the run was produced, to insert into the read group header") val runDate: Option[Iso8601Date] = None,
- @arg(doc="Output gzipped FASTQs (`.fastq.gz`) instead of BAM files") val outputFastqs: Boolean = false,
+ @arg(doc="The type of outputs to produce.") var outputType: OutputType = DemuxFastqs.DefaultOutputType,
+ @deprecated("Use outputType instead.", since="0.5.0")
+ @arg(doc="*** Deprecated: use --output-type instead ***. Output gzipped FASTQs (`.fastq.gz`) instead of BAM files") val outputFastqs: Boolean = false,
```
nh13: I was considering putting a `mutex` on the two options, so that you can specify them both.  What do you think?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -174,16 +180,20 @@ object DemuxFastqs {
       |the FASTQs, to assign each read to a sample.  Reads that do not match any sample within the given error tolerance
       |will be placed in the 'unmatched' file.
       |
-      |The output directory will contain one BAM file per sample in the sample sheet or metadata CSV file, plus a BAM for
-      |reads that could not be assigned to a sample given the criteria.  The output file names will be the concatenation
-      |of sample id, sample name, and sample barcode bases (expected not observed), delimited by `-`.  A metrics file
-      |will also be output providing analogous information to the metric described
+      |The type of output is specified with the `--output-type` option can be BAM (`--output-type BamOnly`), gzipped
```
tfenne: `option an be...` -> `option, and can be...`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -174,16 +180,20 @@ object DemuxFastqs {
...
+      |FASTQ (`--output-type FastqOnly`), or both (`--output-type Both`).
+      |
+      |For BAM output, the output directory will contain one BAM file per sample in the sample sheet or metadata CSV file,
+      |plus a BAM for reads that could not be assigned to a sample given the criteria.  The output file names will be the
+      |concatenation of sample id, sample name, and sample barcode bases (expected not observed), delimited by `-`.  A
+      |metrics file will also be output providing analogous information to the metric described
```
tfenne: I've never liked the link the Picard metric here.  Why not link to our own metric?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -294,26 +304,32 @@ class DemuxFastqs
  @arg(doc="Platform model to insert into the group header (ex. miseq, hiseq2500, hiseqX)") val platformModel: Option[String] = None,
  @arg(doc="Comment(s) to include in the merged output file's header.", minElements = 0) val comments: List[String] = Nil,
  @arg(doc="Date the run was produced, to insert into the read group header") val runDate: Option[Iso8601Date] = None,
- @arg(doc="Output gzipped FASTQs (`.fastq.gz`) instead of BAM files") val outputFastqs: Boolean = false,
+ @arg(doc="The type of outputs to produce.") var outputType: OutputType = DemuxFastqs.DefaultOutputType,
```
tfenne: Is this a `var` for a reason?

tfenne: Also, is the default used in multiple places?  I think it's less clear than just having the value right here.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
tfenne: Yeah, I think probably the right thing to do until you remove the old option would be:
1. Make the old one `Option[Boolean] = None`
2. Make the new one `Option[OutputType] = None`
3. Add a mutex
4. If both are None, default to Bam, else prefer new option, else old option

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -294,26 +304,32 @@ class DemuxFastqs
...
+ @arg(doc="*** Deprecated: use --output-type instead ***. Output gzipped FASTQs (`.fastq.gz`) instead of BAM files") val outputFastqs: Boolean = false,
  @arg(doc="Output FASTQs according to Illumina naming standards, for example, for upload to the BaseSpace Sequence Hub") val illuminaStandards: Boolean = false
 ) extends FgBioTool with LazyLogging {
 
   import DemuxFastqs._
 
   private[fastq] val metricsPath = metrics.getOrElse(output.resolve(DefaultDemuxMetricsFileName))
 
+  // assumes the default for outputFastqs is false
+  if (this.outputFastqs && this.outputType == OutputType.BamOnly) {
```
tfenne: See suggestion above.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -403,27 +419,39 @@ class DemuxFastqs
     val isUnmatched = sample.sampleName == UnmatchedSampleId
     val prefix = toSampleOutputPrefix(sample, isUnmatched, illuminaStandards, output, this.unmatched)
 
-    if (this.outputFastqs) {
-      new FastqRecordWriter(prefix, pairedEnd, illuminaStandards)
+    val writers = new ListBuffer[DemuxWriter]()
+
+    this.outputType match {
```
tfenne: These matches seem to me like they're be more at home as:
`if (this.outputType == FastqOnly || this.outputType == Both)` since you don't use the matched value _and_ you then have add a second default case.

You could simplify a little further by having the `OutputType` enum have a `.producesFastq` and `.producesBam` pair of values ;)

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -403,27 +419,39 @@ class DemuxFastqs
...
+          val header: SAMFileHeader = new SAMFileHeader
+          header.addReadGroup(readGroup)
+          header.setSortOrder(sortOrder)
+          comments.foreach(header.addComment)
+
+          writers += new SamRecordWriter(prefix, header, this.umiTag, numSamples)
+      case _ => Unit
+    }
+
+    new DemuxWriter {
```
tfenne: I think the answer is no, but do we need to worry about memory requirements when writing both?  Or is the case that BAM sorts and fastq doesn't, so it's not double?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -668,3 +696,11 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
     DemuxResult(sampleInfo=sampleInfo, numMismatches=numMismatches, records=records)
   }
 }
+
+sealed trait OutputType extends EnumEntry
+object OutputType extends FgBioEnum[OutputType] {
```
tfenne: I would suggest `Bam`, `Fastq` and `BamAndFastq` for the values for the enum. Or ... actually thinking about it more another option would be to just have `Bam` and `Fastq` and then let the user specify `--output-type Bam Fastq` to get both.  Just a thought.  I think that's cuter, but perhaps less obvious to users.

PR #344, opened 2017-12-21
Better handling of multi-line failure messages.

tfenne:
Nice little one-line PR for you.  The idea being that if you invoke:

```scala
fail("Found a bunch of problems:\n" + errors.mkString("\n"))
```

That all the lines are prefixed the same amount, instead of having a big out-dent for lines 2-n.


nh13:
I'll take your word for it. 👍

PR #345, opened 2017-12-23
Adding in options to DemuxFastqs for asynchronous reading and writing

nh13:
* Added the --async-reading option to specify if and how to perform
  asynchronous reading of the inputs
* Added the --num-writing-threads to specify if and how many threads
  to use when writing asynchronously
* Deprecate the --threads option, use --async-reading and
  --num-writing-threads instead
* Added a number of classes to handle the asynchronous reading and
  writing
Depends on:
* https://github.com/fulcrumgenomics/commons/pull/22
* https://github.com/fulcrumgenomics/commons/pull/23

Click the link below to see some in-depth analysis of the various added options.
<details>

I looked at the six various combination of options:
1. "memory": the amount of memory used by the tool (values: 8g, 16g, 32g)
2. "output type": the type of output files for the `--output-type` option (values: `Bam`, `Fastq`, or `BamAndFastq`)
3. "threads in master": the # of threads for the `--threads` option in the `master` branch (values: `0`, `1`, `2`, `4`, `8`, `16`, `32`).
4. "async reading in devel": the type of asynchronous reading for the `--async-reading` option for the branch in this PR (values: `None`, `SingleThreads`, `ThreadPerReader`).
5. "threads for async writing in devel": the # of threads to use when writing output(s) for the `--async-writing-threads` (values: `0`, `1`, `2`, `4`, `8`, `16`, `32`).
6. "threads for demux in devel": whether or not to use a separate thread for asynchronous demultiplexing (values `true` or `false`).

I ran the tool for all combinations of six values, and I plotted the elapsed time in six plots to examine the affect of run time for each option.  I sub-sorted by "output-type" to show the affect on each relative to outputting BAM, FASTQs or both, since I found that this was informative.

My observations are:
* memory has no real effect
* writing both BAMs and FASTQs causes a significant spike in run time; using increasing # of threads for writing mitigates this issue
* threading in the master branch improves performance of the demultiplexing step (computational portion), but has a lower limit imposed by IO (hence the addition of asynchronous reading and writing)
* interestingly for asynchronous reading, having a single thread read all the inputs is faster than a thread-per-input;  I am not sure why.
* multiple threads for asynchronous writing really improves performance, and makes writing BAMs equivalent to writing both BAMs and FASTQs, but writing just FASTQs is slightly faster
* having a separate thread for the demultiplexing step (computational portion) has some very slight gains.  I did not try multiple threads for the demultiplexing step like in `master`.

In summary, asynchronous writing has huge benefits, which makes sense when we have 96 or 384 samples (and as many output files), but 4-8 thread maxes out write performance.  Asynchronous reading has a small benefit since we are just decompressing 2-4 gzipped FASTQs and are likely limited by a read speed anyhow.  Asynchronously barcode matching (the demultiplexing step) has some small benefits, but perhaps using multiple threads could yield benefits.  I can see retaining all three options (async-reading type, # of async-writing threads, and whether or not to async demux), but they could be confusing to the user since there are so many already.  I'd like to think about this further.

plots

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/Writer.scala
+++ b/src/main/scala/com/fulcrumgenomics/Writer.scala
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2017 Fulcrum Genomics LLC
+ * Copyright (c) 2017 Fulcrum Genomics
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -20,31 +20,11 @@
  * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
+ *
  */
 
 package com.fulcrumgenomics
 
-import java.io.Closeable
-
-import com.fulcrumgenomics.FgBioDef._
-
 /** Writer trait to standardize how to interact with classes that write out objects. */
-trait Writer[A] extends Closeable {
-  /** Writes an individual item. */
-  def write(item: A): Unit
-
-  /** Writes out one or more items in order. */
-  def write[B <: A](items: TraversableOnce[B]): Unit = items.foreach(write)
-
-  /** Writes an item and returns a reference to the writer. */
-  def +=(item: A): this.type = {
-    write(item)
-    this
-  }
-
-  /** Writes an item and returns a reference to the writer. */
-  def ++=[B <: A](items: TraversableOnce[B]): this.type = {
-    write(items)
-    this
-  }
-}
+@deprecated("Use  com.fulcrumgenomics.commons.io.Writer", since="0.5.0")
+trait Writer[A] extends com.fulcrumgenomics.commons.io.Writer[A]
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -25,13 +25,12 @@
 
 package com.fulcrumgenomics.fastq
 
-import java.io.Closeable
-
 import com.fulcrumgenomics.FgBioDef.{FgBioEnum, unreachable}
 import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.commons.CommonsDef.{DirPath, FilePath, PathPrefix, PathToFastq}
-import com.fulcrumgenomics.commons.io.PathUtil
+import com.fulcrumgenomics.commons.async.{AsyncIterator, AsyncWriterPool}
+import com.fulcrumgenomics.commons.io.{PathUtil, Writer}
 import com.fulcrumgenomics.commons.util.{LazyLogging, Logger}
 import com.fulcrumgenomics.fastq.FastqDemultiplexer.{DemuxRecord, DemuxResult}
 import com.fulcrumgenomics.illumina.{Sample, SampleSheet}
@@ -39,14 +38,25 @@ import com.fulcrumgenomics.sopt.{arg, clp}
 import com.fulcrumgenomics.umi.ConsensusTags
 import com.fulcrumgenomics.util.ReadStructure.SubRead
 import com.fulcrumgenomics.util.{ReadStructure, SampleBarcodeMetric, _}
+import enumeratum.EnumEntry
 import htsjdk.samtools.SAMFileHeader.SortOrder
 import htsjdk.samtools._
 import htsjdk.samtools.util.{Iso8601Date, SequenceUtil}
-import enumeratum.EnumEntry
 
 import scala.collection.immutable.IndexedSeq
 import scala.collection.mutable.ListBuffer
-import scala.concurrent.forkjoin.ForkJoinPool
+
+/** Controls how we read from the inputs in [[DemuxFastqs]].*/
+sealed trait DemuxFastqsAsyncReadingType extends EnumEntry
+case object DemuxFastqsAsyncReadingType extends FgBioEnum[DemuxFastqsAsyncReadingType] {
+  def values: IndexedSeq[DemuxFastqsAsyncReadingType] = findValues
+  /** Read in a synchronous fashion */
+  case object NoAsync extends DemuxFastqsAsyncReadingType
+  /** Read all the inputs in a single thread */
+  case object SingleThread extends DemuxFastqsAsyncReadingType
+  /** Read each input in its own thread */
+  case object ThreadPerReader extends DemuxFastqsAsyncReadingType
+}
 
 object DemuxFastqs {
 
@@ -56,15 +66,17 @@ object DemuxFastqs {
   /** The name of the sample for unmatched reads. */
   val UnmatchedSampleId: String = "unmatched"
 
+  /** Contains the parameters used to control asynchronous computation and IO. */
+  private case class DemuxFastqsAsyncOptions(asyncReadingType: DemuxFastqsAsyncReadingType,
+                                             numWriterThreads: Int,
+                                             asyncDemux: Boolean,
+                                             demuxBatchSize: Int = 10e5.toInt) {
+    require(numWriterThreads >= 0)
+    require(!asyncDemux || demuxBatchSize > 0)
+  }
+
   /** The maximum # of records in RAM per SAM/BAM writer. */
   private[fastq] val MaxRecordsInRam: Int = 5e6.toInt
-
-  /** Decides whether or not to use asynchronous IO for the writers. This has a big performance benefit at the cost of
-    * some RAM. RAM may balloon if there is a need to sort the output. */
-  private[fastq] val UseAsyncIo: Boolean  = true
-
-  /** The number of records to batch when demultiplexing in parallel. */
-  private val DemuxBatchRecordsSize = 1e5.toInt
 
   /** Creates the sample output BAM path for the given sample. */
   private[fastq] def outputPrefixFrom(output: DirPath, sample: Sample): PathPrefix = {
@@ -134,36 +146,43 @@ object DemuxFastqs {
     }
   }
 
-  /** Creates a demultiplexing iterator that performs demultiplexing in parallel.
+  /** Creates a demultiplexing iterator.
     *
     * @param sources the FASTQ sources, one per read.
     * @param demultiplexer the demultiplexer to use.  The demultiplexer's [[com.fulcrumgenomics.fastq.FastqDemultiplexer.demultiplex()]]
     *                      method expects the same number of reads as sources.
     */
-  def demultiplexingIterator(sources: Seq[FastqSource],
+  def demultiplexingIterator(sources: Seq[Iterator[FastqRecord]],
                              demultiplexer: FastqDemultiplexer,
-                             threads: Int,
-                             batchSize: Int = DemuxBatchRecordsSize): Iterator[DemuxResult] = {
-
+                             asyncOptions: DemuxFastqsAsyncOptions): Iterator[DemuxResult] = {
     require(demultiplexer.expectedNumberOfReads == sources.length,
       s"The demultiplexer expects ${demultiplexer.expectedNumberOfReads} reads but ${sources.length} FASTQ sources given.")
 
-    val zippedIterator = FastqSource.zipped(sources)
-    if (threads > 1) {
+    val demuxSource = asyncOptions.asyncReadingType match {
+      case DemuxFastqsAsyncReadingType.NoAsync         => FastqSource.zipped(sources)
+      case DemuxFastqsAsyncReadingType.SingleThread    => AsyncIterator(FastqSource.zipped(sources), bufferSize=Some(1024))
+      case DemuxFastqsAsyncReadingType.ThreadPerReader => FastqSource.zipped(sources.map { source => AsyncIterator(source, bufferSize=Some(1024))})
+    }
+
+    if (asyncOptions.asyncDemux) {
+      AsyncIterator[DemuxResult](demuxSource.map { readRecords => demultiplexer.demultiplex(readRecords: _*) }, bufferSize=Some(asyncOptions.demuxBatchSize))
+      // NB: We could enable multiple threads for demultiplexing with this in the future.
       // Developer Note: Iterator does not support parallel operations, so we need to group together records into a
       // [[List]] or [[Seq]].  A fixed number of records are grouped to reduce memory overhead.
+      /*
       import com.fulcrumgenomics.FgBioDef.ParSupport
-      val pool = new ForkJoinPool(threads)
-      zippedIterator
-        .grouped(batchSize)
+      val pool = new ForkJoinPool(demuxThreads)
+      demuxSource
+        .grouped(demuxBatchSize)
         .flatMap { batch =>
           batch
             .parWith(pool=pool)
             .map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
         }.seq // Developer Note: toStream ensures that the only parallelism is within the flatMap
+        */
     }
     else {
-      zippedIterator.map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
+      demuxSource.map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
     }
   }
 }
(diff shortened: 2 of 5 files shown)


nh13:
will eventually depend on https://github.com/fulcrumgenomics/commons/pull/22

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -26,28 +26,180 @@
...
 import htsjdk.samtools._
-import htsjdk.samtools.util.{Iso8601Date, SequenceUtil}
-import enumeratum.EnumEntry
+import htsjdk.samtools.util.{AbstractAsyncWriter, Iso8601Date, SequenceUtil}
 
 import scala.collection.immutable.IndexedSeq
 import scala.collection.mutable.ListBuffer
 import scala.concurrent.forkjoin.ForkJoinPool
 
+private object Async {
```
nh13: Has been moved over to https://github.com/fulcrumgenomics/commons/pull/22.  I will update when that PR gets merged.

src/main/scala/com/fulcrumgenomics/Writer.scala
```diff
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2017 Fulcrum Genomics LLC
+ * Copyright (c) $year Fulcrum Genomics
```
nh13: doh!

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -625,6 +719,23 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
     case n => throw new IllegalArgumentException(s"$n template reads defined. Can't process > 2 template reads.")
   }
 
+  /** Gets the two items [[T]] with the largest associated values ([[Int]]). */
+  private def getTopTwoByMaximum[T](elems: TraversableOnce[(T, Int)]): (Option[(T, Int)], Option[(T, Int)]) = {
```
nh13: I found this sped things up 10-20%!

nh13:
bump

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -25,28 +25,38 @@
...
 import scala.collection.immutable.IndexedSeq
 import scala.collection.mutable.ListBuffer
-import scala.concurrent.forkjoin.ForkJoinPool
+
+/** Controls how we read from the inputs in [[DemuxFastqs]].*/
+sealed trait DemuxFastqsAsyncReadingType extends EnumEntry
+case object DemuxFastqsAsyncReadingType extends FgBioEnum[DemuxFastqsAsyncReadingType] {
+  def values: IndexedSeq[DemuxFastqsAsyncReadingType] = findValues
+  /** Read in a synchronous fashion */
+  case object NoAsync extends DemuxFastqsAsyncReadingType
```
tfenne: Rather than `NoAsync` how about  `Synchronous` or `Serial`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -25,28 +25,38 @@
...
+  /** Read all the inputs in a single thread */
+  case object SingleThread extends DemuxFastqsAsyncReadingType
```
tfenne: Thinking about how this'll look on the command line I don't think this is super clear that it means a single _extra_ thread.  Maybe `SingleAsyncThread`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -25,28 +25,38 @@
...
+  /** Read each input in its own thread */
+  case object ThreadPerReader extends DemuxFastqsAsyncReadingType
```
tfenne: Would `ThreadPerFile` make more sense to users?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -56,16 +66,18 @@ object DemuxFastqs {
   /** The name of the sample for unmatched reads. */
   val UnmatchedSampleId: String = "unmatched"
 
+  /** Contains the parameters used to control asynchronous computation and IO. */
+  private case class DemuxFastqsAsyncOptions(asyncReadingType: DemuxFastqsAsyncReadingType,
+                                             numWriterThreads: Int,
+                                             asyncDemux: Boolean,
+                                             demuxBatchSize: Int = 10e5.toInt) {
+    require(numWriterThreads >= 0)
+    require(!asyncDemux || demuxBatchSize > 0)
```
tfenne: Since you default to > 0, maybe remote the `!asyncDemux` as it's a little confusing.

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -59,14 +59,14 @@ object FastqSource {
     * @param sources a Seq of one or more FastqSource objects
     * @return an Iterator that returns a Seq of FastqRecord, one per source
     */
-  def zipped(sources: Seq[FastqSource]): Iterator[Seq[FastqRecord]] = new Iterator[Seq[FastqRecord]] {
```
tfenne: I'm pretty sure I use this method elsewhere.  Can you keep both please and have vector this one to the one that takes iterators?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -134,36 +146,43 @@ object DemuxFastqs {
...
-
+                             asyncOptions: DemuxFastqsAsyncOptions): Iterator[DemuxResult] = {
     require(demultiplexer.expectedNumberOfReads == sources.length,
       s"The demultiplexer expects ${demultiplexer.expectedNumberOfReads} reads but ${sources.length} FASTQ sources given.")
 
-    val zippedIterator = FastqSource.zipped(sources)
-    if (threads > 1) {
+    val demuxSource = asyncOptions.asyncReadingType match {
+      case DemuxFastqsAsyncReadingType.NoAsync         => FastqSource.zipped(sources)
+      case DemuxFastqsAsyncReadingType.SingleThread    => AsyncIterator(FastqSource.zipped(sources), bufferSize=Some(1024))
```
tfenne: Perhaps `1024` could either be a constant somewhere _or_ another value in the async options case class, or use the batch size from that class?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -134,36 +146,43 @@ object DemuxFastqs {
...
-      zippedIterator
-        .grouped(batchSize)
+      val pool = new ForkJoinPool(demuxThreads)
+      demuxSource
+        .grouped(demuxBatchSize)
         .flatMap { batch =>
           batch
             .parWith(pool=pool)
             .map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
         }.seq // Developer Note: toStream ensures that the only parallelism is within the flatMap
```
tfenne: The developer note here is out of date I think?  `toStream` isn't used?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -134,36 +146,43 @@ object DemuxFastqs {
...
+      AsyncIterator[DemuxResult](demuxSource.map { readRecords => demultiplexer.demultiplex(readRecords: _*) }, bufferSize=Some(asyncOptions.demuxBatchSize))
+      // NB: We could enable multiple threads for demultiplexing with this in the future.
       // Developer Note: Iterator does not support parallel operations, so we need to group together records into a
       // [[List]] or [[Seq]].  A fixed number of records are grouped to reduce memory overhead.
+      /*
       import com.fulcrumgenomics.FgBioDef.ParSupport
-      val pool = new ForkJoinPool(threads)
-      zippedIterator
-        .grouped(batchSize)
+      val pool = new ForkJoinPool(demuxThreads)
```
tfenne: Does using > 1 extra thread here help throughput significantly?  If not you could just wrap this in an AsyncIterator, but I'm guessing more threads must be worth the complexity of batching?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -267,6 +286,25 @@ object DemuxFastqs {
       |
       |1. The file extension will be `_R1_001.fastq.gz` for read one, and `_R2_001.fastq.gz` for read two (if paired end).
       |2. The per-sample output prefix will be `<SampleName>_S<SampleOrdinal>_L<LaneNumber>` (without angle brackets).
+      |
+      |## Multi-threading and Asynchronous IO
+      |
+      |A number of options allow fine-grain control over multi-threading and asynchronous input and output.
+      |
+      |* The `--async-reading-type-type` option allows reading from the input(s) asynchronously.
```
tfenne: `type-type`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -267,6 +286,25 @@ object DemuxFastqs {
...
+      |  * `NoAsync`: reads the input(s) synchronously (no threading).
+      |  * `SingleThread`: reads the input(s) all in a single thread.
```
tfenne: single [additional] thread

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -267,6 +286,25 @@ object DemuxFastqs {
...
+      |  * `ThreadPerReader`: reads each input in a seperate thread.
+      |* The `--async-writing-threads` option enables writing the outputs in seperate thread(s).  If zero, the outputs
+      |  are written synchronously (no threading).  Otherwise, the outputs are grouped into `N` groups, and a thread per
+      |  group is created to write to the outputs in the given group.
+      |* The `--async-demux` option enables the sample barcode matching (demultiplexing) to occur in a seperate thread.
```
tfenne: So you don't expose the ability to use > 1 threads even though it's supported?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -308,13 +345,33 @@ class DemuxFastqs
...
+ @arg(doc="Enable a separate thread for the demultiplexing operation (not IO).") asyncDemux: Boolean = false
 ) extends FgBioTool with LazyLogging {
 
   import DemuxFastqs._
 
   private[fastq] val metricsPath = metrics.getOrElse(output.resolve(DefaultDemuxMetricsFileName))
 
+  private val asyncOptions = threads match {
+    case Some(thr)            =>
+      validate(this.asyncReadingType.isEmpty, "--async-reading-type cannot be used with --threads")
```
tfenne: Do you want to also print a warning here that `threads` is deprecated?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -354,17 +411,48 @@ class DemuxFastqs
...
+    val sortProgress:  ProgressLogger = ProgressLogger(logger, verb="Sorted", unit=2e6.toInt)
+    val writeProgress: ProgressLogger = ProgressLogger(logger, verb="Wrote ", unit=2e6.toInt)
+
     Io.mkdirs(this.output)
     Io.assertCanWriteFile(this.metricsPath)
 
+    // Some logging information about async reading and writing
+    {
+      val readerMessage = this.asyncOptions.asyncReadingType match {
+        case DemuxFastqsAsyncReadingType.NoAsync         => "not use async reading"
```
tfenne: Since this enum exists purely for specifying options here, why not just put these strings inside the case objects?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -447,34 +533,63 @@ class DemuxFastqs
...
-  def add(rec: DemuxRecord): Unit
+/** Writes a [[DemuxResult]] to its associated writer given a map between the sample index and it's [[Writer]]. */
+private class DemuxResultsBySampleIndexWriter(val sampleIndexToWriterMap: Map[Int, Writer[DemuxResult]],
+                                              private[this] val source: Option[{ def close(): Unit }] = None
+                                             ) extends Writer[(DemuxResult, Int)] {
+  def write(itemAndIndex: (DemuxResult, Int)): Unit = {
+    val (item, sampleIndex) = itemAndIndex
+    sampleIndexToWriterMap(sampleIndex).write(item)
+  }
+  def close(): Unit = source match {
```
tfenne: Missing newline.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -447,34 +533,63 @@ class DemuxFastqs
...
+private class DemuxResultWriter(val writer: DemuxRecordWriter, val qualityEncoding: QualityEncoding) extends Writer[DemuxResult] {
+  def write(item: DemuxResult): Unit = item.records.foreach { rec =>
+      this.writer.write(rec.copy(quals=qualityEncoding.toStandardAscii(rec.quals)))
+  }
+  def close(): Unit = this.writer.close()
+}
+
+/** A writer that writes [[DemuxRecord]]s */
+private trait DemuxRecordWriter extends Writer[DemuxRecord] {
+  def write(rec: DemuxRecord): Unit
```
tfenne: Is this not defined in `Writer`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
tfenne: Nice!  I think this is a perfect example of where a tiny case class e.g. `case class IndexAndMatches(index: String, matches: Int)` would make this signature and a bunch of the matches below _way_ clearer.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -635,6 +754,23 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
     case n => throw new IllegalArgumentException(s"$n template reads defined. Can't process > 2 template reads.")
   }
 
+  /** Gets the two items [[T]] with the largest associated values ([[Int]]). */
+  private def getTopTwoByMaximum[T](elems: TraversableOnce[(T, Int)]): (Option[(T, Int)], Option[(T, Int)]) = {
+    val startBest: Option[(T, Int)] = None
```
tfenne: I'm assuming you're doing this because in-lining `None` doesn't infer the type(s) correctly?  I think it would be clearer if you did:

```scala
val empty: Option[(T, Int)] = None
elems.foldLeft((empty, empty)) { case ....
```

src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -59,14 +59,14 @@ object FastqSource {
...
+  def zipped(sources: Seq[Iterator[FastqRecord]]): Iterator[Seq[FastqRecord]] = new Iterator[Seq[FastqRecord]] {
     require(sources.nonEmpty, "No sources provided")
 
     def hasNext(): Boolean = sources.exists(_.hasNext)
 
     def next(): Seq[FastqRecord] = {
       if (!this.hasNext) throw new NoSuchElementException("Calling next() when hasNext() is false.")
-      require(sources.forall(_.hasNext) == sources.head.hasNext, "Sources are out of sync.")
+      require(sources.forall(_.hasNext) == sources.head.hasNext, s"Sources are out of sync: ${sources.map(_.hasNext).mkString(", ")}")
```
tfenne: I tend to think messages in require() statements should make sense without having to look at code - e.g. a user could see this if they feed in a truncated fastq file.  If you want to add more detail you could figure out which one is out of sync and say something like `first and nth fastq file are out of sync`.

src/main/scala/com/fulcrumgenomics/fastq/FastqWriter.scala
```diff
@@ -37,16 +37,16 @@ object FastqWriter {
   def apply(path: Path): FastqWriter = apply(Io.toWriter(path))
 
   /** Constructs a FastqWriter from a Writer. */
-  def apply(writer: java.io.Writer): FastqWriter = {
-    if (writer.isInstanceOf[BufferedWriter]) new FastqWriter(writer.asInstanceOf[BufferedWriter])
-    else new FastqWriter(new BufferedWriter(writer))
+  def apply(writer: java.io.Writer): FastqWriter = writer match {
```
jacarey: Apply takes a writer and either makes sure it is a BufferedWriter or turns it into one. Is there any reason not to have this just require a BufferedWriter?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -447,34 +533,63 @@ class DemuxFastqs
...
+      def write(rec: DemuxRecord): Unit = writers.foreach { writer => writer.write(rec) }
       override def close(): Unit = writers.foreach(_.close())
     }
   }
 }
 
-/** A writer than writes [[DemuxRecord]]s */
-private trait DemuxWriter extends Closeable {
-  def add(rec: DemuxRecord): Unit
+/** Writes a [[DemuxResult]] to its associated writer given a map between the sample index and it's [[Writer]]. */
```
jacarey: its

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -447,34 +533,63 @@ class DemuxFastqs
...
+    val (item, sampleIndex) = itemAndIndex
+    sampleIndexToWriterMap(sampleIndex).write(item)
+  }
+  def close(): Unit = source match {
+    case None      => this.sampleIndexToWriterMap.values.foreach(_.close())
+    case Some(src) => src.close()
+  }
+}
+
+/** A write that writes the records in [[DemuxResult]] to the given [[DemuxRecordWriter]]. */
```
jacarey: write -> writer

nh13:
Closing, use `fqtk demux` insead!

PR #347, opened 2017-12-27
Adding an option to output all bases to the FASTQ.

nh13:
This will include any sample barcode, molecular barcode, skipped, and template
bases.  The motivation is sometimes we want the reads in their original form, for example, since [Basespace wont accept paired end FASTQs where R1 and R2 have different lengths](https://groups.google.com/forum/#!topic/basespace-developers/W6YDxjb8kVs).

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -287,7 +287,7 @@ class DemuxFastqs
       |Illumina for pipeline 1.3 and above (phred scaling + 64) or Standard
       |for phred scaled scores with a character shift of 33.  If this value
       |is not specified, the quality format will be detected automatically.
-  """)
+    """)
  val qualityFormat: Option[QualityEncoding] = None,
  @arg(flag='t', doc="The number of threads to use while de-multiplexing. The performance does not increase linearly with the # of threads and seems not to improve beyond 2-4 threads.") val threads: Int = 1,
  @arg(doc="Maximum mismatches for a barcode to be considered a match.") val maxMismatches: Int = 1,
@@ -304,7 +304,14 @@ class DemuxFastqs
  @arg(doc="The type of outputs to produce.", mutex=Array("outputFastqs")) var outputType: Option[OutputType] = None,
  @deprecated("Use outputType instead.", since="0.5.0")
  @arg(doc="*** Deprecated: use --output-type instead ***. Output gzipped FASTQs (`.fastq.gz`) instead of BAM files", mutex=Array("outputType")) val outputFastqs: Option[Boolean] = None,
- @arg(doc="Output FASTQs according to Illumina naming standards, for example, for upload to the BaseSpace Sequence Hub") val illuminaStandards: Boolean = false
+ @arg(doc="Output FASTQs according to Illumina naming standards, for example, for upload to the BaseSpace Sequence Hub") val illuminaStandards: Boolean = false,
+ @arg(
+   doc=
+     """Output all bases (i.e. all sample barcode, molecular barcode, skipped,
+        and template bases) for every read with template bases (ex. read one
+        and read two) as defined by the corresponding read structure(s).
+     """)
+ val includeAllBasesInFastqs: Boolean = false
 ) extends FgBioTool with LazyLogging {
 
   import DemuxFastqs._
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -600,7 +609,8 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
                                  val umiTag: String = ConsensusTags.UmiBases,
                                  val maxMismatches: Int = 2,
                                  val minMismatchDelta: Int = 1,
-                                 val maxNoCalls: Int = 2) {
+                                 val maxNoCalls: Int = 2,
+                                 val includeOriginal: Boolean = false) {
```
tfenne: Could you add a `@param` to the scaladoc please?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -675,21 +685,43 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
...
+    val molecularBarcode = bases(SegmentType.MolecularBarcode)
+    val sampleBarcode    = bases(SegmentType.SampleBarcode)
+
+    val demuxRecords = reads.zip(this.variableReadStructures)
+      .map { case (read, rs) => rs.extract(read.bases, read.quals) }
+      .filter { segments => segments.exists(_.kind == SegmentType.Template) }
+      .zipWithIndex
+      .map { case (segments, readIndex) =>
+        val readNumber = readIndex + 1
+        val comment    = if (comments.isEmpty) None else Some(comments(readIndex))
```
tfenne: I _think_ this could be a problem, at least in theory.  If I provide 3 fastqs, and only fastq 2 has comments in the read headers, this will try to ask for `comments(0)`, `comments(1)` and `comments(2)` but comments will have `length==1`.

Above, why not just define comments as `reads.map(_.comment)`, and then you can reference it directly here by the index and get either a `None` or `Some`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -675,21 +685,43 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
...
+      val b = subReads.filter(_.kind == segmentType).map(_.bases).mkString("-")
+      if (b.isEmpty) None else Some(b)
     }
 
+    // Get the molecular and sample barcodes
+    val molecularBarcode = bases(SegmentType.MolecularBarcode)
+    val sampleBarcode    = bases(SegmentType.SampleBarcode)
+
+    val demuxRecords = reads.zip(this.variableReadStructures)
+      .map { case (read, rs) => rs.extract(read.bases, read.quals) }
```
tfenne: I think you could make this series of calls a little simpler, and by doing so make it essentially free to include the original bases and quals below, so you don't have to make it optional (in this part of the code, obviously you do elsewhere).  Here's my suggestion:

```scala
val demuxRecords = reads.zip(this.variableReadStructures)
  .filter { case (read, rs) => rs.exists(_.kind == SegmentType.Template) }
  .zipWithIndex
  .map { case ((read, rs), index) => 
    val segments   = rs.extract(read.bases, read.quals)
    val readNumber = readIndex + 1
    val template   = segments.find(_.kind == SegmentType.Template).getOrElse(unreachable("Bug: should contain a template segment"))
    DemuxRecord(
      name             = read.name,
      bases            = template.bases,
      quals            = template.quals,
      molecularBarcode = molecularBarcode,
      sampleBarcode    = sampleBarcode,
      readNumber       = readNumber,
      pairedEnd        = this.pairedEnd,
      comment          = read.comment,
      originalBases    = read.bases,
      originalQuals    = read.quals
    )
  }
```

It also removes the need for the seq of comments above.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -675,21 +685,43 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
...
+      .filter { segments => segments.exists(_.kind == SegmentType.Template) }
+      .zipWithIndex
+      .map { case (segments, readIndex) =>
+        val readNumber = readIndex + 1
+        val comment    = if (comments.isEmpty) None else Some(comments(readIndex))
+        val read       = segments.find(_.kind == SegmentType.Template).getOrElse(unreachable("Bug: should contain a template segment"))
```
tfenne: Do you validate elsewhere that each read can contain only 0-1 template segments?  If not, maybe here it should validate that you get exactly one?

PR #352, opened 2018-01-11
Added the ability to do glocal (as well as global) NW alignments.

tfenne:


src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -25,21 +25,36 @@
...
+
+  /** Enum to represent alignment modes supported by the NeedlenameWunschAligner */
+  object Mode extends FgBioEnum[Mode] {
+    /** Alignment mode for global pairwise alignment. */
+    case object Global extends Mode
+    /** Alignment mode for global alignment of query sequence to local region of target sequence. */
+    case object Glocal extends Mode
+
+    override def values: immutable.IndexedSeq[Mode] = findValues
+  }
```
nh13: Did you test this on the command line (i.e. specify the value for the mode)?  Last time I had an EnumEntry inside the object, it didn't work, so I needed to move it out of the object.  This only affects when the `Mode` type is used as the type for a command line argument.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -25,21 +25,36 @@
...
 import com.fulcrumgenomics.alignment.NeedlemanWunschAligner._
+import enumeratum.EnumEntry
 import htsjdk.samtools.CigarOperator
 
-import scala.collection.mutable
+import scala.collection.{immutable, mutable}
 import scala.math.max
 
 object NeedlemanWunschAligner {
+
```
nh13: Not to be too pedantic, but having a glocal mode makes it no longer Needleman-Wunsch.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -77,6 +94,7 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
     * @return an [[Alignment]] object describing the optimal global alignment of the two sequences
     */
   def align(query: Array[Byte], target: Array[Byte]): Alignment = {
+    require(query.length <= target.length || this.mode == Mode.Global, "For glocal, query length must be <= target length.")
```
nh13: Can you add that to the scaladoc?

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -147,13 +173,40 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
     val elems = new mutable.ArrayBuffer[CigarElem]()
 
-    while (i != 0 || j != 0) {
+    // Always start at the end of the query
+    var i = query.length
+
+    // For the target sequence, start at the end for Global, or the highest scoring cell in the last row for Glocal
+    var j = this.mode match {
+      case Global =>
+        target.length
```
nh13: My style, which is awesome, is to put this on the previous line, so it looks like:
```
case Global => target.length
case Glocal =>
  var ...
```
It's like having a `return` statement at the top of your function, it looks clearer to me.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -147,13 +173,40 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
+      case Glocal =>
+        var (maxScore, maxIndex) = (Int.MinValue, -1)
+        forloop(from=0, until=target.length) { i =>
+          val score = scoring(query.length-1, i)
+          if (score > maxScore) {
+            maxScore = score
+            maxIndex = i
+          }
+        }
```
nh13: How's this for an ugly alternative:
`Stream.range(start=0, end=target.length).zipWithIndex.maxBy(tup => scoring(query.length, tup._1))._2`

If you really want to make it fancy, you can make this into an impicit method (`indexOfMax`) on a `TraversableOnce`.

Or you can keep it readable and as is.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -147,13 +173,40 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
+
+        maxIndex + 1
+    }
+
+    // The score is always the score from the starting cell
+    val score = scoring(i,j)
```
nh13: space after comma

src/test/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAlignerTest.scala
```diff
@@ -35,11 +36,18 @@ class NeedlemanWunschAlignerTest extends UnitSpec {
...
 
+  /** Asserts a couple of things about an alignment that should be constant for glocal alignments. */
+  def assertValidGlocalAlignment(alignment: Alignment): Unit = {
+    alignment.queryStart shouldBe 1
+    alignment.queryEnd   shouldBe  alignment.query.length
+    alignment.targetStart should be >= 1
+    alignment.targetEnd   should be <= alignment.target.length
+  }
+
   "NeedlemanWunschAligner.align" should "align two identical sequences with all matches" in {
```
nh13: `NeedlemanWunschAligner.align` -> `NeedlemanWunschAligner.align(Global)`

src/test/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAlignerTest.scala
```diff
@@ -235,4 +243,39 @@ class NeedlemanWunschAlignerTest extends UnitSpec {
...
+    val result = NeedlemanWunschAligner(1, -1, -3, -1, mode=Glocal).align(s("CCGG"), s("AACCGGTT"))
+    assertValidGlocalAlignment(result)
+    result.queryStart shouldBe 1
+    result.targetStart shouldBe 3
+    result.cigar.toString() shouldBe "4="
+    result.score shouldBe 4
+  }
+
+  it should "correctly align an a sub-sequence with a mismatch" in {
+    val result = NeedlemanWunschAligner(1, -1, -3, -1, mode=Glocal).align(s("-------CGCGTCGTATACGTCGTT"), s("AAGATATCGCGTCGTATACGTCGTA"))
```
nh13: what's with the leading dashes?  Do they get ignored?  Or is this in effect "matches nothing"?  If the latter, can you put "N"s instead.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -25,21 +25,36 @@
...
 object NeedlemanWunschAligner {
+
+  sealed trait Mode extends EnumEntry
+
+  /** Enum to represent alignment modes supported by the NeedlenameWunschAligner */
+  object Mode extends FgBioEnum[Mode] {
+    /** Alignment mode for global pairwise alignment. */
+    case object Global extends Mode
+    /** Alignment mode for global alignment of query sequence to local region of target sequence. */
+    case object Glocal extends Mode
```
nh13: I see four possible modes:
1. global (full query to full target)
2. glocal (full query in a subsequence of target)
3. gLeftLocal (prefix of query in a subsequence of target)
4. gRightLocal (suffix of query in a subsequence of target)

I don't think it would be hard to add those four modes in now since glocal already requires the start/end rules, whereas 3-4 only applies the glocal start/end rule respectively.

src/test/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAlignerTest.scala
tfenne: That `s()` function strips out any `-`s and upper-cases the whole string.  The idea being that I can insert `-`s to make it easier to see where the gaps are and how the strings align, but then they're stripped out before being fed into the aligner.

nh13: I missed the `s` function, sorry.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Yeah, it's true.  But renaming seems like a big hassle.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: I don't disagree, but I don't (yet) see the utility for 3 & 4, so I'm going to skip for now.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: I didn't.  I almost made this a straight up scala Enumeration, because I don't see it being something that's command line exposed (i.e. in most applications you'll want one mode of the other, not a user-choice).  I'll move it to the top-level though just in case.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Good catch.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Ha!  I go back and forth, and generally try to keep all the `case`s the same style, but I'm happy to change it.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Lol.  Cryptic and likely several orders of magnitude slower due to all the intermediate object creation.

PR #353, opened 2018-01-17
Adding validation web-pages.

nh13:
Adding in a page to validate read structures see [here](https://nh13.github.io/fgbio/validate-read-structure.html)

Adding a page to validate a sample sheet for DemuxFastqs see [here](https://nh13.github.io/fgbio/validate-sample-sheet.html)

The main page has been updated to link to the two pages above, see [here](https://nh13.github.io/fgbio/)


index.md
```diff
@@ -48,6 +48,8 @@ Documentation is also available online:
 * Tool usage documentation is available [here](tools/latest)
 * Documentation of the various metrics files is available [here](metrics/latest)
 * API documentation for developers is available through [javadoc.io](http://www.javadoc.io/doc/com.fulcrumgenomics/fgbio_2.12).
+* Web utility for validating [Read Structures](https://github.com/fulcrumgenomics/fgbio/wiki/Read-Structures) is available [here](validate-read-structure.html)
```
tfenne: Maybe we should also add a link to the page explaining read structures here?  (And other wiki pages?)

validate-read-structure.html
```diff
@@ -0,0 +1,67 @@
...
+	<noscript>This page requires JavaScript. Please enable it in your browser settings.</noscript>
+
+	<body>
+		<div class="wrapper">
+			{% include header.html %}
+			<h2>Validate Read Structures</h2>
+			<section>
+				<p>This utility validates one or more <a href="https://github.com/fulcrumgenomics/fgbio/wiki/Read-Structures">Read Structures</a> for use with <a href="http://fulcrumgenomics.github.io/fgbio/">fgbio</a>.</p>
+
+				<p>Enter in a read structure below. Multiple read structures can be entered separated by a space.</p>
```
tfenne: I could not get multiple read structures separated by spaces to work.
Also probably "by a space" -> "by spaces".

index.md
nh13: The page for the read structure is linked on the same line, but I see what you mean.  I'll defer for now.

validate-read-structure.html
nh13: Fixed.

nh13:
I have made most of the suggested fixes, and responded below.  Care to take one last look?

> * RS page: says multiple read structures can be entered separated by a space, but when I do this I get an error "missing length information" and the space is bracketed
Fixed.

> RS page: leading or trailing spaces also trigger the above error, maybe trim() the string before parsing?
Done, I had a different version of `trim()` that I used elsewhere where `trim()` wasn't supported, but it works here so I'll use it.

> RS page: would be really nice if hitting enter after typing in a read structure triggered the Validate button
Done

> RS page: inputting "-7T" gives "Read structure missing length information". Might be nicer to say negative lengths not supported?
Done

> Is all that CSS (and header/footer html) just copy/pasted from our theme? Is there no way to just hot-link to the theme CSS?
Good point, I could hot-link the css, but not the header/footer html.

> Was the JS code all manually translated, or did you use any scala->js tricks? Mostly I'm concerned about how we'll keep this up to date if the scala code changes
It was manually translated.  Your concern is valid, but I think the page is useful enough that I would like exist.

tfenne:
Looks good @nh13.  Two minor thing you can choose to address or not:

1. If you enter read structures separated by tabs (e.g. if copy/pasting from excel), e.g. `76T	76T`, it treats it as one read structure with an invalid character in it
2. If you enter multiple read structures separated by multiple spaces, e.g `76T   76T`, it complains about empty read structures.

PR #355, opened 2018-01-24
Adding Pair and All read types to ClipBam metrics.

nh13:


src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -135,6 +135,14 @@ class ClipBam
       out += rec
     }
 
+    // Set the "All" and "Pair" read types
+    metricsMap.foreach { metrics: Map[ReadType, ClippingMetrics] =>
```
tfenne: I keep getting confused by the fact that `metricsMap` is actually an `Option[Map[A,B]]`.  How would you feel about simplifying things by having the declaration be:

`val metricsMap = if (this.metrics.isEmpty) Map.Empty() else ....`

So that you either have an empty map, or a populated map?  Then where you need to query it you can just call `metricsMap.get()` which will return you an `Option[ClippingMetrics]`?

tfenne: It would also be nice here since you won't have to match it out to `metrics`, which fwiw is shadowing `this.metrics` :/

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -135,6 +135,14 @@ class ClipBam
...
+      val typesForPair: Set[ReadType] = Set(ReadType.ReadOne, ReadType.ReadTwo)
+      val typesForAll: Set[ReadType]  = typesForPair ++ Set(ReadType.Fragment)
+      metrics(ReadType.Pair).add(metrics.filter { case (tpe, _) => typesForPair.contains(tpe) }.values.toSeq:_*)
+      metrics(ReadType.All).add(metrics.filter { case (tpe, _) => typesForAll.contains(tpe) }.values.toSeq:_*)
+    }
```
tfenne: I found this block a little hard to follow.  It may be personal preference, but I offer the below alternative for your consideration:

```scala
import ReadType._
metricsMap.foreach { 
    case (Fragment, metric) => metricMap(All).add(metric)
    case (ReadOne | ReadTwo, metric) => Seq(Pair, All).foreach { r => metricMap(r).add(metric) }
    case _ => Unit
}

File: src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala [hunk 3]
```diff
@@ -289,4 +299,20 @@ case class ClippingMetrics
       if (rec.unmapped && additionalClippedBases > 0) this.reads_unmapped += 1
     }
   }
+  
+  def add(metric: ClippingMetrics*): Unit = {
+    this.reads                     += metric.map(_.reads).sum
```
tfenne: Perfect place to use your new `metric.sumBy(_.reads)`

src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
```diff
@@ -239,6 +239,21 @@ class ClipBamTest extends UnitSpec with ErrorLogLevel with OptionValues {
...
+  "ClippingMetrics.add" should "add two metrics" in {
+    val fragment = ClippingMetrics(ReadType.Fragment, 1, 2,  3,  4,  5,  6, 7,  8,   9, 10, 11, 12, 13)
+    val readOne  = ClippingMetrics(ReadType.ReadTwo,  2, 3,  4,  5,  6,  7, 8,  9,  10, 11, 12, 13, 14)
+    val readTwo  = ClippingMetrics(ReadType.ReadTwo,  3, 4,  5,  6,  7,  8, 9,  10, 11, 12, 13, 14, 15)
+    val all      = ClippingMetrics(ReadType.All,      6, 9, 12, 15, 18, 21, 24, 27, 30, 33, 36, 39, 42)
+
+    // Check that adding fragment, readOne, and readTwo is correct
+    val added    = ClippingMetrics(ReadType.All)
+    added.add(fragment, readOne, readTwo)
+    added.productIterator.toSeq should contain theSameElementsInOrderAs all.productIterator.toSeq
```
tfenne: Can this not simply be `added shouldBe all`?

PR #359, opened 2018-01-31
Added the ability to specify an assembly name to ReferenceSetBuilder.

tfenne:


src/main/scala/com/fulcrumgenomics/testing/ReferenceSetBuilder.scala
```diff
@@ -35,7 +35,7 @@ import htsjdk.samtools.{SAMFileHeader, SAMSequenceDictionary, SAMSequenceRecord,
 import scala.collection.mutable.ListBuffer
 
 /** Class to programatically build up a reference sequence. */
-class ReferenceSetBuilder {
+class ReferenceSetBuilder(assembly: Option[String] = Some("testassembly")) {
```
nh13: I think this is fine, but technically we could have multiple organisms in one reference so it is technically correct to instead  add this to the `add()` method instead, so you can set it per-contig.  You could also create a method on `ReferenceBuilder` to set such meta data, since a reference to a `ReferenceBuilder` is returned by the `ReferenceSetBuilder.add` method.  I guess having one assembly for all the contigs bugs me since I have lately been analyzing datasets where I map to multiple organisms (ex. microbiome).  Up to you though since I have no need and can change it later to prefer a per-contig assembly over the global one.

PR #361, opened 2018-02-06
Consensus calling does not trim correctly by insert size.

nh13:
I improved the trimming of source reads in consensus calling based on their insert size. I adjusted (increased) the insert size if I saw any 5' soft clipping (5' in sequencing order).  I waffled also adding any 5' soft clipping on the mate of the read (5' sequencing order), so I left it out for now.  I can see it work either way, but my reasoning not to include the mate's soft clipping was I don't want any bases in R1 to extend beyond mapped bases in R2.  Happy to discuss the implementation.

I also added a few more tests, since didn't have any tests for trimming based on insert size.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
@@ -216,7 +216,13 @@ trait UmiConsensusCaller[C <: SimpleRead] {
     // Find the last non-N base of sufficient quality in the record, starting from either the
     // end of the read, or the end of the insert, whichever is shorter!
     val len = {
-      var index = if (rec.isFrPair) min(abs(rec.insertSize), trimToLength) - 1 else trimToLength - 1
+      var index = if (!rec.isFrPair) trimToLength - 1 else {
+        // Adjust the insert size based on clipping on the 5' end of the given read.
+        // This *does not* consider any clipping on the 5' end of its mate.
+        val fivePrimeClipping  = cigar.takeWhile(_.operator.isClipping).filter(_.operator == CigarOperator.S).sumBy(_.length)
+        val adjustedInsertSize = abs(rec.insertSize) + fivePrimeClipping
+        min(adjustedInsertSize, trimToLength) - 1
+      }
       while (index >= 0 && (bases(index) == NoCall)) index -= 1
       index + 1
     }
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -216,7 +216,14 @@ trait UmiConsensusCaller[C <: SimpleRead] {
     // Find the last non-N base of sufficient quality in the record, starting from either the
     // end of the read, or the end of the insert, whichever is shorter!
     val len = {
-      var index = if (rec.isFrPair) min(abs(rec.insertSize), trimToLength) - 1 else trimToLength - 1
+      // Adjust the insert size based on clipping on the 5' end of the given read.  This *does not* consider any clipping
+      // on the 5' end of its mate.
+      val adjustedInsertSize = if (!rec.isFrPair) bases.length else {
```
tfenne: I don't love how this has a second check for `rec.isFrPair`, and then passes back the read length.  How about this instead for calculating `index`:

```scala
  // Adjust the insert size based on clipping on the 5' end of the given read.
  // This *does not* consider any clipping on the 5' end of its mate.
  val fivePrimeClipping  = cigar.takeWhile(_.operator.isClipping).filter(_.operator == CigarOperator.S).sumBy(_.length)
  val adjustedInsertSize = abs(rec.insertSize) + fivePrimeClipping
  min(adjustedInsertSize, trimToLength) - 1
}
```

tfenne: Also, note that this removes the `val cigar = ...` because that exact thing is already don on lines 199-203.

src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -88,8 +88,8 @@ class VanillaUmiConsensusCallerTest extends UnitSpec with OptionValues {
     val caller = cc(cco(errorRatePreUmi=PhredScore.MaxValue, minReads=1, minConsensusBaseQuality=0.toByte))
     val consensus = caller.consensusCall(Seq(source))
     consensus shouldBe 'defined
-    consensus.get.bases shouldBe source.bases
-    consensus.get.quals shouldBe source.quals
+    consensus.value.bases shouldBe source.bases
```
tfenne: Why make these changes from `get` to `value`?  That's invoking a scalatest-specific implicit to allow you to call `.value` on `Option` since it's only defined on `Some` not `None`.

PR #367, opened 2018-02-21
Improve error messages in ExtractUmisFromBam

nh13:


src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -161,7 +170,7 @@ class ExtractUmisFromBam
           r1.name = r1.name + "+" + bases1 + bases2
           r2.name = r2.name + "+" + bases1 + bases2
         }
-        assert(r1.name.equals(r2.name))
+        assert(r1.name.equals(r2.name), s"Mismatching read name.  R1: '$r1' R2: '$r2'")
```
tfenne: Just my 2c, but I feel like `assert` is for cases where you are asserting something that should _always_ be true and a failure indicates a bug, vs. `require` for cases where you are requiring a condition that could be made to fail by being given incorrect input.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -126,10 +126,19 @@ class ExtractUmisFromBam
     if (molecularIndexTags.contains(tag)) invalid(s"$tag specified as single-tag and also per-index tag")
   }
 
-  override def execute(): Unit = {
-    assert(molecularIndexTags1.length == rs1.molecularBarcodeSegments.length)
-    rs2.foreach(rs => assert(molecularIndexTags2.length == rs.molecularBarcodeSegments.length))
+  // Validate that the # of molecular barcodes in the read structure matches the tags
+  {
+    def validateMolecularBarcodesInReadStructureAndTag(rs: ReadStructure, tags: Seq[String]): Unit = {
+      validate(molecularIndexTags1.length == rs.molecularBarcodeSegments.length,
```
tfenne: `molecularIndexTags1.length` should be `tags` here

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -126,10 +126,19 @@ class ExtractUmisFromBam
     if (molecularIndexTags.contains(tag)) invalid(s"$tag specified as single-tag and also per-index tag")
   }
 
-  override def execute(): Unit = {
-    assert(molecularIndexTags1.length == rs1.molecularBarcodeSegments.length)
-    rs2.foreach(rs => assert(molecularIndexTags2.length == rs.molecularBarcodeSegments.length))
+  // Validate that the # of molecular barcodes in the read structure matches the tags
+  {
+    def validateMolecularBarcodesInReadStructureAndTag(rs: ReadStructure, tags: Seq[String]): Unit = {
```
tfenne: This (and the prior check) will fail under what appear to be valid circumstances.  E.g. if I run the program with `--read-structure 8M4S8M80T --molecular-index-tags RX` all the constructor validation will pass but it will create `molecularIndexTags1 == Seq("RX")`, which would then cause it to fail here.  I think line 117 needs adjusting to repeat the tag `rs.size` times.

tfenne: Not sure if it's clearer or not, but instead of a block with a function you could write something like:

```scala
for (stuff <- Seq(Some(rs1, molecularIndexTags1), rs2.map(r => (r, molecularIndexTags2))); (rs, tags) <- stuff) {
  validate(molecularIndexTags1.length == rs.molecularBarcodeSegments.length,
    s"Read structure '$rs' had '${rs.molecularBarcodeSegments.length}' molecular barcodes" +
    s"but '${tags.length}' molecular barcode tags expected (${tags.mkString(", ")}")
}
```

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
tfenne: Also, I think `validate` only works how you want it to if it's in the constructor code.  If we want it to work the same in the execute() method then I think we need to add another case to the catch in `FgBioMain` around ~104.

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
nh13: 1. I think we do want to support using `--molecular-index-tags` to store all the molecular barcodes in one tag, so I changed the validation and added a test. 
2. I don't like any of our solutions, but yours consumes the fewest characters so I went with that.

Issue #369, opened 2018-02-26
SampleBarcodeMetric should handle large volumes of data


nh13:
The `Int`s need to be `Long`.

nh13:
@tfenne looks like there are a lot of places (ex `FastqRecord`) where we have "read number" stored as integers.  A recent experience with a NovaSeq run has made this more urgent to not have overflow.

tfenne:
Two thoughts:

1. I'm trying to use `Metric.Count` going forward instead of `Int` or `Long` (and `Metric.Proportion` instead of `Float` or `Double`.  I think it's clearer for users (e.g. to see the type is a Count), and it's a typedef to Long so we can't forget and use Int.
2. FastqRecord's `readNumber` is meant to be 1 or 2 to indicate first or second of pair.  Unless Novaseq has some amazing new chemistry that should be ok as an `Int` 😉

nh13:
(2) doh, that's right!

PR #370, opened 2018-02-26
Use Metric.Count instead of Int for SampleBarcodeMetric.

nh13:
This adds support for there being more than Int.MaxValue reads, which
causes the proportions to be incorrectly computed.


nh13:
@tfenne I tested by bringing up the scala console with sbt, then loading in the metrics that had zero for the proportions (i.e. had overflow when counting the total # of reads in `SampleBarcodeMetric.finalizeMetric`), calling `SampleBarcodeMetric.finalizeMetric` again, then writing them out.

nh13:
Fixes #369

Issue #368, opened 2018-02-26
Verify data is sorted in TemplateCoordinate order in consensus calling tools


nh13:
- [ ] CallMolecularConsensusReads
- [ ] CallDuplexConsensusReads

We could also have it be a warning, or add an option to assume sorted...

PR #371, opened 2018-02-27
Added checks for sort order to consensus calling tools.

tfenne:
Fix for #368.  Checks that the sort order in the header makes sense for consensus calling.

I updated GroupReadsByUmi to correctly set the sort order in the header, it was only doing it partially before.  Because of that there are three levels of checking:

1. If the header matches what GroupReadsByUmi now produces, everything proceeds normally
2. If the header matches what GroupReadsByUmi used to produce, a _warning_ is emitted that the file _may not be_ sorted correctly
3. If the header matches any other sort/group order the programs fail with a clear message.

For (2) the output looks like this:
```
[2018/02/27 08:07:10 | CallMolecularConsensusReads | Warning] File /tmp/old_group_reads.bam may not be sorted correctly for consensus read generation.
```

For (3) the output looks like this:
```
[2018/02/27 08:16:45 | FgBioMain | Fatal] ########################################################################################################################################
[2018/02/27 08:16:45 | FgBioMain | Fatal] Execution failed!
[2018/02/27 08:16:45 | FgBioMain | Fatal] File tmp/resorted_by_samtools.bam is not sorted correctly. Please sort with fgbio SortBam -s TemplateCoordinate.
[2018/02/27 08:16:45 | FgBioMain | Fatal] ########################################################################################################################################
[2018/02/27 08:16:45 | FgBioMain | Info] CallDuplexConsensusReads failed. Elapsed time: 0.04 minutes.
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -473,8 +473,7 @@ class GroupReadsByUmi
     // Output the reads in the new ordering
     logger.info("Assigning reads to UMIs and outputting.")
     val outHeader = header.clone()
-    outHeader.setSortOrder(SortOrder.unsorted)
-    outHeader.setGroupOrder(GroupOrder.query)
+    SamOrder.TemplateCoordinate.applyTo(outHeader)
     val out = SamWriter(output, outHeader)
 
     val iterator = sorter.iterator.grouped(2).map { case Seq(x,y) => if (x.firstOfPair) (x, y) else (y, x) }.buffered // consume in pairs
--- a/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
@@ -123,12 +123,14 @@ object UmiConsensusCaller {
   }
 
   /**
-    * Helper method to check that the input BAM is in the correct order for consensus calling.
+    * Helper method to check that the input BAM is in the correct order for consensus calling. Will call
+    * the warn function if the sort order looks like it's probably compatible but it's not 100% sure.
+    * Will invoke the error function in cases where the sort order is definitely incompatible.
     *
     * @param header the header of the BAM file to be used for consensus calling
     * @param source a path or string representing the source of the header
     * @param warn a function to be called when any warnings are detected/emitted
-    * @param error a function to be called when any errors are encountered
+    * @param error a function to be called when any errors are encountered; should probably throw an exception!
     */
   def checkSortOrder(header: SAMFileHeader, source: Any, warn: String => Unit, error: String => Unit): Unit = {
     // Check that the SAM file is sorted appropriately
--- a/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
@@ -30,7 +30,7 @@ import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord}
 import com.fulcrumgenomics.commons.util.{Logger, SimpleCounter}
 import com.fulcrumgenomics.umi.UmiConsensusCaller._
 import com.fulcrumgenomics.util.NumericTypes.PhredScore
-import htsjdk.samtools.SAMFileHeader.GroupOrder
+import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
 import htsjdk.samtools.util.{Murmur3, SequenceUtil, TrimmingUtil}
 import htsjdk.samtools._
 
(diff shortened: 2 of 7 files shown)


tfenne:
Oops, forgot to commit the tests!  Should be fixed now.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -121,6 +121,27 @@ object UmiConsensusCaller {
...
+
+  /**
+    * Helper method to check that the input BAM is in the correct order for consensus calling.
+    *
+    * @param header the header of the BAM file to be used for consensus calling
+    * @param source a path or string representing the source of the header
+    * @param warn a function to be called when any warnings are detected/emitted
+    * @param error a function to be called when any errors are encountered
+    */
+  def checkSortOrder(header: SAMFileHeader, source: Any, warn: String => Unit, error: String => Unit): Unit = {
```
nh13: Could you add a return value (`Boolean`)?  While you pass in handlers (`warn`/`error`) for some cases, it would just be nice to have a return value that returns `true` when it _definitely_ is in the correct sort order.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -121,6 +121,27 @@ object UmiConsensusCaller {
...
     outHeader
   }
+
+  /**
+    * Helper method to check that the input BAM is in the correct order for consensus calling.
+    *
+    * @param header the header of the BAM file to be used for consensus calling
+    * @param source a path or string representing the source of the header
+    * @param warn a function to be called when any warnings are detected/emitted
+    * @param error a function to be called when any errors are encountered
```
nh13: It's a nice to have to describe when a warning/error occurs.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -473,8 +473,7 @@ class GroupReadsByUmi
     // Output the reads in the new ordering
     logger.info("Assigning reads to UMIs and outputting.")
     val outHeader = header.clone()
-    outHeader.setSortOrder(SortOrder.unsorted)
-    outHeader.setGroupOrder(GroupOrder.query)
+    SamOrder.TemplateCoordinate.applyTo(outHeader)
```
nh13: Just so I am clear, it was the sub-sort that was missing?

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -121,6 +121,27 @@ object UmiConsensusCaller {
...
+    if (!SamOrder(header).contains(SamOrder.TemplateCoordinate)) {
+      // Group reads used to output the header without the sub-sort, so allow this for now
+      if (header.getSortOrder == SortOrder.unsorted && header.getGroupOrder == GroupOrder.query) {
+        warn(s"File $source may not be sorted correctly for consensus read generation.")
+      }
+      else {
+        error(s"File $source is not sorted correctly. Please sort with fgbio SortBam -s TemplateCoordinate.")
+      }
+    }
+  }
```
nh13: Can you ada few tests?
- when the header is in `TemplateCoordinate` order returns `true`
- when the header is unsorted/query, returns `false` and `warn` is called
- otherwise, returns `false` and `error` is called

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
nh13: Also, what do you think defaulting `warn` and `error`:  `warn: String => Unit = logger.warn, error: String => Unit = fail`?  You'd need to mix in `LazyLogging` into the `object` if you'd agree.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Yeah.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
tfenne: I don't like returning Boolean because it's really three-state: ok, probably-ok and definitely-not-ok.  I honestly thought about just copy/pasting the code between the two consensus caller CLPs and creating a super-trait for the two of them, but it all seemed a bit over-engineered to get re-use of ~5 lines of code.

On the defaults ... `fail()` is a clp/FgBioMain thing, so it's not readily available in non CLP classes :/

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
tfenne: Done!

PR #372, opened 2018-02-27
Fix ExtractUmisFromBam when only one read has a UMI in it.

tfenne:
The code as it was previously would write a UMI with a trailing `-` in it in the new test.  This is because, when giving a single tag it would create `molecularIndexTags1` and `molecularIndexTags2` as `Seq(tag)`, and then compute an empty-string as the molecular barcode for the second read.

I think what I've done is a little clearer up front, but you can be the judge of that!

### DIFF
--- a/src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
@@ -29,6 +29,7 @@ import com.fulcrumgenomics.bam.api.{SamRecord, SamWriter}
 import com.fulcrumgenomics.sopt.cmdline.ValidationException
 import com.fulcrumgenomics.testing.{SamBuilder, UnitSpec}
 import com.fulcrumgenomics.util.ReadStructure
+import htsjdk.samtools.SAMRecord
 import org.scalatest.OptionValues
 
 /**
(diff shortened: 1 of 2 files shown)


src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
```diff
@@ -29,6 +29,7 @@ import com.fulcrumgenomics.bam.api.{SamRecord, SamWriter}
 import com.fulcrumgenomics.sopt.cmdline.ValidationException
 import com.fulcrumgenomics.testing.{SamBuilder, UnitSpec}
 import com.fulcrumgenomics.util.ReadStructure
+import htsjdk.samtools.SAMRecord
```
nh13: I don't think we need this.

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
```diff
@@ -383,6 +384,23 @@ class ExtractUmisFromBamTest extends UnitSpec with OptionValues {
...
+
+    new ExtractUmisFromBam(input=builder.toTempFile(), output=output, readStructure=Seq("8M+T", "+T"), molecularIndexTags=Seq("A1")).execute()
+    val recs = readBamRecs(output)
+    recs should have size 2
+    val r1Out = recs.find(_.firstOfPair).getOrElse(fail("Couldnt find R1 in the output"))
+    val r2Out = recs.find(_.secondOfPair).getOrElse(fail("Couldnt find R2 in the output"))
+    r1Out.length shouldBe 42
+    r2Out.length shouldBe 50
+    r1Out.get("A1") shouldBe Some(umi)
+    r2Out.get("A1") shouldBe Some(umi)
```
nh13: I'd prefer mixing in scala test `OptionValues` into the test class and then using:
```
r1Out.get("A1").value shouldBe umi
r2Out.get("A1").value shouldBe umi
```

I find using `.value` when we expect the `Option` to be defined to present a clearer error message when the test fails.

PR #376, opened 2018-02-28
Adding contains and getOrElse methods to SamRecord.

nh13:
Also added a few more tests

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -184,10 +184,13 @@ trait SamRecord {
   @inline final def readGroup: SAMReadGroupRecord = getReadGroup
 
   // Use apply/update for tag attributes
-  @inline final def apply[A](name: String): A              = getAttribute(name).asInstanceOf[A]
-  @inline final def get[A](name: String): Option[A]        = Option(apply(name))
-  @inline final def update(name: String, value: Any): Unit = setAttribute(name, value)
-  @inline final def attributes: Map[String,Any] = getAttributes.map(x => x.tag -> x.value).toMap
+  @inline final def apply[A](name: String): A                    = getAttribute(name).asInstanceOf[A]
+  @inline final def get[A](name: String): Option[A]              = Option(apply(name))
+  @inline final def update(name: String, value: Any): Unit       = setAttribute(name, value)
+  @inline final def attributes: Map[String,Any]                  = getAttributes.map(x => x.tag -> x.value).toMap
+  @inline final def getOrElse[A](name: String, default: => A): A = get(name).getOrElse(default)
+  @inline final def contains(name: String): Boolean              = hasAttribute(name)
+  @inline final def remove(name: String): Unit                   = setAttribute(name, null)
 
   // transient attributes
   @inline final def transientAttrs: TransientAttrs = new TransientAttrs(this)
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -184,10 +184,12 @@ trait SamRecord {
...
   // Use apply/update for tag attributes
-  @inline final def apply[A](name: String): A              = getAttribute(name).asInstanceOf[A]
-  @inline final def get[A](name: String): Option[A]        = Option(apply(name))
-  @inline final def update(name: String, value: Any): Unit = setAttribute(name, value)
-  @inline final def attributes: Map[String,Any] = getAttributes.map(x => x.tag -> x.value).toMap
+  @inline final def apply[A](name: String): A                 = getAttribute(name).asInstanceOf[A]
+  @inline final def get[A](name: String): Option[A]           = Option(apply(name))
+  @inline final def update(name: String, value: Any): Unit    = setAttribute(name, value)
+  @inline final def attributes: Map[String,Any]               = getAttributes.map(x => x.tag -> x.value).toMap
+  @inline final def getOrElse[A](name: String, default: A): A = get(name).getOrElse(default)
```
tfenne: do we want default to have type `=> A` so that it mimics `Option` and also allows you to do things like `rec.getOrElse("RG", throw new IllegalStateException("Booo!"))`?

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -184,10 +184,12 @@ trait SamRecord {
...
+  @inline final def contains(name: String): Boolean           = hasAttribute(name)
```
tfenne: While you're in here how about adding:
`@inline final def remove(name: String): Unit = setAttribute(name, null)`
to hide that ugliness from users of the API?

src/test/scala/com/fulcrumgenomics/bam/api/SamRecordTest.scala
```diff
@@ -27,9 +27,10 @@ package com.fulcrumgenomics.bam.api
 import com.fulcrumgenomics.alignment.Cigar
 import com.fulcrumgenomics.testing.SamBuilder.{Minus, Plus}
 import com.fulcrumgenomics.testing.{SamBuilder, UnitSpec}
-import htsjdk.samtools.{SAMRecord, TextCigarCodec}
+import htsjdk.samtools.TextCigarCodec
+import org.scalatest.OptionValues
 
-class SamRecordTest extends UnitSpec {
+class SamRecordTest extends UnitSpec with OptionValues {
```
tfenne: Should we just addd `OptionValues` into `UnitSpec`?

src/test/scala/com/fulcrumgenomics/bam/api/SamRecordTest.scala
```diff
@@ -106,12 +107,34 @@ class SamRecordTest extends UnitSpec {
     rec.asSam.getCigar shouldBe       null
   }
 
-  "SamRecord.apply" should "not return values for attributes that exist and not for those that don't exist" in {
+  it should "return values and information about SAM tags" in {
```
tfenne: Switching to `it` here is going to list test under `"SamRecord.cigar"` :/  How about:

```scala
"SamRecord" should "provide easy and safe access to extended attributes" in {
```

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: I debated having a static variable named `val NoTag: String = null` but that seemed lame.

src/test/scala/com/fulcrumgenomics/bam/api/SamRecordTest.scala
nh13: I am hesitant to roll to much into UnitSpec.  Lets leave as is for now.

PR #377, opened 2018-03-01
Add a method to expose the output sort of GroupReadsByUmi

nh13:
Also, `TemplateCoordinate` is not the right sort order for the output of `GroupReadsByUmi`, so we should also update the various consensus calling tests that use it, although there is not `SamOrder` for it, so we may be stuck.

### DIFF
--- a/src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
@@ -170,27 +170,6 @@ class GroupReadsByUmiTest extends UnitSpec {
     }
   }
 
-  {
-    "TemplateCoordinate SamOrder" should "sort pairs by the 'lower' 5' position of the pair" in {
-      val builder = new SamBuilder(readLength=100, sort=Some(SamOrder.Coordinate))
-      val exp = ListBuffer[SamRecord]()
-      // Records are added to the builder in the order that we expect them to be sorted, but the builder
-      // will coordinate sort them for us, so we can re-sort them and test the results
-      exp ++= builder.addPair("q1", contig=0, start1=100, start2=300)
-      exp ++= builder.addPair("q2", contig=0, start1=106, start2=300, cigar1="5S95M") // effective=101
-      exp ++= builder.addPair("q3", contig=0, start1=102, start2=299)
-      exp ++= builder.addPair("q4", contig=0, start1=300, start2=110, strand1=Minus, strand2=Plus)
-      exp ++= builder.addPair("q5", contig=0, start1=120, start2=320)
-      exp ++= builder.addPair("q6", contig=1, start1=1,   start2=200)
-
-      // Order they are added in except for q4 gets it's mate's flipped because of strand order
-      val expected = List("q1/1", "q1/2", "q2/1", "q2/2", "q3/1", "q3/2", "q4/2", "q4/1", "q5/1", "q5/2", "q6/1", "q6/2")
-      val actual   = builder.toList.sortBy(r => SamOrder.TemplateCoordinate.sortkey(r)).map(_.id)
-
-      actual should contain theSameElementsInOrderAs expected
-    }
-  }
-
   // Test for running the GroupReadsByUmi command line program with some sample input
   "GroupReadsByUmi" should "group reads correctly" in {
     val builder = new SamBuilder(readLength=100, sort=Some(SamOrder.Coordinate))
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
@@ -24,6 +24,7 @@
 
 package com.fulcrumgenomics.bam.api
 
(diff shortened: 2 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -104,6 +105,42 @@ object GroupReadsByUmi {
...
+      val r = ReadInfo(rec)
+      GroupReadsByUmiSortKey(refIndex=r.refIndex, start1=r.start1, start2=r.start2, strand1=r.strand1, strand2=r.strand2, library=r.library,
+        mid=rec(ConsensusTags.MolecularId), name=rec.name, secondOfPair=rec.secondOfPair)
+    }
+  }
+
+  // NB: use secondOfPair since false sorts before true
+  /** The sort key for [[GroupReadsByUmi]]. */
+  private case class GroupReadsByUmiSortKey(refIndex: Int, start1: Int, start2: Int, strand1: Boolean, strand2: Boolean, library: String,
+                                            mid: String, name: String, secondOfPair: Boolean)
```
nh13: NB: I could have `ReadInfo` as a value here, since I couldn't figure out some scala "diverging implicits" error.  Hence, I recreated all the values.  It would have been nicer to make `ReadInfo` extend `Ordered[ReadInfo]` and just have a reference to a `ReadInfo` object here, but alas, only so many hours.

src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
```diff
@@ -301,4 +301,38 @@ class GroupReadsByUmiTest extends UnitSpec {
...
+      val moleclarIdentifiers = seq(4, "1/A") ++ seq(4, "1/B") ++ seq(2, "2/A") ++ seq(2, "2/B") ++ seq(2, "0/A") ++ seq(2, "0/B")
+      recs.map(_.apply[String]("MI")) should contain theSameElementsInOrderAs moleclarIdentifiers
+      val names = Seq("ab1", "ab2", "ba1", "ba2", "ab3", "ba3", "ab0", "ba0").flatMap { name => seq(2, name)}
+      recs.map(_.name) should contain theSameElementsInOrderAs names
+      recs.grouped(2).foreach { case Seq(first, second) =>
+          first.firstOfPair shouldBe true
+          second.secondOfPair shouldBe true
+          first.name shouldBe second.name
+      }
+
```
nh13: This is for you.

tfenne:
I'm lacking a bunch of context here.  What's going on?

nh13:
@tfenne sometimes I would like to re-run consensus calling from a grouped BAM, but the grouped BAM I have is coordinate sorted order.  I'd rather avoid running `GroupReadsByUmi` again to just get the reads in the same order.  Additionally, I thought it was confusing to make a full-fledged `SamOrder`.  Hence just a method here like in `Bams`.

tfenne:
I see.  I think this is a more general problem though - it would be useful to take grouped data that has been sorted, and be able to consensus call it.  I think there area  couple of other ways to approach this that would be nicer:

1. We could just adapter the consensus callers so that internally they take data in `TemplateCoordinate` order (which GroupReadsByUmi output satisfies), consumes reads that have the same template coordinates, and the groups them by `MI` for consensus calling.
2. We could extend the definition of `TemplateCoordinate` to also include the `MI` tag if it's present on the reads, since `TemplateCoordinate` was created more or less for this purpose - to document and expose the sort order we need for grouping and consensus building.

nh13:
`TemplateCoordinate` sort inserts the read name as a key prior to library, where as for "group-reads-by-umi" order we want to sort by molecular identifier, name, then first/second of pair.  So your second suggestion doesn't technically work without either breaking backwards compatibility with `TemplateCoordinate`, and making it not really `template-coordinate`, since we could have reads with the same name ("template") not next to each other, though you could argue all reads with the same MI are from the same template. 

How about I pursue your first suggestion?

tfenne:
I'm open to either route.  I would point out though that according to the SAM spec any reads with the same name are presumed to be from the same template, therefore the same library.  So only a non-compliant SAM would care whether library or read name comes first in the sorting.

nh13:
@tfenne I took a shot at fixing the `TemplateCoordinate` `SamOrder` to match the output of `GroupReadsByUmi`.  For reads that have the same template coordinates, I found that `GroupReadsByUmi` would output the reads from the same library consecutively.  Similarly, for reads that have the same template coordinates, I found that `GroupReadsByUmi` would output the reads with the same molecular ID consecutively.  We discussed this later issue, but this former issue is also fixed.  You can confirm the fix by looking at how `ReadInfo` was structured (it includes library as part of its key), and look at when `GroupReadsByUmi` receives the batch of records who have the same template coordinates, how it sorts by molecular id length then value.

nh13:
Also, I took at look at using an implicit `Ordering`, but was dismayed to find that (1) there is no `Tuple10` in [Ordering.scala](https://github.com/scala/scala/blob/v2.12.4/src/library/scala/math/Ordering.scala#L1), and (2) that the implementation of the `compare` method is the same as you had implemented already.  So some unnecessary object creation (tuples).

PR #379, opened 2018-03-02
Allow glocal alignment of longer query to shorter target.

tfenne:
@nh13 I thought I was being clever requiring the query to be <= the target for glocal alignment, and that it would help catch bugs where query and target got reversed.  But of course it is valid to align a longer query to a shorter target, and this comes up frequently in the case on long insertions.

### DIFF
--- a/src/test/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAlignerTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAlignerTest.scala
@@ -278,7 +278,10 @@ class NeedlemanWunschAlignerTest extends UnitSpec {
     result.score shouldBe 4
   }
 
-  it should "fail if the target sequence is shorter than the query sequence" in {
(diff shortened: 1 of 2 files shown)


src/test/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAlignerTest.scala
```diff
@@ -278,7 +278,10 @@ class NeedlemanWunschAlignerTest extends UnitSpec {
     result.score shouldBe 4
   }
 
-  it should "fail if the target sequence is shorter than the query sequence" in {
-    an[Exception] shouldBe thrownBy { NeedlemanWunschAligner(1, -1, -3, -1, mode=Glocal).align("ACGTACGTACGT", "ACGT") }
+  it should "align a longer query to the whole target with an insertion" in {
```
nh13: `align a longer query with an insertion to the whole target`?  It have to convince myself an insertion is always query relative to target otherwise.

tfenne: I'll add some commas ;)

PR #382, opened 2018-03-03
ReviewConsensusVariants updates

nh13:
Adds NA for empty filters: this makes it easier for Unix utilities to consume.

Properly format the output base

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
@@ -80,7 +80,7 @@ object ReviewConsensusVariants {
   ( // First set is all the same for each variant
     chrom: String, pos: Int, ref: String, genotype: String, filters: String, A: Int, C: Int, G: Int, T: Int, N: Int,
     // Second set is per-consensus reads
-    consensus_read: String, consensus_insert: String, consensus_call: String, consensus_qual: Int, a: Int, c: Int, g: Int, t: Int, n: Int
+    consensus_read: String, consensus_insert: String, consensus_call: Char, consensus_qual: Int, a: Int, c: Int, g: Int, t: Int, n: Int
   ) extends Metric
 
   /** Extracts the molecular identifier minus any trailing /A etc. */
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -253,21 +253,27 @@ class ReviewConsensusVariants
...
+            filters  = filters,
             A = consensusCounts.a,
             C = consensusCounts.c,
             G = consensusCounts.g,
             T = consensusCounts.t,
             N = consensusCounts.n,
             consensus_read = consensusReadName,
             consensus_insert = toInsertString(rec),
-            consensus_call = c.getReadBase.toString.toUpperCase,
+            consensus_call = c.getReadBase.toChar.toUpper,
```
nh13:
```
scala> 84.toByte.toString.toUpperCase
res0: String = 84
scala> 84.toChar.toUpper
res1: Char = T
```

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
```diff
@@ -253,21 +253,27 @@ class ReviewConsensusVariants
           val mi = toMi(rec)
           val consensusReadName = c.getRecord.getReadName + readNumberSuffix(rec)
           val rawCounts = BaseCounts(rawByMiAndReadNum(mi + readNumberSuffix(rec)))
-
-          val m =  ConsensusVariantReviewInfo(
+          val filters = {
+            variant.filters.getOrElse("") match {
+              case "" => "NA"
```
tfenne: I don't love this.  I usually pop these files open in excel, and it's much easier to differentiate empty vs. non-empty.

Also, if we're going to put anything here I would think `PASS` or `.` to mirror what VCF does would be clearer.

tfenne: Also, isn't this a much more verbose way of writing `variant.filters.getOrElse("NA")`?  If so, you could just do that and move it back inline.  Of it `variant.filters` sometimes contains `Some("")` then I'd say we should fix that and or do `variant.filters.filter(_.nonEmpty).getOrElse("NA")`.

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
tfenne: Huh, how did that ever work?

src/main/scala/com/fulcrumgenomics/umi/ReviewConsensusVariants.scala
nh13: The latter is the problem I had, which was `variant.filters` sometimes returns `Some("")`, so I have changed it to return "PASS"

nh13:
back to you @tfenne

PR #384, opened 2018-03-04
Made sure 0.0 always formats as 0.0, not 0E0.

tfenne:


src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -198,7 +198,8 @@ trait Metric extends Product with Iterable[(String,String)] {
     case d: Iso8601Date => d.toString
     case d: Date        => Metric.DateFormat.synchronized { Metric.DateFormat.format(d) }
     case f: Float       => formatValue(f.toDouble)
-    case d: Double if d < 0.00001 => Metric.SmallDoubleFormat.synchronized { Metric.SmallDoubleFormat.format(d) }
+    case d: Double if d != 0.0 && d < 0.00001 && d > -0.00001 =>
+      Metric.SmallDoubleFormat.synchronized { Metric.SmallDoubleFormat.format(d) }
```
nh13: Can you fix the indent on 203? I won’t pretend to know why BigDoubleFormat formats it as zero but ok.

PR #386, opened 2018-03-09
Fix the NeedlemanWunschAligner to support affine gap penalties.

nh13:
Using only one scoring matrix causes us to not align gaps correctly and
return a sub-optimal alignment.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
+++ b/src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
@@ -66,7 +66,27 @@ object NeedlemanWunschAligner {
   val Up  : Direction     = 2.toByte
   val Diagonal: Direction = 3.toByte
   val Done: Direction     = 4.toByte
+
+  // NB: the order of LeftAndDiagonal and UpAndDiagonal matters when breaking ties!
+  val AllDirections: Seq[Direction]   = Seq(Diagonal, Left, Up)
+  val LeftAndUp: Seq[Direction]       = Seq(Left, Up)
+  val LeftAndDiagonal: Seq[Direction] = Seq(Diagonal, Left)
+  val UpAndDiagonal: Seq[Direction]   = Seq(Diagonal, Up)
+
+  /** The minimum score allowed to start an alignment.  This prevents underflow. */
+  val MinStartScore: Int = Int.MinValue / 2
 }
+
+
+object AlignmentMatrix {
+  def apply(direction: Direction, queryLength: Int, targetLength: Int): AlignmentMatrix = {
+    AlignmentMatrix(direction=direction, scoring=Matrix[Int](queryLength+1, targetLength+1), trace=Matrix[Direction](queryLength+1, targetLength+1))
+  }
+}
+
+/** A single alignment matrix for a given [[Direction]] storing both the scoring and traceback matrices produce by the aligner. */
+case class AlignmentMatrix(direction: Direction, scoring: Matrix[Int], trace: Matrix[Direction])
+
 
 /**
   * Implementation of the Needleman-Wunsch algorithm for global alignment of two sequences with support for an
@@ -98,8 +118,8 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
     * @return an [[Alignment]] object describing the optimal global alignment of the two sequences
     */
   def align(query: Array[Byte], target: Array[Byte]): Alignment = {
-    val (scoring, trace) = buildMatrices(query, target)
-    generateAlignment(query, target, scoring, trace)
+    val matrices = buildMatrices(query, target)
+    generateAlignment(query, target, matrices)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


tfenne:
@nh13 I've added three commits as I make the code progressively more verbose and also faster.  

I'm using the Global mode aligner in a program where I'm aligning tens of thousands or ~100bp queries to a ~100bp target, using multiple threads.  Before this PR I was at 3-4 seconds per 25k alignments.  After the first commit it shot up to 40-45s per 25k.  By my last commit I'm back down to 8-9s per 25k.  So somewhere between 2-3 slower, which seems reasonable to me since we're building up three sets of matrices now.

tfenne:
Oh, also, fwiw, over 90% of the compute time is spent in self-time of `buildMatrices()` which indicates it is in the filling in of the core (not the top or left-most rows).  So unless you see a way to make that faster, I think we're at or near the limit.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -116,66 +136,132 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
     * @param query the query sequence
     * @param target the target sequence
-    * @return a tuple of the scoring matrix and the traceback matrix
+    * @return an array of alignment matrices, where the indices to the array are the Directions
     */
-  protected def buildMatrices(query: Array[Byte], target: Array[Byte]): (Matrix[Int], Matrix[Direction]) = {
-    val scoring = Matrix[Int](query.length + 1, target.length + 1)
-    val trace   = Matrix[Direction](query.length + 1, target.length + 1)
+  protected def buildMatrices(query: Array[Byte], target: Array[Byte]): Array[AlignmentMatrix] = {
+    val matrices = Range.inclusive(0, 2).map(i => AlignmentMatrix(direction=i.toByte, queryLength=query.length, targetLength=target.length)).toArray
```
nh13: Why `Range.inclusive` and not `AllDirections`?

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -116,66 +136,132 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
+
+    val (leftScoreMatrix, leftTraceMatrix, upScoreMatrix, upTraceMatrix, diagScoreMatrix, diagTraceMatrix) = {
```
nh13: Perhaps a comment to say that having direct references is faster for the core loop below?

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -116,66 +136,132 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
-          trace(i,j)   = Done
+    var i = 1
+    while (i <= query.length) {
+      var j = 1
+      while (j <= target.length) {
+        { // Diagonal matrix can come from the previous diagonal, up, or left
+          val addend = scoringFunction(query(i-1), target(j-1))
+          val dScore = diagScoreMatrix(i-1, j-1) + addend
+          val lScore = leftScoreMatrix(i-1, j-1) + addend
+          val uScore = upScoreMatrix(i-1, j-1)   + addend
```
nh13: You can move the `+ addend` to the respective if/else if/else if/ etc. Just make  sure that `dScore < 0` becomes `dScore + addend < 0` in the local mode `if`.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -116,66 +136,132 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
-        // For local mode, we can start anywhere
-        if (mode == Local && scoring(i,j) < 0) {
-          scoring(i,j) = 0
-          trace(i,j)   = Done
+    var i = 1
+    while (i <= query.length) {
+      var j = 1
+      while (j <= target.length) {
+        { // Diagonal matrix can come from the previous diagonal, up, or left
+          val addend = scoringFunction(query(i-1), target(j-1))
```
nh13: Could cache/store `query(i-1)` in a local variable outside the inner loop.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -116,66 +136,132 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
+          else {
+            diagScoreMatrix(i, j) = uScore
+            diagTraceMatrix(i, j) = Up
+          }
+        }
+
+
+        { // Up matrix can come from diagonal or up
+          val dScore = gapOpen + gapExtend + diagScoreMatrix(i-1, j)
+          val uScore =           gapExtend + upScoreMatrix(i-1, j)
```
nh13: Can defer the `gapExtend` addition until you update the `upScoreMatrix`.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -116,66 +136,132 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
+          else {
+            upScoreMatrix(i, j) = uScore
+            upTraceMatrix(i, j) = Up
+          }
+        }
+
+
+        { // Left matrix can come from diagonal or left
+          val dScore = gapOpen + gapExtend + diagScoreMatrix(i, j-1)
+          val lScore =           gapExtend + leftScoreMatrix(i, j-1)
```
nh13: ditto, could defer one addition here.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
```diff
@@ -116,66 +136,132 @@ class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
...
+          if (dScore >= lScore) {
+            leftScoreMatrix(i, j) = dScore
+            leftTraceMatrix(i, j) = Diagonal
+          }
+          else {
+            leftScoreMatrix(i, j) = lScore
+            leftTraceMatrix(i, j)   = Left
```
nh13: misalignment

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Doh.  Because I did something else first and didn't revisit.  Fixed.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: I'm going to leave this one as is as.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Ditto!

src/main/scala/com/fulcrumgenomics/alignment/NeedlemanWunschAligner.scala
tfenne: Fixed.

PR #388, opened 2018-03-10
Rename NeedlemanWunschAligner to Aligner since it does so much more now.

tfenne:
@nh13 Just a bit of tidying up to rename NW to `Aligner`.  I've used a package object to maintain backwards compatibility, and verified that the tests continued to work _before_ refactoring those to use `Aligner`.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
+++ b/src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
@@ -26,17 +26,16 @@ package com.fulcrumgenomics.alignment
 
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.alignment.Mode.{Global, Glocal, Local}
-import com.fulcrumgenomics.alignment.NeedlemanWunschAligner._
+import com.fulcrumgenomics.alignment.Aligner._
 import enumeratum.EnumEntry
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
```diff
@@ -49,61 +48,65 @@ object Mode extends FgBioEnum[Mode] {
...
   *                      else use the M operator for both.
   * @param mode alignment mode to use when generating alignments
   */
-class NeedlemanWunschAligner(val scoringFunction: (Byte,Byte) => Int,
-                             val gapOpen: Int,
-                             val gapExtend: Int,
-                             useEqualsAndX: Boolean = true,
-                             val mode: Mode = Global) {
+class Aligner(val scoringFunction: (Byte,Byte) => Int,
+              val gapOpen: Int,
```
tfenne: Question for you @nh13.  Do you think there's ever a valid case for providing `gapOpen` or `gapExtend` > 0?  I frequently get confused as some aligners take the penalties as positive numbers and subtract them, and some (as we do) take negative numbers and add them.  I'd like to add a pair of:

```scala
require(gapOpen <= 0, "...")
require(gapExtend <= 0, "...")
```

but wanted to see if you thought that was a bad idea for any reason?

PR #408, opened 2018-06-01
Fixed FastqSource to handle read numbers > 2.

tfenne:
A tiny little PR for you @nh13.  I was running into problems with picard's `IlluminaBasecallsToFastq` generating files with a `/3` suffix, and our `FastqToBam` was claiming they were out of sync with the other files with `/1` and `/2` suffixes.  The real change is in `FastqSource`, with a small _personal_ CLP that removes read numbers when they are inconvenient ;)

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
@@ -112,7 +112,8 @@ class FastqSource private(val lines: Iterator[String],
           case i  => (header.substring(1, i), Some(header.substring(i+1)))
         }
 
-        val (name, readNumber) = fullName.endsWith("/1") || fullName.endsWith("/2") match {
+        val suffix = fullName.takeRight(2)
+        val (name, readNumber) = suffix.length == 2 && suffix(0) == '/' && suffix(1).isDigit match {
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/fastq/FastqSource.scala
```diff
@@ -112,7 +112,8 @@ class FastqSource private(val lines: Iterator[String],
           case i  => (header.substring(1, i), Some(header.substring(i+1)))
         }
 
-        val (name, readNumber) = fullName.endsWith("/1") || fullName.endsWith("/2") match {
+        val suffix = fullName.takeRight(2)
+        val (name, readNumber) = suffix.length == 2 && suffix(0) == '/' && suffix(1).isDigit match {
           case true  => (fullName.dropRight(2), Some(fullName.last.asDigit))
           case false => (fullName, None)
         }
```
nh13: How about defining a val in the object:
```
val ReadNumberPattern = """.*(/\d)$""".r
```
then on line 115:
```
val (name, readNumber) = fullName match {
  case FastqSource.ReadNumberPattern(_name, _readNumber) => (_name, Some(_readNumber))
  case _ => (fullName, None)
}
```

PR #412, opened 2018-06-26
Improve the error message when validating a sample sheet.

nh13:
In particular, when we have two or more samples with the same
combination of sample name, library id, and lane.

Also added in a few more unit tests to cover such cases.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
+++ b/src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
@@ -152,13 +152,18 @@ object SampleSheet {
   protected[illumina] val SplitRegex = ",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)"
 
   /** Validate that the sample ids are unique, and that the combination of sample name and library id are unique. */
-  private def validateSamples(samples: Seq[Sample]): Unit = {
+  private def validateSamplesFromTheSameLane(samples: Seq[Sample]): Unit = {
+    // All samples should either be from the same lane or have no lane
+    require(samples.flatMap(_.lane).distinct.lengthCompare(1) <= 0)
+    /** We include the lane if present in the sample key for debugging purposes*/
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -153,12 +153,14 @@ object SampleSheet {
 
   /** Validate that the sample ids are unique, and that the combination of sample name and library id are unique. */
   private def validateSamples(samples: Seq[Sample]): Unit = {
+    def sampleKey(s: Sample): String = s"${s.sampleName} ${s.libraryId}" + s.lane.map(" " + _).getOrElse("")
```
nh13: If the lane is present on the sample, print it out on the error message.  Note that due to how this method is called in the `SampleSheet` constructor (183-186), only samples from the same lane are validated together.

tfenne: Is this used only for debugging?  If so a one-line scaladoc explaining that would be good.

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -153,12 +153,14 @@ object SampleSheet {
...
+
     // Validate some properties for a sample sheet
     require(samples.map(_.sampleId).toSet.size == samples.length,
       "Sample identifiers were not unique: " + samples.groupBy(_.sampleId).filter(_._2.length > 1).keys.mkString(", "))
-    require(samples.map(s => s"${s.sampleName} ${s.libraryId}").toSet.size == samples.length,
+    require(samples.map(sampleKey).toSet.size == samples.length,
```
tfenne: This would probably be neater as:

```scala
samples.groupBy(sampleKey).filter(_._2.length > 1).keys match {
  case Seq() => Unit
  case dupes => illegal(s"Sample name and library identifier combinations were not unique: $dupes")
}
```

Issue #416, opened 2018-07-06
java.lang.AssertionError: assertion failed: Reads out of order


ProfessionalFarmer:
Hi, dear developers.
I met some error when using fgbio's GroupReadsByUmi. It prompted that reads out of order. Even I processed data by Picard's RevertSam with SANITIZE=true and all the REMOVE/RESTORE options set to false, the error still exist.

Exception in thread "main" java.lang.AssertionError: assertion failed: Reads out of order @ M4988:25:000000000-G244L:1:2103:19288:2268/2 + M4988:25:000000000-G244L:1:2101:2860:20442/2
        at scala.Predef$.assert(Predef.scala:219)
        at com.fulcrumgenomics.umi.GroupReadsByUmi.$anonfun$execute$16(GroupReadsByUmi.scala:486)
        at com.fulcrumgenomics.umi.GroupReadsByUmi.$anonfun$execute$16$adapted(GroupReadsByUmi.scala:485)
        at scala.collection.immutable.List.foreach(List.scala:389)
        at com.fulcrumgenomics.umi.GroupReadsByUmi.execute(GroupReadsByUmi.scala:485)
        at com.fulcrumgenomics.cmdline.FgBioMain.makeItSo(FgBioMain.scala:99)
        at com.fulcrumgenomics.cmdline.FgBioMain.makeItSoAndExit(FgBioMain.scala:80)
        at com.fulcrumgenomics.cmdline.FgBioMain$.main(FgBioMain.scala:48)
        at com.fulcrumgenomics.cmdline.FgBioMain.main(FgBioMain.scala)

The reads information in bam as bellow:
[...]

Issue #413, opened 2018-07-02
Assertion error: reads out of order 


sheenams:
I'm running into this error with GroupReadsByUmi after using UmiAwareMarkDuplicatesWithMateCigar. I followed the steps from the closed issue #87 , but am having slightly different issues than that user. 

I've run MergeBamAlignment on the unaligned and aligned bam, to create a 'UMI-marked-bam'. Then, I can take that UMI-marked-bam straight to GroupReadsByUmi and it works. However, if I use UmiAwareMarkDuplicatesWithMateCigar first (to create a duplicate-marked-bam, duplicates only marked, not removed), it fails with the following error:

```[2018/06/30 01:06:30 | FgBioMain | Info] Executing GroupReadsByUmi from fgbio version 0.6.1 as sheenams@larf on JRE 1.8.0_171-8u171-b11-1~deb9u1-b11 with snappy
[2018/06/30 01:06:30 | GroupReadsByUmi | Info] Filtering and sorting input.
[2018/06/30 01:06:42 | GroupReadsByUmi | Info] Sorted     1,000,000 records.  Elapsed time: 00:00:12s.  Time for last 1,000,000:   12s.  Last read position: 9:21,968,715
[2018/06/30 01:06:55 | GroupReadsByUmi | Info] Sorted     2,000,000 records.  Elapsed time: 00:00:24s.  Time for last 1,000,000:   12s.  Last read position: 22:29,091,612
[2018/06/30 01:06:55 | GroupReadsByUmi | Info] Accepted 2,049,604 reads for grouping.
[2018/06/30 01:06:55 | GroupReadsByUmi | Info] Filtered out 388,430 non-PF reads.
[2018/06/30 01:06:55 | GroupReadsByUmi | Info] Filtered out 117,332 reads that were not part of a high confidence FR mapped read pair.
[2018/06/30 01:06:55 | GroupReadsByUmi | Info] Filtered out 0 reads that contained one or more Ns in their UMIs.
[2018/06/30 01:06:55 | GroupReadsByUmi | Info] Assigning reads to UMIs and outputting.
[2018/06/30 01:06:56 | FgBioMain | Info] GroupReadsByUmi failed. Elapsed time: 0.49 minutes.
Exception in thread "main" java.lang.AssertionError: assertion failed: Reads out of order @ MA0337:1:1107:25428:6890/2 + MA0337:1:1105:21989:24121/2
        at scala.Predef$.assert(Predef.scala:219)
[...]

Issue #394, opened 2018-04-03
FgBioMain | Info] GroupReadsByUmi failed.


LeanGreen:
After picard, fgbio SetMateInformation, when I use **## fgbio-0.5.1.jar GroupReadsByUmi** -s Adjacency, it broke.
Like this:
```
[2018/04/03 11:16:51 | FgBioMain | Info] Executing GroupReadsByUmi from fgbio version 0.5.1 as jbk@localhost.localdomain on JRE 1.8.0_162-b12 with snappy
[2018/04/03 11:16:51 | GroupReadsByUmi | Info] Filtering and sorting input.
[2018/04/03 11:16:51 | FgBioMain | Info] GroupReadsByUmi failed. Elapsed time: 0.07 minutes.
Exception in thread "main" java.lang.NullPointerException
        at com.fulcrumgenomics.umi.GroupReadsByUmi.$anonfun$execute$8(GroupReadsByUmi.scala:457)
        at com.fulcrumgenomics.umi.GroupReadsByUmi.$anonfun$execute$8$adapted(GroupReadsByUmi.scala:457)
        at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:502)
        at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:500)
        at scala.collection.Iterator.foreach(Iterator.scala:929)
        at scala.collection.Iterator.foreach$(Iterator.scala:929)
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1417)
        at com.fulcrumgenomics.umi.GroupReadsByUmi.execute(GroupReadsByUmi.scala:465)
        at com.fulcrumgenomics.cmdline.FgBioMain.makeItSo(FgBioMain.scala:99)
        at com.fulcrumgenomics.cmdline.FgBioMain.makeItSoAndExit(FgBioMain.scala:80)
        at com.fulcrumgenomics.cmdline.FgBioMain$.main(FgBioMain.scala:48)
        at com.fulcrumgenomics.cmdline.FgBioMain.main(FgBioMain.scala)
```
How can I solve this problem?

nh13:
Can you confirm your reads have the `RX` tag present, or the custom tag if you specified the raw-tag with `--raw-tag`?  I'd bet a $1 the `RX` tag is missing from one of your reads.  On the other hand, the tool should do a better job of providing an error message (it was supposed to); I have a fix here: https://github.com/fulcrumgenomics/fgbio/pull/406

PR #419, opened 2018-07-09
Improve error message when GroupReadsByUmi produces out of order reads.

nh13:
See:
- https://github.com/fulcrumgenomics/fgbio/issues/413
- https://github.com/fulcrumgenomics/fgbio/issues/416
- https://github.com/fulcrumgenomics/fgbio/issues/394

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -427,7 +427,8 @@ class GroupReadsByUmi
 )extends FgBioTool with LazyLogging {
   import GroupReadsByUmi._
 
-  require(this.minUmiLength.forall(_ => this.strategy != Strategy.Paired), "Paired strategy cannot be used with --min-umi-length")
+  validate(this.minUmiLength.forall(_ => this.strategy != Strategy.Paired), "Paired strategy cannot be used with --min-umi-length")
+  validate(this.rawTag != this.assignTag, s"The input raw UMI tag '$rawTag' cannot be the same as the output assigned UMI TAG")
 
   private val assigner = strategy.newStrategy(this.edits)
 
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -514,6 +514,20 @@ class GroupReadsByUmi
     }
   }
 
+  /** The error message when reads are out of order. */
+  private def getReadsOutOfOrderErrorMessage(r1: SamRecord, r2: SamRecord): String = {
+    val errorMessage = if (this.rawTag == "MI") {
+      s"Using 'MI' tag as the 'raw tag', are you sure this is correct?  Did you mean the 'RX' tag?"
```
tfenne: Should we just check this at startup and emit a warning?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -514,6 +514,20 @@ class GroupReadsByUmi
...
+    }
+    else if (this.rawTag == this.assignTag) {
```
tfenne: This should, I think, be a validation error at startup.  I don't think we should allow this.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
fleharty: @tfenne What UmiAwareMarkDuplicatesWithMateCigar is doing does appear to be very wrong based on what you've shown.  Could you share that bam, I would like to try to fix this issue in Picard.

tfenne: @fleharty I don't think I can share that BAM, but I can probably use what I figured out to generate a small test case that I can pass along.  I'll try and do that in the next couple of days.

fleharty: @tfenne I just wanted to follow up on this.  Could you generate the small test case?
Thanks!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: Done

tfenne:
I think the upshot of this PR is that the problem lies elsewhere.  I'm closing this PR because I don't think it solves the stated problem, that problem lies in another program not giving the same `MI` value to both ends of a pair.

fleharty:
@tfenne If you get a chance could you show me an example of where this happens, I've not seen this happen so if there is an example somewhere I may be able to fix it.

PR #421, opened 2018-07-28
Various updates

nh13:
@tfenne here are some various updates I have been accumulating.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
@@ -40,7 +40,7 @@ case class ProgressLogger(logger: Logger,
   }
 
   /** Method to use to record progress when genome location isn't available or relevant. */
-  def record(): Unit = record(null, -1)
+  def record(): Boolean = record(null, -1)
 
-  def record(rec: SamRecord): Unit = record(rec.asSam)
+  def record(rec: SamRecord): Boolean = record(rec.asSam)
 }
--- a/src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
@@ -43,4 +43,8 @@ case class ProgressLogger(logger: Logger,
   def record(): Boolean = record(null, -1)
 
   def record(rec: SamRecord): Boolean = record(rec.asSam)
+
+  /** Logs the last record if it wasn't already logged. */
+  def logLast(): Boolean = super.log() // Calls the super's log() method, not log(message: String*)
+
 }
--- a/src/main/scala/com/fulcrumgenomics/util/Sequences.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/Sequences.scala
@@ -27,6 +27,9 @@
 package com.fulcrumgenomics.util
 
 import com.fulcrumgenomics.FgBioDef._
+import htsjdk.samtools.util.SequenceUtil
+
+
 /**
   * Utility methods for working with DNA or RNA sequences
   */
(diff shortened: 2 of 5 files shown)


src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -55,25 +58,73 @@ object Sequences {
...
+    * @param s a DNA or RNA sequence
+    * @return the offset and length of the longest dinucleotide sequence
+    */
+  def maxDinuc(s: String) : OffsetAndLength = {
+    val bs = s.toUpperCase.getBytes
+    var (bestStart, bestLength) = (0,0)
+    forloop(0, bs.length-1) { start =>
+      val (b1, b2) = (bs(start), bs(start+1))
+      var i = start
+      while (i < s.length-1 && SequenceUtil.basesEqual(b1, bs(i)) && SequenceUtil.basesEqual(b2, bs(i+1))) i += 2
```
tfenne: Two thoughts:
1. If you upper case like you're doing on 105 you don't really need to use `SequenceUtil.basesEqual()` here.
2. If you continue to use `SequenceUtil.basesEqual()` I'd be tempted to define and use the following within this class:

```scala
/** Checks to see if the two characters are identical, ignoring case. */
@inline final private def same(a: Char, b: Char): Boolean = {
  Character.toUpperCase(a) == Character.toUpperCase(b)
}
```

This has the advantage that a) it'll work for all characters so if you try to use it on proteins it won't fail, and also will just make the usage more compact and easier to see what's going on.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -55,25 +58,73 @@ object Sequences {
...
+    OffsetAndLength(bestStart, bestLength)
+  }
+
+  /** Reverse complements a string of bases. */
+  def reverseComplement(s: String): String = SequenceUtil.reverseComplement(s)
+
+  /** Returns the sequence that is the complement of the provided sequence. */
+  def complement(s: String): String = {
+    val bs = s.getBytes
+    for (i <- 0 until bs.length) bs(i) = SequenceUtil.complement(bs(i))
```
tfenne: I think `forloop(from=0, until=bs.length)` will be much faster here.

src/test/scala/com/fulcrumgenomics/util/SequencesTest.scala
```diff
@@ -48,25 +49,46 @@ class SequencesTest extends UnitSpec {
...
   }
 
-  it should "fail when passed a zero length string" in {
-    an[Throwable] shouldBe thrownBy { Sequences.longestHomopolymer("") }
+  "Sequences.maxPolyX" should "fail when passed a null String" in {
+    an[Throwable] shouldBe thrownBy { Sequences.maxPolyX(null) }
+  }
+
+  it should "return zero when passed a zero length string" in {
+    Sequences.maxPolyX("") shouldBe OffsetAndLength(0, 0)
```
tfenne: Why would you want this behavior?  Also, I don't think it's reasonable to deprecate one method in favor of another if the behavior is different.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -55,25 +58,73 @@ object Sequences {
...
+  case class OffsetAndLength(offset: Int, length: Int)
+
+  /**
+    * Returns the offset (0-based) and length of the longest homopolymer in the sequence. In the case
+    * that there are multiple homopolymers of the same length, the earliest one is returned.
+    *
+    * @param s a DNA or RNA sequence
+    * @return the offset and length of the longest homopolymer
+    */
+  def maxPolyX(s: String) : OffsetAndLength = {
```
tfenne: Two thoughts on naming (apply to maxDinux as well):
1) I think longest is clearer than max.  Longest can only really be interpreted one way.  Max could be interpreted to mean "longest single", or "most abundant" such that in a sequence like `ACACGTGTGTACAC` you might say `AC` is the "max dinuc" because there is more of it.
2) Homopolymer is a well defined term whereas  `polyX` isn't really.

src/test/scala/com/fulcrumgenomics/util/SequencesTest.scala
```diff
@@ -48,25 +49,46 @@ class SequencesTest extends UnitSpec {
...
+    Sequences.maxDinuc("GGGGGGGG") shouldBe OffsetAndLength(0, 8)
+    Sequences.maxDinuc("AGCGtagCGCGCgcGCTCTCTatCGCGCA") shouldBe OffsetAndLength(6, 10)
+  }
+
+  "Sequences.complement" should "return the complement of sequences" in {
+    Sequences.complement("AAA") shouldBe "TTT"
+    Sequences.complement("ACAC") shouldBe "TGTG"
+    Sequences.complement("") shouldBe ""
+    Sequences.complement("AACCGGTGTG") shouldBe "TTGGCCACAC"
   }
```
tfenne: I know it's only a call to a htsjdk function, but could you add a few basic tests for `reverseComplement()` please?  That way if the library function gets broken/changed unexpectedly we'll catch it, or if we decide to reimplement we won't be caught short.

src/test/scala/com/fulcrumgenomics/util/SequencesTest.scala
nh13: 1. Sometimes I have case classes with an empty sequence, so I don't want it to blow up when calling `maxPolyX`: 
```
things.map(maxPolyX(_.sequence))
```
The longest homopolymer in an empty sequence is the empty string isn't it?
2. I don't think you should be using this method to catch the case (i.e. throw an exception) when you have an empty string.  I think it's wrong, so I'd like to fix it. I really don't want to have to write guards everywhere in my code to ignore feeding empty strings to `longestHomopolymer`.  One could argue that asking for the longest homopolymer implies that there is at least one homopolymer, but I think that's just an argument for argument's sake.  

Any objections to fixing this?   Do you have code that will break (i.e. relies on the exception being thrown here)?

src/test/scala/com/fulcrumgenomics/util/SequencesTest.scala
nh13: Done.

src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
```diff
@@ -40,7 +40,11 @@ case class ProgressLogger(logger: Logger,
   }
 
   /** Method to use to record progress when genome location isn't available or relevant. */
-  def record(): Unit = record(null, -1)
+  def record(): Boolean = record(null, -1)
+
+  def record(rec: SamRecord): Boolean = record(rec.asSam)
+
+  /** Logs the last record if it wasn't already logged. */
+  def logLast(): Boolean = super.log()
```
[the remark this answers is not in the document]
nh13: Done

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -55,25 +58,73 @@ object Sequences {
...
-    val xs = s.toUpperCase
-    val len = xs.length
-    assert(len > 0, "Cannot compute longest homopolymer of a zero length string")
+    assert(s.length > 0, "Cannot compute longest homopolymer of a zero length string")
+    val result = maxPolyX(s)
+    (result.offset, result.length)
+  }
+
+
+  case class OffsetAndLength(offset: Int, length: Int)
```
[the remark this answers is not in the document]
nh13: Done

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
nh13: Agreed, fixed.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -55,25 +58,73 @@ object Sequences {
...
+
+  /**
+    * Returns the offset (0-based) and length of the longest homopolymer in the sequence. In the case
+    * that there are multiple homopolymers of the same length, the earliest one is returned.
+    *
+    * @param s a DNA or RNA sequence
+    * @return the offset and length of the longest homopolymer
+    */
+  def maxPolyX(s: String) : OffsetAndLength = {
+    val bs = s.getBytes
```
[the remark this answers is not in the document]
nh13: Done

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -55,25 +58,73 @@ object Sequences {
...
+    var (bestStart, bestLength) = (0,0)
+    var start = 0
+    while (start < bs.length) {
```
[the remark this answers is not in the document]
nh13: Done

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -55,25 +58,73 @@ object Sequences {
...
+      val firstBase = s.charAt(start).toByte
```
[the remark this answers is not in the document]
nh13: Done

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -55,25 +58,73 @@ object Sequences {
...
-        foundLength = newLength
+  /**
+    * Returns the offset (0-based) and length of the longest dinucleotide run in the sequence. In the case
+    * that there are multiple dinucleotide runs of the same length, the earliest one is returned.
+    *
+    * @param s a DNA or RNA sequence
+    * @return the offset and length of the longest dinucleotide sequence
+    */
+  def maxDinuc(s: String) : OffsetAndLength = {
+    val bs = s.toUpperCase.getBytes
```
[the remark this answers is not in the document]
nh13: Done

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
nh13: Done, but "I did it my way"

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
nh13: It is, fixed.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -55,25 +58,73 @@ object Sequences {
...
+        bestLength = length
       }
     }
 
-    (foundIndex, foundLength)
+    OffsetAndLength(bestStart, bestLength)
+  }
+
+  /** Reverse complements a string of bases. */
+  def reverseComplement(s: String): String = SequenceUtil.reverseComplement(s)
```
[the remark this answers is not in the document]
nh13: Feeling good, you?  Fixed.

PR #422, opened 2018-08-08
A tool to identify primers and collect related metrics

nh13:
@tfenne could you please take a look, and if you see any major objections, please stop and ping me via Slack to discuss?  I'd like to get a first pass review before finishing all the tests.


src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+    |location of their respective primers.
+    |2. Assign to a non-canonical primer pair if the 5' end of the mates agree with the mapping location of their
+    |respective primers, and the primer and read sequences "match" respectively for each primer.
+    |3. Assign each mate of a read pair to a primer based on fully-gapped alignment.
+    |
+    |## Paired-End Matching (5' and 3')
+    |
+    |The `--three-prime` option can be used to also search the 3' end of every read for a primer sequence as follows:
+    |
+    |1. If a primer pair is identified using the 5' end, that pair is searched on the 3' end.
```
tfenne: "that pair..." -> "the expected paired primer is searched for at the 3' end"

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+    |2. Otherwise, assign a primer based on fully-gapped alignment.
+    |
+    |The primer matched on the 5' end of read one must match the primer matched on the 3' end of read two, and the
```
tfenne: This read like "Hu's on first".  How about "when a primer match is found at both 5' and 3' ends of reads, the tool enforces that the 5' end of R1 and the 3' end of R2 match the same primer, and vice versa."

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+    |primer matched on the 3' end of read one must match the primer matched on the 5' end of read two.  This will only
+    |be enforced if both ends of a read match a primer.
+    |
+    |## Fragment Matching (5' only)
```
tfenne: How are PE reads with one or both ends unmapped handled?  Still as pairs, or as fragments?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+    |
+    |If the input are fragment reads, then only the 5' end of a read is used to match against the input list of primers:
```
tfenne: This disagrees with the next section.  Maybe you meant "when only" instead of "then only"?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+    |Install [the ksw executable](https://github.com/nh13/ksw) manually or via conda (`conda install -c bioconda ksw`)
+    |and supply the path to the `ksw` executable via `--ksw`.
+  """)
+class IdentifyPrimers
+(@arg(flag='i', doc="Input BAM file.")  val input: PathToBam,
+ @arg(flag='p', doc="File containing information about the primers.") val primers: FilePath,
+ @arg(flag='o', doc="Output BAM file.") val output: PathToBam,
+ @arg(flag='m', doc="Path prefix for the metrics files.") val metrics: PathPrefix,
+ @arg(flag='S', doc="Match to primer locations +/- this many bases.") val slop: Int = 5,
+ @arg(          doc="Search for primers on the 3' end of each read as well (5' only by default).") val threePrime: Boolean = false,
```
tfenne: `flag=3`?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+ @arg(          doc="Maximum mismatches for a primer to be considered a match.") val maxMismatches: Int = 1,
```
tfenne: Is this used only for ungapped alignment?  Or in all cases?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+ @arg(          doc="The minimum alignment score for fully gapped alignment for a primer to be considered a match.") val minAlignmentScore: Int = 10,
```
tfenne: I'm not convinced that users will be good at supplying a cutoff alignment score.  Though I guess if you don't get the mismatches and indels, there's not much else you can do.

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+ @arg(          doc="The match score to use for aligning to primer sequences.") val matchScore: Int = 1,
+ @arg(          doc="The mismatch score to use for aligning to primer sequences.") val mismatchScore: Int = -4,
+ @arg(          doc="The gap open score to use for aligning to primer sequences.") val gapOpen: Int = -6,
+ @arg(          doc="The gap extension score to use for aligning to primer sequences.") val gapExtend: Int = -1,
+ @arg(          doc="The minimum difference in mismatches between the best match and second best match to accept a primer match.  Not used for full alignment.") val minMismatchDelta: Int = 1,
+ @arg(flag='t', doc="The number of threads to use.") val threads: Int = 1,
+ @arg(flag='n', doc="Examine the first N templates.") val numTemplates: Option[Int] = None,
```
tfenne: Should this be a hidden option?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+ @arg(          doc="Skip full-alignment matching") val skipFullAlignment: Boolean = false,
+ @arg(          doc="Path to the ksw aligner.") val ksw: Option[String] = None,
+ @arg(          doc="The maximum number of templates in memory.") val maxTemplatesInRam: Option[Int] = None,
```
tfenne: Are these last two options really needed?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+ @arg(          doc="The number of templates to process at a time per thread.") val templatesPerThread: Int = 1000
+) extends FgBioTool with LazyLogging {
+
+  import IdentifyPrimers._
+
+  private val kswExecutable: Option[FilePath] = this.ksw.map(Paths.get(_)).flatMap {
+    case p if Files.exists(p) => Some(p)
+    case name =>
```
tfenne: this branch will also get triggered if the user typos, e.g. `/usr/locla/bin/ksw`, and that's going to be very confusing.  Could you instead just try executing the executable given and see if it works?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+  }
+
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+  Io.assertCanWriteFile(metrics)
+  kswExecutable.foreach(Io.assertReadable)
+
+  validate(maxMismatches >= 0,   "--max-mismatches must be >= 0")
+  validate(minMismatchDelta > 0, "--min-mismatch-delta must be > 0")
+  validate(matchScore > 0,       "--match-score must be > 0")
```
tfenne: Rather than validating these four, I wonder if we should just `abs()` all of them, and then negate those that should be negative?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+  validate(mismatchScore < 0,    "--mismatch-score must be < 0")
+  validate(gapOpen < 0,          "--gap-open must be < 0")
+  validate(gapExtend < 0,        "--gap-extend must be < 0")
+  maxTemplatesInRam.foreach { m => validate(templatesPerThread < m, "--max-templates-in-ram must be greater than or equal to --templates-per-thread")}
+
+  private val aligner: Aligner = Aligner(matchScore = matchScore, mismatchScore = mismatchScore, gapOpen = gapOpen, gapExtend = gapExtend, mode = AlignmentMode.Glocal)
```
tfenne: Is this just used to get the final alignment with cigar etc?  If so could you comment that here please?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+
+  private val (fwdMatcher, revMatcher) = PrimerMatcher.build(
```
tfenne: No apply()?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+    aligner           = aligner,
+    slop              = slop,
+    maxMismatches     = maxMismatches,
+    minMismatchDelta  = minMismatchDelta,
+    minAlignmentScore = minAlignmentScore,
+    withFullAlignment = !skipFullAlignment,
+    revComp           = true
+  )
+
+  var numAlignments: AtomicLong = new AtomicLong(0)
```
tfenne: Should be a val.

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+      "The pf/pr tags are formatted as follow: <pair_id>,<primer_id>,<refName>:<start>-<end>,<strand>,<5-prime-offset>,<match-type>,<match-type-info>.",
+      s"The match-type is 'location', 'mismatch', or 'aligmnent' based on if the match was found using the location, mismatch-based alignment, or full-alignment.",
+      s"The pp tag is either 'canonical', 'non-canonical', 'single', or '$NoPrimerMatchInfo' based on if the matches for the primer pair agree, disagree, or are missing."
+    )
+  }
+
+  override def execute(): Unit = {
+    this.kswExecutable match {
+      case Some(p) => logger.info(s"Using ksw aligner: $p")
+      case None    => logger.info("Using the scala aligner.")
```
tfenne: How about `logger.info("Using the internal aligner; install ksw for faster operation.")`

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+      SamWriter(path = output, header = header, sort = Some(SamOrder.Coordinate))
+    }
+
+    // Get the number of templates to process in a single thread
+    val batchSize = numTemplates.map(_ / threads) match {
+      case None    => templatesPerThread
+      case Some(n) => math.max(1, math.min(n, templatesPerThread))
+    }
+    // The number of templates to batch across threads
+    val majorBatchSize  = math.min(batchSize * threads * 4, maxTemplatesInRam.getOrElse(Int.MaxValue))
```
tfenne: Why is the `* 4` here?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+      val header = {
+        val h = in.header.clone()
+        comments.foreach { comment => h.addComment(comment) }
+        h
+      }
+      SamWriter(path = output, header = header, sort = Some(SamOrder.Coordinate))
+    }
+
+    // Get the number of templates to process in a single thread
+    val batchSize = numTemplates.map(_ / threads) match {
```
tfenne: I don't think `numTemplates` should be used here.  I see why you're doing it, but it makes this whole section harder to follow, and it's essentially a debugging option in the first place, so if you're going to use that can we assume you're smart enough to set `templatesPerThread` and `threads` to sensible values?

I'd also ditch `maxTemplatesInRam` and just state that the program will keep up to `templatesPerThead *  readsPerTemplate * threads` reads in RAM.  Then you only need to calculate `majorBatchSize` and everything else is already defined.

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+      case None    => templatesPerThread
+      case Some(n) => math.max(1, math.min(n, templatesPerThread))
+    }
+    // The number of templates to batch across threads
+    val majorBatchSize  = math.min(batchSize * threads * 4, maxTemplatesInRam.getOrElse(Int.MaxValue))
+    val readingProgress = ProgressLogger(this.logger, verb="read", unit=batchSize*threads*2)
```
tfenne: Isn't this going to lead to a _ton_ of progress logging?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+    }
+
+    // NB: Add comments explaining tags in the output writer
+    val out: SamWriter = {
+      val header = {
+        val h = in.header.clone()
+        comments.foreach { comment => h.addComment(comment) }
+        h
+      }
+      SamWriter(path = output, header = header, sort = Some(SamOrder.Coordinate))
```
tfenne: Why assume coordinate vs. the input sort order?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+    Metric.write(PathUtil.pathTo(this.metrics + ".summary.txt"), IdentifyPrimersMetric(metricCounter))
+  }
+
+  /** Creates a new [[InteractiveAlignmentScorer]]. */
+  private def newAligner: InteractiveAlignmentScorer[Primer, SamRecordAlignable] = {
+    InteractiveAlignmentScorer(matchScore, mismatchScore, gapOpen, gapExtend, AlignmentMode.Glocal, this.kswExecutable)
+  }
+
+  // NB: batching alignment inputs to the aligner (i.e. ksw) is empirically faster than given them all or just one at a time.
+  private def completePrimerMatch(matchOption: Option[Either[PrimerMatch, Seq[AlignmentTask[Primer, SamRecordAlignable]]]],
```
tfenne: Wouldn't this be clearer as two methods rather that one which takes this crazy `Option[Either[PrimerMatch, Seq[AlignmentTask[Primer, SamRecordAlignable]]]]`?

tfenne: Wait .. so it doesn't even do anything if you pass in a `Some(Left(PrimerMatch))`?  Why is that even here then?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+        val alignmentResults = alignmentInputs.toIterator.zip(aligner.iterator)
+          // Keep results that meet the minimum score and align with a minimum # of query bases
+          .filter { case (alignmentInput, alignmentResult) =>
+            val minQueryEnd = alignmentInput.queryLength - slop
+            alignmentResult.score >= minAlignmentScore && alignmentResult.queryEnd >= minQueryEnd
+          }
+          // now just keep the primer and score
+          .map { case (alignmentInput, alignmentResult) => PrimerAndScore(alignmentInput.query, alignmentResult.score) }
+          // get the best two alignment results by score
+          .maxTwoBy(_.score)
```
tfenne: Why not just `.sort(-_.score).take(2)`?  Your input is already a Seq, so you have it all in memory.

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+  def processBatch(templates: Seq[Template],
+                   metricCounter: SimpleCounter[TemplateTypeMetric],
+                   progress: ProgressLogger): Seq[Template] = {
+
+    val counter = new SimpleCounter[TemplateTypeMetric]()
+    val aligner  = newAligner
+
+    templates.foreach { template =>
+      // match up to but not including full alignment
+      val r1MatchOption = template.r1.map { r1 => fwdMatcher.matchUntilAlignment(r1) }
```
tfenne: Maybe just rename the variable in the closure to `r`.  It's confusing on the next line when you map `r2` and call it `r1`.

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+      val r2MatchOption = template.r2.map { r1 => revMatcher.matchUntilAlignment(r1) }
```
tfenne: The name `matchUntilAlignment` confused me - I thought you were trying to say you were going to do a _sequence match_ of the bases before r1's alignment started (i.e. soft-clipped bases).  But that seemed crazy.  I see what it really means is `matchBasedOnPositionOrUngappedAlignment`.

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+
+      // full in the alignment
```
tfenne: full -> fill

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+          }
+          val mType = formatPair(r1, r2, r1Match, r2Match, r1ThreePrimeMatch, r2ThreePrimeMatch)
+          TemplateTypeMetric(rType, mType, r1Match, r2Match)
+        case (Some(r1), None) =>
+          require(!r1.paired, s"Found paired read but missing R2 for ${r1.name}")
+          val rType = if (r1.mapped) TemplateType.MappedFragment else TemplateType.UnmappedFragment
+          val mType = formatFragment(r1, r1Match, r1ThreePrimeMatch)
+          TemplateTypeMetric(rType, mType, r1Match, r2Match)
+        case _ =>
+          unreachable(s"Template did not have an R1: ${template.name}")
```
tfenne: I don't think I'd use `unreachable` here.  Probably just `throw new IllegalStateException`, as users are plenty capable of creating BAMs that are missing R1s, and `unreachable` is meant for places that are truly unreachable.

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+          val rate = numAlignments.get() / progress.getElapsedSeconds.toDouble
+          logger.info(f"Performed ${numAlignments.get()}%,d full alignments so far ($rate%,.2f alignments/second).")
+        }
+      }
+    }
+
+    templates
+  }
+
+  // FIXME: add to SimpleCounter in commons
```
tfenne: Yes please.  I'm always surprised we don't have `+=` in SimpleCounter.

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+      new IdentifyPrimersMetric(
+        templates                 = readTypeCounter.total,
+        pairs                     = mapped_pairs + unpaired + unmapped_pairs,
+        fragments                 = mapped_fragments + unmapped_fragments,
+        mapped_pairs              = mapped_pairs,
+        unpaired                  = unpaired,
+        unmapped_pairs            = unmapped_pairs,
+        mapped_fragments          = mapped_fragments,
+        unmapped_fragments        = unmapped_fragments,
+        total_primer_pair_types   = matchTypeCounter.total,
```
tfenne: I don't get what this is?  Isn't this going to be the same as `templates`?

src/main/scala/com/fulcrumgenomics/bam/IdentifyPrimers.scala
```diff
@@ -0,0 +1,1027 @@
...
+        if (!SequenceUtil.readBaseMatchesRefBaseWithAmbiguity(bases(index), primer(index))) count += 1
+      }
+      count
+      //bases.zip(primer).count { case (l, r) => !SequenceUtil.readBaseMatchesRefBaseWithAmbiguity(l.toByte, r.toByte) }
+    }
+  }
+
+  /** A little class that wraps a [[SamRecord]] to make it [[Alignable]] and also to defer extracting the bases to align.
+    * NB: end is exclusive, and start and end are zero-based. */
+  case class SamRecordAlignable(rec: SamRecord, start: Int, end: Int) extends Alignable {
```
tfenne: For zero based I think `offset` and `length` are clearer.

nh13:
@tfenne I think this is ready for a re-review.  I have decided to move out the code that attempts to classify primer pair matches as dimer/not-a-dimer/something-else, since it is very experiment specific. I apologize for some really ugly code in advance around the batching and parallelizing of data.  Just a reminder, we do so in two ways:
1. by asynchronously performing gapped alignment when necessary.
2. batching of templates for embarrassingly parallel multi-threading.

src/main/scala/com/fulcrumgenomics/bam/identifyprimers/PrimerMatcher.scala
```diff
@@ -0,0 +1,325 @@
...
+private[identifyprimers] class UngappedAlignmentBasedPrimerMatcher
+(val primers: Seq[Primer],
+ val slop: Int,
+ val maxMismatchRate: Double,
+ val kmerLength: Option[Int] = None
+ ) extends PrimerMatcherWithKmerFilter with UngappedAlignment with LazyLogging {
+
+  /** Examines all primers to find the best mismatch-based alignment match. */
+  def find(rec: SamRecord): Option[UngappedAlignmentPrimerMatch] = {
+    val alignments = getPrimersForAlignmentTasks(rec).flatMap { p => mismatchAlign(rec, p) }
```
tfenne: Might be nice to do `.iterator.flatMap( ... ).filter(_.mismatchRate <= maxMismatchRate).toSeq` to avoid having all the alignments in memory at the same time.

src/main/scala/com/fulcrumgenomics/bam/identifyprimers/PrimerMatcher.scala
```diff
@@ -0,0 +1,325 @@
...
+    getBestUngappedAlignment(alignments)
+  }
+
+  /** Examines all primers to find the best gapped-alignment-based match. */
+  private[identifyprimers] def getBestUngappedAlignment(alignments: Seq[UngappedAlignmentPrimerMatch]): Option[UngappedAlignmentPrimerMatch] = {
+    if (alignments.isEmpty) None
+    else {
+      alignments.sortBy(_.numMismatches).take(2) match {
```
tfenne: It's unlikely you'd have primers with a 2-fold variation in length, but shouldn't this be by _rate_ not number?  Or perhaps by number of matching bases instead?

src/main/scala/com/fulcrumgenomics/bam/identifyprimers/TemplateType.scala
```diff
@@ -0,0 +1,64 @@
...
+
+
+private[identifyprimers] sealed trait TemplateType extends EnumEntry
+
+/** The read structure of a template.  Distinguishes between fragment and reads pairs, and mapped and unmapped. */
+private[identifyprimers] object TemplateType extends FgBioEnum[TemplateType] {
+  /** A read pair where both ends are mapped. */
+  case object MappedPair extends TemplateType
+  /** A read pair where exactly one end is mapped. */
+  case object Unpaired extends TemplateType
```
tfenne: PartiallyMappedPair?

nh13:
This will be moved to its own repo.

PR #427, opened 2018-08-20
Allow fgbio tools to access meta information about themselves.

nh13:
This allows SAM program records to be created.


src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -25,10 +25,39 @@ package com.fulcrumgenomics.cmdline
 
 import com.fulcrumgenomics.cmdline.FgBioMain.FailureException
 import com.fulcrumgenomics.sopt.cmdline.ValidationException
+import com.fulcrumgenomics.util.Metric
+import htsjdk.samtools.{SAMFileHeader, SAMProgramRecord}
+
+import scala.annotation.tailrec
+
+/** Stores meta information about the command line use to invoke a tool. */
+case class FgBioToolInfo(name: String, commandLine: String, description: String, version: String) extends Metric {
```
nh13: I'd be ok with having both the full command line (with default args included) and short command line (without default args included).  Right now, I am providing the former only.

src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
```diff
@@ -39,6 +39,7 @@ import com.fulcrumgenomics.commons.CommonsDef._
 
 import scala.reflect.runtime.universe.TypeTag
 import scala.reflect.ClassTag
+import scala.tools.cmd.CommandLineParser
```
tfenne: I feel like this import is erroneous.  In fact I think all three are unused.

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
tfenne: I think it would be nice, and relatively easy, to do both.  In fact I would suggest we have:

1. `args: Seq[String]` which stores the actual set of args passed to Main
2. `def commandLineWithoutDefaults: String = args.mkString(" ")`
3. `val commandLineWithDefaults` which is what you have as `commandLine` now.

src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
```diff
@@ -85,16 +86,24 @@ class FgBioMain extends LazyLogging {
...
-      case Sopt.SubcommandSuccess(command, subcommand) =>
+      case Sopt.SubcommandSuccess(_, subcommand) =>
         val name = subcommand.getClass.getSimpleName
         try {
+          parser.commandLine.foreach { commandLine =>
+            subcommand.toolInfo = FgBioToolInfo(
+              name        = name,
+              commandLine = commandLine,
+              description = parser.formatShortDescription(Sopt.inspect(subcommand.getClass).description),
+              version     = this.version
```
tfenne: I think to be safe here (and the other place `version` is used, it should probably be implemented as `subcommand.getClass.getPackage.getImplementationVersion`.  That way if the sub-command is in another project that's using fgbio, it'll get the version for the sub-command, not the version for fgbio.

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -25,10 +25,39 @@ package com.fulcrumgenomics.cmdline
...
+  /** Adds a program group to the SAMFileHeader returning the ID. */
+  def applyTo(header: SAMFileHeader): String = {
```
tfenne: It feels a bit weird to have this here in FgBioTool, but I agree it's useful.

tfenne: Also, how about:
1. Calling this `addProgramGroupTo`?
2. Taking either an optional ID or ID prefix?
3. Any reason to return the ID and not the whole record?

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -25,10 +25,39 @@ package com.fulcrumgenomics.cmdline
...
+    // Get the id
+    val id = {
+      @tailrec
+      def getPgId(intId: Int): String = if (header.getProgramRecord(intId.toString) == null) intId.toString else getPgId(intId + 1)
+      getPgId(1)
+    }
+    val pg = new SAMProgramRecord(id)
+    pg.setProgramName(name)
+    pg.setCommandLine(commandLine)
+    pg.setAttribute("DS", description)
```
tfenne: This could be really long, no?  And have newlines.  I think I'd leave this out.

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -45,4 +74,10 @@ trait FgBioTool {
 
   /** Generates a validation exception if the test value is false. */
   def validate(test: Boolean, message: => String) = if (!test) throw new ValidationException(message)
+
+  /** Gets line used to invoke this tool, or [[None]] if unset. */
+  def toolInfo: Option[FgBioToolInfo] = _toolInfo
+
+  /** Sets the command line used to invoke this tool. */
+  private[cmdline] def toolInfo_=(commandLine: FgBioToolInfo): Unit = this._toolInfo = Some(commandLine)
```
tfenne: parameter name is misleading (out of date?).  Perhaps just `toolInfo`?

nh13:
@tfenne this PR got a bit bigger in scope: I added support for collections in `Metric` classes.  It will join values via the comma separator by default, though I don't love the way I have allowed the "collection delimiter" to be exposed (should I even do that now?).

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -26,9 +26,51 @@ package com.fulcrumgenomics.cmdline
...
+  * @param description the description of the tool
+  * @param version the version of the tool.
+  */
+case class FgBioToolInfo(name: String, args: Seq[String], commandLineWithDefaults: String, description: String, version: String) extends Metric {
+
+  /** The command line as given to the tool, without any defaults added.  This will include arguments given to
+    * [[FgBioMain]] in [[FgBioCommonArgs]].*/
+  def commandLineWithoutDefaults: String = args.mkString(" ")
+
+  /** Adds a program group to the SAMFileHeader returning the ID. */
```
clintval, line 50:
```suggestion
  /** Adds a program group to the SAMFileHeader and returns the newly created program group. */
```

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -26,9 +26,51 @@ package com.fulcrumgenomics.cmdline
...
+
+import scala.annotation.tailrec
+
+/** Stores meta information about the command line use to invoke a tool.
+  *
+  * @param name the name of the tool.
+  * @param args the list of arguments as given on the command line; this will contain any arguments also given to
+  *             [[FgBioMain]] in [[FgBioCommonArgs]].
+  * @param commandLineWithDefaults the command line as given to the tool, along with the defaults for all other
+  *              arguments.  This will include arguments given to [[FgBioMain]] in [[FgBioCommonArgs]].
```
clintval, line 40: The definitions between these two seem very similar.

I think the `args` _**does not**_ have any arguments also given to `FgBioMain` in `FgBioCommonArgs`?

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -26,9 +26,51 @@ package com.fulcrumgenomics.cmdline
...
+      def getPgId(intId: Int): String = if (header.getProgramRecord(intId.toString) == null) intId.toString else getPgId(intId + 1)
+      getPgId(1)
+    }
+    val pg = new SAMProgramRecord(pgId)
+    pg.setProgramName(name)
+    pg.setCommandLine(commandLineWithDefaults)
+    pg.setProgramVersion(version)
+    header.addProgramRecord(pg)
+    pg
+  }
```
clintval, line 64: Unit tests needed which ideally test both codepaths for determining the next program group ID.

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -26,9 +26,51 @@ package com.fulcrumgenomics.cmdline
...
+  def commandLineWithoutDefaults: String = args.mkString(" ")
+
+  /** Adds a program group to the SAMFileHeader returning the ID. */
+  def addProgramGroupTo(header: SAMFileHeader, id: Option[String]=None): SAMProgramRecord = {
+    // Get the id
+    val pgId = id.getOrElse {
+      @tailrec
+      def getPgId(intId: Int): String = if (header.getProgramRecord(intId.toString) == null) intId.toString else getPgId(intId + 1)
+      getPgId(1)
+    }
```
clintval, line 57: I'm assuming you've done it this way instead of looping over all program groups because the keys for accessing program groups are strings?

src/test/scala/com/fulcrumgenomics/cmdline/ClpTests.scala
```diff
@@ -61,4 +68,32 @@ class ClpTests extends UnitSpec {
...
+    metric.commandLineWithDefaults should include (" --compression 5") // default argument _prior_ to the tool namee
+    metric.commandLineWithDefaults should include ("TestClp")
+    metric.commandLineWithDefaults should include (s"--info-path $tmpPath")
+    metric.commandLineWithDefaults should include ("--print-me :none:") // a default argument
+
+    // description
+    metric.description shouldBe "A test class"
+
+    // version
+    //metric.version shouldBe "null" // not set in scalatest
```
clintval, line 97: Is this not possible to test?

src/test/scala/com/fulcrumgenomics/cmdline/ClpTests.scala
```diff
@@ -61,4 +68,32 @@ class ClpTests extends UnitSpec {
...
+
+    // commandLineWithDefaults
+    metric.commandLineWithDefaults should include ("fgbio --async-io true") // argument set _prior_ to the tool name
+    metric.commandLineWithDefaults should include (" --compression 5") // default argument _prior_ to the tool namee
+    metric.commandLineWithDefaults should include ("TestClp")
+    metric.commandLineWithDefaults should include (s"--info-path $tmpPath")
+    metric.commandLineWithDefaults should include ("--print-me :none:") // a default argument
+
+    // description
+    metric.description shouldBe "A test class"
```
clintval, line 94: What happens for multi-line descriptions?

clintval (commented):
I have a few questions and recommended some fixups too.

Let me know when you are ready for a re-review. It's close!

src/test/scala/com/fulcrumgenomics/cmdline/ClpTests.scala
nh13: Exactly, it gets put in the manifest in the JAR, which scalatest doesn't have access to

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
nh13: Exactly

src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
nh13: Done

clintval (approved):

PR #435, opened 2018-10-02
Add an option not to redirect error in DnaFoldPredictor

nh13:
@tfenne something that has bugged me for a while, especially if I am creating many threads to run vienna rna


nh13:
Punting this for now since this error is not fun:
```

[error] /home/travis/build/fulcrumgenomics/fgbio/src/main/java/com/fulcrumgenomics/util/DnaFoldPredictor.java:93:1:  scala.Function1 is not a functional interface
[error]     multiple non-overriding abstract methods found in interface scala.Function1
[error]                 new AsyncStreamSink(this.process.getErrorStream(), (String s) -> { this.err.write(s); return BoxedUnit.UNIT; });
[info] Some messages have been simplified; recompile with -Xdiags:verbose to get full output
[error] (Compile / compileIncremental) javac returned non-zero exit code
```
We should just re-write this class in scala

nh13:
Ok, converted this all to scala, yay!

PR #438, opened 2018-10-16
Tool to detect templates with strand-switch events in them.

tfenne:
@nh13: I still need to write tests for all of this, but I wanted to make sure you had a chance to look through and make sure you agree with the general direction, naming of things, etc. before I got too much further.  Do you mind?  I'll request a re-review once I have tests too.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
@@ -49,18 +49,8 @@ object FindSwitchbackReads {
   /** The name of the BAM extended attribute tag used to store switchback information. */
   val SwitchTag = "sb"
 
-  /** The type code used when a switchback is identified within the sequence of a read. In this case the `sb` tag
-    * will have the format `sb:Z:r,{read|mate},{offset},{length}`.
-    */
-  val SwitchTypeRead = "r"
-
-  /** The type code used when a switchback event is detected via the tandem orientation of a read pair.
-    * In this case the `sb` tag will have the format: `sb:Z:t,{gap}`.
-    */
-  val SwitchTypeTandem = "t"
-
   /** Parent trait for the different kinds of hits we can find. */
-  sealed trait SwitchHit
+  sealed trait SwitchHit { def code: Char }
 
   /**
     * Information about a switchback found by looking at soft-clipped complementary sequence within
@@ -69,8 +59,11 @@ object FindSwitchbackReads {
     * @param offset the offset between the reference position at the last base read on the original strand and
     *               the first base read on the opposite strand.
     * @param length the length of the post-switch sequence within the read
+    * @param read optionally the read in which the hit was found (generally filled in later)
     */
-  case class ReadBasedHit(var offset: Int, length: Int, var read: Option[SamRecord] = None) extends SwitchHit
+  case class ReadBasedHit(var offset: Int, length: Int, var read: Option[SamRecord] = None) extends SwitchHit {
+    override val code: Char = 'r'
+  }
 
   /**
     * Information about a switchback found by examining a template with a tandem (i.e. FF or RR) pair orientation.
@@ -83,7 +76,9 @@ object FindSwitchbackReads {
     *     ------>          gap = -2
     * @param gap the unsequenced gap between the the 3' end of the "earlier" read and the 5' end of the "later" read.
     */
-  case class TandemBasedHit(gap: Int) extends SwitchHit
+  case class TandemBasedHit(gap: Int) extends SwitchHit {
+    override val code: Char = 't'
+  }
 
   /**
     * Summary metrics regarding switchback reads found.
@@ -231,17 +226,17 @@ object FindSwitchbackReads {
     |
     |Templates that contain strand switch events (switch-backs) are found by this tool in two different ways:
     |
-    |1. By looking at reads that contain soft-clipping at their 5' end that matches the genome, after
-    |   being reverse complemented, proximal to the 5'-most mapped base of the read.  We call these matches
+    |1. By looking at reads that contain soft-clipped bases at their 5' end that, when reverse complemented, matches the
+    |   genome proximal to the 5'-most mapped base of the read.  We call these matches
     |   "read based switchbacks".  Finding read based switchbacks is based on several parameters:
     |
-    |   1. `max-offset` controls how far away the reverse-complemented sequence is searched for.  The default value of
+    |   1. `max-offset` controls how far away to search for the reverse-complemented sequence.  The default value of
     |      `35` allows matches to be found when the soft-clipped sequence matches the genome _starting_ at most 35bp
     |      from the 5' mapped position of the read, and reading in the opposite direction.
     |   2. `min-length` controls the minimum number of soft-clipped bases that must exist to trigger the search.
     |      Given that the search looks at `2 * max-offset` locations (default=70) it is important that `min-length`
     |      is set such that `4^min-length >> 2 * `max-offset` in order to avoid false positives.
-    |   3. `max-error-rate` allows for some mismatches to exist between the soft-clipped sequence and the genome when matching
+    |   3. `max-error-rate` allows for some mismatches to exist between the soft-clipped sequence and the genome when matching.
     |
     |2. By identifying templates with `FF` or `RR` (aka tandem) orientation where it is surmised that the template
     |   switch occurred in the un-sequenced region of the template between R1 and R2.  We call these `tandem based
@@ -338,7 +333,7 @@ class FindSwitchbackReads
             Unit
           case Some(hit: TandemBasedHit) =>
             switchbackGaps.count(hit.gap)
-            val tagValue = s"${SwitchTypeTandem},${hit.gap}"
+            val tagValue = s"${hit.code},${hit.gap}"
             template.allReads.foreach(r => r(SwitchTag) = tagValue)
           case Some(hit: ReadBasedHit) =>
             switchbackLengths.count(hit.length)
@@ -346,7 +341,7 @@ class FindSwitchbackReads
             val foundInReadNum = readNum(hit.read.getOrElse(unreachable("read should be populated!")))
             template.allReads.foreach { r =>
               val num = readNum(r)
-              r(SwitchTag) = s"${SwitchTypeRead},${if (num == foundInReadNum) "read" else "mate"},${hit.offset},${hit.length}"
+              r(SwitchTag) = s"${hit.code},${if (num == foundInReadNum) "read" else "mate"},${hit.offset},${hit.length}"
             }
         }
 
@@ -381,7 +376,12 @@ class FindSwitchbackReads
 
       Metric.write(PathUtil.pathTo(prefix + ".summary.txt"), m)
 
-      for ((suffix, counter, header) <- Seq(("lengths", switchbackLengths, "length"), ("offsets", switchbackOffsets, "offset"), ("gaps", switchbackGaps, "gap_length"))) {
+
+      Seq(
+        ("lengths", switchbackLengths, "length"),
+        ("offsets", switchbackOffsets, "offset"),
+        ("gaps",    switchbackGaps,    "gap_length")
+      ).foreach { case (suffix, counter, header) =>
         val writer = Io.toWriter(PathUtil.pathTo(s"$prefix.$suffix.txt"))
         writer.write(s"$header\tcount\n")
         counter.foreach { case (x, count) => writer.write(s"$x\t$count\n")}
--- a/src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
@@ -153,7 +153,8 @@ object FindSwitchbackReads {
         val windowStart = max(1, rec.start - maxOffset)
         val windowEnd   = min(ref.length(), CoordMath.getEnd(rec.start, bases.length + maxOffset))
         val refBases    = new String(ref.getBases, windowStart - 1, CoordMath.getLength(windowStart, windowEnd))
-        find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(hit.offset + windowStart - rec.start); hit }
+        find(bases, refBases, maxErrorRate, preferLaterMatches=false)
+          .map { hit => hit.read = Some(rec); hit.offset = -(hit.offset + windowStart - rec.start); hit }
       }
       else if (rec.negativeStrand && rec.cigar.last.operator == CigarOperator.S && rec.cigar.last.length >= minLength) {
         val ref         = refs(rec.refName)
@@ -161,7 +162,8 @@ object FindSwitchbackReads {
         val windowStart = max(1, rec.end - bases.length - maxOffset + 1)
         val windowEnd   = min(ref.length(), rec.end + maxOffset)
         val refBases    = new String(ref.getBases, windowStart - 1, windowEnd - windowStart + 1)
-        find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(rec.end - (hit.offset + windowStart + bases.length - 1)); hit }
+        find(bases, refBases, maxErrorRate, preferLaterMatches=true)
+          .map { hit => hit.read = Some(rec); hit.offset = -(rec.end - (hit.offset + windowStart + bases.length - 1)); hit }
       }
       else {
         None
@@ -175,7 +177,7 @@ object FindSwitchbackReads {
     *
     * @param template the template to be examined.
     * @param maxGap the maximum allowable gap size
-    * @return either `Some(TandemBasedHit)` if a hit can be found, else `None``
+    * @return either `Some(TandemBasedHit)` if a hit can be found, else `None`
     */
   private[bam] def findTandemSwitchback(template: Template, maxGap: Int) : Option[TandemBasedHit] =
     if (maxGap <=0 ) None else template.pairOrientation match {
@@ -201,7 +203,7 @@ object FindSwitchbackReads {
     *                           to the start
     * @return a [[ReadBasedHit]] if a match is found, otherwise None.
     */
-  def find(query: String, target: String, maxErrorRate: Double, preferLaterMatches: Boolean = false): Option[ReadBasedHit] = {
+  def find(query: String, target: String, maxErrorRate: Double, preferLaterMatches: Boolean): Option[ReadBasedHit] = {
     val maxMismatches = Math.floor(query.length * maxErrorRate)
     val range = {
       if (preferLaterMatches) Range.inclusive(target.length-query.length, 0, step = -1)
--- a/src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
@@ -38,16 +38,16 @@ import com.fulcrumgenomics.util._
 import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
 import htsjdk.samtools.SamPairUtil.PairOrientation
 import htsjdk.samtools.reference.ReferenceSequence
-import htsjdk.samtools.util.SequenceUtil
+import htsjdk.samtools.util.{CoordMath, SequenceUtil}
 import htsjdk.samtools.{CigarOperator, SAMFileHeader, SAMReadGroupRecord}
 
 
 object FindSwitchbackReads {
   /** Resource path for R script for plotting. */
-  val PlottingScript = "com/fulcrumgenomics/bam/FindSwitchbackReads.R"
+  private val PlottingScript = "com/fulcrumgenomics/bam/FindSwitchbackReads.R"
 
   /** The name of the BAM extended attribute tag used to store switchback information. */
-  val SwitchTag = "sb"
+  val SwitchTag: String = "sb"
 
   /** Parent trait for the different kinds of hits we can find. */
   sealed trait SwitchHit { def code: Char }
@@ -117,36 +117,56 @@ object FindSwitchbackReads {
     * @param template the template to be examined.
     * @param minLength the minimum length of switchback to be detected.
     * @param maxOffset the maximum offset on the reference between the end of the pre-switch sequence and the start
-    *                  of the post-switch sequence
+    *                  of the post-switch sequence. A value of `0` will cause this method to return `None`.
     * @param maxErrorRate the maximum rate of mismatches between the switched-back sequence and the reference
-    * @return Some(ReadBasedHit) if a hit is found else None.
+    * @return `Some(ReadBasedHit)` if a hit is found else `None`.
     */
   private[bam] def findReadBasedSwitchback(refs: Map[String, ReferenceSequence],
-                                      template: Template,
-                                      minLength: Int,
-                                      maxOffset: Int,
-                                      maxErrorRate: Double): Option[ReadBasedHit] = {
+                                           template: Template,
+                                           minLength: Int,
+                                           maxOffset: Int,
+                                           maxErrorRate: Double): Option[ReadBasedHit] = {
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // Implementation notes:
+    //
+    // Read based switchbacks are detected by looking for soft-clipped sequence at the 5' end of the
+    // read - i.e. the start of F reads and the end of R reads.  To trigger further checks we require
+    // that soft-clipping exist and be at least as long as `minLength`.
+    //
+    // The second part of the check is that the soft-clipped sequence matches the reverse-complement
+    // of the reference, and that the match _starts_ reading on the opposite strand within `maxOffset`
+    // of the 5'-most mapped base of the non-clipped sequence.  How we do this differs by strand.
+    //
+    // A key value we wish to determine is the `offset` - the gap between where we leave off reading
+    // one strand, and where we start reading on the other strand.  This value is positive if
+    // post-switch the enzyme starts outside the sequence read on the original strand and moves back
+    // towards the last base sequenced on the original strand, and negative if the enzyme starts
+    // reading the second strand recessed within the already-read sequence of the first strand.
+    // The `find` method returns the offset within the reference window given, which is then
+    // translated into the offset from the 5' alignment position in a strand dependent manner.
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+
     if (maxOffset <= 0) None else template.primaryReads.filter(_.mapped).flatMap { rec =>
-        if (rec.positiveStrand && rec.cigar.head.operator == CigarOperator.S && rec.cigar.head.length >= minLength) {
-          val ref = refs(rec.refName)
-          val bases = Sequences.revcomp(rec.basesString.take(rec.cigar.head.length))
-          val windowStart = max(1, rec.start - maxOffset)
-          val windowEnd = min(ref.length(), rec.start + bases.length + maxOffset)
-          val refBases = new String(ref.getBases, windowStart - 1, windowEnd - windowStart + 1)
-          find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(hit.offset + windowStart - rec.start); hit }
-        }
-        else if (rec.negativeStrand && rec.cigar.last.operator == CigarOperator.S && rec.cigar.last.length >= minLength) {
-          val ref = refs(rec.refName)
-          val bases = Sequences.revcomp(rec.basesString.takeRight(rec.cigar.last.length))
-          val windowStart = max(1, rec.end - bases.length - maxOffset + 1)
-          val windowEnd = min(ref.length(), rec.end + maxOffset)
-          val refBases = new String(ref.getBases, windowStart - 1, windowEnd - windowStart + 1)
-          find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(rec.end - (hit.offset + windowStart + bases.length - 1)); hit }
-        }
-        else {
-          None
-        }
-      }.find(_ => true)
+      if (rec.positiveStrand && rec.cigar.head.operator == CigarOperator.S && rec.cigar.head.length >= minLength) {
+        val ref         = refs(rec.refName)
+        val bases       = Sequences.revcomp(rec.basesString.take(rec.cigar.head.length))
+        val windowStart = max(1, rec.start - maxOffset)
+        val windowEnd   = min(ref.length(), CoordMath.getEnd(rec.start, bases.length + maxOffset))
+        val refBases    = new String(ref.getBases, windowStart - 1, CoordMath.getLength(windowStart, windowEnd))
+        find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(hit.offset + windowStart - rec.start); hit }
+      }
+      else if (rec.negativeStrand && rec.cigar.last.operator == CigarOperator.S && rec.cigar.last.length >= minLength) {
+        val ref         = refs(rec.refName)
+        val bases       = Sequences.revcomp(rec.basesString.takeRight(rec.cigar.last.length))
+        val windowStart = max(1, rec.end - bases.length - maxOffset + 1)
+        val windowEnd   = min(ref.length(), rec.end + maxOffset)
+        val refBases    = new String(ref.getBases, windowStart - 1, windowEnd - windowStart + 1)
+        find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(rec.end - (hit.offset + windowStart + bases.length - 1)); hit }
+      }
+      else {
+        None
+      }
+    }.find(_ => true)
   }
 
   /**
@@ -155,7 +175,7 @@ object FindSwitchbackReads {
     *
     * @param template the template to be examined.
     * @param maxGap the maximum allowable gap size
-    * @return either Some(TandemBasedHit) if a hit can be found, else None
+    * @return either `Some(TandemBasedHit)` if a hit can be found, else `None``
     */
   private[bam] def findTandemSwitchback(template: Template, maxGap: Int) : Option[TandemBasedHit] =
     if (maxGap <=0 ) None else template.pairOrientation match {
@@ -164,12 +184,7 @@ object FindSwitchbackReads {
           case (Some(r1), Some(r2)) =>
             val (earlier, later) = if (r1.start <= r2.start) (r1, r2) else (r2, r1)
             val gap = later.start - earlier.end - 1
-            if (abs(gap) <= maxGap) {
-              Some(TandemBasedHit(later.start - earlier.end - 1))
-            }
-            else {
-              None
-            }
+            if (abs(gap) <= maxGap) Some(TandemBasedHit(later.start - earlier.end - 1)) else None
         }
       case _ =>
         None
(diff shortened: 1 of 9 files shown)


src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+  /** Parent trait for the different kinds of hits we can find. */
+  sealed trait SwitchHit
+
+  /**
+    * Information about a switchback found by looking at soft-clipped complementary sequence within
+    * a single read.
+    *
+    * @param offset the offset between the reference position at the last base read on the original strand and
+    *               the first base read on the opposite strand.
+    * @param length the length of the post-switch sequence within the read
```
nh13: scaladoc for `read`

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+    */
+  val SwitchTypeRead = "r"
+
+  /** The type code used when a switchback event is detected via the tandem orientation of a read pair.
+    * In this case the `sb` tag will have the format: `sb:Z:t,{gap}`.
+    */
+  val SwitchTypeTandem = "t"
+
+  /** Parent trait for the different kinds of hits we can find. */
+  sealed trait SwitchHit
```
nh13:
```suggestion
  sealed trait SwitchHit(switchType: Char)
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+
+  /**
+    * Information about a switchback found by looking at soft-clipped complementary sequence within
+    * a single read.
+    *
+    * @param offset the offset between the reference position at the last base read on the original strand and
+    *               the first base read on the opposite strand.
+    * @param length the length of the post-switch sequence within the read
+    */
+  case class ReadBasedHit(var offset: Int, length: Int, var read: Option[SamRecord] = None) extends SwitchHit
```
nh13:
```suggestion
  case class ReadBasedHit(var offset: Int, length: Int, var read: Option[SamRecord] = None) extends SwitchHit('r')
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+    * The offset in this case is the distance between the end of one read and the beginning of the next.  E.g.
+    *
+    * 01234567890123456789
+    * ----->  ----->       gap = 2
+    *  <----- <-----       gap = 1
+    * ----->
+    *     ------>          gap = -2
+    * @param gap the unsequenced gap between the the 3' end of the "earlier" read and the 5' end of the "later" read.
+    */
+  case class TandemBasedHit(gap: Int) extends SwitchHit
```
nh13:
```suggestion
  case class TandemBasedHit(gap: Int) extends SwitchHit('t')
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+  }
+}
+
+@clp(group=ClpGroups.Personal, description=
+  """
+    |Finds reads where a template switch occurred during library construction.  Some library construction methods,
+    |notably ultra-low-input shotgun methods, are prone to template switching events that create molecules
+    |(templates, inserts) that instead of being a linear copy of a segment of genomic DNA, instead are chimeras formed
+    |by starting on one strand of DNA and then, switching to the opposite strand.  Frequently when the switch occurs
+    |there may be a small offset between where the first strand was departed and the opposite strand joined.
```
nh13: Add this? "This offset may be due to skipping bases on the first strand after the switch point, or skipping bases on the opposite strand after the switch point."

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+    |
+    |## Algorithm
+    |
+    |Templates that contain strand switch events (switch-backs) are found by this tool in two different ways:
+    |
+    |1. By looking at reads that contain soft-clipping at their 5' end that matches the genome, after
+    |   being reverse complemented, proximal to the 5'-most mapped base of the read.  We call these matches
```
nh13: huh? so do you look at the 5' end of the read before or after reverse complementing.

nh13: Do you mean:
```
By looking at reads that contain soft-clipping at their 5' end that matches the genome when reverse complemented.
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+    |   "read based switchbacks".  Finding read based switchbacks is based on several parameters:
+    |
+    |   1. `max-offset` controls how far away the reverse-complemented sequence is searched for.  The default value of
```
nh13:
```suggestion
    |   1. `max-offset` controls how far away to search for the reverse-complemented sequence.
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+    |      `35` allows matches to be found when the soft-clipped sequence matches the genome _starting_ at most 35bp
+    |      from the 5' mapped position of the read, and reading in the opposite direction.
+    |   2. `min-length` controls the minimum number of soft-clipped bases that must exist to trigger the search.
+    |      Given that the search looks at `2 * max-offset` locations (default=70) it is important that `min-length`
+    |      is set such that `4^min-length >> 2 * `max-offset` in order to avoid false positives.
+    |   3. `max-error-rate` allows for some mismatches to exist between the soft-clipped sequence and the genome when matching
```
nh13:
```suggestion
    |   3. `max-error-rate` allows for some mismatches to exist between the soft-clipped sequence and the genome when matching.
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+    |```sb:Z:t,{gap}```
+    |
+    |## Inputs and Outputs
+    |
+    |The tool takes as input a SAM or BAM file, and by default consumes from `stdin`.  The primary output is also a
+    |SAM or BAM file, and defaults to compressed BAM on `stdout`.  This allows the tool to be run immediately after
+    |an aligner in a pipe, e.g. `bwa mem ref.fa r1.fq r2.fq | fgbio -Xmx8g --ref=ref.fa | ...`.
+    |
+    |If the input is neither `queryname` sorted nor `queryname` grouped (i.e. all reads with the same name grouped
+    |together) it will be sorted into `queryname` order by the tool before processing.
```
nh13: perhaps mention `fgbio --tmp-dir`?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+            val foundInReadNum = readNum(hit.read.getOrElse(unreachable("read should be populated!")))
+            template.allReads.foreach { r =>
+              val num = readNum(r)
+              r(SwitchTag) = s"${SwitchTypeRead},${if (num == foundInReadNum) "read" else "mate"},${hit.offset},${hit.length}"
+            }
+        }
+
+        if (result.isDefined && !this.dontUnmap) template.unmapped else template
+      }
+    }
```
nh13: this is easily parallelizable if you want it to go faster

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,395 @@
...
+        read_based_switchbacks   = switchbackLengths.total,
+        mean_length              = switchbackLengths.mean(),
+        mean_offset              = switchbackOffsets.mean(),
+        tandem_based_switchbacks = switchbackGaps.total,
+        mean_gap                 = switchbackGaps.mean()
+      )
+
+      Metric.write(PathUtil.pathTo(prefix + ".summary.txt"), m)
+
+      for ((suffix, counter, header) <- Seq(("lengths", switchbackLengths, "length"), ("offsets", switchbackOffsets, "offset"), ("gaps", switchbackGaps, "gap_length"))) {
```
nh13: just a style thing, but 
```suggestion
Seq(
    ("lengths", switchbackLengths, "length"), 
    ("offsets", switchbackOffsets, "offset"),
    ("gaps", switchbackGaps, "gap_length")
).foreach { case suffix, counter, header) =>
```

nh13: Just a style thing for me, since now it is much easier to see how things align

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+                                      maxOffset: Int,
+                                      maxErrorRate: Double): Option[ReadBasedHit] = {
+    if (maxOffset <= 0) None else template.primaryReads.filter(_.mapped).flatMap { rec =>
+        if (rec.positiveStrand && rec.cigar.head.operator == CigarOperator.S && rec.cigar.head.length >= minLength) {
+          val ref = refs(rec.refName)
+          val bases = Sequences.revcomp(rec.basesString.take(rec.cigar.head.length))
+          val windowStart = max(1, rec.start - maxOffset)
+          val windowEnd = min(ref.length(), rec.start + bases.length + maxOffset)
+          val refBases = new String(ref.getBases, windowStart - 1, windowEnd - windowStart + 1)
+          find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(hit.offset + windowStart - rec.start); hit }
```
nh13: If you pass in `rec` to `find`, then `hit.read` can be a `val`

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+  /**
+    * Information about a switchback found by looking at soft-clipped complementary sequence within
+    * a single read.
+    *
+    * @param offset the offset between the reference position at the last base read on the original strand and
+    *               the first base read on the opposite strand.
+    * @param length the length of the post-switch sequence within the read
+    * @param read optionally the read in which the hit was found (generally filled in later)
+    */
+  case class ReadBasedHit(var offset: Int, length: Int, var read: Option[SamRecord] = None) extends SwitchHit {
```
nh13: I'd prefer `offset` and `read` to be `val`s.  Is it possible to update the method that creates a "hit" to make it so?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+    * @param maxErrorRate the maximum rate of mismatches between the switched-back sequence and the reference
+    * @return Some(ReadBasedHit) if a hit is found else None.
+    */
+  private[bam] def findReadBasedSwitchback(refs: Map[String, ReferenceSequence],
+                                      template: Template,
+                                      minLength: Int,
+                                      maxOffset: Int,
+                                      maxErrorRate: Double): Option[ReadBasedHit] = {
+    if (maxOffset <= 0) None else template.primaryReads.filter(_.mapped).flatMap { rec =>
+        if (rec.positiveStrand && rec.cigar.head.operator == CigarOperator.S && rec.cigar.head.length >= minLength) {
```
nh13: I am finding all of this really hard to follow, some commentary would be appreciated.

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
nh13: Can you add a comment explaining `-(hit.offset + windowStart - rec.start)` for me and our future selves?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+          val refBases = new String(ref.getBases, windowStart - 1, windowEnd - windowStart + 1)
+          find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(hit.offset + windowStart - rec.start); hit }
+        }
+        else if (rec.negativeStrand && rec.cigar.last.operator == CigarOperator.S && rec.cigar.last.length >= minLength) {
+          val ref = refs(rec.refName)
+          val bases = Sequences.revcomp(rec.basesString.takeRight(rec.cigar.last.length))
+          val windowStart = max(1, rec.end - bases.length - maxOffset + 1)
+          val windowEnd = min(ref.length(), rec.end + maxOffset)
+          val refBases = new String(ref.getBases, windowStart - 1, windowEnd - windowStart + 1)
+          find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(rec.end - (hit.offset + windowStart + bases.length - 1)); hit }
```
nh13: same comments as the `if` branch, minus the one about the minus.

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+        }
+        else {
+          None
+        }
+      }.find(_ => true)
```
nh13:
```suggestion
      }.headOption
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+              None
+            }
+        }
+      case _ =>
+        None
+  }
+
+  /**
+    * Attempts to find the first ungapped match of the query sequence within the target sequence. "First" is either
+    * from the start of the target sequence, or if `preferLaterMatches` is true, from the end of the target sequence.
```
nh13: I don't see `preferLaterMatches` used anywhere, remove it?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+
+    range.iterator.flatMap { targetOffset =>
+      var mismatches = 0
+      forloop (from=0, until=query.length) { i =>
+        if (!SequenceUtil.basesEqual(query(i).toByte, target(targetOffset+i).toByte)) mismatches += 1
+      }
+
+      if (mismatches <= maxMismatches) Some(ReadBasedHit(offset=targetOffset, length=query.length))
+      else None
+    }.find(_ => true)
```
nh13:
```suggestion
    }.headOption
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+          find(bases, refBases, maxErrorRate).map { hit => hit.read = Some(rec); hit.offset = -(rec.end - (hit.offset + windowStart + bases.length - 1)); hit }
+        }
+        else {
+          None
+        }
+      }.find(_ => true)
+  }
+
+  /**
+    * Attempts to find a switchback based on the orientation of the primary reads (FF or RR) and the distance between
```
nh13:
```suggestion
    * Attempts to find a tandem switchback based on the orientation of the primary reads (FF or RR) and the distance between
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+    * the end of one read (in sequencing order) and the start of the next read (in sequencing order), i.e. the gap size.
+    *
+    * @param template the template to be examined.
+    * @param maxGap the maximum allowable gap size
+    * @return either Some(TandemBasedHit) if a hit can be found, else None
```
nh13:
```suggestion
    * @return either `Some(TandemBasedHit)` if a hit can be found, else `None`
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+    */
+  private[bam] def findTandemSwitchback(template: Template, maxGap: Int) : Option[TandemBasedHit] =
+    if (maxGap <=0 ) None else template.pairOrientation match {
```
nh13:
```suggestion
    if (maxGap <= 0) None else template.pairOrientation match {
```

nh13: I think you could also do:
```
 if (maxGap <= 0) None else template.pairOrientation.collect { case _: PairOrientation.TANDEM =>
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+      case Some(PairOrientation.TANDEM) =>
+        (template.r1, template.r2) match {
+          case (Some(r1), Some(r2)) =>
```
nh13: part of me thinks you already did this matching in `template.pairOrientation`, so can't you just do `template.r1.get`?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+            val (earlier, later) = if (r1.start <= r2.start) (r1, r2) else (r2, r1)
+            val gap = later.start - earlier.end - 1
+            if (abs(gap) <= maxGap) {
+              Some(TandemBasedHit(later.start - earlier.end - 1))
+            }
+            else {
+              None
+            }
```
nh13: if/else could be a one-liner

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+        }
+      case _ =>
+        None
```
nh13: move to previous line?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+  @arg(flag='r', doc="Reference genome fasta file.") val ref: PathToFasta,
+  @arg(flag='O', doc="Maximum offset between end the two segments of the read on the reference. Set to 0 to disable read-based checks.") val maxOffset: Int = 35,
+  @arg(flag='g', doc="Maximum gap between R1 and R2 of tandem reads to call a template a switchback. Set to 0 to disable tandem-based checks.") val maxGap: Int = 500,
+  @arg(flag='l', doc="Minimum match length of the switched back segment.") val minLength: Int = 6,
+  @arg(flag='e', doc="Maximum error rate of switchback match to genome.") val maxErrorRate: Double = 0.1,
+  @arg(flag='d', doc="IF true, do NOT unmap reads from switchback templates.") val dontUnmap: Boolean = false
+) extends FgBioTool with LazyLogging {
+  import FindSwitchbackReads._
+  Io.assertReadable(input)
+  Io.assertReadable(ref)
```
nh13: `metrics.map(m => Io.canWriteFile(m)` or equivalent?  Same for `output`?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+class FindSwitchbackReads
+( @arg(flag='i', doc="Input BAM file.") val input: PathToBam = Io.StdIn,
+  @arg(flag='o', doc="Output BAM file.") val output: PathToBam = Io.StdOut,
+  @arg(flag='m', doc="Metrics output file.") val metrics: Option[PathPrefix] = None,
+  @arg(flag='s', doc="Output sort order.") val sortOrder: Option[SamOrder] = None,
+  @arg(flag='r', doc="Reference genome fasta file.") val ref: PathToFasta,
+  @arg(flag='O', doc="Maximum offset between end the two segments of the read on the reference. Set to 0 to disable read-based checks.") val maxOffset: Int = 35,
+  @arg(flag='g', doc="Maximum gap between R1 and R2 of tandem reads to call a template a switchback. Set to 0 to disable tandem-based checks.") val maxGap: Int = 500,
+  @arg(flag='l', doc="Minimum match length of the switched back segment.") val minLength: Int = 6,
+  @arg(flag='e', doc="Maximum error rate of switchback match to genome.") val maxErrorRate: Double = 0.1,
```
nh13:
```suggestion
  @arg(flag='e', doc="Maximum mismatch error rate of switchback match to genome.") val maxErrorRate: Double = 0.1,
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+
+    val progress = ProgressLogger(logger)
+    var total   = 0L
+    var aligned = 0L
+    val switchbackGaps    = new NumericCounter[Int]()
+    val switchbackLengths = new NumericCounter[Int]()
+    val switchbackOffsets = new NumericCounter[Int]()
+
+    val in = SamSource(input)
+    val iterator = Bams.templateIterator(in)
```
nh13: Just a thought.  Perhaps `Bams.templateIterator` should return a `SelfClosingIterator[SamRecord]`?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+    val outHeader = in.header.clone()
+    sortOrder match {
+      case Some(so) => so.applyTo(outHeader)
+      case None     =>
+        if (in.header.getSortOrder != SortOrder.queryname && in.header.getGroupOrder != GroupOrder.query) {
+          SamOrder.Queryname.applyTo(outHeader)
+        }
+    }
+    sortOrder.foreach(_.applyTo(outHeader))
+    val out = SamWriter(output, header=outHeader, sort=sortOrder)
```
nh13: Just my thought on style and scoping, but I'd prefer this:
```
val out = {
    val outHeader = in.header.clone()
    sortOrder match {
      case Some(so) => so.applyTo(outHeader)
      case None     =>
        if (in.header.getSortOrder != SortOrder.queryname && in.header.getGroupOrder != GroupOrder.query) {
          SamOrder.Queryname.applyTo(outHeader)
        }
    }
    sortOrder.foreach(_.applyTo(outHeader))
    SamWriter(output, header=outHeader, sort=sortOrder)
}
```
For me, it keeps objects tightly scoped.

nh13: If you want to go all the way, which you don't, you could do:
```
val out = {
    val outHeader = {
        val header = in.header.clone()
        sortOrder match {
          case Some(so) => so.applyTo(header)
          case None     =>
            if (in.header.getSortOrder != SortOrder.queryname && in.header.getGroupOrder != GroupOrder.query) {
              SamOrder.Queryname.applyTo(header)
            }
        }
        sortOrder.foreach(_.applyTo(header))
    }
    SamWriter(output, header=outHeader, sort=sortOrder)
}
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+
+    val outHeader = in.header.clone()
+    sortOrder match {
+      case Some(so) => so.applyTo(outHeader)
+      case None     =>
+        if (in.header.getSortOrder != SortOrder.queryname && in.header.getGroupOrder != GroupOrder.query) {
+          SamOrder.Queryname.applyTo(outHeader)
+        }
+    }
+    sortOrder.foreach(_.applyTo(outHeader))
```
nh13: I think this is unnecessary since you already applied it on 316

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+    val out = SamWriter(output, header=outHeader, sort=sortOrder)
+
+    logger.info("Examining reads for switchbacks.")
+    iterator.map { template =>
+      total += 1
+      if (!template.primaryReads.exists(_.mapped)) template else {
```
nh13: `template.primaryReads.forall(_.unmapped)`?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+        aligned += 1
+
+        val result: Option[SwitchHit] = findReadBasedSwitchback(refs, template, minLength, maxOffset, maxErrorRate)
+          .orElse(findTandemSwitchback(template, maxGap))
+
+        result match {
```
nh13: why not 
```
result.foreach {
   case hit: TandemBasedHit => ...
   case hit: ReadBasedHit => ....
}
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+            template.allReads.foreach { r =>
+              val num = readNum(r)
+              r(SwitchTag) = s"${hit.code},${if (num == foundInReadNum) "read" else "mate"},${hit.offset},${hit.length}"
+            }
+        }
+
+        if (result.isDefined && !this.dontUnmap) template.unmapped else template
+      }
+    }
+    .foreach { template =>
```
nh13: why does this need another `foreach` on the templates, why not just put it after the `if/else` that starts on 328?

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
```diff
@@ -0,0 +1,398 @@
...
+        read_based_switchbacks   = switchbackLengths.total,
+        mean_length              = switchbackLengths.mean(),
+        mean_offset              = switchbackOffsets.mean(),
+        tandem_based_switchbacks = switchbackGaps.total,
+        mean_gap                 = switchbackGaps.mean()
+      )
+
+      Metric.write(PathUtil.pathTo(prefix + ".summary.txt"), m)
+
+
```
nh13: bingo

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -227,6 +229,14 @@ trait SamRecord {
       refIndex == mateRefIndex &&
       SamPairUtil.getPairOrientation(this) == PairOrientation.FR
   }
+
+  /** Exposes clone() publicly for use. */
```
nh13:
```suggestion
  /** Exposes [[SAMRecord.clone()]] publicly for use. */
```

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -227,6 +229,14 @@ trait SamRecord {
...
+  override def clone(): SamRecord = {
+    val r = super.clone().asInstanceOf[SamRecord]
+    r.bases = this.bases.clone()
+    r.quals = this.quals.clone()
```
nh13: does `SAMRecord.clone()` not do this?  Yikes.

src/test/scala/com/fulcrumgenomics/bam/FindSwitchbackReadsTest.scala
```diff
@@ -0,0 +1,350 @@
...
+import com.fulcrumgenomics.util.Metric.{Count, Proportion}
+import htsjdk.samtools.reference.ReferenceSequenceFileFactory
+import org.scalatest.OptionValues
+import org.scalatest.OptionValues.convertOptionToValuable
+
+class FindSwitchbackReadsTest extends UnitSpec with OptionValues {
+  private val (ref, refPath) = {
+    val builder = new ReferenceSetBuilder(assembly=Some("TF"))
+    val seq = builder.add("tf")
+    // Following sequence is 70bp per line
```
nh13: verified!

src/test/scala/com/fulcrumgenomics/bam/FindSwitchbackReadsTest.scala
```diff
@@ -0,0 +1,350 @@
...
+      |TCGAGGCAAGAAGTCCTGCCACACGGGTCTAGGCAGGTCCGCTGGGTGGAACATCCCCATAGGCTTACTT
+      |TACTGTGACTTACCTGAGCCACGTAAACCTCTTGAGAAAGCAGTGGCCAATTTCTTCTCGGGCAGCTGTG
+      |CCCCTTGTGCGGATGGGACGGACTTCCCCCAGCTGTGTCAACTGTGTCCAGGGTGTGGCTGCTCCACCCT
+      |TAACCAATACTTCGGCTACTCGGGAGCCTTCAAGTGTCTGAAGGATGGTGCTGGGGATGTGGCCTTTGTC
+      |AAGCACTCGACTATATTTGAGAACTTGGCAAACAAGGCTGACAGGGACCAGTATGAGCTGCTTTGCCTGG
+      |ACAACACCCGGAAGCCGGTAGATGAATACAAGGACTGCCACTTGGCCCAGGTCCCTTCTCATACCGTCGT
+      |GGCCCGAAGTATGGGCGGCAAGGAGGACTTGATCTGGGAGCTTCTCAACCAGGCCCAGGAACATTTTGGC
+      |AAAGACAAATCAAAAGAATTCCAACTATTCAGCTCTCCTCATGGGAAGGACCTGCTGTTTAAGGACTCTG
+      |CCCACGGGTTTTTAAAAGTCCCCCCCAGGATGGATGCCAAGATGTACCTGGGCTATGAGTATGTCACTGC
+      |CATCCGGAATCTACGGGAAGGCACATGCCCAGAAGCCCCAACAGATGAATGCAAGCCTGTGAAGTGGTGT
```
nh13: well, you see that second `C`, that's your problem right there.

src/test/scala/com/fulcrumgenomics/bam/FindSwitchbackReadsTest.scala
```diff
@@ -0,0 +1,350 @@
...
+    val builder = new SamBuilder(readLength=50, sd=Some(ref.getSequenceDictionary))
+    val ff = Template(builder.addPair(contig=0, start1=100, start2=160, strand1=Plus, strand2=Plus).iterator)
+    FindSwitchbackReads.findTandemSwitchback(ff, maxGap=200) shouldBe defined
+
+    ff.r2.value.refIndex = 1
+    ff.r1.value.mateRefIndex = 1
+    FindSwitchbackReads.findTandemSwitchback(ff, maxGap=200) shouldBe None
+  }
+
+  it should "not call FR or RF pairs switchbacks" in {
```
nh13:
```suggestion
  it should "not call FR or RF pairs tandem switchbacks" in {
```

src/test/scala/com/fulcrumgenomics/bam/FindSwitchbackReadsTest.scala
```diff
@@ -0,0 +1,350 @@
...
+    (Seq(-201, 201) ++ Range(-200, 200, step=20)).foreach { gap =>
+      val ff  = Template(builder.addPair(contig=0, start1=start, start2=start+readLen+gap, strand1=Plus, strand2=Plus).iterator)
+      val hit = FindSwitchbackReads.findTandemSwitchback(ff, maxGap=200)
+
+      if (gap <= 200 && gap >= -200) hit.value.gap shouldBe gap
+      else hit shouldBe None
+    }
+  }
+
+  it should "not call tandem pairs switchbacks if the reads are on different chromosomes" in {
```
nh13:
```suggestion
  it should "not call pairs tandem switchbacks if the reads are on different chromosomes" in {
```

src/test/scala/com/fulcrumgenomics/bam/FindSwitchbackReadsTest.scala
```diff
@@ -0,0 +1,350 @@
...
+  "FindSwitchbackReads.findTandemSwitchback" should "call FF and RR templates with small enough gaps as switchbacks" in {
+    val builder = new SamBuilder(readLength=50, sd=Some(ref.getSequenceDictionary))
+    val ff = Template(builder.addPair(contig=0, start1=100, start2=160, strand1=Plus, strand2=Plus).iterator)
+    val rr = Template(builder.addPair(contig=0, start1=100, start2=160, strand1=Minus, strand2=Minus).iterator)
+
+    FindSwitchbackReads.findTandemSwitchback(ff, maxGap=200).value.gap shouldBe 10
+    FindSwitchbackReads.findTandemSwitchback(rr, maxGap=200).value.gap shouldBe 10
+  }
+
+  it should "not call tandem pairs switchbacks if the gap is too big" in {
```
nh13:
```suggestion
  it should "not call pairs tandem switchbacks if the gap is too big" in {
```

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
tfenne: Done.  Although I think it's overkill since it always calls `queryGroupedIterator` which yields a `SelfClosingIterator`.

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
tfenne: Ah, thanks.  Yes, this changed around a bunch and this is a hangover.

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
tfenne: Changed, but not sure how I feel about using `.foreach { case ...` for more than just value decomposition.  It feels odd to have multiple `case`s without a `match` keyword nearby.

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
tfenne: I just think it's clearer this way - since the output template varies based on the path above, it's easier/clear to `map().foreach()`.  The other option is to catch the return value of the `if` on line 328, and convert this into a single `foreach`.  I don't think that's any better.

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
tfenne: you sunk my battleship!

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
tfenne: It doesn't.  Which is pretty janky.  I think they were trying to save time/memory when just modifying some other things, but it's nasty.

src/test/scala/com/fulcrumgenomics/bam/FindSwitchbackReadsTest.scala
tfenne: I prefer the original wording.  They are 100% "tandem pairs" and the outcome I want is that they are identified as switchbacks via the tandem-finding method.

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
nh13: I don't think it'd be _that_ complicated, since your just adding some parameters to a function or two, but since I am just a commentator, I'll let it go.

src/main/scala/com/fulcrumgenomics/bam/FindSwitchbackReads.scala
nh13: To be clear, there are no calls to `find` that change the value of `preferLaterMatches` from` false.  I can only see `find` being called in `findReadBasedSwitchback`, which does not provide a value for `preferLaterMatches`

tfenne:
Ah - that's a good find @nh13.  The else for the reverse strand reads should be passing `preferLaterMatches=true`, so that both invocations search in the same direction relative to the original sequencing strand..  I've added that.  It only makes a difference if the clipped sequence can match in multiple places, which shouldn't happen too often.

PR #443, opened 2018-11-05
Added test of streaming support to FastqIOTest

gpratt:
Not quite sure if this should be a PR, but the added failing test shows `FastqSource` doesn't support streaming.


nh13:
Fixed in https://github.com/fulcrumgenomics/commons/pull/36.  It's better to have the unit tests there, though we could consider this an integration tests.  Lets close it for now.

PR #445, opened 2018-11-13
Added Allele Depths option to VariantContextSetBuilder

gpratt:
VariantContextSetBuilder might need the ability to specify allele depths explicitly.  This adds the feature.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
+++ b/src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
@@ -92,7 +92,9 @@ class VariantContextSetBuilder(sampleNames: Seq[String] = List("Sample")) extend
                  genotypeAlleles: List[String] = List.empty,
                  genotypeAttributes: Map[String,Any] = Map.empty,
                  sampleName: Option[String] = None,
-                 phased: Boolean = false): this.type = {
+                 phased: Boolean = false,
+                 ad: List[Int] = List.empty
+                ): this.type = {
     if (!sampleName.forall { sn => this.header.getGenotypeSamples.contains(sn)}) {
       throw new IllegalArgumentException(s"Sample with name $sampleName not found in the VCF header.")
     }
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -92,7 +92,9 @@ class VariantContextSetBuilder(sampleNames: Seq[String] = List("Sample")) extend
                  genotypeAlleles: List[String] = List.empty,
                  genotypeAttributes: Map[String,Any] = Map.empty,
                  sampleName: Option[String] = None,
-                 phased: Boolean = false): this.type = {
+                 phased: Boolean = false,
+                 ad: Option[List[Int]] = None
+                ): this.type = {
```
nh13: move this to the previous line

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -92,7 +92,9 @@ class VariantContextSetBuilder(sampleNames: Seq[String] = List("Sample")) extend
                  genotypeAlleles: List[String] = List.empty,
                  genotypeAttributes: Map[String,Any] = Map.empty,
                  sampleName: Option[String] = None,
-                 phased: Boolean = false): this.type = {
+                 phased: Boolean = false,
+                 ad: Option[List[Int]] = None
```
nh13: `ad` -> `alleleDepth` 

Avoid using one or two letter parameters in public APIs, and use them only in very small scopes.

nh13: Why can we just have `alleleDepths: List[Int] = List.empty`?  What's the added value of an option versus an empty list?

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -129,6 +131,7 @@ class VariantContextSetBuilder(sampleNames: Seq[String] = List("Sample")) extend
       case gAlleles => new GenotypeBuilder(name, toAlleles(gAlleles, referenceAllele=referenceAllele).asJava)
     }
     genotypeBuilder.phased(phased)
```
nh13: Do you want to make sure that the length of alleleDepths (when >0) is the same as the # of alleles?

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
gpratt: The allele depth is automatically set by the builder when you don't define it, so I thought it would be better as an `Option` to indicate that you have control, rather than it just being empty by default. 

I've changed it to a List.empty for now, but happy to change it back.

PR #448, opened 2018-11-28
Improve genotype attribute support in VariantContextSetBuilder

nh13:
Here's my attempt at https://github.com/fulcrumgenomics/fgbio/pull/445


src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -142,6 +151,27 @@ class VariantContextSetBuilder(sampleNames: Seq[String] = List("Sample")) extend
     yieldingThis(this.variants.append(ctxBuilder.make()))
   }
 
+  /** Returns `v` as type `T`.  Throws an exception if `v` is not assignable to type `T`.*/
+  private def anyToType[T : ClassTag](k: String, v: Any, typeName: String): T = {
```
tfenne: Why do all this vs. just having `v.asInstanceOf[Int]` inline?  Seems overkill.

src/main/scala/com/fulcrumgenomics/testing/VariantContextSetBuilder.scala
```diff
@@ -142,6 +151,27 @@ class VariantContextSetBuilder(sampleNames: Seq[String] = List("Sample")) extend
...
+  /** Returns `v` as type `Array[T]`.  Throws an exception if `v` is not assignable to `TraversableOnce[_]` or `Array[_]`,
+    * or if any element in `v` is not of type `T` .*/
+  private def anyToArray[T : ClassTag](k: String, v: Any, typeName: String): Array[T] = {
+    val values = v match {
+      case _v: TraversableOnce[_] => _v.toArray[Any]
+      case _v: Array[_]           => _v
+      case _                      =>
+        throw new IllegalArgumentException(s"$k attribute value must be an TraversableOnce[$typeName], found ${v.getClass.getSimpleName}")
+    }
+    values.map(anyToType[T](k, _, typeName))
```
tfenne: I think you could also just do `k.asInstanceOf[T]` here.

PR #452, opened 2018-11-30
Change aligner to generate cigar matches when bases are compatible

tfenne:
And also a tiny bit more performance tuning.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
+++ b/src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
@@ -27,6 +27,7 @@ package com.fulcrumgenomics.alignment
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.alignment.Mode.{Global, Glocal, Local}
 import com.fulcrumgenomics.alignment.Aligner._
+import com.fulcrumgenomics.util.Sequences
 import enumeratum.EnumEntry
 import htsjdk.samtools.CigarOperator
 
@@ -97,6 +98,12 @@ object Aligner {
   * A scoring function (`scoringFunction`) is taken to score pair-wise aligned bases. A default
   * implementation is supplied via the companion [[Aligner]] object which uses a fixed match
   * score and mismatch penalty.
+  *
+  * When generating CIGARs for alignments the aligner uses the [[isMatch()]] method to determine
+  * whether to treat an aligned pair of bases as a match or mismatch.  The default implementation
+  * of this method treats U bases as T bases to allow DNA/RNA alignments.  It also will identify
+  * as matches any pair of bases (including IUPAC ambiguity codes) that share at least one base
+  * in common.  This behaviour can be modified by overriding the [[isMatch()]] method.
   *
   * @param scoringFunction a function to score the alignment of a pair of bases. Parameters
   *                        are the query and target bases as bytes, in that order.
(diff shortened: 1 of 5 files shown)


src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
```diff
@@ -389,9 +395,15 @@ class Aligner(val scoringFunction: (Byte,Byte) => Int,
         currLength   = 1
       }
 
-      if (done()) elems += CigarElem(currOperator, currLength)
+      nextD = matrices(curD).trace(curI,curJ)
```
nh13: space between commas

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -284,8 +274,22 @@ case class Alignment(query: Array[Byte],
...
+          op match {
+            case CigarOperator.EQ => alignBuffer.append(matchChar)
+            case CigarOperator.X  => alignBuffer.append(mismatchChar)
+            case CigarOperator.M  => alignBuffer.append(if (q == t) matchChar else mismatchChar)
+            case other            => throw new IllegalStateException(s"Unsupported cigar operator: $other")
+          }
+
+          qOffset += 1
+          tOffset += 1
+        }
```
nh13: why was this moved to the last `case` statement?  I would think it happens the most frequently, so it would be best to have it first?

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -37,6 +37,54 @@ object Sequences {
   /** Common contig/chrom names for non-autosomal sequences in mammals. */
   val CommonNonAutosomalContigNames = Seq("M", "chrM", "MT", "X", "chrX", "Y", "chrY")
 
+  /** Array for determining what bases are compatible. Indices are all DNA/RNA bases. Contents
+    * are Int masks with a bit set for each possible base.  U is treated as identical to T.
+     */
```
nh13: indent

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -37,6 +37,54 @@ object Sequences {
...
+    *   - `compatible('A', 'R') == true`
+    *   - `compatible('M', 'S') == true`
+    *   - `compatible('A', 'C') == false`
+    *   - `compatible('M', 'K') == false`
+    *
+    * @param base1 the first base to be compared
+    * @param base2 the second base to be compared
+    * @return true if the bases share at least one concrete base in common, false otherwise
+    */
+  final def compatible(base1: Byte, base2: Byte): Boolean = (IupacMasks(base1) & IupacMasks(base2)) > 0
```
nh13: should `compatible('?'.toByte, ?'.toByte)` be allowed (i.e. should it throw an exception?  You could check if IupacMasks(base1/2) returns 0.  Up to you.

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
tfenne: I had to either move it to the end, or turn it into three cases, because I want it to match all three of `=/X/M`, but also to know which one it was.  I've converted the code to match just on the operator so it should be more efficient, to make up for the change.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
tfenne: I waffled on this.  I've changed it to `base1 == base2 || ...` which will allow exact matches for characters we don't expect, and probably also make it more efficient when comparing mostly-identical sequences.

PR #453, opened 2018-11-30
Added revcomp() functions to Sequences that support ambiguous bases.

tfenne:


src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -150,15 +152,74 @@ object Sequences {
...
     val bs = s.getBytes
-    forloop(from=0, until=bs.length) { i =>  bs(i) = SequenceUtil.complement(bs(i)) }
+    complement(bs)
     new String(bs)
   }
 
+  /** Reverse complements an array of bases in place. See [[complement()]] for how bases are complemented. */
+  def revcomp(bs: Array[Byte]): Unit = {
+    SequenceUtil.reverse(bs, 0, bs.length)
+    complement(bs)
```
nh13: I'm torn, since you can't use `SequenceUtil.reverseComplement` since you want IUPAC support, but you make two passes through the array.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -150,15 +152,74 @@ object Sequences {
...
+        case 'A' => 'T'
+        case 'C' => 'G'
+        case 'G' => 'C'
+        case 'T' => 'A'
+        case 'U' => 'A'
+        case 'M' => 'K'
+        case 'R' => 'Y'
+        case 'W' => 'S'
+        case 'S' => 'W'
+        case 'Y' => 'R'
```
nh13: can you move this to after line 190 to keep IUPAC codes that represent two bases together? Same for K/M, R/Y, and W/S.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -150,15 +152,74 @@ object Sequences {
...
+        case 'K' => 'M'
+        case 'V' => 'B'
```
nh13: Could you document that a complement of an IUPAC character is complement of each base it represents?  You could misinterpret complement as the inverse of the IUPAC base, namely V's complement is T.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -150,15 +152,74 @@ object Sequences {
...
+        case 'H' => 'D'
+        case 'D' => 'H'
+        case 'B' => 'V'
```
nh13: Same, B->V and V->B, just have them one after the other.  Helps check correctness.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -150,15 +152,74 @@ object Sequences {
...
+        case 'N' => 'N'
+        case 'a' => 't'
```
nh13: you could just check if the character is lower case before the match, then upper case it for the match, then lower case the result if necessary.

nh13: Since it's less likely to have a lot of IUPAC bases, if you stick with listing all the upper and lower case bases, how about moving the a/c/g/t/u cases to under the A/C/G/T/U cases?

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
tfenne: Eh, fine, I've implemented it as single pass.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
tfenne: Heh, my first implementation did the upper/lower casing, but I think it's more efficient to have a single match/switch.  Also the order shouldn't really matter as it's getting compiled to a switch which is just a big jump table.

PR #456, opened 2019-01-07
Add test cases for whitespace padded Sample.extendedAttributes values

clintval:


src/main/scala/com/fulcrumgenomics/illumina/Sample.scala
```diff
@@ -74,10 +74,9 @@ case class Sample(sampleOrdinal: Int,
 
   /** Gets an extended attribute for the given column name.  The column name can be any case. */
   def extendedAttribute(name: String): Option[String] = {
-    extendedAttributes.get(name.toUpperCase) match {
-      case None => None
-      case Some(str) if str.isEmpty => None
-      case Some(str) => Some(str.trim)
+    extendedAttributes.get(name.toUpperCase).map(_.trim) match {
+      case None | Some("") => None
+      case Some(str)       => Some(str)
```
nh13: forgot the `str.trim`!

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -138,10 +138,9 @@ object SampleSheet {
 
   /** Gets the trimmed value of the given key (`name`) from the map.  If empty or not found, returns None */
   protected def getStringField(sampleDatum: Map[String, String], name: String): Option[String] = {
-    sampleDatum.get(name.toUpperCase) match {
-      case None => None
-      case Some(str) if str.isEmpty => None
-      case Some(str) => Some(str.trim)
+    sampleDatum.get(name.toUpperCase).map(_.trim) match {
+      case None | Some("") => None
+      case Some(str)       => Some(str)
```
nh13: forgot the `str.trim`!

src/test/scala/com/fulcrumgenomics/illumina/SampleTest.scala
```diff
@@ -41,13 +41,16 @@ class SampleTest extends UnitSpec with OptionValues{
   }
 
   it should "support extended attributes" in {
-    val sample = new Sample(0, "ID", "NAME", "LIBRARY", extendedAttributes=Seq(("FOO", "1"), ("BAR", "2  ")).toMap)
+    val sample = new Sample(0, "ID", "NAME", "LIBRARY", extendedAttributes=Seq(("FOO", "1"), ("BAR", " 2  "), ("BAZ", "  ")).toMap)
```
nh13: how about an attribute with only spaces?

clintval: Only spaces for the key, or value? The latter was added `("BAZ", "  ")`.

I think a requirement of the spec. would be to have no key-less values (whitespace only). I could add that requirement here?

https://github.com/fulcrumgenomics/fgbio/blob/3183c233307250564088561fad22a83f368e4cda/src/main/scala/com/fulcrumgenomics/illumina/Sample.scala#L61

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
clintval: Line 141 trims, then matches. Not good?

tfenne:
@nh13 I think the change is good and I'm fine with the small API/contract change.

PR #462, opened 2019-02-11
Add unmatchedBamFileFrom() to get the unmatched BAM file paths

clintval:
We have a helper function for grabbing sample/lane BAMs, it would be convenient to have one for the unmatched BAMs as well.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParamsForPicard.scala
+++ b/src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParamsForPicard.scala
@@ -179,8 +179,8 @@ object BasecallingParams {
 
   /** Produces the path to the BAM file for the given sample and lane.
     *
-    * The BAM file name will <sample-name>.<barcode-seq>.<lane>.bam.  The resulting file name will be sanitized to
-    * remove problematic characters.
+    * The BAM file name will be of the form <sample-name>.<barcode-seq>.<lane>.bam.  The resulting file name will be
+    * sanitized to remove problematic characters.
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParamsForPicard.scala
```diff
@@ -192,4 +192,18 @@ object BasecallingParams {
...
+  /** Produces the path to the unmatched BAM file for the given lane.
+    *
+    * The BAM file name will be of the form "unmatched.<lane>.bam".  The resulting file name will be sanitized to
+    * remove problematic characters.
+    *
+    * @param output the base output directory.
+    * @param lane the lane number.
+    * @return the path to the unmatched BAM file for the given lane.
+    */
+  def unmatchedBamFileFrom(output: DirPath, lane: Int): PathToBam = {
```
nh13: Looks good.  Could you plumb this in here: https://github.com/fulcrumgenomics/fgbio/blob/e1ac3b157c6abb2992a88d1a31f677329c8c05fc/src/main/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParamsForPicard.scala#L150

That way if we change this method, it will update the params file too?

nh13: Bonus points for making the name `"unmatched"` a static variable in the object and using it instead of hard-code strings.

clintval:
Went for the bonus points, back to you!

PR #467, opened 2019-02-27
WIP: Add some tools to reformat based on alternate names

nh13:


nh13:
I think I'll add a wiki page with something along the lines of this script when this gets merged: https://gist.github.com/nh13/37a829af9d7ae3a9805caa0151f45211

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -0,0 +1,175 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
```
jacarey: Shouldn't we keep the base repo license up to date and not add it to each individual file?

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -0,0 +1,175 @@
...
+  case object GenBankAccession extends AssemblyReportColumn { val key: String = "GenBank-Accn" }
+  case object RefSeqAccession  extends AssemblyReportColumn { val key: String = "RefSeq-Accn" }
+  case object UcscName         extends AssemblyReportColumn { val key: String = "UCSC-style-name" }
+
+  val MissingValue: String   = "na"
+  val SequenceRole: String   = "Sequence-Role"
+  val SequenceLength: String = "Sequence-Length"
+}
+
+/** Trait that entries in SequencRole will extend. */
```
jacarey: Sequenc -> Sequence

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -0,0 +1,175 @@
...
+  case object UnlocalizedScaffold extends SequenceRole { val key: String = "unlocalized-scaffold"; val primary: Boolean = false }
+  case object UnplacedScaffold    extends SequenceRole { val key: String = "unplaced-scaffold"; val primary: Boolean = false }
+}
+
+
+@clp(description =
+  """
+    |Collates the alternate contig names from an NCBI assembly report.
+    |
+    |The input is be the `*.assembly_report.txt` obtained from NCBI.
```
jacarey: "The input is be the" ?

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -0,0 +1,175 @@
...
+    |
+    |The output will have the first column is the primary name and the second column is an alternative name.  If there
```
jacarey: "The output will have primary name as the first column and alternate name as the second column." Reads a little better to me

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -0,0 +1,175 @@
...
+    |is more than one alternate name, each alternate name will be on a separate line.  The primary name should be
+    |specified with `--primary` while the alternate name(s) should be specified with `--alternates`.
+    |
+    |First, sequences with the Sequence-Role "assembled-molecule" will be outputted.  Next, the remaining sequences will
+    |be sorted by length, with smallest sequence output first.
```
jacarey: "sorted by descending length"

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -0,0 +1,175 @@
...
+      if (sequenceRoles.isEmpty || sequenceRoles.exists(_.key == d(AssemblyReportColumn.SequenceRole))) {
+        Some(d)
+      } else {
+        None
+      }
+    }.toSeq
+
+    // Re-order the entries based on primary vs non primary sequence roles, then based on increasing sequence length
+    val primary   = maps.filter { d => SequenceRole(d(AssemblyReportColumn.SequenceRole)).primary }
+    val secondary = maps.filter { d => !SequenceRole(d(AssemblyReportColumn.SequenceRole)).primary}.sortBy { d => d(AssemblyReportColumn.SequenceLength)}
```
jacarey: Will this sort be lexographic if the AssemblyReportColumn.SequenceLength is a string?

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -0,0 +1,106 @@
...
+        out.write(bases(baseIdx))
+        progress.record(targetName, baseIdx + 1)
+        baseCounter += 1
+        if (baseCounter >= lineLength) {
+          out.newLine()
+          baseCounter = 0
+        }
+      }
+      if (baseCounter > 0) out.newLine()
+      seen.add(srcName)
```
jacarey: Isn't this redundant with line 76?

src/main/scala/com/fulcrumgenomics/util/UpdateGffContigNames.scala
```diff
@@ -0,0 +1,151 @@
...
+ @arg(doc="Skip missing contigs.") val skipMissing: Boolean = false
+) extends FgBioTool with LazyLogging {
+
+  Io.assertReadable(input)
+  Io.assertReadable(Seq(input, mapping))
+  Io.assertCanWriteFile(output)
+
+  private val sequenceRegionKey = "##sequence-region"
+
+
```
jacarey: extra newline

src/main/scala/com/fulcrumgenomics/util/UpdateGffContigNames.scala
```diff
@@ -0,0 +1,151 @@
...
+  override def execute(): Unit = {
+    val progress = ProgressLogger(logger, noun="features", verb="written", unit=10e5.toInt)
+
+    val srcToTarget = Io.readLines(mapping).map { line =>
+      val fields = line.split('\t')
+      require(fields.length <= 2, s"Malformed line: expected at least two columns: $line")
+      (fields(0), fields(1))
+    }.toMap
```
jacarey: This validation is similar to to UpdateFastaContigNames.scala:70 and I think these inputs are the same. The only difference being that one adds to "seen" set. Seems like it could be refactored though.

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -0,0 +1,106 @@
...
+import htsjdk.samtools.reference.{FastaSequenceIndex, ReferenceSequenceFileFactory}
+
+@clp(description =
+  """
+    |Updates the sequence names in a FASTA.
+    |
+    |The first column of the input is the source name and the second column is an target name.  If there
+    |are more than one target names (ex. multiple alternates), each alternate name may be on a separate line or as
+    |additional columns.  Only the first target name will be considered.  This is mainly to support the output of
+    |`CollectAlternateContigNames`.
```
jacarey: This fasta will need to be reindexed right? Should we mention that here or maybe do that inline?

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
sstadick: Yes it would be, I'm going to convert it to an Int. @nh13, does that sound correct to you?

src/main/scala/com/fulcrumgenomics/util/UpdateGffContigNames.scala
sstadick: I've added a 'ContigNameMapping' class to parse the mapping file and handle many to one src-> target mappings.

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
sstadick: After looking at the tests, this does seem to actually want to be in lexicographical order

src/main/scala/com/fulcrumgenomics/util/ContigNameMapping.scala
```diff
@@ -0,0 +1,43 @@
...
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.util
+
+import com.fulcrumgenomics.FgBioDef._
+
+object  ContigNameMapping {
+    /**
+      * Parses a contig name Mappings file that is `source`, `List[target]`.
```
nh13, line 31: Document how they are delimited.  Perhaps even inline an example file (first 5 lines?).

src/main/scala/com/fulcrumgenomics/util/ContigNameMapping.scala
```diff
@@ -0,0 +1,43 @@
...
+      *
+      * Multiple targets can be specified on one line, or on multiple lines.
+      */
+    def parse(nameMapping: FilePath): Map[String, List[String]] = {
+
```
nh13, line 36: extra newline

src/main/scala/com/fulcrumgenomics/util/ContigNameMapping.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.fulcrumgenomics.FgBioDef._
+
+object  ContigNameMapping {
+    /**
+      * Parses a contig name Mappings file that is `source`, `List[target]`.
+      *
+      * Multiple targets can be specified on one line, or on multiple lines.
+      */
+    def parse(nameMapping: FilePath): Map[String, List[String]] = {
```
nh13, line 35: Can be a `Seq` instead of `List`

src/main/scala/com/fulcrumgenomics/util/ContigNameMapping.scala
```diff
@@ -0,0 +1,43 @@
...
+
+        Io.readLines(nameMapping).map { line =>
+            val fields = line.split('\t')
+            require(fields.length >= 2, s"Malformed line: expected at least two columns: $line")
+            (fields.head, fields.tail)
```
nh13, line 40: up to you, but you could also match on `line.split('\t')`

src/main/scala/com/fulcrumgenomics/util/UpdateBedContigNames.scala
```diff
@@ -0,0 +1,72 @@
...
+import com.fulcrumgenomics.util.ContigNameMapping
+
+@clp(description =
+  """
+    |Updates the sequence names in a BED file.
+    |
+    |The first column of the input is the source name and the second column is an target name.  If there
+    |are more than one target names (ex. multiple alternates), each alternate name may be on a separate line or as
+    |additional columns.  Only the first target name will be considered.  This is mainly to support the output of
+    |`CollectAlternateContigNames`.
```
nh13, line 42: How about we change this tool to `UpdateIntervaListContigNames`?  I don't like BED files, and have a semi-dogmatic view that fgbio should operate on interval lists and not BED files to force folks to use interval lists.  If you must have a BED, they can use `picard IntervalListToBed`

src/main/scala/com/fulcrumgenomics/util/UpdateBedContigNames.scala
```diff
@@ -0,0 +1,72 @@
...
+  Io.assertReadable(Seq(input, mapping))
+  Io.assertCanWriteFile(output)
+
+  override def execute(): Unit = {
+    val progress = ProgressLogger(logger, noun="bases", verb="written", unit=10e7.toInt)
+    val out      = Io.toWriter(output)
+
+    val srcToTarget = ContigNameMapping.parse(mapping)
+
+    Io.readLines(input).foreach{ line =>
```
nh13, line 62:
```suggestion
    Io.readLines(input).foreach { line =>
```

tfenne:
I haven't had time to go through this in detail yet, but I have some thoughts I wanted to share before this got too close to being merged.  I think this ties in with some discussion we had on implementing our own `ReferenceFile` stuff, and some other work I've done elsewhere where I needed to do similar things.

I tend to think that we should:

1. Implement our own set of classes for a `SequenceDictionary` and `SequenceRecord` (better names welcome).  These classes should have first-class support for sequence aliases (`AN` in the .dict file), by which I mean `SequenceRecord` should have a vals/defs for `name`, `aliases` and `allNames` which gives the primary and all aliases.  And `SequenceDictionary` should allow lookup of sequences by primary name and or alias via the same mechanism (e.g. `dict("1")` and `dict("chr1")` should yield the same result).

2. Make it really easy to take in a FASTA (or per-chrom fastas?), rename the contigs, re-order the contigs, normalize (i.e. fix line lengths), update contig names, and then emit a target fasta file and possibly also sequence dictionary (and .fai)?

3. Have all of the `UpdateContigNamesIn...`, with the exception of the FASTA one, use the sequence dictionary to drive the process.

E.g. attached is an HG38 sequence dictionary ([hg38.dict.txt](https://github.com/fulcrumgenomics/fgbio/files/4523421/hg38.dict.txt)) I generated for this exact purpose (also by parsing the NCBI assembly report!).

PR #470, opened 2019-03-20
Various typo and style fixups

nh13:
Only one change that I think is consequential; I'll add a comment highlight ing it.

image


src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
```diff
@@ -426,16 +426,16 @@ class Aligner(val scorer: AlignmentScorer,
 
     while (nextD != Done) {
       val op    = (curD: @switch) match {
+        case Diagonal =>
+          curI -= 1
+          curJ -= 1
+          if (isMatch(query(curI), target(curJ))) this.matchOp else this.mismatchOp
```
nh13: I am assuming the `Diagonal` case occurs most frequently, so it should help to have that branch first.

tfenne: I'm not sure it matters since it's a `@switch` with a jump table, but it can't hurt either.

PR #473, opened 2019-03-29
Tool to sort fastq files.

tfenne:


src/main/scala/com/fulcrumgenomics/fastq/SortFastq.scala
```diff
@@ -0,0 +1,89 @@
...
+/** Companion object for the SortFastq tool. */
+object SortFastq {
+  private class FastqCodec extends Sorter.Codec[FastqRecord] {
+    /** Encodes the record directly to fastq. */
+    override def encode(rec: FastqRecord): Array[Byte] = rec.toString.getBytes
+
+    /** Decodes the record from the fastq string. */
+    override def decode(bs: Array[Byte], start: Int, length: Int): FastqRecord = {
+      val lines = new String(bs, start, length).lines
+      val name       = lines.next().substring(1)
```
tfenne: I know this is a little suspect in that it'll parse the entire header line into "name", including any read number and comment string, but since all we do is spit them back out another file it doesn't actually hurt.

PR #483, opened 2019-04-12
Remove silly ERROR logging message due to Intel compression additions

nh13:
And the JAR gets bigger :/


nh13:
The message was:
```
ERROR StatusLogger Log4j2 could not find a logging implementation. Please add log4j-core to the classpath. Using SimpleLogger to log to the console...
```

nh13:
Without a config file we get this message:
```
ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console.
```

PR #489, opened 2019-06-13
Bug fix: output consensus UMI bases (RX) was wrongly computed

nh13:
@tfenne could you take a close look at this.  I've read through `GroupReadsByUmi` and convinced myself of this, and also tested it on a BAM to make sure it does the right thing.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
@@ -296,8 +296,8 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
         (duplexR1, duplexR2) match {
           case (Some(r1), Some(r2)) =>
             Seq(
-              createSamRecord(r1, FirstOfPair, toUmiBases(duplexR1Sources)),
-              createSamRecord(r2, SecondOfPair, toUmiBases(duplexR2Sources))
+              createSamRecord(r1, FirstOfPair, toUmiBasesForConsensusUmiCalling(duplexR1Sources, firstOfPair=true)),
+              createSamRecord(r2, SecondOfPair, toUmiBasesForConsensusUmiCalling(duplexR2Sources, firstOfPair=false))
             )
           case _                    =>
             // NB: some reads may have been rejected already in filterToMostCommonAlignment, so just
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -263,7 +263,12 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
     // The source read may not have the RX tag, and if so, we ignore them through
     // the use of .flatMap and .get, with the latter returning an `Some(bases)` if the
     // UMI bases are present, `None` otherwise.
-    reads.flatMap(_.sam).flatMap(_.get[String](ConsensusTags.UmiBases))
+    reads.flatMap(_.sam).flatMap { rec =>
+      rec.get[String](ConsensusTags.UmiBases).map { umi =>
+        // when combining umi bases across multiple strands, we will need to reverse one or the other.
```
nh13: delete this comment

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -263,7 +263,12 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+        if (rec.firstOfPair) umi else umi.split('-').reverse.mkString("-")
```
tfenne: Having looked through things a bunch I think this makes sense.  However, I was initially worried because I thought this method had much broader usage.  I think the code fix is right, but I also think we should:

1. Rename this method to `toUmiBasesForConsensusUmiCalling()`
2. Add some explanation of how and why this method works

Because if we tried to use this elsewhere to just grab and examine the UMIs with the expectation that they'd conform to the usual duplex layout we'd make a big mess.

src/test/scala/com/fulcrumgenomics/umi/DuplexConsensusCallerTest.scala
```diff
@@ -289,13 +289,8 @@ class DuplexConsensusCallerTest extends UnitSpec {
...
       val mi = rec[String](MI)
-      // first of pair ABs and second of pair BAs
-      if ((rec.firstOfPair && mi.endsWith("/A")) || (rec.secondOfPair && mi.endsWith("/B"))) {
-        rec(RX) = "AAT-CCG"
-      }
-      else {
-        rec(RX) = "CCG-AAT"
-      }
+      // all "/A" have the same UMI bases, while the "/B" bases have them swapped
+      if (mi.endsWith("/A")) rec(RX) = "AAT-CCG" else rec(RX) = "CCG-AAT"
```
tfenne: Does this test still pass?  I see below that we check the RX tag on the output reads and expect them to differ.  Wouldn't we expect them to be the same after this change?

PR #492, opened 2019-06-21
Bug fix:  post-labeling error rate was not used

nh13:
It should be used when calculating the final base quality: https://github.com/fulcrumgenomics/fgbio/wiki/Calling-Consensus-Reads#adjusting-the-input-base-qualities


src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -148,7 +148,7 @@ class ConsensusCaller(errorRatePreLabeling:  PhredScore,
           //               + Pr(pre-UMI error AND error in consensus, that do not give us the correct bases)
           // The last term tries to capture the case where a pre-UMI error modifies the base (ex. A->C) but a sequencing
           // error calls it the correct base (C->A).  Only 2/3 times will the two errors result in the incorrect base.
-          val p = LogProbability.probabilityOfErrorTwoTrials(LnErrorRatePreLabeling, pConsensusError)
+          val p = LogProbability.probabilityOfErrorTwoTrials(LnErrorRatePostLabeling, pConsensusError)
```
nh13: Note: `LnErrorRatePostLabeling` was not being used anywhere in the file, this is how I found it.

nh13:
Not actually a bug, it's just that `this.errorRatePostLabeling` is used in pre-computing `phredToAdjustedLogProbError`.  We could update it there.

PR #493, opened 2019-06-21
Speeding up the duplex consensus caller

nh13:
This can speed up the consensus caller by 1.5-2.15x for 2 and 4 threads respectively.  Speeds up by 1.15x for a single thread.  If there is _very_ high per-molecule coverage, this can speed up by 20x or more (single threaded).

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
@@ -26,62 +26,120 @@ package com.fulcrumgenomics.umi
 
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.bam.api.SamRecord
+import com.fulcrumgenomics.commons.async.AsyncIterator
+import com.fulcrumgenomics.commons.util.LazyLogging
+import com.fulcrumgenomics.umi.UmiConsensusCaller.SimpleRead
 import com.fulcrumgenomics.util.ProgressLogger
 
-import scala.collection.mutable
+import scala.concurrent.forkjoin.ForkJoinPool
 
 /**
-  * An iterator that consumes from an incoming iterator of SAMRecords and generates consensus
-  * read SAMRecords using the supplied caller.
+  * An iterator that consumes from an incoming iterator of [[SamRecord]]s and generates consensus
+  * read [[SamRecord]]s using the supplied consensus caller.
   *
-  * @param sourceIterator an iterator of SAMRecords
-  * @param caller   the consensus caller to use to call consensus reads
+  * @param sourceIterator the iterator over input [[SamRecord]]s.
+  * @param caller the consensus caller to use to call consensus reads
   * @param progress an optional progress logger to which to log progress in input reads
+  * @param threads the number of threads to use.
+  * @param maxRecordsInRam the approximate maximum number of input records to store in RAM across multiple threads.
   */
-class ConsensusCallingIterator(sourceIterator: Iterator[SamRecord],
-                               val caller: UmiConsensusCaller[_],
-                               val progress: Option[ProgressLogger] = None
-                              ) extends Iterator[SamRecord] {
+class ConsensusCallingIterator[ConsensusRead <: SimpleRead](sourceIterator: Iterator[SamRecord],
+                               caller: UmiConsensusCaller[ConsensusRead],
+                               progress: Option[ProgressLogger] = None,
+                               threads: Int = 1,
+                               maxRecordsInRam: Int = 128000)
+  extends Iterator[SamRecord] with LazyLogging {
 
-  private val input = sourceIterator.bufferBetter
-  private val outputQueue: mutable.Queue[SamRecord] = mutable.Queue[SamRecord]()
-
-    /** True if there are more consensus reads, false otherwise. */
-  def hasNext(): Boolean = this.outputQueue.nonEmpty || (this.input.nonEmpty && advance())
-
-  /** Returns the next consensus read. */
-  def next(): SamRecord = {
-    if (!this.hasNext()) throw new NoSuchElementException("Calling next() when hasNext() is false.")
-    this.outputQueue.dequeue()
+  private val progressIterator = progress match {
+    case Some(p) => sourceIterator.map { r => p.record(r); r }
+    case None    => sourceIterator
   }
 
-  /**
-    * Consumes the next group of records from the input iterator, based on molecule id
-    * and returns them as a Seq.
-    */
-  private def nextGroupOfRecords(): Seq[SamRecord] = {
-    if (this.input.isEmpty) {
-      Nil
+  protected val iterator: Iterator[SamRecord] = {
+    if (threads <= 1) {
+      val groupingIterator = new SamRecordGroupedIterator(progressIterator, caller.sourceMoleculeId)
+      groupingIterator.flatMap(caller.consensusReadsFromSamRecords)
     }
     else {
-      val idToMatch = this.caller.sourceMoleculeId(this.input.head)
-      this.input.takeWhile(this.caller.sourceMoleculeId(_) == idToMatch).toSeq
+      val halfMaxRecords   = maxRecordsInRam / 2
+      val groupingIterator = new SamRecordGroupedIterator(new AsyncIterator(progressIterator, Some(halfMaxRecords)).start(), caller.sourceMoleculeId)
+      val pool             = new ForkJoinPool(threads - 1, ForkJoinPool.defaultForkJoinWorkerThreadFactory, null, true)
+      val bufferedIter     = groupingIterator.bufferBetter
+      val callers          = new IterableThreadLocal[UmiConsensusCaller[ConsensusRead]](() => caller.emptyClone())
+
+      // Create an infinite length iterator that will pull input records (up to the half the maximum number in memory)
+      // to be consensus called.  Each chunk of records will then be called in parallel.  If no more input records are
+      // available, an empty sequence will be returned by the iterator.
+      val iterator = Iterator.continually {
+        // Pull input reads to be consensus called.  We have to be careful to pull all the raw reads for each source
+        // molecule, so we pull from a [[SamRecordGroupedIterator]] above.
+        var total = 0L
+        bufferedIter
+          .takeWhile { chunk => if (halfMaxRecords <= total) false else {total += chunk.length; true } }
+          .toSeq
+          .parWith(pool)
+          .flatMap { records =>
+            val caller = callers.get()
+            caller.synchronized { caller.consensusReadsFromSamRecords(records) }
+          }
+          .seq
+      }
+
+      // Take from the iterator until an empty sequence of records is returned, which indicates that there were no more
+      // input records on which to call consensus reads.  When this happens, update the consensus calling statistics
+      // and stop iteration.
+      iterator.takeWhile { records =>
+        if (records.nonEmpty) true else {
+          // add the statistics to the original caller since there are no more reads
+          require(bufferedIter.isEmpty, "Bug: input is not empty")
+          callers.foreach(caller.addStatistics)
+          false
+        }
+      }.flatten
+
     }
   }
+  override def hasNext: Boolean = this.iterator.hasNext
+  override def next(): SamRecord = this.iterator.next
+}
 
-  /** Consumes input records until one or more consensus reads can be created, or no more input records are available.
-    * Returns true if a consensus read was created and enqueued, false otherwise. */
-  @annotation.tailrec
+// TODO: migrate to the commons version of this class after the next commons release
+// TODO: https://github.com/fulcrumgenomics/commons/pull/51
+private class IterableThreadLocal[A](factory: () => A) extends ThreadLocal[A] with Iterable[A] {
+  private val all = new java.util.concurrent.ConcurrentLinkedQueue[A]()
+
+  override def initialValue(): A = {
+    val a = factory()
+    all.add(a)
+    a
+  }
+
+  /** Care should be taken accessing the iterator since objects may be in use by other threads. */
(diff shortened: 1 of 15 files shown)


src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -26,62 +26,95 @@ package com.fulcrumgenomics.umi
...
+    }
+  }
+  override def hasNext: Boolean = this.iterator.hasNext
+  override def next(): SamRecord = this.iterator.next
+}
 
-  private val input = sourceIterator.bufferBetter
-  private val outputQueue: mutable.Queue[SamRecord] = mutable.Queue[SamRecord]()
+/** Groups consecutive records based on a method to group records. */
+private class SamRecordGroupedIterator[Key](sourceIterator: Iterator[SamRecord],
```
tfenne: Ugh, I didn't realize `Iterator.groupBy` doesn't exist :(  I wonder if we'd be better served adding a `groupBy` to `BetterBufferedIterator` and making this more generic?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -117,23 +117,55 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
   /**
-    * Returns the MI tag minus the trailing suffix that identifies /A vs /B
+    * Returns the MI tag **with** the trailing suffix that identifies /A vs /B
     */
-  override protected[umi] def sourceMoleculeId(rec: SamRecord): String = {
-    rec.get[String](ConsensusTags.MolecularId) match {
-      case None =>
+  private def sourceMoleculeAndStrandId(rec: SamRecord): String = {
+    // Optimization: speed up retrieving this tag by storing it in the transient attributes
+    rec.transientAttrs.getOrElse[String](ConsensusTags.MolecularId, {
```
tfenne: I don't love `getOrElse` usage where the `else` is side-effecting.  I think it's confusing, and also multi-line default values are hard to read.  Is this slower when structured as a match?  if you really must do it this way can we add a `getOrElseUpdate` to the `TransientAttrs`?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -117,23 +117,55 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
-        require(index > 0, s"Read ${rec.name}'s $ConsensusTags tag doesn't look like a duplex id: $mi")
-        mi.substring(0, index)
-    }
+      )
+      rec.transientAttrs(ConsensusTags.MolecularId) = mi
+      mi
+    })
+  }
+
+  private val MolecularIdNoTrailingSuffix: String = "__" + ConsensusTags.MolecularId + "__"
```
tfenne: This could use a comment since I had to stare at it to figure out what it was!

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -191,6 +223,19 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
     }
   }
 
+  /** Split records into those that should make a single-end consensus read, first of pair consensus read,
+    * and second of pair consensus read, respectively.  For [[DuplexConsensusCaller]], no fragment reads should be
+    * given to this method, so none will be returned.
+    */
+  override protected def subGroupRecords(records: Seq[SamRecord]): (Seq[SamRecord], Seq[SamRecord], Seq[SamRecord]) = {
```
tfenne: Can you comment that this is presumably overridden this way for performance reasons?  Otherwise it would seem much simpler to just let the super class do this.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -191,6 +223,19 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+    // NB: the input records should not have fragments
+    val (firstOfPair, secondOfPair) = records.partition { r =>
+      assert(r.paired, "Fragment reads should not be given to subGroupRecords in DuplexConsensusCaller.")
```
tfenne: require instead of assert?

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -191,6 +223,19 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
+      r.firstOfPair
+    }
+    (Seq.empty, firstOfPair, secondOfPair)
```
tfenne: If you really want every last iota of performance you might cache `val NoSamRecords = Seq.empty[SamRecord]` in the object and reference it here.  `Seq.empty` constructs a new one each time.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -238,8 +283,8 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
         // Call the duplex reads
         val duplexR1Sources = filteredAbR1s ++ filteredBaR2s
         val duplexR2Sources = filteredAbR2s ++ filteredBaR1s
-        val duplexR1 = duplexConsensus(abR1Consensus, baR2Consensus, duplexR1Sources)
-        val duplexR2 = duplexConsensus(abR2Consensus, baR1Consensus, duplexR2Sources)
+        val duplexR1 = duplexConsensus(abR1Consensus, baR2Consensus, duplexR1Sources.toArray[SourceRead])
```
tfenne: The method being called still takes `Seq[SourceRead`, so this is confusing.  I think either the the signature on `duplexConsensus` needs to change, or this should remain a `Seq`, and if you want arrays for performance, you can do that internally to `duplexConsensus`.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -305,10 +350,19 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
...
           val (base, qual) = if (aBase == NoCall || bBase == NoCall || rawQual == PhredScore.MinValue) (NoCall, NoCallQual) else (rawBase, rawQual)
 
-          bases(i)  = base
-          quals(i)  = qual
-
-          errors(i) = min(sourceReads.count(s => s.length > i && isError(s.bases(i), rawBase)), Short.MaxValue).toShort
+          bases(i) = base
+          quals(i) = qual
+
+          // NB: optimized based on profiling; was previously:
```
tfenne: This makes me sad.  Is it really faster-enough to matter?

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -159,7 +158,7 @@ trait UmiConsensusCaller[C <: SimpleRead] {
 
   // vars to track how many reads meet various fates
   private var _totalReads: Long = 0
-  private val filteredReads = new SimpleCounter[String]()
+  private[umi] val _filteredReads = new SimpleCounter[String]()
```
tfenne: Why now `private[umi]`?  It looks like all accesses are still within `UmiConsensusCaller`.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -313,36 +320,33 @@ trait UmiConsensusCaller[C <: SimpleRead] {
     * NOTE: filtered out reads are sent to the [[rejectRecords]] method and do not need further handling
     */
   protected[umi] def filterToMostCommonAlignment(recs: Seq[SourceRead]): Seq[SourceRead] = {
```
tfenne: I find this new implementation impossible to follow.  I've literally been staring at it side-by-side in IntelliJ with the old implementation for ~10 minutes and I still struggle to follow it.  Two thoughts:

1. Unless this makes a very significant different I would just revert to the old method that I think was easier to follow.
2. If it does make a significant difference, then I think we need to find a clearer implementation.  I wonder if something like the following would work:

```scala
case class AlignmentGroup(cigar: Cigar, reads: mutable.Buffer[SourceRead])

protected[umi] def filterToMostCommonAlignment(recs: Seq[SourceRead]): Seq[SourceRead] = {
  val groups = new ArrayBuffer[AlignmentGroup]
  recs.sortBy(r => -r.length).foreach { rec =>
    val simpleCigar = simplifyCigar(rec.cigar)
    var found = false
    groups.foreach { g => if (simpleCigar.isPrefixOf(g.cigar) { g.reads += rec; found = true } }
    if (!found) {
      val newGroup = AlignmentGroup(simpleCigar, new ArrayBuffer[SourceRead](recs.size))
      newGroup += rec
      groups += newGroup
    }
  }

  if (groups.isEmpty) {
    Seq.empty
  }
  else {
    val sorted  = groups.sortBy(g => - g.size)
    val keepers = sorted.head
    val rejects = recs.filter(r => !keepers.contains(r))
    rejectRecords(rejects.flatMap(_.sam), FilterMinorityAlignment)
   keepers
  }
}
```

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -200,10 +208,12 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
           }
         }
 
+        val contributions = builder.contributions // NB: cache this value, as it is re-computed each time
```
tfenne: You know, you really don't need to prefix your comments with "NB" - comments _are_ notes to our future selves 😛

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -212,7 +222,7 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
         consensusQuals(positionInRead) = qual
 
         // Generate the values for depth and count of errors
-        val depth  = builder.contributions
+        val depth  = contributions
```
tfenne: Maybe just move this line up to where you define `contributions` now, and use `depth` in the other spot too?

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -242,7 +252,7 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
   /** If a reject writer was provided, emit the reads to that writer. */
   override protected def rejectRecords(recs: Traversable[SamRecord], reason: String): Unit = {
     super.rejectRecords(recs, reason)
-    this.rejects.foreach(rej => rej ++= recs)
+    this.rejects.foreach(rej => rej ++= recs) // NB: this may not be thread-safe
```
tfenne: Nowhere does this class say that it's threadsafe, so this comment is a little odd.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
nh13: I'll change it to use `match` statements.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
nh13: ctrl-f for the win?  I'll add a comment too.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
nh13: Yup.

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
nh13: Some of us have style Tim

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
nh13: This comment was a note to our future selves that if/when we want to parallelize `CallMolecularConsensusReads`, this will need to be thread-safe.  I'll remove it.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
nh13: The problem with the previous and your implementation is the `keepers.contains` method can be really really slow if we have many raw reads with the same cigar (think 20 cigar groups, each with 1000 reads).  This is the point of optimizing, and it really does make a difference. I'll try to clean it up.

tfenne: I see.  I'm curious if you tried either a) replacing the `recs.filter` with a `recs.diff(keepers)` or be creating a `val keepSet = keepers.toSet` and then calling contains on that?  You'd still pay the cost of the conversion to a set, but then the lookup time would be constant.

nh13:
A TODO includes a reference to https://github.com/fulcrumgenomics/commons/pull/51

PR #497, opened 2019-06-26
UmiConsensusCaller now merges Platform tags case-insensitively

gpratt:
Due to historical reasons, some BAM files that UmiConsensusCaller will create consensus reads from may have multiple read groups with platform (or `PL` tags) that are inconsistently uppercase and lower case.  

This PR now merges the `PL` tag in a case insensitive manner, reporting only the final, merged uppercase `PL` tag.

I'm also happy to add an error message if there is more than one `PL` tag still in a `UmiConsensusCaller` output BAM, but feel that may be more appropriately handled by a downstream `ValidateSamFile` call.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
@@ -107,7 +107,7 @@ object UmiConsensusCaller {
     rg.setDescription     (collapse(_.getDescription))
     rg.setLibrary         (collapse(_.getLibrary))
     rg.setSample          (collapse(_.getSample))
-    rg.setPlatform        (collapse(_.getPlatform))
+    rg.setPlatform        (collapse(x => Option(x.getPlatform).map(_.toUpperCase).orNull)) // NB: this to ensure that platforms are all upper-case; not all tools are modern or well-behaved
     rg.setPlatformUnit    (collapse(_.getPlatformUnit))
     rg.setSequencingCenter(collapse(_.getSequencingCenter))
 
--- a/src/test/scala/com/fulcrumgenomics/umi/UmiConsensusCallerTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/umi/UmiConsensusCallerTest.scala
@@ -27,6 +27,7 @@ package com.fulcrumgenomics.umi
 import com.fulcrumgenomics.bam.api.SamOrder
 import com.fulcrumgenomics.testing.{SamBuilder, UnitSpec}
 import htsjdk.samtools.SAMFileHeader.GroupOrder
+import htsjdk.samtools.SAMReadGroupRecord
 import org.scalatest.OptionValues
 
 class UmiConsensusCallerTest extends UnitSpec with OptionValues {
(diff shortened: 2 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -107,7 +111,7 @@ object UmiConsensusCaller {
     rg.setDescription     (collapse(_.getDescription))
     rg.setLibrary         (collapse(_.getLibrary))
     rg.setSample          (collapse(_.getSample))
-    rg.setPlatform        (collapse(_.getPlatform))
+    rg.setPlatform        (collapse(_.getPlatform, toUpper = true))
```
nh13: I think you can get rid of your modifications above and just inline this (since you only do it once).  My philosophy only generalize if you need to do it 2+ times.

```suggestion
    rg.setPlatform        (collapse(_.getPlatform.toUpper))  // NB: this to ensure that platforms are all upper-case; not all tools are modern or well-behaved
```

Feel free to remove my snarky comment.

gpratt: This approach fails if any platform is null. The only reason I just caught that is because its an undocumented  part of the added unit test.  

There are actually 3 read groups that are created.  The original RG as part of the `SamBuilder` class, and the two for testing purposes.  We already filter out null values as part of the `collapse` function.  

I tried this as a solution:

```scala
rg.setPlatform        (collapse(_.getPlatform.filter(_ != null).toUpperCase)) // NB: this to ensure that platforms are all upper-case; not all tools are modern or well-behaved
 ```

but I that also failed the unit test.

I can keep on plugging away if we really want to don't want parameterize the call.

Comment (nh13) [reply src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala]: `   rg.setPlatform        (collapse(x => Option(x.getPlatform).map(_.toUpperCase).orNull))`

**Comment by gpratt** [comment]
Back to you @nh13

File: src/test/scala/com/fulcrumgenomics/umi/UmiConsensusCallerTest.scala [hunk 2]
```diff
@@ -57,4 +58,23 @@ class UmiConsensusCallerTest extends UnitSpec with OptionValues {
     warning shouldBe None
     error.value.indexOf("foo.bam") should be >= 0
   }
+
+  it should "header should output correct platform id" in {
```
nh13:
```suggestion
  "UmiConsensusCaller.outputHeader" should "output the correct platform id" in {
```

1. The `it` should be more specific, as it's testing a specific method, not the end-to-end tool.
2. `it should` will write "UmiConsensusCaller should ...." so no need to write "header should..." as then it would write "UmiConsensusCaller should header should output..." which makes no sense!

src/test/scala/com/fulcrumgenomics/umi/UmiConsensusCallerTest.scala
```diff
@@ -57,4 +58,23 @@ class UmiConsensusCallerTest extends UnitSpec with OptionValues {
...
+      return rg
+    }
+
+    val builder = new SamBuilder(sort=Some(SamOrder.TemplateCoordinate))
+
+    builder.header.addReadGroup(buildRg(id = "B", sample ="Sample", platform = "ILLUMINA"))
+    builder.header.addReadGroup(buildRg(id = "C", sample ="Sample2", platform = "illumina"))
+    val outHeader = UmiConsensusCaller.outputHeader(in=builder.header, readGroupId = "A")
+    println("foo")
+    println(outHeader.getReadGroup("A").getPlatform)
```
nh13: remove the `println`s

gpratt:
Thanks for the checks! They have been fixed.

nh13:
I am closing in favor of https://github.com/fulcrumgenomics/fgbio/pull/499

PR #501, opened 2019-07-08
Scala VCF API

tfenne:
A complete wrapper around the HTSJDK VCF classes.  The goal here is to completely hide the HTSJDK implementations - the scala versions are standalone classes and there is a single `VcfConversions` class that translates back and forth between the HTSJDK objects and the scala versions when reading/writing VCFs and BCFs.

I'm slightly worried about the performance of this approach, but I think it's likely we won't notice for smaller VCFs, and if we end up dealing with multi-TB VCFs using this API we can revisit and either do some performance tuning _or_ do some of our own parsing at least of rows in the VCF.

For anyone else looking, @nh13 and I had a lot of discussion about how to facilitate easy access to common INFO and sample fields, given that all INFO and sample fields are strictly optional, but most of the time you're working with VCFs that have a raft of common values.  We've decided for now to leave this up to the user, using the provided `apply[A](key)` and `get[A](key)` methods, and that once we have some road time with the API we will loop back on this front.

A few thoughts on where things are at:

1. Do we like the naming on all the classes, types, etc?
2. Do we like that in `Genotype` the numeric alleles (i.e. `0/1`) are second class citizens to the `Allele` objects which include the bases/IDs.
3. Is there any functionality obviously missing?
4. Do any of us understand the ALT, PEDIGREE and SAMPLE header lines well enough to make classes for those and handle translation from HTSJDK headers?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
+++ b/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
@@ -128,7 +128,7 @@ class FgBioMain extends LazyLogging {
     }
 
     val startTime = System.currentTimeMillis()
-    val exit      = Sopt.parseCommandAndSubCommand[FgBioCommonArgs,FgBioTool](name, args, Sopt.find[FgBioTool](packageList)) match {
+    val exit      = Sopt.parseCommandAndSubCommand[FgBioCommonArgs,FgBioTool](name, args.toIndexedSeq, Sopt.find[FgBioTool](packageList)) match {
       case Sopt.Failure(usage) =>
         System.err.print(usage())
         1
@@ -147,7 +147,7 @@ class FgBioMain extends LazyLogging {
             val banner = "#" * ex.message.map(_.length).getOrElse(80)
             logger.fatal(banner)
             logger.fatal("Execution failed!")
-            ex.message.foreach(msg => msg.lines.foreach(logger.fatal))
+            ex.message.foreach(msg => msg.linesIterator.foreach(logger.fatal))
             logger.fatal(banner)
             printEndingLines(startTime, name, false)
             ex.exit
--- a/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
+++ b/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
@@ -36,6 +36,7 @@ import com.fulcrumgenomics.commons.util.{LazyLogging, LogLevel, Logger}
 import com.fulcrumgenomics.sopt.{Sopt, arg}
 import com.fulcrumgenomics.sopt.cmdline.CommandLineProgramParserStrings
 import com.fulcrumgenomics.util.Io
+import com.fulcrumgenomics.vcf.api.VcfWriter
 import com.intel.gkl.compression.{IntelDeflaterFactory, IntelInflaterFactory}
 import htsjdk.samtools.{SAMFileWriterFactory, ValidationStringency}
 import htsjdk.samtools.util.{BlockCompressedOutputStream, BlockGunzipper, IOUtil, SnappyLoader}
@@ -98,6 +99,7 @@ class FgBioCommonArgs
 
   SamSource.DefaultUseAsyncIo = asyncIo
   SamWriter.DefaultUseAsyncIo = asyncIo
+  VcfWriter.DefaultUseAsyncIo = asyncIo
 
   SamWriter.DefaultCompressionLevel = compression
   BlockCompressedOutputStream.setDefaultCompressionLevel(compression)
--- a/src/main/scala/com/fulcrumgenomics/testing/VcfBuilder.scala
+++ b/src/main/scala/com/fulcrumgenomics/testing/VcfBuilder.scala
@@ -46,7 +46,7 @@ object VcfBuilder {
       contigs = contigs,
       infos   = Seq(
         VcfInfoHeader(id="AC",  count=VcfCount.OnePerAltAllele, kind=VcfFieldType.Integer, description="Alternate allele counts in genotypes."),
-        VcfInfoHeader(id="DP",  count=VcfCount.Fixed(1),        kind=VcfFieldType.Integer, description="Depth across all samples."),
+        VcfInfoHeader(id="DP",  count=VcfCount.Fixed(1),        kind=VcfFieldType.Integer, description="Depth across all samples.")
       ),
       formats = Seq(
         VcfFormatHeader(id="GT", count=VcfCount.Fixed(1),       kind=VcfFieldType.String, description="Genotype string."),
--- a/src/main/scala/com/fulcrumgenomics/testing/VcfBuilder.scala
+++ b/src/main/scala/com/fulcrumgenomics/testing/VcfBuilder.scala
@@ -1,0 +1,232 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2019 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.testing
+
+import java.nio.file.Files
+
+import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.testing.VcfBuilder.Gt
+import com.fulcrumgenomics.vcf.api.Allele.NoCallAllele
+import com.fulcrumgenomics.vcf.api._
+
+import scala.collection.immutable.ListMap
+import scala.collection.mutable
+import scala.util.Try
+
+object VcfBuilder {
+  /** The default header that is used when making a new VcfBuilder. */
+  val DefaultHeader: VcfHeader = {
+    val contigs = new SamBuilder().dict.getSequences
+      .map(s => VcfContigHeader(s.getSequenceIndex, s.getSequenceName, length=Some(s.getSequenceLength), assembly=Option(s.getAssembly)))
+      .toIndexedSeq
+
+    VcfHeader(
+      contigs = contigs,
+      infos   = Seq(
+        VcfInfoHeader(id="AC",  count=VcfCount.OnePerAltAllele, kind=VcfFieldType.Integer, description="Alternate allele counts in genotypes."),
+        VcfInfoHeader(id="DP",  count=VcfCount.Fixed(1),        kind=VcfFieldType.Integer, description="Depth across all samples."),
+      ),
+      formats = Seq(
+        VcfFormatHeader(id="GT", count=VcfCount.Fixed(1),       kind=VcfFieldType.String, description="Genotype string."),
+        VcfFormatHeader(id="AD", count=VcfCount.OnePerAllele,   kind=VcfFieldType.Integer, description="Depth per allele."),
+        VcfFormatHeader(id="GQ", count=VcfCount.Fixed(1),       kind=VcfFieldType.Integer, description="Genotype quality."),
+        VcfFormatHeader(id="PL", count=VcfCount.OnePerGenotype, kind=VcfFieldType.Integer, description="Phred scaled genotype likelihoods.")
+      ),
+      filters = Seq(
+        VcfFilterHeader(id="LowQD", description="Low Quality/Depth value"),
+        VcfFilterHeader(id="LowAB", description="Low/poor allele balance.")
+      ),
+      others  = Seq(
+        VcfGeneralHeader(headerType="ALT",      id="NON_REF", data=Map("Description" -> "Represents any non-reference allele."))
+      ),
+      samples = IndexedSeq()
+    )
+  }
+
+  /**
+    * Convenience case class used to build up genotypes.
+    *
+    * @param sample the name of the sample being genotyped
+    * @param gt a genotype either in VCF format (e.g. 0/1) or using actual alleles (e.g. A/T)
+    * @param dp optional depth (DP)
+    * @param ad optional set of allele depths (AD)
+    * @param pl optional set of phred-scale genotype likelihoods (PL)
+    * @param attrs extended attributes (must be in FORMAT headers)
+    */
+  case class Gt(sample: String,
+                gt: String,
+                dp: Int = -1,
+                ad: Seq[Int] = Seq.empty,
+                pl: Seq[Int] = Seq.empty,
+                attrs: Map[String,Any] = Map.empty
+               ) {
+
+    def allAttrs: Map[String, Any] = {
+      attrs ++
+        (if (dp < 0) Map.empty else Map("DP" -> dp)) ++
+        (if (ad.isEmpty) Map.empty else Map("AD" -> ad.toIndexedSeq)) ++
+        (if (pl.isEmpty) Map.empty else Map("DP" -> pl.toIndexedSeq))
+    }
+  }
+
+  /** Constructs a VcfBuilder using the supplied header. */
+  def apply(header: VcfHeader): VcfBuilder = new VcfBuilder(header)
+
+  /** Constructs a VcfBuilder using the [[DefaultHeader]] and the set of samples provided. */
+  def apply(samples: Seq[String]): VcfBuilder = {
+    require(samples.distinct.size == samples.size, s"${samples.mkString(",")} contains duplicate sample names.")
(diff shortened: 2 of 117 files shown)


src/main/scala/com/fulcrumgenomics/vcf/api/Allele.scala
```diff
@@ -0,0 +1,60 @@
...
+import com.fulcrumgenomics.FgBioDef._
+import htsjdk.samtools.util.SequenceUtil
+
+sealed trait Allele { }
+
+object Allele {
+  // TODO: have optimized lookup for base A/C/G/T and maybe dinuc alleles?
+  def apply(value: String): Allele = value match {
+    case "." => NoCallAllele
+    case "*" => SpannedAllele
```
nh13: use `SpannedAllele.toString` instead of `"*"`?

src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala
```diff
@@ -0,0 +1,183 @@
...
+  def get[A](key: String): Option[A] = attributes.get(key).asInstanceOf[Option[A]]
+
+  /** Retrieves an optional value from the INFO map.  Will return `default` if the key does not exist. */
+  def getOrElse[A](key: String, default: => A): Option[A] = attributes.getOrElse(key, default).asInstanceOf[Option[A]]
+
+  /** Object that provides convenient accessors to common, spec-defined, attributes from the INFO field.
+    * Methods on this object return the expected type directly and will throw exceptions if the keys
+    * do not exist in the info map.
+    */
+  final object INFO {
```
tfenne: @nh13 So this seems to work.  I'm having a hard time coming up with names for the pair of objects here and again in `Genotype`, so suggestions are welcome.

I renamed `info` to `attributes` in the constructor for a couple of reasons:
1) So I could use `info` here
2) So it was less obvious when using the object; it does need to be exposed though otherwise it makes calling the copy() methods really awkward

src/main/scala/com/fulcrumgenomics/vcf/api/Allele.scala
```diff
@@ -0,0 +1,120 @@
...
+    override def hashCode(): Int = this.bases.toUpperCase.hashCode
+  }
+
+  /** Class for symbolic alleles. */
+  case class SymbolicAllele private (id: String) extends Allele {
+    override def value: String = id
+  }
+
+  /** Class for break-end alleles. */
+  case class BreakendAllele private (override val value: String) extends Allele // TODO: what does this need to have
```
tfenne: I looked at the VCF spec and breakend stuff is super-complicated.  For now it's just pushing it into a separate type but not trying to parse the alleles at all.

src/main/scala/com/fulcrumgenomics/vcf/api/Genotype.scala
```diff
@@ -0,0 +1,91 @@
...
+final case class Genotype(alleles: AlleleSet,
+                          sample: String,
+                          calls: IndexedSeq[Allele],
+                          phased: Boolean = false,
+                          attrs: Map[String, Any] = Variant.EmptyGtAttrs
+                         ) {
+  require(calls.nonEmpty, "Genotype must have ploidy of at least 1!.")
+
+  /** The indices of the calls within the AlleleSet. If any allele is no-called, that is returned as -1. */
+  val callIndices: IndexedSeq[Int] = calls.map(c => if (c == NoCallAllele) -1 else alleles.indexOf(c))
```
tfenne: Should this be private?

src/main/scala/com/fulcrumgenomics/vcf/api/Genotype.scala
```diff
@@ -0,0 +1,91 @@
...
+  /** True if none of [[calls]] are no-call alleles. */
+  def isFullyCalled: Boolean = !calls.contains(Allele.NoCallAllele)
+
+  /** True if all [[calls]] are the reference allele. */
+  def isHomRef: Boolean = calls.forall(_ == alleles.ref)
+
+  /** True if all [[calls]] are the same non-reference allele. */
+  def isHomVar: Boolean = isFullyCalled && calls(0) != alleles.ref && calls.forall(_ == calls(0))
+
+  /** True if the genotype contains at least two called alleles that are different from one another. */
```
tfenne: I _think_ this definition makes most sense given ploidy can be any value, but take a look at the tests and see if you agree.  Same for isHetNonRef.

src/main/scala/com/fulcrumgenomics/vcf/api/Genotype.scala
```diff
@@ -0,0 +1,91 @@
...
+  def apply[A](key: String): A = attrs(key).asInstanceOf[A]
+
+  /** Retrieves an optional value from the INFO map.  Will return [[None]] if the key does not exist. */
+  def get[A](key: String): Option[A] = attrs.get(key).asInstanceOf[Option[A]]
+
+  /** Retrieves an optional value from the INFO map.  Will return `default` if the key does not exist. */
+  def getOrElse[A](key: String, default: => A): Option[A] = attrs.getOrElse(key, default).asInstanceOf[Option[A]]
+
+  /** Yields a genotype string using bases, e.g. A/C, or CTTT|C. */
+  def gtWithBases: String = calls.map(_.toString).mkString(separator)
```
tfenne: Very much open to better names for this and `gtVcfStyle`

src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala
```diff
@@ -0,0 +1,88 @@
...
+
+object Variant {
+  /** Value used in VCF for values that are missing. */
+  val Missing: String = "."
+
+  /** The value stored for fields of type `Flag` when stored in a Map. */
+  val FlagValue: String = Missing
+
+  /** The type used in attribute maps (INFO, genotype) to store multi-valued attributed. */
+  type ArrayAttribute[A] = IndexedSeq[A]
```
tfenne: Should we pick a shorter name for this type?

src/main/scala/com/fulcrumgenomics/vcf/api/VcfHeader.scala
```diff
@@ -0,0 +1,175 @@
...
+  /** Character field type, represented internally as [[Char]]. */
+  case object Flag      extends VcfFieldType { override def parse(s: String): Any = "." }
+
+  override def values: immutable.IndexedSeq[VcfFieldType] = findValues
+}
+
+/** Trait representing an entry/line in a VCF header. */
+sealed trait VcfHeaderEntry {}
+
+// TODO add subclasses for ALT and PEDIGREE lines
```
tfenne: I think I could probably handle ALT.  But SAMPLE and PEDIGREE lines are so under-specified and I don't have experience with them, I'm not sure what those should look like.

src/main/scala/com/fulcrumgenomics/vcf/api/VcfSource.scala
```diff
@@ -0,0 +1,98 @@
...
+  override def iterator: VariantIterator = wrap(reader.iterator())
+
+  /**
+    * Returns an iterator over variants overlapping the specified genomic region.
+    *
+    * The returned iterator may be be closed by invoking `close()` on it, and will automatically close itself
+    * when exhausted.  Only a single iterator at a time is supported per [[VcfSource]], including iterators
+    * returned from [[iterator()]].
+    */
+  def query(chrom: String, start: Int, end: Int): Iterator[Variant] = wrap(reader.query(chrom, start, end))
```
tfenne: Do we also want a query method that takes in > 1 range and provides a single iterator of unique variants coming back?  This might be tricky to do as the functionality isn't in the lower level library _and_ I think we can only have one active query at a time.

src/main/scala/com/fulcrumgenomics/vcf/api/VcfConversions.scala
```diff
@@ -0,0 +1,330 @@
...
+        if (g.hasDP) buffer.append("DP" -> g.getDP)
+        if (g.hasGQ) buffer.append("GQ" -> g.getGQ)
+        if (g.hasPL) buffer.append("PL" -> g.getPL.toIndexedSeq)
+
+        g.getExtendedAttributes.keySet().foreach { key =>
+          val value = g.getExtendedAttribute(key)
+
+          header.format.get(key) match {
+            case Some(hd) => buffer.append(key -> toTypedValue(value, hd.kind, hd.count))
+            case None     => throw new IllegalStateException(s"Format field $key not described in header.")
```
tfenne: Here and again for the INFO map below, we could decide to do a best-effort conversion if the header line is missing.  I'm thinking we'd basically do the following:
1. Split the value on commas
2. Attempt to parse all values as Int, then Float then String, picking the type where all values are valid
3. If there's a single value, return the single value; if there are multiple return them as an ArrayAttribute

It wouldn't be perfect, but presumably better than blowing up?

src/main/scala/com/fulcrumgenomics/vcf/api/VcfHeader.scala
```diff
@@ -0,0 +1,175 @@
...
+
+/** Trait for the set of acceptable atomic types that can appear in VCF's INFO and genotype fields. */
+sealed trait VcfFieldType extends EnumEntry {
+  /** Must parse a String into a single value of the type given. */
+  def parse(s: String): Any
+}
+
+object VcfFieldType extends FgBioEnum[VcfFieldType] {
+  /** Integer field type, represented internally as [[Int]]. */
+  case object Integer   extends VcfFieldType { override def parse(s: String): Any = if (s == ".") -1 else s.toInt }
```
tfenne: This is horrifying.  Apparently it's ok to have an `Int` type field in VCF that takes multiple values and put in things like `1,2,.,4` - which makes it very hard to present typed values for array types.
Perhaps the thing to do here is replace the use of `IndexedSeq` with our own `MultiValuedAttr` type that wraps an array and also tracks missing values?  Ugh.

src/main/scala/com/fulcrumgenomics/vcf/api/VcfSource.scala
```diff
@@ -0,0 +1,114 @@
...
+    * @param path the path to a VCF, gzipped VCF or BCF file
+    * @return a VariantSource for reading from the path given
+    */
+  def apply(path: PathToVcf): VcfSource = {
+    val codec  = if (PathUtil.extensionOf(path).contains(".bcf")) {
+      new BCF2Codec
+    }
+    else {
+      val c = new VCFCodec
+      c.disableOnTheFlyModifications()
```
tfenne: This is the only way to stop HTSJDK borking up the headers to match standard GATK headers. So we end up having to use the tribble readers directly instead of `VCFFileReader`.  Sigh.

tfenne:
@nh13 Just a note that you're free to review here, but it'll probably need a little work after the 2.13 cross-build PR is merged.

src/main/scala/com/fulcrumgenomics/vcf/api/VcfConversions.scala
```diff
@@ -0,0 +1,339 @@
...
+    case (_, VcfFieldType.Flag, _       )              => Some(Variant.FlagValue)
+    case (_, _,                 Fixed(0))              => Some(Variant.FlagValue)
+    case (s: String, _,         Fixed(1))              => if (s == ".") None else Some(kind.parse(s))
+    case (s: String, _,         _       )              => val xs = s.split(","); if (xs.forall(_ == ".")) None else Some(ArrayAttr(xs.map(kind.parse)))
+    case (l: JavaList[String @unchecked], _, Fixed(1)) => if (l.get(0) == ".") None else Some(kind.parse(l.get(0)))
+    case (l: JavaList[String @unchecked], _, _)        => if (l.forall(_ == ".")) None else Some(ArrayAttr(l.map(kind.parse)))
+  }
+
+  /**
+    * Converts a scala [[Variant]] back into a [[VariantContext]].
```
nh13:
```suggestion
    * Converts a Scala [[Variant]] back into a [[VariantContext]].
```

src/main/scala/com/fulcrumgenomics/vcf/api/VcfConversions.scala
```diff
@@ -0,0 +1,339 @@
...
+    */
+  private def toJavaAttributeMap(attrs: Map[String,Any]): java.util.LinkedHashMap[String,Any] = {
+    val out = new util.LinkedHashMap[String,Any]()
+    attrs.foreach {
+      case (key, value: Seq[Any]) => out.put(key, value.toArray)
+      case (key, value)           => out.put(key, value)
+    }
+
+    out
+  }
```
nh13: thanks for writing these converters; boy are they ugly.

src/main/scala/com/fulcrumgenomics/vcf/api/VcfSource.scala
nh13: Nah, it shouldn't be hard to do it manually for now.

src/main/scala/com/fulcrumgenomics/vcf/api/VcfSource.scala
nh13: awesome

src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
```diff
@@ -98,6 +99,7 @@ class FgBioCommonArgs
 
   SamSource.DefaultUseAsyncIo = asyncIo
   SamWriter.DefaultUseAsyncIo = asyncIo
+  VcfWriter.DefaultUseAsyncIo = asyncIo
```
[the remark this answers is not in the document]
tfenne: Not that I could see.

src/main/scala/com/fulcrumgenomics/testing/VcfBuilder.scala
```diff
@@ -0,0 +1,232 @@
...
+  def apply(header: VcfHeader): VcfBuilder = new VcfBuilder(header)
+
+  /** Constructs a VcfBuilder using the [[DefaultHeader]] and the set of samples provided. */
+  def apply(samples: Seq[String]): VcfBuilder = {
+    require(samples.distinct.size == samples.size, s"${samples.mkString(",")} contains duplicate sample names.")
+    new VcfBuilder(DefaultHeader.copy(samples=samples.toIndexedSeq))
+  }
+}
+
+/** Class for building VCFs for testing purposes. */
```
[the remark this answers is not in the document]
tfenne: Hrm.  Do you think it should move out of `testing` and into `vcf.api` then?

src/main/scala/com/fulcrumgenomics/testing/VcfBuilder.scala
```diff
@@ -0,0 +1,232 @@
...
+  private type Location = (Int, Int)
+  private var _header: VcfHeader = initialHeader
+  private val variants: mutable.Map[Location, Variant] = mutable.TreeMap()
+
+  /** Provides access to the header that will be written to the VCF. */
+  def header: VcfHeader = _header
+
+  /** Adds one or more INFO headers to the VCF Header. */
+  def withInfoHeaders(headers: VcfInfoHeader*): this.type = {
+    this._header = this._header.copy(infos = this._header.infos ++ headers)
```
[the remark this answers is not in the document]
tfenne: Then you'll have duplicate INFO lines in your VCF.

src/main/scala/com/fulcrumgenomics/testing/VcfBuilder.scala
```diff
@@ -0,0 +1,232 @@
...
+    * The genotypes are specified using instance of the [[Gt]] helper class.  The genotype strings within the [[Gt]]
+    * objects may be either numeric like in a VCF (e.g. `0/1`) or use allele strings (e.g. `A/T`.)
+    *
+    * The variant must also be valid, e.g. not reference INFO, FILTER or FORMAT fields that are no in the header,
+    * and not have alleles in genotypes that are not in the set of alleles for the variant.
+    *
+    * If the header contains more samples than there are genotype give, simple diploid no-call genotypes will
+    * be inserted for the remaining samples.
+    */
+  def add(chrom: String = this.header.contigs.head.name,
```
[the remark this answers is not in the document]
tfenne: I feel like you have a lot of comments on `VcfBuilder` based on your suggestion that it should be for more than just testing.  Right now, this is intended solely for testing, based on it being in the `testing` package.  If we want to make it robust enough to handle all user inputs, and have it be for more testing we can do that, but I think that's a separate PR because that's a significant amount of work.

src/main/scala/com/fulcrumgenomics/vcf/api/ArrayAttr.scala
```diff
@@ -0,0 +1,82 @@
...
+  override def length: Int = values.length
+
+  /** Accesses the value at the specified index. May return a Missing value if no value is defined at the index.
+    * To avoid dealing with Missing values use [[isMissing(idx)]] or [[isDefined(idx)]] prior to accessing the
+    * element, or use [[get()]] instead.
+    */
+  override def apply(idx: Int): A = this.values(idx)
+
+  def get(idx: Int): Option[A] = if (isDefined(idx)) Some(apply(idx)) else None
+}
```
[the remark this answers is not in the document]
tfenne: Let's wait and see if we feel the need for this.  I think for these types we'd probably want it to be `Iterator[Int, Option[A]` because you'd probably want the index too?  So let's see what we need and add it then.

src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala
```diff
@@ -0,0 +1,120 @@
...
+    * To check for this value you must use `value.equals(MissingFloat)` NOT `value == MissingFloat` as the latter
+    * will always return false.
+    */
+  val MissingFloat: Float = {
+    import java.lang.{Long => JLong, Float => JFloat}
+    val l = JLong.parseLong("7F800001", 16)
+    JFloat.intBitsToFloat(l.intValue())
+  }
+
+  private val MissingFloatIntBits: Int = java.lang.Float.floatToRawIntBits(MissingFloat)
```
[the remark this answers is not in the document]
tfenne: Ugh, it's super-complicated.  But basically in BCF Heng had this idea to use a non-canonical `NaN` value to represent missing.  In Java/scala any `NaN` will always return `false` from any equality test.  So to test for this value we have to convert it to it's raw int bits, and compare against that.

src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  /** Returns true if the passed value is one of the missing values. */
+  def isMissingValue(x: Any): Boolean = x match {
+    case ch: Char   => ch == '.'
+    case s: String  => s == "."
+    case i: Int     => i == MissingInt
+    case f: Float   => java.lang.Float.floatToRawIntBits(f) == MissingFloatIntBits
```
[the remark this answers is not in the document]
tfenne: Won't work due to being an `NaN`.

PR #503, opened 2019-07-18
Attempts to make the Duplex caller even faster, and maintain readability

tfenne:


src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -50,32 +51,38 @@ class ConsensusCallingIterator[C <: SimpleRead](sourceIterator: Iterator[SamReco
...
+      val groupingIterator = new SamRecordGroupedIterator(progressIterator, caller.sourceMoleculeId)
       groupingIterator.flatMap(caller.consensusReadsFromSamRecords)
     }
     else {
-      val pool          = new ForkJoinPool(threads, ForkJoinPool.defaultForkJoinWorkerThreadFactory, null, true)
-      val bufferedIter  = groupingIterator.bufferBetter
-      val callersFactor = 1000 // the number of callers per thread to create. This to reduce chance of picking the same caller
-      val callers       = Array.range(start=0, threads*callersFactor).map { _ => caller.emptyClone() }
+      val halfMaxRecords   = maxRecordsInRam / 2
+      val groupingIterator = new SamRecordGroupedIterator(new AsyncIterator(progressIterator, Some(halfMaxRecords)).start(), caller.sourceMoleculeId)
```
tfenne: Putting the async iterator here sped up one test case from 5.04 minutes to 4.45 minutes.  Not that it is also not increasing the total number of threads or records in RAM, but "borrowing" from the specified totals.

PR #507, opened 2019-07-20
Moving cross-building to 2.12 and 2.13

tfenne:
This one is uglier.  I had to resort to having a single 2.12 vs. 2.13 specific source file to work around the fact that `IterableView` is gone in 2.13 and `View` didn't exist in `2.12`.

There are still a ton of deprecation warnings too.  But things are at least working, and we can chip away at the deprecations.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -83,8 +83,8 @@ lazy val commonSettings = Seq(
   organizationName     := "Fulcrum Genomics LLC",
   homepage             := Some(url("http://github.com/fulcrumgenomics/fgbio")),
   startYear            := Some(2015),
-  scalaVersion         := "2.12.8",
-  crossScalaVersions   :=  Seq("2.11.12", "2.12.8"),
+  scalaVersion         := "2.13.0",
+  crossScalaVersions   :=  Seq("2.12.8", "2.13.0"),
   scalacOptions        ++= Seq("-target:jvm-1.8", "-deprecation", "-unchecked"),
   scalacOptions in (Compile, doc) ++= docScalacOptions,
   scalacOptions in (Test, doc) ++= docScalacOptions,
@@ -122,17 +122,19 @@ lazy val root = Project(id="fgbio", base=file("."))
     libraryDependencies ++= Seq(
       "org.scala-lang"            %  "scala-reflect"  % scalaVersion.value,
       "org.scala-lang"            %  "scala-compiler" % scalaVersion.value,
-      "org.scala-lang.modules"    %% "scala-xml"      % "1.0.6",
-      "com.fulcrumgenomics"       %% "commons"        % "0.7.0",
-      "com.fulcrumgenomics"       %% "sopt"           % "0.7.0",
+      "org.scala-lang.modules"    %% "scala-xml"      % "1.2.0",
+      "org.scala-lang.modules"    %% "scala-collection-compat" % "2.1.1",
+      "com.fulcrumgenomics"       %% "commons"        % "0.8.0-87f4b88-SNAPSHOT",
+      "com.fulcrumgenomics"       %% "sopt"           % "0.8.0-6330673-SNAPSHOT",
       "com.github.samtools"       %  "htsjdk"         % "2.16.1" excludeAll(htsjdkExcludes: _*),
       "org.apache.commons"        %  "commons-math3"  % "3.6.1",
-      "com.beachape"              %% "enumeratum"     % "1.5.12",
+      "com.beachape"              %% "enumeratum"     % "1.5.13",
       "com.intel.gkl"             %  "gkl"            % "0.8.6",
 
       //---------- Test libraries -------------------//
-      "org.scalatest"             %% "scalatest"     % "3.0.4"  % "test->*" excludeAll ExclusionRule(organization="org.junit", name="junit")
+      "org.scalatest"             %% "scalatest"     % "3.0.8"  % "test->*" excludeAll ExclusionRule(organization="org.junit", name="junit")
     ))
+
 
 ////////////////////////////////////////////////////////////////////////////////////////////////
 // Merge strategy for assembly
--- a/src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
+++ b/src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
@@ -415,7 +415,7 @@ class Aligner(val scorer: AlignmentScorer,
                                   location: MatrixLocation): Alignment = {
     var currOperator: CigarOperator = null
     var currLength: Direction = 0
-    val elems = new mutable.ArrayBuffer[CigarElem]()
+    val elems = IndexedSeq.newBuilder[CigarElem]
     var MatrixLocation(curI, curJ, curD) = location
 
     // The score is always the score from the starting cell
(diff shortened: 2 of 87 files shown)


build.sbt
```diff
@@ -122,18 +122,31 @@ lazy val root = Project(id="fgbio", base=file("."))
     libraryDependencies ++= Seq(
       "org.scala-lang"            %  "scala-reflect"  % scalaVersion.value,
       "org.scala-lang"            %  "scala-compiler" % scalaVersion.value,
-      "org.scala-lang.modules"    %% "scala-xml"      % "1.0.6",
-      "com.fulcrumgenomics"       %% "commons"        % "0.7.0",
-      "com.fulcrumgenomics"       %% "sopt"           % "0.7.0",
+      "org.scala-lang.modules"    %% "scala-xml"      % "1.2.0",
+      "org.scala-lang.modules"    %% "scala-collection-compat" % "2.1.1",
+      "com.fulcrumgenomics"       %% "commons"        % "0.8.0-c052fc9-SNAPSHOT",
```
nh13: confirming that we'll use https://github.com/fulcrumgenomics/commons/pull/54 when it is merged?  In particular `View`.

src/main/scala-2.13-/com/fulcrumgenomics/util/CrossCompatView.scala
```diff
@@ -0,0 +1,5 @@
+package com.fulcrumgenomics.util
+
+trait CrossCompatView[A, Coll] extends scala.collection.IterableView[A, Coll] {
```
nh13: This goes away with https://github.com/fulcrumgenomics/commons/pull/54?

src/main/scala/com/fulcrumgenomics/Writer.scala
```diff
@@ -24,27 +24,6 @@
...
-  }
-
-  /** Writes an item and returns a reference to the writer. */
-  def ++=[B <: A](items: TraversableOnce[B]): this.type = {
-    write(items)
-    this
-  }
-}
+@deprecated(message="Use com.fulcrumgenomics.commons.io.Writer directly instead.")
+trait Writer[A] extends com.fulcrumgenomics.commons.io.Writer[A]
```
nh13: I think master already has this.

src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
```diff
@@ -512,7 +512,7 @@ class Aligner(val scorer: AlignmentScorer,
   private def findByScore(matrices: Array[AlignmentMatrix], minScore: Int): Seq[MatrixLocation] = {
     val qLen = matrices(0).queryLength
     val tLen = matrices(0).targetLength
-    val hits = new ArrayBuffer[MatrixLocation]()
```
nh13: Can you explain this change for my own learning?  I would have expected `ArrayBuffer.newBuilder` (maybe)?  Why doesn't `new ArrayBuffer[]()` work anymore?  It looks like they're still here: https://docs.scala-lang.org/overviews/collections/concrete-mutable-collection-classes.html#array-buffers

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -386,13 +386,13 @@ class CollectDuplexSeqMetrics
       case _         => unreachable(s"Found ${ssGroups.size} single strand families in a double strand family!")
     }
 
-    val umi1s = new ArrayBuffer[String] // ab UMI 1 (and ba UMI 2) sequences
-    val umi2s = new ArrayBuffer[String] // ab UMI 2 (and ba UMI 1) sequences
+    val umi1s = Seq.newBuilder[String] // ab UMI 1 (and ba UMI 2) sequences
+    val umi2s = Seq.newBuilder[String] // ab UMI 2 (and ba UMI 1) sequences
```
nh13: Arrays are fast, is this fast?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -205,19 +205,19 @@ object GroupReadsByUmi {
 
     /** Assigns IDs to each UMI based on the root to which is it mapped. */
     protected def assignIdsToNodes(roots: Seq[Node]): Map[Umi, MoleculeId] = {
-      val mappings = mutable.Buffer[(Umi,MoleculeId)]()
+      val mappings = Map.newBuilder[Umi,MoleculeId]
```
nh13: presumably we used `mutable.Buffer` for performance?

src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
tfenne: The problem stems from the fact that all the type aliases in `scala.` that are automatically available have been changed.  They be things like `type Seq[+A] = scala.collection.Seq[A]`, but they have been changed in 2.13, to point instead to the immutable version of the traits.  That leaves us with a choice between two routes:

1) Explicitly import `scala.collection.Seq` and shadow `scala.Seq` when we want to specify _either mutable or immutable_, mirroring the 2.12 behaviour. 
2) Change our code to produce immutable `Seq`s (and `IndexedSeq`s etc.) so that they are now of type `scala.Seq`.

Since we're generally treating `Seq` and `IndexedSeq` as immutable types, I thought it was better go with with option (2).  The problem this causes is that we used to get away with using things like `ArrayBuffer` and `ListBuffer` and rely on the fact that they implemented `Seq`, but they implement `mutable.Seq`.  So I _think_ that a call to `ArrayBuffer.toSeq` will cause a second builder to be made to make the immutable Seq.

I'm assuming that using the builders efficient, but haven't tested.  Note that this is what any `.map()`, `.filter()` etc uses to build the new collection, so it's not like we haven't been relying on them all along.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I think we used it because it was the _obvious_ way to build up a map mutably.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
tfenne: Ok, so a little poking, and in 2.12 and 2.13, `Seq.newBuilder` yields a `ListBuffer`, and `IndexedSeq.newBuilder` yields a `VectorBuilder` which uses arrays under the covers and is fairly smart about not re-allocating stuff to make the immutable collection.  I'll change over to `IndexedSeq.newBuilder` where I had `Seq.newBuilder.`

tfenne:
@nh13 I did some more work this morning to remove as many deprecations as I could.  There are two major sources of deprecation warnings left:

1. `String.sliding(n)` has been deprecated, and the warning says to use `String.toSeq.sliding(n).map(_.unwrap)` which a) is super ugly and b) doesn't work under 2.12 because it would appear that the return type of `String.toSeq` changed between 2.12 and 2.13 :/
2. Use of java collection converters - 2.13 really wants you to use `scala.jdk.CollectionConverters`, which didn't exist in 2.12.  

It would be really nice to resolve both of these, but they require a little more time and thought than I want to spend right now.

I also removed the deprecated options in `ClipBam` and `DemuxFastqs` - see the last commit.  They've been deprecated for quite some time now and we're breaking enough else it seemed like a good time.

tfenne:
@nh13 I've squashed this down to 2 commits, one for all the cross-building and deprecation changes, and one for the changes to ClipBam and DemuxFastqs.  I'm going to go ahead and merge so I can rebase the VCF API PR on top of this.  If you want to review post-hoc, I'm happy to make another PR to master to address any concerns.

PR #512, opened 2019-08-07
Add a few more methods on ProgressLogger

nh13:
Supports Variant and Interval, as well as any type.

@tfenne This PR has few things I find useful that should each be judged on their own merits:

1. added `record` methods for `Variant` and `Interval`.  It's about time.
2. added a generic `record` method that accepts any time, and if the type isn't known, fallsback to a given method, which defaults to just calling `record()`
3. an implicit class that adds `progress` to an iterator to easily record progress on an iterator.  It turns this:

```scala
iterator.map { item => p.record(item); item }.foreach { ... }
```

into this
```
import com.fulcrumgenomics.util.ProgressLogger.ProgressLoggingIterator
iterator.progress(p).foreach { ... }
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
@@ -24,8 +24,12 @@
 package com.fulcrumgenomics.util
 
 import com.fulcrumgenomics.bam.api.SamRecord
+import com.fulcrumgenomics.commons.collection.SelfClosingIterator
 import com.fulcrumgenomics.commons.util.Logger
-import htsjdk.samtools.util.AbstractProgressLogger
+import com.fulcrumgenomics.vcf.api.Variant
+import htsjdk.samtools.util.{AbstractProgressLogger, Locatable}
+
+import scala.reflect.runtime.universe._
 
 /**
   * A subclass of HTSJDK's progress logger that uses fgbio's logging system.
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
```diff
@@ -42,9 +43,48 @@ case class ProgressLogger(logger: Logger,
...
+  def record(interval: Interval): Boolean = record(interval.getContig, interval.getStart)
+
+  /** Record progress of a single item.
+    *
+    * @param item the item log
+    * @param fallback the fallback method to record progress if the type is not a type supported by [[ProgressLogger]].
+    * @return
+    */
+  def record[ItemType](item: ItemType,
+                       fallback: (ProgressLogger, ItemType) => Boolean = (p: ProgressLogger, _: ItemType) => p.record()): Boolean = {
```
nh13: If you don't like this method on `ProgressLogger` but like the `ProgressLoggingIterator`, I can move the pattern matching directly into the method in `ProgressLoggingIterator`.

src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
```diff
@@ -42,9 +43,48 @@ case class ProgressLogger(logger: Logger,
...
+  }
 
   /** Logs the last record if it wasn't already logged. */
   def logLast(): Boolean = super.log() // Calls the super's log() method, not log(message: String*)
+}
+
+object ProgressLogger {
 
+  /** An implicit class that wraps an [[Iterator]] and provides a method to record progress as items are consumed. */
+  implicit class ProgressLoggingIterator[ItemType](iterator: Iterator[ItemType]) {
```
tfenne: I feel like there's probably a good way, using implicits, to limit the use of the class to iterators over types that can actually be logged.

src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
```diff
@@ -42,9 +43,48 @@ case class ProgressLogger(logger: Logger,
   /** Method to use to record progress when genome location isn't available or relevant. */
   def record(): Boolean = record(null, -1)
 
-  def record(rec: SamRecord): Boolean = record(rec.asSam)
+  def record(rec: SamRecord): Boolean = super.record(rec.asSam)
+
+  def record(variant: Variant): Boolean = record(variant.chrom, variant.pos)
```
tfenne: :+1:

src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
```diff
@@ -42,9 +43,48 @@ case class ProgressLogger(logger: Logger,
...
+
+  def record(interval: Interval): Boolean = record(interval.getContig, interval.getStart)
```
tfenne: 👍

src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
tfenne: I think it would be better to have this be a private def in the iterator as you suggest.

src/main/scala/com/fulcrumgenomics/util/ProgressLogger.scala
```diff
@@ -42,9 +46,43 @@ case class ProgressLogger(logger: Logger,
...
 
+  /** Provides a method to record items of the given type.  If [[ItemType]] is not one of [[SamRecord]],
+    * `(String, Int)`, [[Variant]], or [[Interval], [[ProgressLogger.record()]] will be used. */
+  sealed abstract class ProgressLoggingHelper[ItemType: TypeTag] {
+    protected val record: (ProgressLogger, ItemType) => Boolean = {
+      typeOf[ItemType] match {
+        case t if t =:= typeOf[SamRecord]     => (p: ProgressLogger, item: ItemType) => p.record(item.asInstanceOf[SamRecord])
+        case t if t =:= typeOf[(String, Int)] => (p: ProgressLogger, item: ItemType) => { val (contig, pos) = item.asInstanceOf[(String, Int)]; p.record(contig, pos) }
+        case t if t =:= typeOf[Variant]       => (p: ProgressLogger, item: ItemType) => p.record(item.asInstanceOf[Variant])
+        case t if t =:= typeOf[Interval]      => (p: ProgressLogger, item: ItemType) => p.record(item.asInstanceOf[Interval])
```
tfenne: I thnk `=:=` require equality of type right?  Perhaps you want to use `t <:< typeOf[Interval]` that way any sub-type of Interval can also be handled?

tfenne: And in fact, then you could use `Locatable` instead of `Interval` and catch a lot more stuff.

PR #513, opened 2019-08-07
Relax requirements on bases and quals in SamBuilder

nh13:
I'll adds if you agree.  I want a few more things to be supported:
1. Bases are present, but missing quals.
2. Bases are missing, but quals are present (less important, but symmetric)

(1) for sure I want and isn't currently supported.  I have to after the fact set the quals to missing when building `SamRecord`s.


tfenne:
Seems reasonable, I think.

nh13:
@tfenne one last look?

PR #514, opened 2019-08-08
Add option to set sampleName in SamBuilder

mjhipp:


src/main/scala/com/fulcrumgenomics/testing/SamBuilder.scala
```diff
@@ -45,11 +45,14 @@ object SamBuilder {
...
   object Plus  extends Strand { val isNegative = false }
   object Minus extends Strand { val isNegative = true }
+  val DefaultSampleName: String = "Sample"
+  val DefaultReadGroupId: String = "A"
 }
 
 /** Class to create sets of [[com.fulcrumgenomics.bam.api.SamRecord]]s for testing. */
 class SamBuilder(val readLength: Int=100,
                  val baseQuality: Int=30,
+                 val sampleName: String = DefaultSampleName,
```
mjhipp: If it is preferred, I can instead use the convention:

```suggestion
                 val sampleName: Option[String] = None,
```
and set it downstream like it is done with `readGroupId`

PR #516, opened 2019-08-13
Add queryname sorted SamRecord and Template iterators

clintval:
I need a queryname sorted iterator of `Template` objects for [a template disambiguation algorithm](https://github.com/clintval/cvbio/blob/master/tools/src/com/cvbio/tools/disambiguate/Disambiguate.scala).

I use the methods in this PR to walk over templates, by name, from many SAM sources ([snippet](https://github.com/clintval/cvbio/blob/bc37a999dabc529cd179ec7c723bcea206d8740f/tools/src/com/cvbio/bam/Bams.scala#L59-L96)).

Any interest in accepting these new features?


src/test/scala/com/fulcrumgenomics/bam/BamsTest.scala
```diff
@@ -85,6 +85,52 @@ class BamsTest extends UnitSpec {
...
+    Bams.querySortedIterator(in=builder.toSource).map(_.name).toSeq shouldBe Seq("p1", "p1", "q1", "q2")
+  }
+
+  it should "use Picard style sorting of querynames" in {
+    val builder = new SamBuilder(sort=Some(SamOrder.Queryname))
+    builder.addFrag(name="A9", start=100)
+    builder.addPair(name="A88", start1=100, start2=300)
+    builder.addFrag(name="A88", start=200)
+
+    Bams.querySortedIterator(in=builder.toSource).map(_.name).toSeq shouldBe Seq("A88", "A88", "A88", "A9")
```
clintval: See: https://github.com/samtools/hts-specs/pull/361#issuecomment-447065728

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -207,17 +207,52 @@ object Bams extends LazyLogging {
...
+      case (Some(Queryname), _) => new SelfClosingIterator(iterator.bufferBetter, () => CloserUtil.close(iterator))
+      case (_, _) =>
+        logger.info(parts = "Sorting into queryname order.")
+        val progress = ProgressLogger(this.logger, "Records", "sorted")
+        val sort     = sorter(Queryname, header, maxInMemory, tmpDir)
+        iterator.foreach { rec =>
+          progress.record(rec)
+          sort.write(rec)
+        }
+        new SelfClosingIterator(sort.iterator, () => sort.close())
```
nh13: `querySortedIterator` looks good to me.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -263,6 +298,54 @@ object Bams extends LazyLogging {
...
+    * in a non-guaranteed sort order.
+    *
+    * @see [[com.fulcrumgenomics.bam.Bams.templateIterator]]
+    *
+    * @param in a SamReader from which to consume records
+    * @param maxInMemory the maximum number of records to keep and sort in memory, if sorting is needed
+    * @param tmpDir a temp directory to use for temporary sorting files if sorting is needed
+    * @return an Iterator of queryname sorted Template objects
+    */
+  def sortedTemplateIterator(in: SamSource,
```
nh13:
```suggestion
  def templateSortedIterator(in: SamSource,
```

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -263,6 +298,54 @@ object Bams extends LazyLogging {
...
+    val _iterator = new Iterator[Template] {
+      override def hasNext: Boolean = queryIterator.hasNext
+      override def next: Template   = {
+        require(hasNext, "next() called on empty iterator")
+        Template(queryIterator)
+      }
+    }
+
+    new SelfClosingIterator(_iterator, () => queryIterator.close())
+  }
```
nh13: So this method is here so that we can have an iterator over `Template`s but guaranteed to be in query-name order?  Can you explain a use-case?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -263,6 +298,54 @@ object Bams extends LazyLogging {
...
+                             tmpDir: DirPath): SelfClosingIterator[Template] = {
+    val queryIterator = querySortedIterator(iterator, header, maxInMemory, tmpDir)
+
+    val _iterator = new Iterator[Template] {
+      override def hasNext: Boolean = queryIterator.hasNext
+      override def next: Template   = {
+        require(hasNext, "next() called on empty iterator")
+        Template(queryIterator)
+      }
+    }
```
nh13: Are you ready for my cute one-liner?
```scala
val _iterator = Iterator.continually { Template(queryIterator } takeWhile { _ => queryIterator.hasNext }
```

clintval: Super cute. But always leaves out the last element of the iterator! Sadly no [`takeTo`](https://github.com/scala/scala-dev/issues/417) in the Scala language yet.

I spent 20 minutes on a one-liner workaround, but did not find a solution worth posting.

clintval:
As far as use cases, I have just one. You [may have skimmed in the past](https://twitter.com/clint_valentine/status/1139075434602160128).

I need to walk through templates from one sequencing experiment that were aligned to completely different references. I want to guarantee that I can accept BAMs in any sort order, but then internally sort by queryname so I can zip through the templates and compare alignment scores.

I am already using these routines in a [template-level multi-species disambiguation utility](https://github.com/clintval/cvbio/blob/master/tools/src/com/cvbio/tools/disambiguate/Disambiguate.scala).

A snippet that uses the `templateSortedIterator`:

```scala
/** Collectively iterate through [[SamSource]] iterators and emit templates of the same query.
  *
  * Reads will be grouped and sorted into queryname order if they are not already. All [[SamSource]] must contain
  * the same templates by query name.
  *
  * @param sources the SAM sources to iterate over
  * @return a self-closing iterator over all templates in all SAM sources sorted by name
  */
def templatesIterator(sources: SamSource*): SelfClosingIterator[Seq[Template]] = {
  val iterator: Iterator[Seq[Template]] = new Iterator[Seq[Template]] {

    /** The underlying template iterators. */
    private val underlying = sources.map(templateSortedIterator(_))

    /** Test all template iterators to see if they have another template to emit. */
    override def hasNext: Boolean = {
      underlying
        .map(_.hasNext)
        .ensuring(_.distinct.length <= 1, "SAM sources do not have the same number of templates!")
        .contains(true)
    }

    /** Advance to the next sequence of templates. */
    override def next(): Seq[Template] = {
      require(hasNext, "next() called on empty iterator")
      underlying
        .map(_.next)
        .ensuring(templates =>
          templates.map(_.name).distinct.length <= 1,
          "Templates with different names found! This can only occur if your SAM sources are queryname sorted using"
            + " different implementations, such as with Picard tools versus Samtools. If you have encountered this"
            + " exception, then please alert the maintainer!"
        )
    }
  }

  new SelfClosingIterator(iterator, () => sources.foreach(_.safelyClose()))
}
```

And I use it like:

```scala
val sources = input.map(bam => SamSource(bam))

templatesIterator(sources: _*).foreach { templates =>
  /** Do work with templates of the same queryname, from different SAM sources. */
  ???
}
```

clintval:
No problem if these routines are too specific to a single example, and therefore better left out of this codebase.

I can maintain fine where the routines currently live. I'm posting to see if you'd like to round out your iterator collections and if these routines may be of use to others.

src/test/scala/com/fulcrumgenomics/bam/BamsTest.scala
```diff
@@ -109,6 +159,44 @@ class BamsTest extends UnitSpec {
     }
   }
 
+  "Bams.sortedTemplateIterator" should "return template objects in order" in {
```
clintval:
```suggestion
  "Bams.templateSortedIterator" should "return template objects in a sorted order" in {
```

nh13:
@tfenne could use a fresh set of eyes given this touches the core API

nh13:
@clintval looks good to me? Let's give @tfenne a week and if I don't hear anything we'll merge?

PR #520, opened 2019-08-19
Don't try to index BAM files that can't be indexed with HTSJDK.

tfenne:
@nh13 The change here is quite simple.  HTSJDK blows up if you ask it to index on the fly and hand it reference sequences that are beyond what BAI can handle.  There's no CSI writing support yet either.

I'd also like to wait until https://github.com/samtools/htsjdk/pull/1410 is merged, and update our HTSJDK dependency in this same PR if possible.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,6 +95,7 @@ lazy val commonSettings = Seq(
   fork in Test         := true,
   resolvers            += Resolver.sonatypeRepo("public"),
   resolvers            += Resolver.mavenLocal,
+  resolvers            += "broad-snapshots" at "https://artifactory.broadinstitute.org/artifactory/libs-snapshot/",
   shellPrompt          := { state => "%s| %s> ".format(GitCommand.prompt.apply(state), version.value) },
   updateOptions        := updateOptions.value.withCachedResolution(true)
 ) ++ Defaults.coreDefaultSettings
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/bam/api/SamWriter.scala
```diff
@@ -83,11 +83,19 @@ object SamWriter extends LazyLogging {
         Some(new Sorter(maxRecordsInRam, new SamRecordCodec(header), so.sortkey))
     }
 
+    // Only index the output file if the sort order is right, indexing is requested, and no chrom is too long
+    val actuallyIndex = {
+      val hasChromsTooLong = header.getSequenceDictionary.getSequences.exists(_.getSequenceLength > GenomicIndexUtil.BIN_GENOMIC_SPAN)
+      val wouldHaveIndexed = header.getSortOrder == SortOrder.coordinate && index
+      if (wouldHaveIndexed && hasChromsTooLong) logger.warning(s"Cannot index $path as one or more chromosomes is too long.")
+      wouldHaveIndexed && !hasChromsTooLong
```
nh13:
```scala
if (header.getSortOrder != SortOrder.coordinate || !index) false else {
  val hasChromsTooLong = header.getSequenceDictionary.getSequences.forall(_.getSequenceLength <= GenomicIndexUtil.BIN_GENOMIC_SPAN)
  if (hasChromsTooLong) logger.warning(s"Cannot index $path as one or more chromosomes is too long.")
  hasChromsTooLong
}

File: build.sbt [hunk 2]
```diff
@@ -95,6 +95,7 @@ lazy val commonSettings = Seq(
   fork in Test         := true,
   resolvers            += Resolver.sonatypeRepo("public"),
   resolvers            += Resolver.mavenLocal,
+  resolvers            += "broad-snapshots" at "https://artifactory.broadinstitute.org/artifactory/libs-snapshot/",
```
nh13: is this to get htsjdk snapshots?  Sucks, but I understand.

tfenne: It is indeed for htsjdk snapshots.

PR #521, opened 2019-08-20
Same fix, in RandomizeBam, that was applied for RandomQuery sorting t…

tfenne:
…o prevent collisions.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/RandomizeBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/RandomizeBam.scala
@@ -58,12 +58,22 @@ class RandomizeBam(
   Io.assertCanWriteFile(output)
 
   override def execute(): Unit = {
-    val extractor  = if (queryGroup) (rec: SamRecord) => rec.name
-                     else            (rec: SamRecord) => rec.name + ":" + rec.basesString + ":" + rec.qualsString
     val in = SamSource(input)
     val hasher = new Murmur3(seed)
-    val keyfunc = (r: SamRecord) => SamOrder.RandomKey(hasher.hashUnencodedChars(extractor.apply(r)), r.asSam.getFlags)
-    val sorter = new Sorter(2e6.toInt, new SamRecordCodec(in.header), keyfunc)
+
+    val sorter = {
+      if (queryGroup) {
+        val f = (r: SamRecord) => SamOrder.RandomQueryKey(hasher.hashUnencodedChars(r.name), r.name, r.asSam.getFlags)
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/bam/RandomizeBam.scala
```diff
@@ -58,12 +58,22 @@ class RandomizeBam(
...
+      if (queryGroup) {
+        val f = (r: SamRecord) => SamOrder.RandomQueryKey(hasher.hashUnencodedChars(r.name), r.name, r.asSam.getFlags)
+        new Sorter(2e6.toInt, new SamRecordCodec(in.header), f)
+      }
+      else {
+        val f = (r: SamRecord) => {
+          val key = s"${r.name}:${r.basesString}:${r.qualsString}"
+          SamOrder.RandomKey(hasher.hashUnencodedChars(key), r.asSam.getFlags)
+        }
+        new Sorter(2e6.toInt, new SamRecordCodec(in.header), f)
```
tfenne: I don't love repeating this here, but after a few minutes of banging my head against the type inference stuff, I gave up trying to extract it.

src/main/scala/com/fulcrumgenomics/bam/RandomizeBam.scala
```diff
@@ -58,12 +58,22 @@ class RandomizeBam(
...
-    val extractor  = if (queryGroup) (rec: SamRecord) => rec.name
-                     else            (rec: SamRecord) => rec.name + ":" + rec.basesString + ":" + rec.qualsString
     val in = SamSource(input)
     val hasher = new Murmur3(seed)
-    val keyfunc = (r: SamRecord) => SamOrder.RandomKey(hasher.hashUnencodedChars(extractor.apply(r)), r.asSam.getFlags)
-    val sorter = new Sorter(2e6.toInt, new SamRecordCodec(in.header), keyfunc)
+
+    val sorter = {
+      if (queryGroup) {
+        val f = (r: SamRecord) => SamOrder.RandomQueryKey(hasher.hashUnencodedChars(r.name), r.name, r.asSam.getFlags)
```
nh13: Why don't you just append the name to the hash?  Then you can go back to the original implementation since `f` has type `SamRecord => SamOrder.RandomKey`?

```suggestion
        val f = (r: SamRecord) => SamOrder. RandomKey(hasher.hashUnencodedChars(r.name) + r.name, r.asSam.getFlags)
```

tfenne: The hash is an `Int` not a `String`.

PR #535, opened 2019-10-08
Add classes for reading and writing interval lists

nh13:
The motivation comes from creating interval lists that cannot fit into memory, as well as supporting gzipped interval lists.


src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+import com.fulcrumgenomics.commons.util.StringUtil
+import htsjdk.samtools.util.{BufferedLineReader, Interval}
+import htsjdk.samtools.{SAMFileHeader, SAMSequenceDictionary, SAMTextHeaderCodec}
+
+import scala.io.Source
+
+object IntervalListSource {
+
+  /** Creates a new interval list source from a sequence of lines. */
+  def apply(lines: Seq[String]): IntervalListSource = new IntervalListSource(lines.iterator)
```
tfenne: Perhaps `Iterable[String]` rather than `Seq[String]`?

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  /** Creates a new interval list source from an iterator of lines. */
+  def apply(lines: Iterator[String]): IntervalListSource = new IntervalListSource(lines)
+
+  /** Creates a new interval list source from an input stream. */
+  def apply(stream: InputStream): IntervalListSource = new IntervalListSource(Source.fromInputStream(stream).getLines(), Some(stream))
```
tfenne: I feel like you probably shouldn't close streams or sources that are handed to you, only ones you create.  Or at the very least warn in the docs that the stream/source will be closed.

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  /** Creates a new interval list source from a source. */
+  def apply(source: Source): IntervalListSource = new IntervalListSource(source.getLines(), Some(source))
+
+  /** Creates a new interval list source from a File. */
+  def apply(file: File): IntervalListSource = apply(path=file.toPath)
+
+  /** Creates a new interval list source from a Path. */
+  def apply(path: PathToIntervals): IntervalListSource = apply(Io.toInputStream(path))
```
tfenne: Could also `apply(Io.readLines(path))`

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+  * Reads intervals from any text based source via a reader.  The underlying reader is closed automatically when EOF is
+  * reached.
+  */
+class IntervalListSource private(lines: Iterator[String],
+                                 private[this] val source: Option[{ def close(): Unit }] = None)
+  extends Iterator[Interval] with Closeable {
```
tfenne: Do you want to mix in `View[Interval]` here like we do with other readers so that map/filter/etc. are lazy?

tfenne: Makes me wonder actually if we should have this in commons:
```
trait Source[A] extends View[A] with Closeable {
  def underlying: Iterable[A] = this
}
```

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  private val iter = lines.bufferBetter
+
+  private var lineNumber = 1L
+
+  // Read the header
+  val header: SAMFileHeader = {
+    val codec = new SAMTextHeaderCodec
+    val headerLines = iter.takeWhile(_.startsWith("@")).toSeq
```
tfenne: FYI I tend to prefer `toIndexedSeq` as `toSeq` will often return a lazy Stream which can be a PITA when debugging.

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+    require(headerLines.nonEmpty, "No header found")
+    lineNumber += headerLines.length
+    val lineReader = BufferedLineReader.fromString(headerLines.mkString("\n"))
+    codec.decode(lineReader, "IntervalListSource")
+  }
+
+  require(!this.dict.isEmpty, "No reference sequences found in the header.")
+
+  def dict: SAMSequenceDictionary= this.header.getSequenceDictionary
```
tfenne: scaladoc for public function

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+  override def hasNext: Boolean = this.iter.nonEmpty
+
+  /** Returns the next record if available, or throws an exception if none is available. */
+  override def next(): Interval = parse(this.iter.next())
+
+  /** Closes the underlying reader; only necessary if EOF hasn't been reached. */
+  override def close(): Unit = this.source.foreach(_.close())
+
+  private def parse(line: String): Interval = {
+    require(StringUtil.split(line, '\t', parseArray) == 5, s"Expected 5 fields on line $lineNumber")
```
tfenne: Could you split this into two lines please?  I.e.
```scala
val fieldCount = StringUtil.split(line, '\t', parseArray)
require(fieldCount == 5, "...")
```
I don't like having side-effecting things in `require()`s

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+        require(start <= end, f"Start is greater than end on line number $lineNumber")
+    }
+
+    val negative = strand match {
+      case "-" => true
+      case "+" => false
+      case _   => throw new IllegalArgumentException(s"Unrecognized strand '$strand' on line number $lineNumber")
+    }
+
+    lineNumber += 1
```
tfenne: I would suggest this belongs in `next()` where you actually consume the line.  E.g. 
```scala
def next(): Interval = yieldAndThen(parse(this.iter.next()) { lineNumber += 1 }
```

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  /** Returns the next record if available, or throws an exception if none is available. */
+  override def next(): Interval = parse(this.iter.next())
+
+  /** Closes the underlying reader; only necessary if EOF hasn't been reached. */
+  override def close(): Unit = this.source.foreach(_.close())
+
+  private def parse(line: String): Interval = {
+    require(StringUtil.split(line, '\t', parseArray) == 5, s"Expected 5 fields on line $lineNumber")
+    val refName     = parseArray(0)
```
tfenne: Could be cute and do:
```scala
val Array(refName, startStr, endStr, strand, name) = parseArray
```

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+    val strand      = parseArray(3)
+    val name        = parseArray(4)
+
+    val start = startString.toInt
+    val end   = endString.toInt
+
+    Option(dict.getSequence(refName)) match {
+      case None => throw new IllegalArgumentException(f"Reference contig '$refName' not found in the sequence dictionary.")
+      case Some(seq) =>
+        require(1 <= start, s"Start is less than 1 on line number $lineNumber")
```
tfenne: Would be nice to emit the string of the line in error also for these errors.

src/main/scala/com/fulcrumgenomics/util/IntervalListSource.scala
```diff
@@ -0,0 +1,126 @@
...
+      case "-" => true
+      case "+" => false
+      case _   => throw new IllegalArgumentException(s"Unrecognized strand '$strand' on line number $lineNumber")
+    }
+
+    lineNumber += 1
+
+    new Interval(refName, start, end, negative, name)
+  }
+}
```
tfenne: How about adding this so that we could use this as our default way to also load interval lists?

```scala
  def toIntervalList: IntervalList = {
    val list = new IntervalList(dict)
    while (hasNext) list.add(next())
    list
  }
```

src/main/scala/com/fulcrumgenomics/util/IntervalListWriter.scala
```diff
@@ -0,0 +1,92 @@
...
+import java.io.{BufferedWriter, Closeable}
+
+import com.fulcrumgenomics.FgBioDef.PathToIntervals
+import com.fulcrumgenomics.commons.io.Writer
+import htsjdk.samtools.{SAMFileHeader, SAMSequenceDictionary, SAMTextHeaderCodec}
+import htsjdk.samtools.util.Interval
+
+
+object IntervalListWriter {
+  /** Constructs a [[IntervalListWriter]] that will write to the provided path. */
```
tfenne: a -> an

src/main/scala/com/fulcrumgenomics/util/IntervalListWriter.scala
```diff
@@ -0,0 +1,92 @@
...
+    val header = new SAMFileHeader()
+    header.setSequenceDictionary(dict)
+    this.apply(writer, header)
+  }
+}
+
+/**
+  * Implements a writer for interval lists.
+  */
+class IntervalListWriter(val out: BufferedWriter, header: SAMFileHeader) extends Closeable with Writer[Interval] {
```
tfenne: `Writer` already extends `Closeable`.

src/main/scala/com/fulcrumgenomics/util/IntervalListWriter.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  // Write out the header
+  {
+    val codec = new SAMTextHeaderCodec
+    codec.encode(out, header)
+  }
```
tfenne: Why not just `new SAMTextHeaderCodec().encode(out, header)`?

Issue #547, opened 2019-10-28
CollectDuplexSeqMetrics needs raw-tag argument


danielnavarrogomez:
Hello!

I use different tags than "RX" on my bam files and that works fine for all the tools I use from fgbio but CollectDuplexSeqMetrics which doesn't have an option to change it from the default one.

As an example, if I use this command:
java -jar fgbio GroupReadsByUmi --input=INPUT.bam --output=OUTPUT.bam --strategy=paired --raw-tag=XD --min-map-q=30

The output file is not usable for CollectDuplexSeqMetrics

nh13:
@danielnavarrogomez see #548

PR #548, opened 2019-10-28
Add options for the RX/MI tags to CollectDuplexSeqMetrics

nh13:
see https://github.com/fulcrumgenomics/fgbio/issues/547


src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -274,6 +273,8 @@ class CollectDuplexSeqMetrics
   @arg(flag='u', doc="If true, produce the .duplex_umi_counts.txt file with counts of duplex UMI observations.") val duplexUmiCounts: Boolean = false,
   @arg(flag='a', doc="Minimum AB reads to call a tag family a 'duplex'.") val minAbReads: Int = 1,
   @arg(flag='b', doc="Minimum BA reads to call a tag family a 'duplex'.") val minBaReads: Int = 1,
+  @arg(flag='t', doc="The tag containing the raw UMI.")  val rawTag: String    = ConsensusTags.UmiBases,
```
tfenne: I know these names mirror the names used in `GroupReadsByUmi` but I think the context make them less clear here.  How about `umiTag` and `miTag`?

nh13:
Will merge once unit tests pass.

PR #550, opened 2019-11-04
First pass at enabling consensus read generation from SE data.

tfenne:


src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
```diff
@@ -176,11 +176,11 @@ object SamOrder {
     override val subSort:    Option[String] = Some("template-coordinate")
     override val sortkey: SamRecord => A = rec => {
       val readChrom = if (rec.unmapped)     Int.MaxValue else rec.refIndex
-      val mateChrom = if (rec.mateUnmapped) Int.MaxValue else rec.mateRefIndex
+      val mateChrom = if (!rec.paired || rec.mateUnmapped) Int.MaxValue else rec.mateRefIndex
```
tfenne: @nh13 What do you think about adding `@inline final def unpaired: Boolean = !paired` to `SamRecord`? I feel like it would be tidier than using `!rec.paired` everywhere.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -83,13 +83,17 @@ object GroupReadsByUmi {
...
+        val lib     = library(rec)
         val chrom   = rec.refIndex
         val recNeg  = rec.negativeStrand
         val recPos  = if (recNeg) rec.unclippedEnd else rec.unclippedStart
-        val mateNeg = rec.mateNegativeStrand
-        val matePos = if (mateNeg) SAMUtils.getMateUnclippedEnd(rec.asSam) else SAMUtils.getMateUnclippedStart(rec.asSam)
-        val lib = library(rec)
+
+        val (mateNeg, matePos) = if (!rec.paired) (false, Int.MaxValue) else {
+          if (rec.paired && rec.refIndex != rec.mateRefIndex) throw new IllegalArgumentException("Mate on different chrom.")
```
nh13:
```suggestion
          if (rec.refIndex != rec.mateRefIndex) throw new IllegalArgumentException("Mate on different chrom.")
```

nh13: `rec.paired` must be true

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -372,8 +376,11 @@ object Strategy extends FgBioEnum[Strategy] {
     |   3. The assigned UMI tag
     |   4. Read Name
     |
-    |Reads are aggressively filtered out so that only high quality read pairs with both ends
-    |mapped are taken forward.  (Note: the `MQ` tag is required on reads with mapped mates).
+    |Reads are aggressively filtered out so that only high quality reads/mappings are taken forward. Single-end
+    |reads must have mapping quality >= `min-map-q`.  Paired-end reads must have both reads mapped to the same
+    | chromosome with both reads having mapping quality >= `min-mapq`.  (Note: the `MQ` tag is required on reads
```
nh13:
```suggestion
    |chromosome with both reads having mapping quality >= `min-mapq`.  (Note: the `MQ` tag is required on reads
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -431,7 +438,12 @@ class GroupReadsByUmi
 
   private val assigner = strategy.newStrategy(this.edits)
 
-  type ReadPair = (SamRecord, SamRecord)
+  /** Checks that the read's MAPQ is over a minimum, and if the read is paired, that the mate mapq is also over the min. */
```
nh13:
```suggestion
  /** Checks that the read's mapq is over a minimum, and if the read is paired, that the mate mapq is also over the min. */
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -431,7 +438,12 @@ class GroupReadsByUmi
...
+  private def mapqOk(rec: SamRecord, minMapQ: Int): Boolean = {
+    rec.mapq >= minMapQ &&
+      (!rec.paired || rec.get[Int](SAMTag.MQ.name()).exists(_ >= minMapQ))
```
nh13: Why `.get` and `.exists` when earlier you said:
> the `MQ` tag is required on reads with mapped mates

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -450,11 +462,11 @@ class GroupReadsByUmi
...
       .filter(r => !r.secondary && !r.supplementary)
-      .filter(r => (includeNonPfReads || r.pf)            || { filteredNonPf += 1; false })
-      .filter(r => (r.paired && r.mapped && r.mateMapped) || { filteredPoorAlignment += 1; false })
-      .filter(r => (r.refIndex == r.mateRefIndex)         || { filteredPoorAlignment += 1; false })
-      .filter(r => (r.mapq >= this.minMapQ && r.get[Int](SAMTag.MQ.name()).exists(_ >= this.minMapQ)) || { filteredPoorAlignment += 1; false })
-      .filter(r => !r.get[String](rawTag).exists(_.contains('N')) || { filteredNsInUmi += 1; false })
+      .filter(r => (includeNonPfReads || r.pf)                      || { filteredNonPf += 1; false })
+      .filter(r => (r.mapped && (!r.paired || r.mateMapped))        || { filteredPoorAlignment += 1; false })
+      .filter(r => (!r.paired || r.refIndex == r.mateRefIndex)      || { filteredPoorAlignment += 1; false })
+      .filter(r => mapqOk(r, this.minMapQ)                          || { filteredPoorAlignment += 1; false })
```
nh13: Can anything that increments `filteredPoorAlignment` just be refactored into one method?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -466,7 +478,7 @@ class GroupReadsByUmi
 
     logger.info(f"Accepted $kept%,d reads for grouping.")
     if (filteredNonPf > 0) logger.info(f"Filtered out $filteredNonPf%,d non-PF reads.")
-    logger.info(f"Filtered out $filteredPoorAlignment%,d reads that were not part of a high confidence mapped read pair.")
+    logger.info(f"Filtered out $filteredPoorAlignment%,d reads due to low mapping quality.")
```
nh13: Well not exactly true, since we could have a really high mapq but the mates are mapped to different chromosomes.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -554,31 +562,39 @@ class GroupReadsByUmi
...
     */
-  private def umiForRead(r1: SamRecord, r2: SamRecord): Umi = r1.get[String](this.rawTag) match {
-    case None | Some("") => fail(s"Record '$r1' was missing the raw UMI tag '${this.rawTag}'")
-    case Some(chs)       =>
-      val umi = chs.toUpperCase
+  private def umiForRead(t: Template): Umi = {
+    // Check that all the primary reads have the UMI defined
+    t.primaryReads.foreach { rec =>
+      val umi = rec.get[String](this.rawTag)
+      if (!umi.exists(_.nonEmpty)) fail(s"Record '$rec' was missing the raw UMI tag '${this.rawTag}'")
```
nh13: Why not `require`?  When is something "non-exceptional"?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -554,31 +562,39 @@ class GroupReadsByUmi
...
-          require(umis.length == 2, s"Paired strategy used but umi did not contain 2 segments: $umi")
-
-          if (r1Lower) paired.lowerReadUmiPrefix  + ":" + umis(0) + "-" + paired.higherReadUmiPrefix + ":" + umis(1)
-          else         paired.higherReadUmiPrefix + ":" + umis(0) + "-" + paired.lowerReadUmiPrefix  + ":" + umis(1)
-        case _ =>
-          umi
-      }
+    if (t.r1.isEmpty) fail(s"Read 1 missing for template ${t.name}.")
+
+    val umi = t.r1.get[String](this.rawTag)
```
nh13: what's the type here?  I thought `Template.r1` was an `Option[SamRecord]`

nh13: Is there an implicit here or something?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -117,10 +117,10 @@ object Template {
...
     var r1: Option[SamRecord] = None
     var r2: Option[SamRecord] = None
-    val r1Supp = ListBuffer[SamRecord]()
-    val r2Supp = ListBuffer[SamRecord]()
-    val r1Sec  = ListBuffer[SamRecord]()
-    val r2Sec  = ListBuffer[SamRecord]()
+    var r1Supp: List[SamRecord] = Nil
+    var r2Supp: List[SamRecord] = Nil
+    var r1Sec : List[SamRecord] = Nil
+    var r2Sec : List[SamRecord] = Nil
```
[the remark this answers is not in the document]
tfenne: It is.  I changed GroupReads to use `Template` instead of `(SamRecord, SamRecord)` to deal with the variable number of reads per template.  That slowed it down a tiny bit and this restores _most_ of the performance by not allocating four list-builders per template.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Good call.  I should add a check earlier on and fail if I have paired reads and MQ is absent, then I can just fetch it directly here.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I just spent 5 minutes trying to do this and I feel like the result is worse.  It's actually helpful to apply the filters stepwise so you don't call things that don't make sense on certain reads.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: `require` and `fail` both throw exceptions.  It's just that in FgBioMain we caatch `FailureException` and print a more nicely formatted message without the stack trace.  So I think `fail` is nicer if the error is likely due to user input.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: This is one of the few cases where I wish there were a better way to differentiate what's happening. This is the equivalent of:

```scala
val r1 = t.r1.get
val umi = r1[String](this.rawTag)
```

I'll change the `.get` to `.getOrElse(unreachable("R1 must be present"))` which should make it a little clearer

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -83,13 +83,17 @@ object GroupReadsByUmi {
...
+        val lib     = library(rec)
         val chrom   = rec.refIndex
         val recNeg  = rec.negativeStrand
         val recPos  = if (recNeg) rec.unclippedEnd else rec.unclippedStart
-        val mateNeg = rec.mateNegativeStrand
-        val matePos = if (mateNeg) SAMUtils.getMateUnclippedEnd(rec.asSam) else SAMUtils.getMateUnclippedStart(rec.asSam)
-        val lib = library(rec)
+
+        val (mateNeg, matePos) = if (!rec.paired) (false, Int.MaxValue) else {
+          if (rec.refIndex != rec.mateRefIndex) throw new IllegalArgumentException("Mate on different chrom.")
```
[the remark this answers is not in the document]
nh13: Also, do you want to add the read name?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -431,7 +438,15 @@ class GroupReadsByUmi
 
   private val assigner = strategy.newStrategy(this.edits)
 
-  type ReadPair = (SamRecord, SamRecord)
+  /** Checks that the read's mapq is over a minimum, and if the read is paired, that the mate mapq is also over the min. */
+  private def mapqOk(rec: SamRecord, minMapQ: Int): Boolean = {
+    val mateMqOk = if (rec.unpaired) true else {
+      val mateMq = rec.get[Int](SAMTag.MQ.name())
+      if (mateMq.isEmpty) fail(s"") else mateMq.forall(_ >= minMapQ)
```
nh13: better fail message since no stack trace is given.  Include the read name.   Why not a match statement?
``` 
rec.get[Int](SAMTag.MQ.name()) match {
    case None      => fail(s"Read did not have the mate mapping quality (MQ tag) set: ${rec.name}")
   case Some(mapq) => mapq >= minMapQ
}
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -450,11 +465,11 @@ class GroupReadsByUmi
...
     in.iterator
       .filter(r => !r.secondary && !r.supplementary)
-      .filter(r => (includeNonPfReads || r.pf)            || { filteredNonPf += 1; false })
-      .filter(r => (r.paired && r.mapped && r.mateMapped) || { filteredPoorAlignment += 1; false })
-      .filter(r => (r.refIndex == r.mateRefIndex)         || { filteredPoorAlignment += 1; false })
-      .filter(r => (r.mapq >= this.minMapQ && r.get[Int](SAMTag.MQ.name()).exists(_ >= this.minMapQ)) || { filteredPoorAlignment += 1; false })
-      .filter(r => !r.get[String](rawTag).exists(_.contains('N')) || { filteredNsInUmi += 1; false })
+      .filter(r => (includeNonPfReads || r.pf)                      || { filteredNonPf += 1; false })
+      .filter(r => (r.mapped && (r.unpaired || r.mateMapped))       || { filteredPoorAlignment += 1; false })
+      .filter(r => (r.unpaired || r.refIndex == r.mateRefIndex)     || { filteredPoorAlignment += 1; false })
```
nh13: Did you want to remove this filter?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I was going to leave it for now as you were a little concerned about it.  We can always remove it in a separate PR.

PR #552, opened 2019-12-03
Do not return zero-length alignments

nh13:
This occurs when performing local alignment and setting the minimum
score to zero or smaller.


nh13:
@tfenne I updated the commit for local alignment in `Aligner.findBest`.  This **_must_** return an alignment, but since local alignment has a minimum score of zero by definition, it will return an "empty" alignment (i.e. the cigar is empty).  It probably would be better to return an `Option[Alignment]`, but this is fine for now.

PR #553, opened 2019-12-06
Store UMI base quals as SAM tag(s) with ExtractUmisFromBam

09aaronl:
- Added an optional '-q' flag (molecularIndexTagsQuals) for storing SAM tag(s) with UMI base quals. Validation checks for molecularIndexTagsQuals parallel molecularIndexTags ('-t').

- Also fixed unexpected behavior when a single tag '-t' is specified, but when there are multiple UMIs in either read1 or read2, e.g., '-t XM -r 9M66T 9M8M58T'. Previously, molecularIndexTags1/2 were filled with duplicate values of '-t' based on rs1/2.molecularBarcodeSegments.length. Now, it takes the minimum value of that (which could be 0 if empty) or 1 to avoid duplicate values of '-t'.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
@@ -81,6 +81,7 @@ class ExtractUmisFromBam
   @arg(flag='o', doc = "Output BAM file.")                                     val output: PathToBam,
   @arg(flag='r', doc = "The read structure, one per read in a template.")      val readStructure: Seq[ReadStructure],
   @arg(flag='t', doc = "SAM tag(s) in which to store the molecular indices.", minElements=0) val molecularIndexTags: Seq[String] = Seq.empty,
+  @arg(flag='q', doc = "SAM tag(s) in which to store the molecular indices quals.", minElements=0) val molecularIndexTagsQuals: Seq[String] = Seq.empty,
   @arg(flag='s', doc = "Single tag into which to concatenate all molecular indices.") val singleTag: Option[String] = None,
   @arg(flag='a', doc = "Annotate the read names with the molecular indices. See usage for more details.") val annotateReadNames: Boolean = false,
   @arg(flag='c', doc = "The SAM tag with the position in read to clip adapters (e.g. `XT` as produced by Picard's `MarkIlluminaAdapters`).") val clippingAttribute: Option[String] = None
@@ -110,8 +111,18 @@ class ExtractUmisFromBam
     if (molecularIndexTags.size > 1 && rsMolecularBarcodeSegmentCount != molecularIndexTags.size) {
       invalid("Either a single SAM tag, or the same # of SAM tags as molecular indices in the read structure,  must be given.")
     }
-  }
 
+    // Same validation for molecularIndexTagsQuals
+    if (!molecularIndexTagsQuals.isEmpty)
+    {
+      molecularIndexTagsQuals.foreach(tag => if (tag.length != 2) invalid("SAM tags must be of length two: " + tag))
+      // ensure we either have one tag, or we have the same # of tags as molecular indices in the read structure.
+      if (molecularIndexTagsQuals.size > 1 && rsMolecularBarcodeSegmentCount != molecularIndexTagsQuals.size) {
+        invalid("Either a single SAM tag, or the same # of SAM tags as molecular indices in the read structure,  must be given.")
+      }
+    }
+  }
+  
   // Verify that if a single tag was specified that it is valid and not also contained in the per-index tags
   singleTag.foreach { tag =>
     if (tag.length != 2) invalid(s"All tags must be two characters: ${tag}")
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -110,8 +111,18 @@ class ExtractUmisFromBam
...
     }
-  }
 
+    // Same validation for molecularIndexTagsQuals
+    if (!molecularIndexTagsQuals.isEmpty)
+    {
+      molecularIndexTagsQuals.foreach(tag => if (tag.length != 2) invalid("SAM tags must be of length two: " + tag))
+      // ensure we either have one tag, or we have the same # of tags as molecular indices in the read structure.
+      if (molecularIndexTagsQuals.size > 1 && rsMolecularBarcodeSegmentCount != molecularIndexTagsQuals.size) {
+        invalid("Either a single SAM tag, or the same # of SAM tags as molecular indices in the read structure,  must be given.")
```
nh13, line 121: update error message to specify its the index _qualities_

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -121,15 +132,34 @@ class ExtractUmisFromBam
...
+  // Same validation for molecularIndexTagsQuals
+  private val (molecularIndexTagsQuals1, molecularIndexTagsQuals2) = molecularIndexTagsQuals match {
+    case Seq(one) =>
+      val t1 = Seq.fill(rs1.molecularBarcodeSegments.length.min(1))(one)
+      val t2 = Seq.fill(rs2.map(_.molecularBarcodeSegments.length).getOrElse(0).min(1))(one)
+      (t1, t2)
+    case tags =>
+      // Validate that the # of molecular barcodes in the read structure matches the tags
+      if(tags.length > 0) 
+      {
```
nh13, line 153: ditto about style, I assume you'll fix up everywhere

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -121,15 +132,34 @@ class ExtractUmisFromBam
...
       validate(tags.length == total, s"Read structures contains $total molecular barcode segments, but ${tags.length} tags provided.")
       (tags.take(rs1.molecularBarcodeSegments.length), tags.drop(rs1.molecularBarcodeSegments.length))
   }
+  // Same validation for molecularIndexTagsQuals
+  private val (molecularIndexTagsQuals1, molecularIndexTagsQuals2) = molecularIndexTagsQuals match {
+    case Seq(one) =>
+      val t1 = Seq.fill(rs1.molecularBarcodeSegments.length.min(1))(one)
+      val t2 = Seq.fill(rs2.map(_.molecularBarcodeSegments.length).getOrElse(0).min(1))(one)
+      (t1, t2)
+    case tags =>
```
nh13, line 150: might be nicer:

```scala
  private val (molecularIndexTagsQuals1, molecularIndexTagsQuals2) = molecularIndexTagsQuals match {
    case Seq(one) =>
      val t1 = Seq.fill(rs1.molecularBarcodeSegments.length.min(1))(one)
      val t2 = Seq.fill(rs2.map(_.molecularBarcodeSegments.length).getOrElse(0).min(1))(one)
      (t1, t2)
    case tags if tags.length == 0 => (Seq.empty, Seq.empty)
    case tags =>
        ...
```

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -121,15 +132,34 @@ class ExtractUmisFromBam
   // Validate the molecular index tags against the read structures and then split them out
   private val (molecularIndexTags1, molecularIndexTags2) = molecularIndexTags match {
     case Seq(one) =>
-      val t1 = Seq.fill(rs1.molecularBarcodeSegments.length)(one)
-      val t2 = Seq.fill(rs2.map(_.molecularBarcodeSegments.length).getOrElse(0))(one)
+      val t1 = Seq.fill(rs1.molecularBarcodeSegments.length.min(1))(one)
+      val t2 = Seq.fill(rs2.map(_.molecularBarcodeSegments.length).getOrElse(0).min(1))(one)
```
nh13, line 136: why the `min(1)` change here and below

src/main/scala/com/fulcrumgenomics/umi/ExtractUmisFromBam.scala
```diff
@@ -121,15 +132,34 @@ class ExtractUmisFromBam
...
+      {
+        val total = rs1.molecularBarcodeSegments.length + rs2.map(_.molecularBarcodeSegments.length).getOrElse(0)
+        validate(tags.length == total, s"Read structures contains $total molecular barcode segments, but ${tags.length} tags provided.")
+        (tags.take(rs1.molecularBarcodeSegments.length), tags.drop(rs1.molecularBarcodeSegments.length))
+      }
+      else
+      {
+        (Seq.empty, Seq.empty)
+      }
+  }
```
nh13, line 162: do you want to validate that the tags for the index bases does not overlap the tags for the index qualities?

nh13:
@09aaronl thank-you for making a PR BTW!

nh13:
@09aaronl also, the code doesn't compile (see the build failures)

PR #560, opened 2020-01-07
Add the `primary` keyword for accessing a SamRecord's secondary flag

rnazad:
This minor change introduces the ability to get and set the secondary bit flag of a `SamRecord` using the `primary` keyword.

For example the following two statements would be equivalent:

```scala
rec.primary = true
```

and 

```scala
rec.secondary = false
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -124,8 +124,8 @@ trait SamRecord {
 
   @inline final def secondary: Boolean = isSecondaryAlignment
   @inline final def secondary_=(secondary: Boolean):Unit = setSecondaryAlignment(secondary)
-  @inline final def primary: Boolean = !isSecondaryAlignment
-  @inline final def primary_=(primary: Boolean):Unit = setSecondaryAlignment(!primary)
+  @inline final def primary: Boolean = !this.secondary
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -124,6 +124,8 @@ trait SamRecord {
 
   @inline final def secondary: Boolean = isSecondaryAlignment
   @inline final def secondary_=(secondary: Boolean):Unit = setSecondaryAlignment(secondary)
+  @inline final def primary: Boolean = !isSecondaryAlignment
+  @inline final def primary_=(primary: Boolean):Unit = setSecondaryAlignment(!primary)
```
nh13:
```suggestion
  @inline final def primary_=(primary: Boolean):Unit = this.secondary(!primary)
```

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -124,6 +124,8 @@ trait SamRecord {
 
   @inline final def secondary: Boolean = isSecondaryAlignment
   @inline final def secondary_=(secondary: Boolean):Unit = setSecondaryAlignment(secondary)
+  @inline final def primary: Boolean = !isSecondaryAlignment
```
nh13:
```suggestion
  @inline final def primary: Boolean = !this.secondary
```

PR #564, opened 2020-02-12
Two fixes to RefFlatSource:

tfenne:
- Exons were not being put into transcripts in transcription order which is required (but not verified in Transcript)
- Gene start/end were being taken as the min of the transcript starts/ends, but for end it should be max


nh13:
Perhaps it’s useful to link to the UCSC schema: http://api.genome.ucsc.edu/list/schema?genome=hg38;track=refFlat

I am still unsure if the Exon ordering as it says:

```
{ "name": "exonStarts", "sqlType": "longblob", "jsonType": "string", "description": "Exon start positions (or end positions for minus strand item)"} , { "name": "exonEnds", "sqlType": "longblob", "jsonType": "string", "description": "Exon end positions (or start positions for minus strand item)"} 
```

It’s unclear that the start posturing means start in transcript order but I can believe it.

tfenne:
My experience is that the UCSC exports are always in genomic order, but that they flip the starts/ends based on transcript strand which is just unhelpful.  Hence my decision to always sort into the desired order here.

PR #568, opened 2020-02-19
Changed the gene annotation case classes and the RefFlatSource to res…

tfenne:
…olve two issues:

1. RefFlatSource would drop transcript mappings if there were mappings to > 1 chrom or > 1 strand for a given gene
2. RefFlatSource would combine transcript mappings at wildly different locations on the same chrom/strand

This necessitated changing the case classes a bit to insert a GeneLocus between Gene and Transcript, as a collection of transcripts at a common locus.

Also performed some other tidying up while I was in there.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
@@ -25,32 +25,92 @@
 
 package com.fulcrumgenomics.util
 
-import htsjdk.samtools.util.{CoordMath, Interval}
+import htsjdk.samtools.util.{CoordMath, Locatable}
+import com.fulcrumgenomics.FgBioDef._
 
 /** Stores classes useful for storing annotation information for genes and their transcripts and exons. */
 object GeneAnnotations {
+  /**
+    * A gene with a collection of gene loci.
+    */
+  case class Gene(name: String, loci: Seq[GeneLocus]) extends Iterable[Transcript] {
+    def iterator: Iterator[Transcript] = this.loci.iterator.flatten
 
-  /** A gene with zero or more transcripts, that all are on the same transcription strand. */
-  case class Gene(contig: String, start: Int, end: Int, negativeStrand: Boolean, name: String, transcripts: Seq[Transcript])
-    extends Interval(contig, start, end, negativeStrand, name) with Iterable[Transcript] {
-    def iterator: Iterator[Transcript] = transcripts.iterator
+    /** Returns the locus with the maximum transcript mappings. */
+    def primaryLocus: GeneLocus = loci.maxBy(_.size)
   }
 
-  /** A transcript associated with a given gene (which stores the transcription strand).  Contains zero or more exons.
-    * The exons should be given in transcript order. */
-  case class Transcript(name: String, start: Int, end: Int, cdsStart: Int, cdsEnd: Int, exons: Seq[Exon]) {
+
+  /**
+    * A gene locus representing a set of transcripts that are mapped to the same location on the genome.
+    *
+    * @param transcripts the set of transcripts at the locus
+    */
+  case class GeneLocus(transcripts: Seq[Transcript]) extends Locatable with Iterable[Transcript] {
+    require(transcripts.nonEmpty, "GeneLocus cannot be generated from an empty set of transcripts.")
+    require(transcripts.forall(t => t.chrom == transcripts.head.chrom))
+    require(transcripts.forall(t => t.negativeStrand == transcripts.head.negativeStrand))
+
+    val chrom: String = transcripts.head.chrom
+    val start: Int = transcripts.minBy(_.start).start
+    val end:   Int = transcripts.maxBy(_.end).end
+    val negativeStrand: Boolean = transcripts.head.negativeStrand
+    def positiveStrand: Boolean = !negativeStrand
+
+    // Locatable implementation
+    override def getContig: String = chrom
+    override def getStart: Int = start
+    override def getEnd: Int = end
+
+    override def iterator: Iterator[Transcript] = this.transcripts.iterator
+  }
+
+
+  /**
+    * A transcript [mapping] associated with a given gene locus.  Contains zero or more exons.
+    * The exons should be given in the order of transcription.
+    */
+  case class Transcript(name: String,
+                        chrom: String,
+                        start: Int,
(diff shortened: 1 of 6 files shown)


src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -25,32 +25,92 @@
...
+    override def iterator: Iterator[Transcript] = this.transcripts.iterator
+  }
+
+
+  /**
+    * A transcript [mapping] associated with a given gene locus.  Contains zero or more exons.
+    * The exons should be given in the order of transcription.
+    */
+  case class Transcript(name: String,
+                        chrom: String,
```
tfenne: Note: I pushed `chrom` and `negativeStrand` down here for a few reasons:
1. It's just convenient to have this info on transcript
2. Most sources of annotation are per transcript mapping, so I feel like it makes sense to have all the details here and then also roll it up higher where we need it
3. It allows `Transcript` to implement Locatable which seems useful

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -25,32 +25,92 @@
...
+                        start: Int,
+                        end: Int,
+                        cdsStart: Option[Int],
```
tfenne: Made these options as there's a lot of non-coding transcripts (and "genes"), and then in the RefFlatSource I detect where there is no defined CDS and set them to None.

src/main/scala/com/fulcrumgenomics/util/RefFlatSource.scala
```diff
@@ -122,13 +123,13 @@ class RefFlatSource private(lines: Iterator[String],
...
     var numTranscripts          = 0
     val _genes = new DelimitedDataParser(lines=_lines, delimiter='\t').flatMap { row =>
-      val geneName       = row[String]("geneName")
-      val transcriptName = row[String]("name")
-      val contig         = row[String]("chrom")
-      val strand         = row[String]("strand")
-      val exonCount      = row[Int]("exonCount")
-      val exonStarts     = row[String]("exonStarts").split(',').filter(_.nonEmpty).map(_.toInt)
-      val exonEnds       = row[String]("exonEnds").split(',').filter(_.nonEmpty).map(_.toInt)
+      val geneName       = row.string("geneName")
```
tfenne: This is pure optimization.  Using the `string()` method is a lot faster when you're mostly just dealing with the columns as strings.

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -25,32 +25,92 @@
...
+import htsjdk.samtools.util.{CoordMath, Locatable}
+import com.fulcrumgenomics.FgBioDef._
 
 /** Stores classes useful for storing annotation information for genes and their transcripts and exons. */
 object GeneAnnotations {
+  /**
+    * A gene with a collection of gene loci.
+    */
+  case class Gene(name: String, loci: Seq[GeneLocus]) extends Iterable[Transcript] {
+    def iterator: Iterator[Transcript] = this.loci.iterator.flatten
```
nh13: this is cute, and since I read this first, I didn't understand that `GeneLocus` implemented `Iterable`.  Perhaps a comment is useful for any future person reading the code?

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -25,32 +25,92 @@
...
+    val chrom: String = transcripts.head.chrom
+    val start: Int = transcripts.minBy(_.start).start
+    val end:   Int = transcripts.maxBy(_.end).end
+    val negativeStrand: Boolean = transcripts.head.negativeStrand
+    def positiveStrand: Boolean = !negativeStrand
+
+    // Locatable implementation
+    override def getContig: String = chrom
+    override def getStart: Int = start
+    override def getEnd: Int = end
```
nh13: It's a new decade so we're not aligning the `=`s anymore?

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -25,32 +25,92 @@
...
+    * The exons should be given in the order of transcription.
+    */
+  case class Transcript(name: String,
+                        chrom: String,
+                        start: Int,
+                        end: Int,
+                        cdsStart: Option[Int],
+                        cdsEnd: Option[Int],
+                        negativeStrand: Boolean,
+                        exons: Seq[Exon]) extends Locatable {
```
nh13: I'd set defaults for `exons`, `cdsStart`, and `cdsEnd`.

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -25,32 +25,92 @@
...
       val exonsOverlap = genomicOrder.sliding(2).map { e => (e.head, e.last) }.exists { case (e1, e2) => CoordMath.overlaps(e1.start, e1.end, e2.start, e2.end) }
       require(!exonsOverlap, s"exons overlap for transcript: $name")
     }
+
+    require(cdsStart.isDefined == cdsEnd.isDefined, s"cdsStart and cdsEnd must either both or neither be defined on tx: $name.")
+
+    // Locatable implementation
+    override def getContig: String = chrom
+    override def getStart: Int = start
+    override def getEnd: Int = end
```
nh13: This should just be a trait somewhere that we can mix in, and then we can standardize on the real names for these.

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
tfenne: I've added `= None` for cdsStart and cdsEnd, but I don't think there is a sensible default for exons.  It doesn't make sense to have a transcript that has no exons (i.e. no transcribed sequence).

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
tfenne: I agree .. and I think I've done that elsewhere.  I'll dig it up and open up a separate PR in the near future.

PR #569, opened 2020-03-05
Basic implementation of an adapter for locatable.

tfenne:
This is a draft PR containing a class I wrote elsewhere.  Intended for discussion (not [yet] nit-picking).  The goal being to standardize how we implement classes in scala that have genomic locations while also adapting them for use in HTSJDK's APIs.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/LocatableAdapter.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/LocatableAdapter.scala
@@ -1,0 +1,59 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2020 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/util/LocatableAdapter.scala
```diff
@@ -0,0 +1,59 @@
...
+import htsjdk.samtools.util.{CoordMath, Locatable}
+
+/**
+  * Trait that can be added to any class that defines chrom/start/end to implement Locatable
+  */
+trait LocatableAdapter[A <: LocatableAdapter[A]] extends Locatable with Ordered[A] {
+  def chrom: String
+  def start: Int
+  def end: Int
+  def name: String
```
tfenne: Should this be here?  Should it have a default `= ""` or perhaps have type `Option[String] = None`?  Or should we have a `NamedLocatable` sub-trait?

src/main/scala/com/fulcrumgenomics/util/LocatableAdapter.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  // Implementations of Locatable interface methods
+  final override def getContig: String = chrom
+  final override def getStart: Int     = start
+  final override def getEnd: Int       = end
+
+  final def length: Int = getLengthOnReference
+
+  /** Calculates the overlap between two locatables assuming that they do in fact overlap. */
+  def overlap(other: Locatable): Int = CoordMath.getOverlap(start, end, other.getStart, other.getEnd)
```
tfenne: This should probably be bullet-proofed to check for overlap first or return 0 if they don't overlap.

nh13 (commented):
1. How about an `encloses` method?
2. It would also be nice to have a scala version of `OverlapDetector`, so we have the nice `apply`/`get`/`update` methods.   I find the `addLhs` method so ugly.
3. How about a `toInterval` method?
4. How about a scala version of `Interval`?   `IntervalList`?

clintval:
I like this! I'm using [`GenomicSpan`](https://github.com/fulcrumgenomics/fgbio/blob/3f90c873aa91ee078be020e71ef5b89283b903a9/src/main/scala/com/fulcrumgenomics/util/GenomicSpan.scala) in a few places, so would it be possible to reconcile before merging?

PR #573, opened 2020-03-25
Added a source class for parsing and reading gene annoations from a N…

tfenne:
…CBI RefSeq .gff file.


src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
```diff
@@ -0,0 +1,266 @@
...
+
+  {
+    val iter = _bufferedLines.filter(l => !l.startsWith("#")).map { line =>
+      val fs = line.split('\t')
+      GffRecord(fs(0), fs(1), fs(2), int(fs(3)), int(fs(4)), fs(6) == "-", if (fs(7) == ".") -1 else int(fs(7)), fs(8))
+    }.bufferBetter
+
+    while (iter.nonEmpty) {
+      // Scan for the next chromosome record, discard any other records along the way
+      iter.find(rec => rec.kind == "region" && rec.accession.startsWith("NC_")) match {
```
jacarey: `startsWith("NC_")` is very specific. the gff spec says:
`a standard chromosome name or an Ensembl identifier such as a scaffold ID, without any additional content such as species or assembly`

would it make more sense to grab on record. parse the accession name for the region and then use that in a takeWhile instead of constructing the inner iter below?

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
```diff
@@ -0,0 +1,266 @@
...
+
+    Transcript(txRec("transcript_id"), chrom, txRec.start, txRec.end, cdsStart, cdsEnd, txRec.negative, exons)
+  }
+
+  /**
+    * Looks up a gene by name. Only recognizes the gene names from the GFF without any aliasing.
+    * @param name the name of the gene
+    * @return either Some(gene) if the gene exists or None otherwise.
+    */
+  def get(name: String): Option[Gene] = this.genes.get(name)
```
jacarey: getGene instead of just get?

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
```diff
@@ -0,0 +1,266 @@
...
+  /**
+    * Parses a single gene and all of it's sub-features from the iterator. After invoking this method
+    * the `iter` passed in will be positioned such that the next record is the record _after_ all
+    * children of the gene record (and their children and so on).
+    *
+    * @return a [[Some(Gene)]] if one of more transcripts are parsed, or [[None]] if no transcripts are
+    *         parsed. A [[None]] is primarily returned if the gene has only experimentally determined
+    *         transcript accessions (X?_*), and [[includeXs]] is false.
+    * */
+  private def parseGene(chrom: String, geneRec: GffRecord, iter: BetterBufferedIterator[GffRecord]): Option[Gene] = {
```
jacarey: I don't love all this passing in (and side-effecting) iterators even though they are private methods. Why not just accumulate records and pass around Seqs? It seems like if the records for a region are sequential then it should be easy to break all the records into Seqs and then process them but maybe I'm missing something.

tfenne: I actually like this way of organizing things.  It's a little complicated because of the fact that there's basically hierarchical data represented in a flat structure.  So each function parses a level of the structure.  I say "it's tricky" because there is no easy way to say, e.g. "consume all the records for this gene".  Instead you have to consume the gene records, then for each transcript read the transcript and find all the records that have the transcript as parent, and then repeat for each transcript until you hit the last transcript for the gene.

I'm going to leave it this way for now because it makes sense to me and is working and I'm not convinced the alternative would be clearer.

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
tfenne: Ok ... I took a shot at allowing mapping of the other accessions using aliases in the sequence dictionary.  It means constructing a sequence dictionary that has the accessions as aliases, but is better I think.

PR #581, opened 2020-04-27
example API for sequence dictionaries

nh13:


src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+  require(length > 0, s"Length must be > 0 for '$name'")
+
+  /** All names, including aliases */
+  @inline final def allNames: Seq[String] = name +: aliases
+  @inline final def apply[T](key: String): T = this.attributes(key).asInstanceOf[T]
+  @inline final def get[T](key: String): Option[T] = this.attributes.get(key).map(_.asInstanceOf[T])
+  @inline final def contains(key: String): Boolean = this.attributes.contains(key)
+
+  @inline final def isAlternate: Boolean = this.contains("AH")
+  def alternate: Option[AlternateLocus] = {
```
nh13, line 56: This could be an `Interval` object too.

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+    this.apply[String]("AH") match {
```
nh13, line 57: Could make our own list of reserved tags

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+      case "*" => None
+      case string =>
+        val Array(refName, rest) = string.split(':')
+        val Array(start, end) = rest.split('-')
```
nh13, line 61: why is there no utility class in fgbio that handles this?

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+  * @param attributes attributes of this sequence
+  */
+case class SequenceInfo(name: String,
+                        length: Int,
+                        aliases: Seq[String] = Seq.empty,
+                        attributes: Map[String, Any] = Map.empty
+                       ) {
+
+  allNames.foreach { name => SAMSequenceRecord.validateSequenceName(name) }
+  require(length > 0, s"Length must be > 0 for '$name'")
```
nh13, line 47: Note: could relax to `>= 0` to allow for missing lengths (or make this an `Option`, but that's so ugly!)

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
tfenne: I think it's nice to have it properly typed - even if it extends some interval/region base.

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  /** All names, including aliases */
+  @inline final def allNames: Seq[String] = name +: aliases
+  @inline final def apply[T](key: String): T = this.attributes(key).asInstanceOf[T]
```
tfenne, line 51: I like these functions ... but does that imply we're going to do type-aware parsing of attributes when we read in dictionaries?  Otherwise maybe these should just be strings?

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
tfenne: Also perhaps this should be a lazy val?

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
tfenne: Seize the day!

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+        val Array(start, end) = rest.split('-')
+        val locus = refName match {
+          case "=" => AlternateLocus(refName=this.name, start=start.toInt, end=end.toInt)
+          case _   => AlternateLocus(refName=refName, start=start.toInt, end=end.toInt)
+        }
+        Some(locus)
+    }
+  }
+
+  @inline final def m5: String = this.apply(SAMSequenceRecord.MD5_TAG)
```
tfenne, line 70: `md5`?

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+/** Stores information about a single Sequence (ex. chromosome, contig)
+  *
+  * @param name the primary name of the sequence
+  * @param length the length of the sequence, or zero if unknown
+  * @param aliases the list of aliases for this sequence
+  * @param attributes attributes of this sequence
+  */
+case class SequenceInfo(name: String,
+                        length: Int,
+                        aliases: Seq[String] = Seq.empty,
```
tfenne, line 42: I'm curious why aliases get elevated out of the attributes, but other extended things don't.  What happens if someone provides `aliases=Seq("foo", "bar"), attributes=Map("AN" -> "notfoo"))`

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
tfenne, line 70: I would think anything beyond name and length should be `Option[T]` unless there is a sensible default value to sub in.

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+  @inline final def description: String = this.apply(SAMSequenceRecord.DESCRIPTION_TAG)
+  @inline final def topology: Topology = Topology.values.find(_.name == this.apply("TP")).get
+
+  /** Returns true if the the sequences share a common reference name (including aliases), have the same length, and
+    * the same MD5 if both have MD5s. */
+  def same(that: SequenceInfo): Boolean = {
+    val thisM5: Option[BigInt] = if (this.contains(SAMSequenceRecord.MD5_TAG)) Some(this.m5Int) else None
+    val thatM5: Option[BigInt] = if (that.contains(SAMSequenceRecord.MD5_TAG)) Some(that.m5Int) else None
+
+    this.allNames.exists(that.allNames.contains) &&
```
tfenne, line 85: Maybe re-order as `if (this.length != that.length) false else { ... }` for efficiency?

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+      this.length == that.length &&
+      thisM5 == thatM5
+  }
+
+}
+
+/** Contains an ordered collection of sequences. */
+class SequenceInfos(infos: IndexedSeq[SequenceInfo]) extends Iterable[SequenceInfo] {
```
tfenne, line 93: How about `SequenceDictionary` and `SequenceMetadata`?

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  private val mapping: Map[String, SequenceInfo] = infos.flatMap { info =>
+    info.allNames.map { name => name -> info }
+  }.toMap
+
+  def apply(name: String): SequenceInfo = this.mapping(name)
+  def get(name: String): Option[SequenceInfo] = this.mapping.get(name)
+  def contains(name: String): Boolean = this.mapping.contains(name)
+  override def iterator: Iterator[SequenceInfo] = this.infos.iterator
+}
```
tfenne, line 112: I know `infos` is public, but I think we'd still want a `def apply(index: int) = this.info(int)`

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
```diff
@@ -0,0 +1,171 @@
...
+      thisM5 == thatM5
+  }
+
+}
+
+/** Contains an ordered collection of sequences. */
+class SequenceInfos(infos: IndexedSeq[SequenceInfo]) extends Iterable[SequenceInfo] {
+  // Ensure that all names, even aliases, are unique.
+  {
+    val duplicateNames: String = infos.flatMap(_.allNames)
```
tfenne, line 96: This might suck for very large sequence dictionaries.

tfenne: Agreed.  Perhaps the compromise would be to do it more procedurally (i.e. create a set, add names to the set, catch when you add a name that's already there) rather than by using `groupBy` etc.?

src/main/scala/com/fulcrumgenomics/fasta/SequenceInfos.scala
tfenne: I think this is tricky though because the header/dictionary format we're parsing doesn't have type tags on the attributes (e.g. doesn't allow `xi:i:273`) so when you're parsing from a file there is no way to know that "273" should be an `Int` and not a `String`.  And so if you allow for non-String types then you end up with problems like:

1. Do you get different types if constructed in memory (e.g. I put in an `Int` during construction) vs. when reading it back from disk?
2. Do you try and be "smart" when reading from disk and parse the attributes as Int/Float/String/List etc.?  But then you get inconsistent results.  What if I have one record with `xi:123` and another with `xi:foo`?  If I call `sequence[Int]("xi")` it will fail on one and if I call `sequence[String]("xi")` on the other it will fail because it is only _casting_ not doing type conversions.

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,180 @@
...
+  *
+  * @param name the primary name of the sequence
+  * @param length the length of the sequence, or zero if unknown
+  * @param attributes attributes of this sequence
+  */
+case class SequenceMetadata(name: String,
+                            length: Int,
+                            attributes: Map[String, Any] = Map.empty) {
+
+  allNames.foreach { name => SAMSequenceRecord.validateSequenceName(name) }
```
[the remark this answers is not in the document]
nh13: Will you ever run validate?  Will you remember to use it?  Probably not, so let's keep it in for now

nh13:
@tfenne and @sstadick finally got this in good shape with tests and incorporated your feedback. Care to take one more review and then I'll merge to complete the target branch?

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -54,4 +54,21 @@ object FgBioDef extends CommonsDef {
       case _                           => PathUtil.basename(input, trimExt=true).toString
     }
   }
+
+  /** Parses a genomic range of the form `<chr>:<start>-<end>`. */
+  def parseRange(range: String): (String, Int, Int) = {
```
tfenne, line 59: How about having this be in a `case class GenomicRange(chrom, start, end)`?

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,206 @@
...
+  * @param index the index in the sequence dictionary
+  * @param attributes attributes of this sequence
+  */
+case class SequenceMetadata(name: String,
+                            length: Int,
+                            index: Int = 0,
+                            attributes: Map[String, String] = Map.empty) {
+  allNames.foreach { name => SAMSequenceRecord.validateSequenceName(name) }
+  require(length >= 0, s"Length must be >= 0 for '$name'")
+  require(attributes.keys.forall(_ != Keys.Name),
```
tfenne, line 64: Why not ` require(!attributes.contains(Keys.Name), "...") and the same for `Length` to avoid the iteration?

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,206 @@
...
+    f"`${Keys.Name}` should not given in the list of attributes")
+  require(attributes.keys.forall(_ != Keys.Length),
+    s"`${Keys.Length}` should not given in the list of attributes")
+
+  @inline final def apply(key: String): String = this.attributes(key)
+  @inline final def get(key: String): Option[String] = this.attributes.get(key)
```
tfenne, line 70: Maybe add `getOrElse(key: String, default: String)`?

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,206 @@
...
+  @inline final def contains(key: String): Boolean = this.attributes.contains(key)
+  @inline final def aliases: Seq[String] = this.get(Keys.Aliases).map(_.split(',').toSeq).getOrElse(Seq.empty[String])
```
tfenne, line 72: For more complicated things like this I _think_ I'd suggest making it a `lazy val`, what do you think?

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,206 @@
...
+    * the same MD5 if both have MD5s. */
+  def sameAs(that: SequenceMetadata): Boolean = {
+    val md5Match = (this.md5Int, that.md5Int) match {
+      case (Some(thisMd5Int), Some(thatMd5Int)) => thisMd5Int == thatMd5Int
+      case _                                    => true
+    }
+
+    this.length == that.length &&
+      (this.name == that.name || this.allNames.exists(that.allNames.contains)) &&
+      md5Match
```
tfenne, line 108: Why not put this first since you've already computed it?

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,206 @@
...
+}
+
+/** Stores information about a single Sequence (ex. chromosome, contig)
+  *
+  * @param name the primary name of the sequence
+  * @param length the length of the sequence, or zero if unknown
+  * @param index the index in the sequence dictionary
+  * @param attributes attributes of this sequence
+  */
+case class SequenceMetadata(name: String,
```
tfenne, line 58: Perhaps this is for a follow-up but would be nice to have an apply() in the object that takes in all the "known" attributes (defaulting most to `None`) and constructs a valid SequenceMetadata.

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,206 @@
...
+
+/** Contains an ordered collection of sequences. */
+case class SequenceDictionary(infos: IndexedSeq[SequenceMetadata]) extends Iterable[SequenceMetadata] {
+
+  private val mapping: Map[String, SequenceMetadata] = {
+    // validates that the same name is not present twice across allNames (so includes aliases)
+    val names: mutable.Set[String] = mutable.HashSet[String]()
+    infos.iterator.zipWithIndex.flatMap { case (info, index) =>
+      require(info.index == index, s"Infos must be given with index set correctly. See ${index}th with name: $names")
+      info.allNames.map { name =>
```
tfenne, line 130: Thanks - I like this a lot better than the previous `groupBy`.

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,206 @@
...
+      }
+    }.toMap
+  }
+
+  def apply(name: String): SequenceMetadata = this.mapping(name)
+  def get(name: String): Option[SequenceMetadata] = this.mapping.get(name)
+  def contains(name: String): Boolean = this.mapping.contains(name)
+  def apply(index: Int): SequenceMetadata = this.infos(index)
+  override def iterator: Iterator[SequenceMetadata] = this.infos.iterator
+  def length: Int = this.infos.length
```
tfenne, line 143: A thought for our future selves.... this would be a nice place to add methods like `genomeSize` that gives you the total of all the sequences.

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,206 @@
...
+}
+
+
+/** Contains useful converters to and from HTSJDK objects. */
+object Converters {
+
+  /** Converter from a [[SequenceMetadata]] to a [[SAMSequenceRecord]] */
+  implicit class ToSAMSequenceRecord(info: SequenceMetadata) {
```
tfenne, line 151: I'm conflicted on naming ... the target class uses `SAM` but our convention is to use `Sam`.

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -0,0 +1,206 @@
...
+object Topology extends FgBioEnum[Topology] {
+  def values: scala.collection.immutable.IndexedSeq[Topology] = findValues
+  /** The sequence is linear. */
+  case object Linear extends Topology
+  /** The sequence is circular. */
+  case object Circular extends Topology
+}
+
+/** Stores information about the coordinates of the alternate locus  */
+case class AlternateLocus(refName: String, start: Int, end: Int)
```
tfenne, line 206: Could you please add scaladoc about how to interpret these?

PR #582, opened 2020-04-27
Added support for Iso8610 dates in RunInfo.xml as Illumina has starte…

tfenne:
…d using that now.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/illumina/RunInfo.scala
+++ b/src/main/scala/com/fulcrumgenomics/illumina/RunInfo.scala
@@ -30,6 +30,7 @@ import com.fulcrumgenomics.FgBioDef.FilePath
 import com.fulcrumgenomics.bam.Template
 import com.fulcrumgenomics.util.SegmentType.SampleBarcode
 import com.fulcrumgenomics.util.{Metric, ReadSegment, ReadStructure, SegmentType}
+import htsjdk.samtools.util.DateParser.InvalidDateException
 import htsjdk.samtools.util.Iso8601Date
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/illumina/RunInfo.scala
```diff
@@ -54,6 +55,14 @@ case class RunInfo
 ) extends Metric
 
 object RunInfo {
+  /** A set of functions to parse dates in various formats. */
+  val DateParsers: Seq[String => Option[Iso8601Date]] = Seq(
+    d => {if (d.length == 6) Some(new Iso8601Date("20" + d.substring(0,2) + "-" + d.substring(2,4) + "-" + d.substring(4))) else None },
+    d => {if (d.length == 8) Some(new Iso8601Date(d.substring(0,4) + "-" + d.substring(4,6) + "-" + d.substring(6))) else None},
+    d => {try {Some(new Iso8601Date(new SimpleDateFormat("M/d/yyyy h:mm:ss a").parse(d)))} catch { case _: ParseException => None }},
+    d => {try {Some(new Iso8601Date(d))} catch {case _: InvalidDateException => None}}
```
nh13, line 63: Not worth it, but it'd be nicer if this was a method that just returned a `Try[Iso8601Date ]`

tfenne: Yeah, it's a mix - because then I think it would require throwing exceptions in the first two cases, which is not great either.

PR #584, opened 2020-05-15
Assigned-Molecules should only output an alias for Assembled-Molecules

nh13:
Also, add an option to skip contigs that are missing aliases.

If you take a look at this [NCBI assembly report](https://ftp.ncbi.nlm.nih.gov/genomes/all/GCF/000/001/405/GCF_000001405.13_GRCh37/GCF_000001405.13_GRCh37_assembly_report.txt) and the `Assigned-Molecule` column,  the molecules with `assembled-molecule` sequencing role have sensical value.  The sequencing roles are either missing (ex. `HSCHRUN_RANDOM_CTG28` has `na`) or refer to the `Assigned-Molecule` to that of a `assembled-molecule` (ex `HSCHR1_RANDOM_CTG5 ` molecule has a sequencing role of `unlocalized-scaffold` and an assigned molecule as `1`).

```
# Sequence-Name       Sequence-Role         Assigned-Molecule  Assigned-Molecule-Location/Type  GenBank-Accn  Relationship  RefSeq-Accn   Assembly-Unit     Sequence-Length  UCSC-style-name
1                     assembled-molecule    1                  Chromosome                       CM000663.1    =             NC_000001.10  Primary Assembly  249250621        chr1
...
HSCHR1_RANDOM_CTG5    unlocalized-scaffold  1                  Chromosome                       GL000191.1    =             NT_113878.1   Primary Assembly  106433           chr1_gl000191_random
...
HSCHRUN_RANDOM_CTG28  unplaced-scaffold     na                 na                               GL000235.1    =             NT_167229.1   Primary Assembly  34474            chrUn_gl000235
...

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
@@ -42,11 +42,15 @@ sealed trait AssemblyReportColumn extends EnumEntry {
   def key: String
   /** The tag to store in the SAM header */
   def tag: String
+  /** True if this columns contains a valid molecule alias, false otherwise. */
+  def alias: Boolean = true
 }
 
 /** Enum to represent columns in a NCBI assembly report. */
 object AssemblyReportColumn extends FgBioEnum[AssemblyReportColumn] {
-  def values: IndexedSeq[AssemblyReportColumn] = findValues
+  // Developer note: we only return those that are "aliases" so that sequence role and sequence length are not
+  // valid options on the command line.
+  def values: IndexedSeq[AssemblyReportColumn] = findValues.filter(_.alias)
 
   /** Allows the column to be build from the Enum name or the actual column name. */
   override def apply(str: String): AssemblyReportColumn = {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala [hunk 1]
```diff
@@ -137,42 +141,47 @@ class CollectAlternateContigNames
...
+      val role   = SequenceRole(dict(AssemblyReportColumn.SequenceRoleKey))
+      val alts   = this.alternates.flatMap {
+        // Developer note: the value of assigned-molecule sequence column is one of the values in the same column for
+        // an assembled-molecule, so only include them for assembled-molecules.  Otherwise, we will have the same alias
+        // across two or molecules.  Perhaps this is better represented as an alternate locus, but that is not
+        // implemented here.
+        case AssemblyReportColumn.AssignedMolecule if role != SequenceRole.AssembledMolecule => None
+        case alt: AssemblyReportColumn =>
+          dict(alt.key) match {
+            case alternate if alternate == AssemblyReportColumn.MissingValueKey =>
```
tfenne, line 160: Can't this just be `case AssemblyReportColumn.MissingValueKey`?

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -54,19 +54,19 @@ object AssemblyReportColumn extends FgBioEnum[AssemblyReportColumn] {
...
   case object SequenceName     extends AssemblyReportColumn { val key: String = "Sequence-Name"; val tag: String = "sn" }
-  case object AssignedMolecule extends AssemblyReportColumn { val key: String = "Assigned-Molecule"; val tag: String = "am"  }
+  case object AssignedMolecule extends AssemblyReportColumn { val key: String = "Assigned-Molecule"; val tag: String = "am" }
   case object GenBankAccession extends AssemblyReportColumn { val key: String = "GenBank-Accn"; val tag: String = "ga"  }
   case object RefSeqAccession  extends AssemblyReportColumn { val key: String = "RefSeq-Accn"; val tag: String = "ra"  }
   case object UcscName         extends AssemblyReportColumn { val key: String = "UCSC-style-name"; val tag: String = "un"  }
 
   /** The missing value for a column */
-  val MissingValue: String = "na"
+  val MissingValueKey: String = "na"
```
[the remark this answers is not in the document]
tfenne: Also, it feels like this doesn't belong inside as enum called "AssemblyReportColumn".

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -54,19 +54,19 @@ object AssemblyReportColumn extends FgBioEnum[AssemblyReportColumn] {
...
 
   /** The column key for Sequence-Role */
-  val SequenceRole: String = "Sequence-Role"
+  val SequenceRoleKey: String = "Sequence-Role"
```
tfenne, line 66: Why not just make this a fully fledge `AssemblyReportColumn`?  You define it's tag below too.

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -54,19 +54,19 @@ object AssemblyReportColumn extends FgBioEnum[AssemblyReportColumn] {
...
 
   /** The column key for Sequence-Length */
-  val SequenceLength: String = "Sequence-Length"
+  val SequenceLengthKey: String = "Sequence-Length"
```
tfenne, line 69: Could also make this a `AssemblyReportColumn` with tag `LN`?

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -112,6 +112,9 @@ object SequenceRole extends FgBioEnum[SequenceRole] {
     |
     |First, contig with the Sequence-Role "assembled-molecule" will be outputted.  Next, the remaining contigs will be
     |sorted by descending length.
+    |
+    |Aliases for values in the Assigned-Molecule column will only be applied to sequences with the "assembled-molecule"
```
tfenne, line 116: Two thoughts:
1. You could make this an option (with default) so it does the right thing for human-like assembly report.  I don't know enough about assembly reports for less finished organisms to know if we'd want to do something different there?  /shrug.
2. If you leave as is, maybe reword to "The `Assigned-Molecule` column, if specified as an `--alternate` will only be used for sequences with `Sequence-Role` `assembled-molecule`.

PR #585, opened 2020-05-16
Add an option to modify an existing sequence dictionary

nh13:
This allows the order of the contigs to be pre-specified and ensures the
contigs we want are extracted from the assembly report.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
@@ -228,7 +228,9 @@ class CollectAlternateContigNames
       case Some(sequenceDictionary) =>
         // re-order the contigs by the existing dictionary
         val infosMap = (primaries ++ secondaries).map { metadata => (metadata.name, metadata) }.toMap
-        sequenceDictionary.map { metadata => infosMap(metadata.name) }.toSeq
+        sequenceDictionary.map { metadata =>
+          infosMap.getOrElse(metadata.name, metadata)
+        }.toSeq
     }
 
     // Write it out!
--- a/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
@@ -29,6 +29,7 @@ import com.fulcrumgenomics.FgBioDef.{FgBioEnum, PathToSequenceDictionary}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.commons.CommonsDef._
 import com.fulcrumgenomics.commons.util.LazyLogging
+import com.fulcrumgenomics.fasta.SequenceMetadata.Keys
 import com.fulcrumgenomics.sopt._
 import com.fulcrumgenomics.util.Io
 import enumeratum.EnumEntry
@@ -76,7 +77,6 @@ object AssemblyReportColumn extends FgBioEnum[AssemblyReportColumn] {
 /** Trait that entries in SequenceRole will extend. */
 sealed trait SequenceRole extends EnumEntry {
   def key: String
-  def primary: Boolean
 }
 
 /** Enum to represent the various types of sequence-roles in NCBI assembly report. */
@@ -88,12 +88,20 @@ object SequenceRole extends FgBioEnum[SequenceRole] {
     values.find(_.key == str).getOrElse(super.apply(str))
   }
 
-  case object AltScaffold         extends SequenceRole { val key: String = "alt-scaffold"; val primary: Boolean = false }
-  case object AssembledMolecule   extends SequenceRole { val key: String = "assembled-molecule"; val primary: Boolean = true }
-  case object FixPatch            extends SequenceRole { val key: String = "fix-patch"; val primary: Boolean = false }
-  case object NovelPatch          extends SequenceRole { val key: String = "novel-patch"; val primary: Boolean = false }
-  case object UnlocalizedScaffold extends SequenceRole { val key: String = "unlocalized-scaffold"; val primary: Boolean = false }
-  case object UnplacedScaffold    extends SequenceRole { val key: String = "unplaced-scaffold"; val primary: Boolean = false }
+  /** Allows the column to be build from the Enum name or the actual sequence-role column name. */
+  def apply(metadata: SequenceMetadata): SequenceRole = {
+    val roleValue = metadata.attributes.getOrElse(AssemblyReportColumn.SequenceRole.tag,
+      throw new IllegalArgumentException(s"Metadata missing tag: ${AssemblyReportColumn.SequenceRole.tag}")
+    )
+    SequenceRole(roleValue)
+  }
+
+  case object AltScaffold         extends SequenceRole { val key: String = "alt-scaffold" }
+  case object AssembledMolecule   extends SequenceRole { val key: String = "assembled-molecule" }
+  case object FixPatch            extends SequenceRole { val key: String = "fix-patch" }
+  case object NovelPatch          extends SequenceRole { val key: String = "novel-patch" }
+  case object UnlocalizedScaffold extends SequenceRole { val key: String = "unlocalized-scaffold" }
+  case object UnplacedScaffold    extends SequenceRole { val key: String = "unplaced-scaffold" }
 }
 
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -124,10 +124,13 @@ class CollectAlternateContigNames
  @arg(flag='a', doc="The assembly report column(s) for the alternate contig name(s)", minElements=1) val alternates: Seq[AssemblyReportColumn],
  @arg(flag='s', doc="Only output sequences with the given sequence roles.  If none given, all sequences will be output.", minElements=0)
   val sequenceRoles: Seq[SequenceRole] = Seq.empty,
+ @arg(flag='d', doc="Add aliases to this existing sequence dictionary file.  The primary contig must match.") val existing: Option[PathToSequenceDictionary] = None,
+ @arg(flag='x', doc="Allow mismatching sequence lengths when using an existing sequence dictionary fil.") val allowMismatchingLengths: Boolean = false,
```
tfenne, line 128: `fil` -> [`file`, ""]

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -124,10 +124,13 @@ class CollectAlternateContigNames
  @arg(flag='a', doc="The assembly report column(s) for the alternate contig name(s)", minElements=1) val alternates: Seq[AssemblyReportColumn],
  @arg(flag='s', doc="Only output sequences with the given sequence roles.  If none given, all sequences will be output.", minElements=0)
   val sequenceRoles: Seq[SequenceRole] = Seq.empty,
+ @arg(flag='d', doc="Add aliases to this existing sequence dictionary file.  The primary contig must match.") val existing: Option[PathToSequenceDictionary] = None,
```
tfenne, line 127: Maybe add a sentence about this in the class level usage too?  There's also more room there to explain "The primary contig must match".  I assume that means that _all_ the sequences in the input dictionary must match to a sequence in the assembly report by `--primary` name but that it's ok for the assembly report to have more records?  Might also explain whether or not you match against all aliases on both sides.

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -124,10 +124,13 @@ class CollectAlternateContigNames
...
+ @arg(flag='x', doc="Allow mismatching sequence lengths when using an existing sequence dictionary fil.") val allowMismatchingLengths: Boolean = false,
  @arg(doc="Skip contigs that have no alternates") val skipMissing: Boolean = true
 ) extends FgBioTool with LazyLogging {
 
   Io.assertReadable(input)
+  existing.map(Io.assertReadable)
```
tfenne, line 133: `existing.foreach()`

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -141,6 +144,9 @@ class CollectAlternateContigNames
     require(iter.hasNext, s"Missing header from $input.")
     val header = iter.next().substring(2).split('\t')
 
+    // read in the existing sequence dictionary, if specified
+    val existingSequenceDictionary: Option[SequenceDictionary] = this.existing.map(SequenceDictionary(_))
```
tfenne, line 148: So nice :)

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -178,10 +184,22 @@ class CollectAlternateContigNames
         AssemblyReportColumn.values.foreach { column =>
           attributes += ((column.tag, dict(column.key)))
         }
+        // get the list of aliases (may update attributes if an existing contig is found)
+        val aliases: Seq[String] = existingSequenceDictionary match {
+          case None           => alts
```
tfenne, line 189: Please either remove extra spaces or align with the `=>` below.

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -178,10 +184,22 @@ class CollectAlternateContigNames
...
+          case Some(sequenceDictionary) =>
```
tfenne, line 190: I don't think this belongs here for a couple of reasons:
1. It's weird to do all this validation when in theory just collecting the list of aliases for a sequence
2. Without more hoop-jumping you can't update a dictionary that doesn't contain the matching set of sequences for the assembly report / sequence-role(s).

I think it would be much cleaner to revert this section and have the main loop just generate a new set of `SequenceMetadata` objects from the assembly report and then do the merging in one single pass down below.

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -192,11 +210,22 @@ class CollectAlternateContigNames
...
-    val infos = (primaries ++ secondaries.sortBy(-_.length))
-      .zipWithIndex
-      .map { case (metadata, index) => metadata.copy(index=index) }
-      .toSeq
+
+    // Get the ordering of the molecules
+    val infos: Seq[SequenceMetadata] = existingSequenceDictionary match {
+      case None =>
+        // Sort the secondary entries based on descending sequence length.
+        (primaries ++ secondaries.sortBy(-_.length))
```
tfenne, line 218: Why do this sort?  The assembly reports are (in my experience) nicely ordered and this shuffles unplaced/alts/randoms together and by length, instead of keeping them separate and ordered karyotypically.

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -192,11 +210,22 @@ class CollectAlternateContigNames
...
+          .zipWithIndex
+          .map { case (metadata, index) => metadata.copy(index=index) }
+          .toSeq
+      case Some(sequenceDictionary) =>
```
tfenne, line 222: I think this needs to be done differently.  The docs say "update an existing sequence dictionary" but really this is just "ordering by an existing sequence dictionary" and will drop any tags (e.g. MD5) not generated by `CollectAlternateContigNames`.  I think this should be implemented to cycle through the input dictionary and generate copies of the _input_ metadatas that:

1. Add any new aliases from the report that are not already contained in the input record
2. Add/update any other tags extended tags on the input record that are defined on the new record

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
nh13: I think this is a hold-over from a genome FASTA that doesn't follow the ordering in the NCBI assembly report.  I suppose the answer is to either manually re-order the input report or manually re-order the output dictionary.

nh13:
@tfenne great suggestions, and I've implemented all suggestions.  One more scan?

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -107,11 +115,13 @@ object SequenceRole extends FgBioEnum[SequenceRole] {
...
     |
-    |First, contig with the Sequence-Role "assembled-molecule" will be outputted.  Next, the remaining contigs will be
-    |sorted by descending length.
-    |
     |The `Assigned-Molecule` column, if specified as an `--alternate`, will only be used for sequences with
     |`Sequence-Role` `assembled-molecule`.
+    |
+    |When updating an existing sequence dictionary with `--existing` the primary contig names must match.  I.e. the
+    |contig name from the assembly report column specified by `--primary` must match the contig name in the existing
+    |sequence dictionary (`@SQ.SN`).  All contigs in the existing sequence report must be present in the assembly
```
tfenne, line 123: sequence report -> sequence dictionary

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -190,21 +201,32 @@ class CollectAlternateContigNames
...
+
+    // Apply to an existing sequence dictionary if necessary
+    val dict: SequenceDictionary = existing match {
+      case None => SequenceDictionary(metadatas.toSeq:_*)
+      case Some(path) =>
+        val assemblyReportMetadataMap = metadatas.map { m => (m.name, m) }.toMap
+        val updatedMetadatas          = SequenceDictionary(path).map { existingMetadata =>
+          // Get the metadata from the assembly report
+          val assemblyReportMetadata: SequenceMetadata = assemblyReportMetadataMap.getOrElse(existingMetadata.name,
+            throw new IllegalArgumentException(s"Could not find contig '${existingMetadata.name}' in the asesmlby report'")
```
tfenne, line 216: asesmlby -> assembly

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -190,21 +201,32 @@ class CollectAlternateContigNames
...
+          )
+          // append new aliases, and add new tags
+          val attributes = existingMetadata.attributes ++ assemblyReportMetadata.attributes.map {
```
tfenne, line 219: I think this works but I'm a bit nervous about iteration order of `map ++ map`.

PR #586, opened 2020-05-16
Tool for updating contig names in a delimited data file

nh13:
First use case is a GTF

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
@@ -33,7 +33,7 @@ import com.fulcrumgenomics.sopt.{arg, clp}
 
 @clp(description =
   """
-    |Updates then contig names in a delimited data file.
+    |Updates the contig names in columns of a delimited data file (e.g. CSV, TSV).
     |
     |The name of each sequence must match one of the names (including aliases) in the given sequence dictionary.  The
     |new name will be the primary (non-alias) name in the sequence dictionary.
@@ -43,10 +43,10 @@ class UpdateDelimitedFileContigNames
 (@arg(flag='i', doc="Input delimited data file.") val input: FilePath,
  @arg(flag='d', doc="The path to the sequence dictionary with contig aliases.") val dict: PathToSequenceDictionary,
  @arg(flag='c', doc="The column indices for the contig names (0-based).") val columns: Seq[Int],
- @arg(flag='o', doc="Output delimited data file.") val output: FilePath,
  @arg(flag='T', doc="The delimiter") val delimiter: Char = '\t',
- @arg(flag='H', doc="Treat lines with this starting string as comments (always printed)") val header: String = "#",
- @arg(doc="Skip missing contigs.") val skipMissing: Boolean = false
+ @arg(flag='H', doc="Treat lines with this starting string as comments (always printed)") val comment: String = "#",
+ @arg(flag='o', doc="Output delimited data file.") val output: FilePath,
+ @arg(flag='n', doc="Output the first `N` lines as-is (always printed).") val outputFirstNumLines: Int = 0
 ) extends FgBioTool with LazyLogging {
 
   Io.assertReadable(input)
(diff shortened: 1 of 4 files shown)


src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
```diff
@@ -0,0 +1,117 @@
...
+        forloop(from=0, until=line.length) { i =>
+          if (line(i) == delimiter) count += 1
+        }
+        count
+      }
+      if (numExpected > numFields) {
+        assert(fields.length < numExpected + fields.map(_.length).sum)
+        fields = Array.fill(numExpected)("")
+        assert(StringUtil.split(line, delimiter, fields) == numExpected)
+      }
```
tfenne, line 90: This is a lot of work to re-use the same array and keep expanding it.  Is it really worth it?  Especially since it requires you to iterate over the input line character-by-character twice for each line.
If you really want to do something like this I would suggest replacing with something like:

```scala
var fields = Array[String](100)
in.foreach { line => 
  var numFields   = StringUtil.split(line, delimiter, fields)
  while (numFields == fields.length) {
    fields = Array[String](fields.length * 2)
    numFields = StringUtil.split(line, delimiter, fields)
  }
```

Better though would be to extend that `split` method to have an option to place _all remaining text_ into the last value in the array instead of discarding extra values.  Then you could just size your array to `max(columns) + 2` and call it a day.

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
```diff
@@ -0,0 +1,117 @@
...
+
+      // update the fields and output
+      columns.foreach { column =>
+        require(column < numFields, f"Too few columns (at least ${column + 1}) on line ${progress.getCount}%,d")
+        dict.get(fields(column)) match {
+          case Some(metadata) => fields(column) = metadata.name
+          case None           =>
+            val message = s"Did not find contig ${fields(column)} in the list of source names."
+            if (skipMissing) {
```
tfenne, line 99: Collapse if/else to a single line?  Or:
```scala
case None if skipMissing => logger.warn("Did not find contig ${fields(column)} in the list of source names.")
case None                => throw new IllegalStateException("Did not find contig ${fields(column)} in the list of source names.")
```

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
```diff
@@ -0,0 +1,117 @@
...
+    |The name of each sequence must match one of the names (including aliases) in the given sequence dictionary.  The
+    |new name will be the primary (non-alias) name in the sequence dictionary.
+  """,
+  group = ClpGroups.Utilities)
+class UpdateDelimitedFileContigNames
+(@arg(flag='i', doc="Input delimited data file.") val input: FilePath,
+ @arg(flag='d', doc="The path to the sequence dictionary with contig aliases.") val dict: PathToSequenceDictionary,
+ @arg(flag='c', doc="The column indices for the contig names (0-based).") val columns: Seq[Int],
+ @arg(flag='o', doc="Output delimited data file.") val output: FilePath,
+ @arg(flag='T', doc="The delimiter") val delimiter: Char = '\t',
```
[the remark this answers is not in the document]
nh13: Keeping `-d` to be consistent with the `Update*ContigNames` tools.

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  override def execute(): Unit = {
+    val dict   = SequenceDictionary(this.dict)
+    val in     = Io.readLines(input).bufferBetter
+    val out    = Io.toWriter(output)
+
+    val progress = ProgressLogger(logger, noun="lines", verb="written")
+
+    // write the header
+    in.takeWhile(_.startsWith(this.header)).foreach { line =>
```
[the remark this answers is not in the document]
nh13: Great ideas all around!

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
nh13: https://github.com/fulcrumgenomics/commons/pull/64

nh13: I suppose I'm over-optimizing here...

nh13:
Fixed all the things, made tests (was a draft!), and depends on https://github.com/fulcrumgenomics/commons/pull/64, which requires an update to `build.sbt` when merged (and in snapshots).

PR #590, opened 2020-06-05
Add the sort and default-contigs options to UpdateFastaContigName

nh13:
The `--sort` option will output the contigs in the sequence dictionary
order.

The `--default-contigs` option will add missing contigs in the input
FASTA to the output FASTA.  This is useful when renaming a FASTA where
some contigs are missing (ex. alt scaffolds in Ensembl vs. UCSC).

---

I have a case where the Ensembl provided VCF needs to have it's contigs renamed, and have missing contigs added from a secondary FASTA.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
@@ -24,13 +24,15 @@
 
 package com.fulcrumgenomics.fasta
 
+import java.io.BufferedWriter
+
 import com.fulcrumgenomics.FgBioDef.PathToSequenceDictionary
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.commons.CommonsDef._
 import com.fulcrumgenomics.commons.util.LazyLogging
 import com.fulcrumgenomics.sopt.{arg, clp}
 import com.fulcrumgenomics.util.{Io, ProgressLogger}
-import htsjdk.samtools.reference.{FastaSequenceIndex, ReferenceSequenceFileFactory}
+import htsjdk.samtools.reference.{FastaSequenceIndex, ReferenceSequence, ReferenceSequenceFile, ReferenceSequenceFileFactory}
 
 @clp(description =
   """
@@ -38,6 +40,12 @@ import htsjdk.samtools.reference.{FastaSequenceIndex, ReferenceSequenceFileFacto
     |
     |The name of each sequence must match one of the names (including aliases) in the given sequence dictionary.  The
     |new name will be the primary (non-alias) name in the sequence dictionary.
+    |
+    |By default, the sort order of the contigs will be the same as the input FASTA.  Use the `--sort` option to sort by
+    |the input sequence dictionary.  Furthermore, the sequence dictionary may contain **more** contigs than the input
+    |FASTA, and they wont be used.  Use the `--skip-missing` option to skip contigs in the input FASTA that cannot be
+    |renamed (i.e. who are not present in the input sequence dictionary).  Finally, use the `--default-contigs` to
+    |append contigs not present in the input FASTA but present in the sequence dictionary.
   """,
   group = ClpGroups.Fasta)
 class UpdateFastaContigNames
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -24,70 +24,157 @@
...
   """,
   group = ClpGroups.Fasta)
 class UpdateFastaContigNames
 (@arg(flag='i', doc="Input FASTA.") val input: PathToFasta,
  @arg(flag='d', doc="The path to the sequence dictionary with contig aliases.") val dict: PathToSequenceDictionary,
  @arg(flag='o', doc="Output FASTA.") val output: PathToFasta,
  @arg(flag='l', doc="Line length or sequence lines.") val lineLength: Int = 100,
- @arg(doc="Skip missing source contigs.") val skipMissing: Boolean = false
+ @arg(doc="Skip missing source contigs.") val skipMissing: Boolean = false,
+ @arg(doc="Sort the contigs based on the input sequence dictionary") sort: Boolean = false,
```
sstadick, line 57: Possibly rename to `sortByDict`. Since `--sort` usually means "output is sorted with this flag, and not sorted without this flag", and that is not the case here.

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -24,70 +24,157 @@
...
 
+  /** Little class to store the sequence metadata and reference sequence to output.
+    *
+    * The reference sequence to output is stored as a method, so we can lazily retrieve it and not incur large memory
+    * overhead.
+    *
+    * @param info the sequence metadata to output
+    * @param toSequence method that returns the reference sequence to output
+    */
+  private case class OutputSequence(info: SequenceMetadata, toSequence: () => ReferenceSequence) {
```
sstadick, line 75: you could make `toSequence` a by-name parameter, which does the same thing but is more readable / usable IMO.

```scala
private case class OutputSequence(info: SequenceMetadata, sequence: => ReferenceSequence)
```
Whatever sequence is isn't evaluated till you use it. So at a call site you can just call a method like normal instead of wrapping things up. 

```scala
OutputSequence(info=info, sequence=defaultRefFile.getSequence(info.name)))
```

nh13: This works, but just be aware we can't use by-name parameters for case classes.

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -24,70 +24,157 @@
...
 @clp(description =
   """
     |Updates the sequence names in a FASTA.
     |
     |The name of each sequence must match one of the names (including aliases) in the given sequence dictionary.  The
     |new name will be the primary (non-alias) name in the sequence dictionary.
+    |
+    |By default, the sort order of the contigs will be the same as the input FASTA.  Use the `--sort-by-dict` option to
+    |sort by the input sequence dictionary.  Furthermore, the sequence dictionary may contain **more** contigs than the
+    |input FASTA, and they wont be used.  Use the `--skip-missing` option to skip contigs in the input FASTA that cannot
```
tfenne, line 46: I think a little more explanation here would be good.  Assuming `--skip-missing` means to not emit the contig (not just not rename it), I'd rewrite the last two sentences as:

```
Use the `--skip-missing` option to skip contigs in the input FASTA that cannot be renamed (i.e. who are not present in the input sequence dictionary); missing contigs will not be written to the output fasta.  Finally, use the `--default-contigs` option to specify an additional FASTA which will be queried to locate contigs not present in the the input FASTA but present in the sequence dictionary.
```

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -24,70 +24,157 @@
...
+ defaultContigs: Option[PathToFasta] = None
 ) extends FgBioTool with LazyLogging {
+  import com.fulcrumgenomics.fasta.Converters.FromSAMSequenceDictionary
 
-  Io.assertReadable(input)
   Io.assertReadable(Seq(input, dict))
+  defaultContigs.foreach(Io.assertReadable)
   Io.assertCanWriteFile(output)
 
+  /** Little class to store the sequence metadata and reference sequence to output.
```
tfenne, line 67: I feel like we need a competition to find the most arcane and hilarious synonym for "little" for use in these situations.

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -24,70 +24,157 @@
...
+    * overhead.
+    *
+    * @param info the sequence metadata to output
+    * @param sequence the reference sequence to output
+    */
+  private class OutputSequence(val info: SequenceMetadata, sequence: => ReferenceSequence) {
+    /** Writes the output sequence */
+    def write(writer: BufferedWriter, progress: ProgressLogger): Unit = {
+      val ref = sequence
+      writer.append('>').append(info.name).append('\n')
```
tfenne, line 79: Here's my version of this ... I particularly like the `grouped(80)` for this:

```scala
      out.write(s">${ref.getName}\n")
      ref.getBaseString.toSeq.grouped(80).foreach { line =>
        out.write(line.unwrap)
        out.write("\n")
      }
```

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -24,70 +24,157 @@
...
-    val srcContigs = refFile.getSequenceDictionary match {
-      case null =>
-        require(refFile.isIndexed,
-          "Reference sequence file must have a sequence dictionary or be indexed.  Try 'picard CreateSequenceDictionary' or 'samtools faidx <ref.fasta>'.")
-        new FastaSequenceIndex(ReferenceSequenceFileFactory.getFastaIndexFileName(this.input)) map(_.getContig)
-      case dict => dict.getSequences.map(_.getSequenceName)
+    val (srcDict, srcRefFile) = getDictAndFile(fasta=this.input)
+    val defaultDictAndRefFile = this.defaultContigs.map(fasta => getDictAndFile(fasta=fasta))
+
+    // Make sure the default FASTA and the input dict are the same
```
tfenne, line 103: Why make this a requirement?  E.g. what if I have a nicely ordered dict that I made, but my default fasta is hastily sourced from UCSC and is in lexicographic order?  Maybe just warn for this?  Or fail later if there's a sequence needed that can't be found within the default?

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -24,70 +24,157 @@
...
+    // Note: we do not pre-load the sequences, as to preserve memory
+    val srcSequences: Seq[OutputSequence] = srcDict.iterator.flatMap { srcInfo =>
+      // Check if the sequence dictionary can rename this contig, and if not, either log or fail depending on --skip-missing
+      val dictInfo = dict.get(srcInfo.name) match {
+        case None if skipMissing => logger.warning(s"Did not find contig ${srcInfo.name} in the sequence dictionary."); None
+        case None                => throw new IllegalStateException(s"Did not find contig ${srcInfo.name} in the sequence dictionary.")
+        case Some(info)          => Some(info)
+      }
+      dictInfo.map(info => new OutputSequence(info=info, sequence=srcRefFile.getSequence(srcInfo.name)))
+    }.toSeq
```
tfenne, line 118: My habit it to use `.toIndexedSeq` to avoid `.toSeq` returning a lazy `Stream` which can make debugging more difficult.

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -24,70 +24,157 @@
...
+    sequences.foreach(_.write(writer=out, progress=progress))
+    logger.info(s"Wrote ${sequences.length} contigs.")
+
+    srcRefFile.safelyClose()
+    defaultDictAndRefFile.foreach(_._2.safelyClose())
     out.close()
   }
 
+
+  /** Gets the sequence dictionary and referene sequence file for a given FASTA */
```
tfenne, line 160: referene -> reference

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
```diff
@@ -24,70 +24,157 @@
...
+    val refFile = ReferenceSequenceFileFactory.getReferenceSequenceFile(fasta, true, true)
+    val dict    = Option(refFile.getSequenceDictionary) match {
+      case Some(dict) => dict.fromSam
+      case None       =>
+        require(refFile.isIndexed,
+          s"Reference sequence file must have a sequence dictionary or be indexed." +
+          s"  Try 'picard CreateSequenceDictionary' or 'samtools faidx <ref.fasta>'. $fasta"
+        )
+        // Build a very basic sequence dictionary from the fasta index
+        val infos = new FastaSequenceIndex(ReferenceSequenceFileFactory
```
tfenne, line 171: Nice!

src/main/scala/com/fulcrumgenomics/fasta/UpdateFastaContigNames.scala
nh13: The comment may be misleading.  I mean to say:
> Make sure the default FASTA and the **_default_** input dict are the same, as the latter should be built from the former

PR #591, opened 2020-06-12
Bugfix: UpdateGffContigNames outputs a misformatted GFF

nh13:
See: https://github.com/arq5x/bedtools2/issues/848

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/UpdateGffContigNames.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/UpdateGffContigNames.scala
@@ -92,7 +92,7 @@ class UpdateGffContigNames
         val position = fields(3).toInt
         out.append(targetName)
         out.append('\t')
-        out.append(line.substring(srcName.length))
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/util/UpdateGffContigNames.scala
```diff
@@ -92,7 +92,7 @@ class UpdateGffContigNames
         val position = fields(3).toInt
         out.append(targetName)
         out.append('\t')
-        out.append(line.substring(srcName.length))
+        out.append(line.substring(srcName.length + 1)) // skip over the previous target name AND trailing tab
         out.append('\n')
```
sstadick, line 96: Actually, while we are in off by one territory, does the BetterBuffereIteror strip the newline? Or are you getting 2 newlines?

nh13: `in` comes all the way back from `Io.readLines`, which strips the newline.

PR #593, opened 2020-06-19
Updated GenomicRange to handle single-bp ranges without end specified…

tfenne:
…, and to handle commas in the coordinates.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/FgBioDef.scala
+++ b/src/main/scala/com/fulcrumgenomics/FgBioDef.scala
@@ -71,23 +71,40 @@ object FgBioDef extends CommonsDef {
     override def toString: String = f"$refName:$start-$end"
   }
 
-  /** Parses a genomic range of the form `<chr>:<start>-<end>`, assuming the range is 1-based inclusive (closed-ended).
+  object GenomicRange {
+    /** Parses a genomic range of the form `<chr>:<start>-<end>` or `<chr>:<start>`, assuming the range is 1-based
+      * inclusive (closed-ended).  If only a start position is included the range includes just the single base.
+      * Coordinates may include commas in the numbers.
+      *
+      * @param range the string to parse.
+      */
+    def apply(range: String): GenomicRange = {
+      val (refName: String, rest: String) = range.indexOf(':') match {
+        case -1  => throw new IllegalArgumentException(f"Missing colon in genomic range: '$range'")
+        case idx =>
+          val (left, right) = range.splitAt(idx)
+          (left, right.drop(1).replace(",", ""))
+      }
+
+      val (start, end) = rest.indexOf('-') match {
+        case -1  => (rest, rest)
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -71,23 +71,40 @@ object FgBioDef extends CommonsDef {
...
+      * Coordinates may include commas in the numbers.
+      *
+      * @param range the string to parse.
+      */
+    def apply(range: String): GenomicRange = {
+      val (refName: String, rest: String) = range.indexOf(':') match {
+        case -1  => throw new IllegalArgumentException(f"Missing colon in genomic range: '$range'")
+        case idx =>
+          val (left, right) = range.splitAt(idx)
+          (left, right.drop(1).replace(",", ""))
```
nh13, line 86: Doing the replace here will make this parseable: `chr1:2-,`.  Perhaps move it to line 93 and apply on the left and right individually?

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -71,23 +71,40 @@ object FgBioDef extends CommonsDef {
...
+  }
+
+  /** Parses a genomic range of the form `<chr>:<start>-<end>` or `<chr>:<start>`, assuming the range is 1-based
+    * inclusive (closed-ended).  If only a start position is included the range includes just the single base.
+    * Coordinates may include commas in the numbers.
     *
     * @param range the string to parse.
+    * @deprecated use GenomicRange(string) instead
     */
   def parseRange(range: String): GenomicRange = {
```
nh13, line 107: do you want a deprecation annotation?

```
@deprecated(message="Use GenomicRange(string) instead.", since="X.Y.Z")
def parseRange(range: String): GenomicRange = {
```

tfenne: Done.

src/main/scala/com/fulcrumgenomics/FgBioDef.scala
tfenne: I added a test to ensure parsing `chr1:2-,` fails and it already fails, so leaving the parsing code as is.

nh13: Doh, how about `chr1:2,`?

tfenne: But that's not what you said 😉 
If you feel strongly I'm happy to update, but my 2c is that for parsing like this that's likely to be used on the command line a lot, I'm happy to parse questionably valid things so long as there's only a single reasonable interpretation of what the right value is (e.g. can you suggest any _reasonable_ parsing of `chr1:2,` that isn't just `chr1:2`?).

PR #598, opened 2020-07-27
Add the option to sort the output in UpdateDelimitedFileContigNames

nh13:
The --sort-order option controls if the output should be sorted by:
1. Contig only - preserves the input order for lines with the same contig
2. Coordinate - preserves the input order for line with the same coordiante

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
@@ -26,10 +26,15 @@
 package com.fulcrumgenomics.util
 
 import com.fulcrumgenomics.FgBioDef.{FilePath, PathToSequenceDictionary, _}
+import com.fulcrumgenomics.bam.Bams.MaxInMemory
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.commons.util.{LazyLogging, StringUtil}
 import com.fulcrumgenomics.fasta.SequenceDictionary
 import com.fulcrumgenomics.sopt.{arg, clp}
+import com.fulcrumgenomics.util.Sorter.Codec
+import enumeratum.EnumEntry
+
+import scala.collection.immutable.IndexedSeq
 
 @clp(description =
   """
@@ -48,7 +53,10 @@ class UpdateDelimitedFileContigNames
  @arg(flag='H', doc="Treat lines with this starting string as comments (always printed)") val comment: String = "#",
  @arg(flag='o', doc="Output delimited data file.") val output: FilePath,
  @arg(flag='n', doc="Output the first `N` lines as-is (always printed).") val outputFirstNumLines: Int = 0,
- @arg(doc="Skip lines where a contig name could not be updated (i.e. missing from the sequence dictionary).") skipMissing: Boolean = false
+ @arg(doc="Skip lines where a contig name could not be updated (i.e. missing from the sequence dictionary).") skipMissing: Boolean = false,
+ @arg(flag='s', doc="Sort the output based on the following order.") sortOrder: SortOrder = SortOrder.Unsorted,
+ @arg(doc="The column index for the contig (0-based) for sorting. Use the first column if not given.") val contig: Option[Int] = None,
+ @arg(doc="The column index for the genomic position (0-based) for sorting by coordinate.") val position: Option[Int] = None
 ) extends FgBioTool with LazyLogging {
 
   Io.assertReadable(input)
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
```diff
@@ -48,7 +53,10 @@ class UpdateDelimitedFileContigNames
  @arg(flag='H', doc="Treat lines with this starting string as comments (always printed)") val comment: String = "#",
  @arg(flag='o', doc="Output delimited data file.") val output: FilePath,
  @arg(flag='n', doc="Output the first `N` lines as-is (always printed).") val outputFirstNumLines: Int = 0,
- @arg(doc="Skip lines where a contig name could not be updated (i.e. missing from the sequence dictionary).") skipMissing: Boolean = false
+ @arg(doc="Skip lines where a contig name could not be updated (i.e. missing from the sequence dictionary).") skipMissing: Boolean = false,
+ @arg(flag='s', doc="Sort the output based on the following order.") sortOrder: SortOrder = SortOrder.Unsorted,
+ @arg(doc="The column index for the contig (0-based) for sorting. Use the first column if not given.") val contig: Option[Int] = None,
```
tfenne, line 58: I would make these 1-based and shift it internally.  I think when talking about columns it's easier to match e.g. what `cut` and `awk` use since that'll be intuitive.

tfenne: But I guess that ship might have already sailed?

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
```diff
@@ -57,11 +65,21 @@ class UpdateDelimitedFileContigNames
 
   this.columns.foreach(column => validate(column >= 0, s"Column index < 0: $column"))
 
+  if (sortOrder == SortOrder.ByCoordinate) {
+    require(position.nonEmpty, "--position is required when sorting by coordinate")
```
tfenne, line 69: Use `validate(position.nonEmpty)`.

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
```diff
@@ -99,17 +117,81 @@ class UpdateDelimitedFileContigNames
...
+            // sorting
+            sorter match {
+              case None => // Write out the fields without sorting
+                forloop(from = 0, until = numFields) { i =>
+                  if (0 < i) out.append(delimiter)
+                  out.write(fields(i))
+                }
+                out.write('\n')
+              case Some(_sorter) => // add then to the sorter, to be written later
+                _sorter += CoordinateAndString(
```
tfenne, line 129: The name `CoordinateAndString` is confusing.  Why not just call it `LineInfo` or `LineAndSortKeys` or something?

tfenne: The way you're constructing this will use 2x the amount of memory for the sorting.  The whole idea of the `Sorter` is to serialize the whole thing as effectively as possible to reduce storage, and then only hold the key in memory.  Since the key has all the data over again, it'll store twice as much.

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
```diff
@@ -99,17 +117,81 @@ class UpdateDelimitedFileContigNames
...
+      }
+      _sorter.close()
+    }
+
     out.close()
   }
+}
+
+/** Trait that entries in SortOrder will extend. */
+sealed trait SortOrder extends EnumEntry
```
tfenne, line 156: Could all of this stuff go into a companion object please?

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
tfenne: I _think_ the best way to do this would be to a) just use the line as the thing put into the sorter and b) have the sorter be able to re-split the line and pull out the contig and position fields, but I see that's rather complicating.

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
nh13: Unfortunately `FgBioEnum`s cannot be in the companion object, but the rest can.

src/main/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNames.scala
nh13: Fixed

PR #601, opened 2020-07-31
Have Metric classes correctly serialize EnumEntry fields to string

clintval:
This section describes the ways to serialize `EnumEntry` enumerations to string:

https://github.com/lloydmeta/enumeratum#enum

This comes up if you want to override the string representations of the enumeration (normally just use case object name).

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/Metric.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/Metric.scala
@@ -35,6 +35,7 @@ import com.fulcrumgenomics.commons.reflect.{ReflectionUtil, ReflectiveBuilder}
 import com.fulcrumgenomics.commons.util.DelimitedDataParser
 import htsjdk.samtools.util.{FormatUtil, Iso8601Date}
 import com.fulcrumgenomics.commons.io.{Writer => CommonsWriter}
+import enumeratum.EnumEntry
 
 import scala.collection.compat._
 import scala.reflect.runtime.{universe => ru}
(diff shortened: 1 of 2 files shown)


nh13:
@clintval how does this compare to just using a FgBioEnum? The latter enables having Enums on the command line and should work for metric classes too? Is there something missing?

clintval:
Not entirely sure I understand your question. This is only needed for serialization to string when you **override the inherit entry name of the enumeration itself** (regardless of if it's `EnumEntry` or `FgBioEnum`). So it's the special case of these enumerations for when you want their string representation to be different than their class/object name as a string.

Normally this isn't needed because `.toString()` returns the class name as a string, which is the default (a wrapper over `EnumEntry#entryName` itself):

Check out the [`source`](https://github.com/lloydmeta/enumeratum/blob/3beb3dd1fe22cda12fe13469c299af5a9c07aa33/enumeratum-core/src/main/scala/enumeratum/EnumEntry.scala#L5-L25) for a better look:

```scala
/**
  * Base type for an enum entry for [[Enum]]
  *
  * By default, the entryName method used for serialising and deserialising Enum values uses
  * toString, but feel free to override to fit your needs.
  *
  * Mix in the supplied stackable traits to convert the entryName to [[EnumEntry.Snakecase Snakecase]],
  * [[EnumEntry.Uppercase Uppercase]], [[EnumEntry.Lowercase Lowercase]] etc.
  */
trait EnumEntry {

  /**
    * String representation of this Enum Entry.
    *
    * Override in your implementation if needed
    */
  def entryName: String = stableEntryName

  private[this] lazy val stableEntryName: String = toString

}
```

nh13:
I’ll review over the weekend, but could you also make sure the tests compile and pass?

PR #607, opened 2020-08-19
Add recommended bwa usage for AssignPrimers

nh13:


src/main/scala/com/fulcrumgenomics/util/AssignPrimers.scala
```diff
@@ -42,6 +42,13 @@ import com.fulcrumgenomics.sopt.{arg, clp}
     |
     |The read sequence of the primer is not checked against the expected reference sequence at the primer's genomic
     |coordinates.
+    |
+    |In some cases, large deletions within one end of a read pair may cause a primary and supplementary alignment to be
```
nh13, line 46:
```suggestion
    |In some cases, large deletions within one end of a read pair may cause a primary and supplementary alignments to be
```

Issue #599, opened 2020-07-28
Add an option to ErrorRateByReadPosition not to collapse substitution types


nh13:
Currently, `ErrorRateByReadPosition` collapses substitution types based on the reference base.

>Substitution types are collapsed based on the reference or expected base, with only six substitution types being reported: A>C, A>G, A>T, C>A, C>G and C>T. For example, T>G is grouped in with A>C.

The example is illustrative.  If we care about variant calling a SNV at a `A` (or `T`) reference positions, then we really don't care if the read (in sequencing order) had a `A>C` or `T>G`, since we can observe that reference base on either strand.

But if we care about the fidelity of sequencing, in sequencing order, we do not want to collapse `A>C` and `T>G`.   It may be the sequencer has much harder time reading through the top strand of the genome versus the bottom strand at the same SNV site above.

I think it should be as simple as adding some logic around these two lines:
https://github.com/fulcrumgenomics/fgbio/blob/c4b778baf0c780fa60a894c578804cec5398b829/src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala#L145-L146

@tfenne any objections?

tfenne:
No real objections.  I'd like the default to remain as it is though please.

I think the bulk of the work is really going to be in writing tests.  I think we essentially need to cover all combinations of:
* Reads mapped to the positive and negative strands
* Reads with A>N, C>N, G>N, T>N
And just make sure they all come out as expected.

PR #608, opened 2020-08-20
[feature] add an option do not collapse substitution types

nh13:
This is by default "off" in `fgbio ErrorRateByReadPosition`.

### DIFF
--- a/src/main/resources/com/fulcrumgenomics/bam/ErrorRateByReadPosition.R
+++ b/src/main/resources/com/fulcrumgenomics/bam/ErrorRateByReadPosition.R
@@ -48,7 +48,7 @@ error_rate_plot = ggplot(data) + aes(x=position) +
   geom_line(aes(y=c_to_a_error_rate, color="C>A")) +
   geom_line(aes(y=c_to_g_error_rate, color="C>G")) +
   geom_line(aes(y=c_to_t_error_rate, color="C>T"))
-if ("collapsed" %in% names(data) && data$collapsed[1] == "true") {
+if ("collapsed" %in% names(data) && data$collapsed[1] == "false") {
   error_rate_plot = error_rate_plot +
     geom_line(aes(y=g_to_a_error_rate, color="G>A")) +
     geom_line(aes(y=g_to_c_error_rate, color="G>C")) +
--- a/src/main/resources/com/fulcrumgenomics/bam/ErrorRateByReadPosition.R
+++ b/src/main/resources/com/fulcrumgenomics/bam/ErrorRateByReadPosition.R
@@ -48,7 +48,7 @@ error_rate_plot = ggplot(data) + aes(x=position) +
   geom_line(aes(y=c_to_a_error_rate, color="C>A")) +
   geom_line(aes(y=c_to_g_error_rate, color="C>G")) +
   geom_line(aes(y=c_to_t_error_rate, color="C>T"))
-if ("collapsed" %in% names(data) && data$collapsed[1] == "false") {
+if ("collapsed" %in% names(data) && data$collapsed[[1]] == "false") {
   error_rate_plot = error_rate_plot +
     geom_line(aes(y=g_to_a_error_rate, color="G>A")) +
     geom_line(aes(y=g_to_c_error_rate, color="G>C")) +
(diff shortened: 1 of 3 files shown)


nh13:
Fixes #599 .  I ran the script by hand to verify it plots the uncollapsed substitution types.

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -197,48 +201,62 @@ class ErrorRateByReadPosition
...
-    case ('C', 'G') => c_to_g_obs += 1
-    case ('C', 'T') => c_to_t_obs += 1
-    case _          => unreachable("Should never be invoked with a case other than the above.")
+/** Counter for substitutions for a given read number and position in that read.
+  *
+  * @param readNumber the read number
+  * @param position the position in the read
+  * @param collapse true to collapse substitution types, false otherwise
+  */
+private class ObsCounter(readNumber: Int, position: Int, collapse: Boolean) extends SimpleCounter[(Char, Char)] {
```
tfenne, line 210: I figured this would be slower, so I ran a little test on a sample BAM that I have kicking around.  The runtime of the metric collection phase (ignoring starting) grew by 50%.  I think this is all due to using SimpleCounter.  I have a suggestion on how to fix and will make a PR as I think that's easiest.

src/main/resources/com/fulcrumgenomics/bam/ErrorRateByReadPosition.R
```diff
@@ -39,18 +39,30 @@ ggplot(data) + aes(x=position) +
...
+# versions of fgbio
+error_rate_plot = ggplot(data) + aes(x=position) +
   geom_line(aes(y=a_to_c_error_rate, color="A>C")) +
   geom_line(aes(y=a_to_g_error_rate, color="A>G")) +
   geom_line(aes(y=a_to_t_error_rate, color="A>T")) +
   geom_line(aes(y=c_to_a_error_rate, color="C>A")) +
   geom_line(aes(y=c_to_g_error_rate, color="C>G")) +
-  geom_line(aes(y=c_to_t_error_rate, color="C>T")) +
+  geom_line(aes(y=c_to_t_error_rate, color="C>T"))
+if ("collapsed" %in% names(data) && data$collapsed[1] == "true") {
```
tfenne, line 51: Shouldn't this be `== "false"`?  Also, are you sure that R isn't going to auto-coerce the values to it's boolean type?

nh13, line 51:
```suggestion
if ("collapsed" %in% names(data) && data$collapsed[1] == "false") {
```

nh13: I ran the script directly, and it worked, so ¯\_(ツ)_/¯?

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -88,7 +88,11 @@ class ErrorRateByReadPosition
   @arg(flag='l', doc="Optional list of intervals to restrict analysis to.") val intervals: Option[PathToIntervals] = None,
   @arg(flag='d', doc="Include duplicate reads, otherwise ignore.") val includeDuplicates: Boolean = false,
   @arg(flag='m', doc="The minimum mapping quality for a read to be included.") val minMappingQuality: Int = 20,
-  @arg(flag='q', doc="The minimum base quality for a base to be included.") val minBaseQuality: Int = 0
+  @arg(flag='q', doc="The minimum base quality for a base to be included.") val minBaseQuality: Int = 0,
+  @arg(doc="Collapse substitution types based on the reference or expected base, with only six substitution" +
```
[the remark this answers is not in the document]
nh13: I like that it's name is the same as in the metric.  Keeping.

src/main/resources/com/fulcrumgenomics/bam/ErrorRateByReadPosition.R
nh13: Let me test it one more time 👍

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
nh13: Merged your PR!

PR #609, opened 2020-08-21
Updates to make the VCF api code considerably faster when reading VCFs with may samples

tfenne:
All of this is to make reading large VCFs with many samples not have significant overhead vs. using the HTSJDK api directly.  

I tested the runtime using a ~9k sample VCF with just GT in the genotype columns.  Before this PR reading with the HTSJDK API directly (and decoding genotypes) took ~10s, and the scala API took approximately 60s.  After these changes the scala API is down to ~15s.

@nh13 I think your review is optional, but I would appreciate a quick skim to see if you find anything objectionable.


nh13:
@tfenne I want to get your thoughts on the VCF API changes I made in https://github.com/fulcrumgenomics/fgbio/pull/610/files. I think it's going to be ok, but there will be conflicts.

PR #610, opened 2020-08-24
Adding a tool to validate VCFs

nh13:
There are other tools out there.  The motivation for this tool are:
1. Validates the VCF header, including against spec-reserved header values (id, type, and count)
2. Validates types and counts of variants (INFO) and genotypes (FORMAT) fields versus (1)
3. Allows for custom validators across variants/genotypes
4. Allows for quickly incorporating other types of validation (all field types, across genotypes/samples/variants/fields etc)
5. Other tools may not validate the spec strictly, as they are part of **G**enome **A**nalysis **T**ool **K**its that produce variants, sometimes in non-spec compliant ways.  We do not need to make those compromises
6. In Scala: succinct, concise, and performant.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/vcf/ValidateVcf.scala
+++ b/src/main/scala/com/fulcrumgenomics/vcf/ValidateVcf.scala
@@ -1,0 +1,193 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2020 Fulcrum Genomics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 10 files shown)


src/main/scala/com/fulcrumgenomics/vcf/ValidateVcf.scala
```diff
@@ -0,0 +1,183 @@
...
+import com.fulcrumgenomics.vcf.validation.{ValidationResult, VcfHeaderValidator, _}
+
+@clp(group=ClpGroups.VcfOrBcf, description=
+  """
+    |Validates a VCF.
+    |
+    |# Header
+    |
+    |Errors if:
+    |- Reserved INFO/FORMAT lines are have the proper type and count
```
yfarjoun, line 46: do not

src/main/scala/com/fulcrumgenomics/vcf/ValidateVcf.scala
```diff
@@ -0,0 +1,183 @@
...
+    |- Duplicate INFO/FORMAT identifiers are found
```
yfarjoun, line 47: filter?

src/main/scala/com/fulcrumgenomics/vcf/ValidateVcf.scala
```diff
@@ -0,0 +1,183 @@
...
+    |
+    |Errors if:
+    |- INFO/FORMAT field has the wrong type compared to the VCF header
+    |- INFO/FORMAT field has the wrong count compared to the VCF header
+    |- INFO/FORMAT field is not defined in the VCF header
+    |
+    |# Future work
+    |
+    |Validate:
+    |- values for specific fields (ex. CIGAR)
```
yfarjoun, line 65: sort order?

PR #612, opened 2020-08-24
Feature/fix vcf phase set

nh13:


src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+  private var variants: Long    = 0
+  private var genotypes: Long   = 0
+  val descriptionCounter: SimpleCounter[String] = new SimpleCounter()
+
+  def update(variant: Variant): Variant = {
+    // Phase sets cannot span contigs!  So empty our mappings if we get to a new one
+    if (variant.chrom != lastChrom) {
+      this.phaseSetToPositionBySample.values.foreach(_.clear())
+      this.lastChrom = variant.chrom
+    }
```
tfenne, line 189: Just looking at this, I'm guessing the solution below requires the VCF to be sorted?  Given that it's borked up in other ways do you want to check/assert that as you're iterating over the variants?  Maybe the chromosome ordering doesn't matter, but that you see each chromosome in a contiguous block of variants and the variants have rising positions?

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+
+    variants += 1
+    val genotypes: Map[String, Genotype] = variant
+      .genotypes
```
tfenne, line 193:
```suggestion
    val genotypes: Map[String, Genotype] = variant.genotypes
```

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+      .map { case (sampleName: String, genotype: Genotype) =>
+        // update the genotype
+        val result = updateGenotype(variant=variant, genotype=genotype)
+        // log the results
+        result match {
+          case NotPhasedWithPhaseSetValue(_)   =>
+            logger.warning(
+              "Genotype had a phase set but was unphased:" +
+                f"${variant.chrom}:${variant.pos}:${variant.id.getOrElse(".")} PS=${genotype("PS")}" +
```
tfenne, line 202:
```suggestion
                f"${variant.chrom}:${variant.pos}:${variant.id.getOrElse(".")} sample=${sampleName} PS=${genotype("PS")}" +
```

tfenne: Are you sure you want to emit this _every_ time?  I'd suggest counting up the number of these occurrences and logging at the end about how many you saw if there were > 0.  Adding to this argument is the fact that you refer to command line options for `FixVcfPhaseSet` even though you're in another class.

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+                "; consider r-running using `-x/--phase-genotypes-with-phase-set`"
+            )
+          case PhasedMissingPhaseSetValue(_) =>
+            logger.warning(
+              f"Genotype had no phase set but was phased: ${variant.chrom}:${variant.pos}:${variant.id.getOrElse(".")}"
```
tfenne, line 207: Similarly, I feel like if a VCF is really messy you'll get thousands or millions of lines of this.

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+        // return the (potentially) new/updated genotype
+        (sampleName, result.genotype)
+      }
+    variant.copy(genotypes=genotypes)
+  }
+
+  def updateGenotype(variant: Variant, genotype: Genotype): Result = {
+    this.genotypes += 1
+
+    (genotype.phased, genotype.get[Any]("PS")) match {
```
tfenne, line 221: I think that since you updated the header you should be able to use `get[String]` reliably here.  If that doesn't can you let me know please because that should get fixed.

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+      case (true, None)  => PhasedMissingPhaseSetValue(genotype) // phased but no phase set, what can we do?
```
tfenne, line 222: I wonder if this should be another option, to de-phase in this case?

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+      case (false, Some(_)) if !phaseGenotypesWithPhaseSet =>    // unphased and a phase set, but we do not want to convert it to phased
+        NotPhasedWithPhaseSetValue(genotype.copy(attrs=genotype.attrs.filterNot(_._1 == "PS"))) // remove the phase set for good measure
+      case (isPhased, Some(oldValue)) => // may or not be phased, but has a phase set
+        // Get the new phase set
+        val phaseSetToValue = this.phaseSetToPositionBySample(genotype.sample)
+        val oldValueString  = oldValue.toString
+        val newValue        = phaseSetToValue.getOrElse(oldValueString, {
+          phaseSetToValue(oldValueString) = variant.pos
+          variant.pos
+        })
```
tfenne, line 233:
```suggestion
        val newValue        = phaseSetToValue.getOrElseUpdate(oldValueString, variant.pos)
```

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+        // Build the new set of attributes
+        val phaseSetAttrs = {
+          if (keepOriginal) Map("PS" -> newValue, "OPS" -> oldValueString)
+          else Map("PS" -> newValue)
+        }
+        // update the genotype
+        val newGenotype = genotype.copy(attrs = genotype.attrs.filterNot(_._1 == "PS") ++ phaseSetAttrs, phased=true)
+        // return based on if we phased the variant and if we added/updated the phase set value
+        (isPhased, oldValue == newValue && isPhased) match {
```
tfenne, line 242: I don't _think_ the `oldValue == newValue` test will ever pass because `oldValue` should always be a `String`.

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+          case (true, true)   => Valid(newGenotype)
+          case (true, false)  => UpdatedPhaseSetValue(newGenotype)
+          case (false, true)  => UpdatedPhaseOnly(newGenotype)
```
tfenne, line 245: How can this be true?  Your match is on `isPhased` which must be false _and_ `oldValue == newValue && isPhased` which must then necessarily also be false.

src/main/scala/com/fulcrumgenomics/vcf/api/VcfSource.scala
```diff
@@ -42,9 +42,10 @@ import htsjdk.variant.vcf.{VCFCodec, VCFFileReader, VCFHeader}
   *
   * @param reader the underlying HTSJDK [[VCFFileReader]]
   */
-class VcfSource private(private val reader: AbstractFeatureReader[VariantContext, _]) extends View[Variant] with Closeable {
+class VcfSource private(private val reader: AbstractFeatureReader[VariantContext, _],
+                        _header: Option[VcfHeader]) extends View[Variant] with Closeable {
```
tfenne, line 46: I get why you need something like this but I'm not a big fan of it.  I'd like to suggest something a bit different:

1. How about we have it take an optional `headerTransformer: VcfHeader => VcfHeader` instead of a replacement header?  I think they're functionally equivalent, but this makes it clear the purpose is to allow you to fix/adjust the header rather than implying things might still work with a wholly different header.
2. We only expose it in the constructor and not the object.apply() so it's semi-hidden, or
3. We put it in _an_ apply but have two separate apply methods so it's super-obvious that you don't need it?

src/main/scala/com/fulcrumgenomics/vcf/api/VcfSource.scala
```diff
@@ -53,7 +54,7 @@ class VcfSource private(private val reader: AbstractFeatureReader[VariantContext
     */
   type VariantIterator = SelfClosingIterator[Variant]
 
-  /** Wraps an iterator provided by HTSJDK into a SelfClosingIterator that transforms VariatnContexts into Variants. */
+  /** Wraps an iterator provided by HTSJDK into a SelfClosingIterator that transforms VariantContexts into Variants. */
```
tfenne, line 57: Oops, thanks.

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+    |The `--keep-original` option may be used to store the original `PS` value in a new `OPS` field.  The type
+    |described in the header will match the original.
+    |
+    |This tool cannot fix phased variants without a phase set, or phased variant sets who have different phase set
+    |values.
+  """)
+class FixVcfPhaseSet
+( @arg(flag='i', doc="Input VCF.") val input: PathToVcf,
+  @arg(flag='o', doc="Output VCFs") val output: PathToVcf,
+  @arg(flag='s', doc="Samples to mix. See general usage for format and examples.", minElements=0) val samples: Seq[String] = Seq.empty,
```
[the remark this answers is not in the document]
nh13: This option should be removed

src/main/scala/com/fulcrumgenomics/vcf/FixVcfPhaseSet.scala
```diff
@@ -0,0 +1,250 @@
...
+  @arg(flag='k', doc="Store the original phase set in the `OPS` field.") val keepOriginal: Boolean = false,
+  @arg(flag='x', doc="Set unphased genotypes with a PS FORMAT value to be phased.") val phaseGenotypesWithPhaseSet: Boolean = false,
+  @arg(flag='z', doc="Remove header lines that do not the required ID.") val removeNoIdHeaderLines: Boolean = false
```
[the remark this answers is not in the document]
nh13: I don't think it helps to have a second tool at this time for this.

PR #614, opened 2020-08-25
Removed scala 2.12 cross-build support.

tfenne:


build.sbt
```diff
@@ -83,8 +83,8 @@ lazy val commonSettings = Seq(
   organizationName     := "Fulcrum Genomics LLC",
   homepage             := Some(url("http://github.com/fulcrumgenomics/fgbio")),
   startYear            := Some(2015),
-  scalaVersion         := "2.13.0",
-  crossScalaVersions   :=  Seq("2.12.12", "2.13.3"),
+  scalaVersion         := "2.13.3",
+  crossScalaVersions   :=  Seq("2.13.3"),
```
jacarey, line 87:
```suggestion
  crossScalaVersions   :=  Seq(scalaVersion),
```
Should we just reference the scalaVersion to keep them in sync?

tfenne: Did this, but slightly differently, by defining a val above all the common settings and then using it in both places.

PR #622, opened 2020-08-28
Add an option to SplitBam to reduce memory usage if the input has many read groups

nh13:


src/main/scala/com/fulcrumgenomics/bam/SplitBam.scala
```diff
@@ -55,77 +55,101 @@ object SplitType extends FgBioEnum[SplitType] {
...
     }
+    progress.logLast()
 
-    writers.values.foreach(_.close())
-    unknownBamAndWriter.close()
     in.safelyClose()
 
-    counter.toSeq sortBy { _._1.name } foreach { case (info, count) =>
-      logger.info(s"Wrote $count records to ${info.bam.getFileName}")
+    writerInfoMap.values.toSeq.sortBy(_.name).distinct.foreach { info =>
```
nh13, line 95: `.distinct` is added for good measure in case.  There could be multiple `WriterInfo`s that write to the same output (ex. `SplitByLibrary`)

src/main/scala/com/fulcrumgenomics/bam/SplitBam.scala
```diff
@@ -55,77 +55,101 @@ object SplitType extends FgBioEnum[SplitType] {
...
+      case SplitType.Library   => header.getReadGroups.toSeq.groupBy { rg => rg.getLibrary }
+      case SplitType.ReadGroup => header.getReadGroups.map(rg => rg.getId -> Seq(rg)).toMap
+    }
+    logger.info(f"Outputting to ${groups.size}%,d files ($splitBy).")
+    groups.flatMap { case (library, readGroups) =>
+      val bam    = toOutput(library)
+      val writer = SamWriter(
+        path   = bam,
+        header = toOutputHeader(header=header, readGroups:_*),
+        async  = !noAsyncWriting
```
nh13, line 124: If `--async-io` is used, then we'll have a thread per `SamWriter`, which for 10K read groups is too many for some systems.  Even when we up the thread-per-process limit, each thread will buffer, and so we use a too much memory.   It runs fast enough if we just disable async io for writing.

src/main/scala/com/fulcrumgenomics/bam/SplitBam.scala
```diff
@@ -55,77 +55,101 @@ object SplitType extends FgBioEnum[SplitType] {
...
-    outputDir.resolve(PathUtil.sanitizeFileName(s"$prefix.$name.bam"))
+  /** Stores the path to the BAM and the associated writer */
+  private case class WriterInfo(name: String, bam: PathToBam, writer: Writer[SamRecord], var count: Long = 0) extends Closeable {
+    def close(): Unit = writer.close()
+  }
+
+  /** Efficiently creates an output header with a single read group from the input header.
+    *
+    * Developer note: `SamFileHeader.clone` is very memory inefficient.  It creates a `String` from the header to clone,
+    * the decodes it!
```
nh13, line 138: Imagine having to parse the header for 10K `SamWriters`!  This takes forever.

src/main/scala/com/fulcrumgenomics/bam/SplitBam.scala
```diff
@@ -55,77 +55,101 @@ object SplitType extends FgBioEnum[SplitType] {
...
+    // HD
+    outputHeader.setAttribute(SAMFileHeader.VERSION_TAG, header.getVersion)
+    // SD
+    outputHeader.setSequenceDictionary(header.getSequenceDictionary)
+    // RG
+    outputHeader.setReadGroups(readGroup.iterator.toJavaList)
+    // PG
+    outputHeader.setProgramRecords(header.getProgramRecords)
+    // Comment
+    outputHeader.setComments(header.getComments)
```
nh13, line 151: I am not sure how to add non-standard header lines with htsjdk.  Do we care?

src/main/scala/com/fulcrumgenomics/bam/SplitBam.scala
nh13: Note that previously the output would contain all the read groups in the header.  In this version, only the single read group being output is added to the output header.

src/main/scala/com/fulcrumgenomics/bam/SplitBam.scala
```diff
@@ -55,77 +55,115 @@ object SplitType extends FgBioEnum[SplitType] {
...
-  private[bam] def toOutput(name: String): PathToBam = {
-    val outputDir = output.getParent
-    val prefix    = output.getFileName
-    outputDir.resolve(PathUtil.sanitizeFileName(s"$prefix.$name.bam"))
+  /** Stores the path to the BAM and the associated writer */
+  private case class WriterInfo(name: String, bam: PathToBam, writer: Writer[SamRecord], var count: Long = 0) extends Closeable {
+    def close(): Unit = writer.close()
+  }
+
+  /** Efficiently creates an output header with a single read group from the input header.
```
tfenne, line 144:
```suggestion
  /** Efficiently creates an output header with a subset of read groups from the input header.
```

tfenne (approved):
A few minor comments/suggestions.  Good to merge once you've fixed up.

PR #627, opened 2020-09-02
[feature] Add a tool to get percentage of bases falling in each interval by name.

jacarey:
Currently this expects single ended reads or paired end template reads to be full enclosed within an interval by name in order for it to be counted for that interval.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
+++ b/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
@@ -86,12 +86,13 @@ class AnnotateByIntervalList
     // an interval for each template. For single ended reads the interval is just the read.
     val templateIntervals = records.filter { record =>
       if (record.mateMapped) {
-        val equalStartTieBreaker =
+        val equalStartTieBreaker = {
           if (Math.min(record.start, record.end) == Math.min(record.mateStart, record.mateEnd.getOrElse(Int.MaxValue))) {
             record.firstOfPair
           } else {
             true
           }
+        }
         // Take only the first read by coordinate order
         Math.min(record.start, record.end) < Math.min(record.mateStart, record.mateEnd.getOrElse(Int.MaxValue)) &&
         // if mate start == start pick first of pair
--- a/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
+++ b/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
@@ -17,7 +17,7 @@ import htsjdk.samtools.util.{CoordMath, Interval, IntervalList, OverlapDetector}
 @clp(group = ClpGroups.Fastq, description =
   """
     |Collects base and read coverage statistics across intervals by name.
-    |A regular expression is used to classify the name of the intervals to be counted and how they should be grouped.
+    |A regular expression may used to classify the name of the intervals to be counted and how they should be grouped.  Otherwise, the name is used as-is.
     |
     |For example for the following interval list:
     |
--- a/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
+++ b/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
@@ -35,7 +35,7 @@ import htsjdk.samtools.util.{CoordMath, Interval, IntervalList, OverlapDetector}
     |
     |Using the the regex `([^;]*).*` would produce 3 separate interval counts CDS, utr and intron.
     |
-    |Counts template bases (paired-end) or read bases (single end) that are fully enclosed within an interval.
+    |Template bases (paired-end) or read bases (single end) are only counted if they fully enclosed within an interval.
     |By default if a read is enclosed by multiple intervals it will be counted for both intervals. This behavior
     |can be changed using a different `OverlapAssociationStrategy`
   """)
--- a/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
+++ b/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
@@ -105,8 +105,8 @@ class AnnotateByIntervalList
       // Interval should be end of read one to the beginning of read two
       new Interval(
         record.refName,
-        Math.max(record.start, record.end) + 1,
-        Math.min(record.mateStart, record.mateEnd.getOrElse(Int.MaxValue)) - 1
+        Math.max(record.start, record.mateStart) + 1,
+        Math.max(record.mateStart, record.mateEnd.getOrElse(exception) - 1
       )
     )
 
--- a/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
+++ b/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
@@ -84,7 +84,7 @@ class AnnotateByIntervalList
 
     // First we filter to ensure that we only pick one read in a template and then iterate over those reads to create
     // an interval for each template. For single ended reads the interval is just the read.
-    val templateIntervals = records.filter(record => {
+    val templateIntervals = records.filter { record =>
       if (record.mateMapped) {
         val equalStartTieBreaker =
           if (Math.min(record.start, record.end) == Math.min(record.mateStart, record.mateEnd.getOrElse(Int.MaxValue))) {
--- a/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
+++ b/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
@@ -16,7 +16,7 @@ import htsjdk.samtools.util.{CoordMath, Interval, IntervalList, OverlapDetector}
 
 @clp(group = ClpGroups.Fastq, description =
   """
-    |Get percentage of bases falling in each interval by name.
+    |Collects base and read coverage statistics across intervals by name.
     |A regular expression is used to classify the name of the intervals to be counted and how they should be grouped.
     |
     |For example for the following interval list:
--- a/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
+++ b/src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
@@ -1,0 +1,149 @@
+package com.fulcrumgenomics.rnaseq
+
+import java.util
+
+import com.fulcrumgenomics.bam.api.SamSource
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.commons.CommonsDef.{FilePath, PathToBam, PathToIntervals}
+import com.fulcrumgenomics.commons.util.{LazyLogging, SimpleCounter}
+import com.fulcrumgenomics.sopt.{arg, clp}
+import htsjdk.samtools.util.{Interval, IntervalList, OverlapDetector, SamLocusIterator}
+import com.fulcrumgenomics.FgBioDef.IteratorToJavaCollectionsAdapter
+import com.fulcrumgenomics.FgBioDef.javaIterableToIterator
+import com.fulcrumgenomics.commons.io.Io
+import com.fulcrumgenomics.util.{Metric, ProgressLogger}
+import htsjdk.samtools.SAMSequenceDictionary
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+  Io.assertReadable(this.input)
+  Io.assertReadable(this.intervals)
+  Io.assertCanWriteFile(this.output)
+
+  override def execute(): Unit = {
+    val progress = ProgressLogger(logger, verb = "Processed", noun = "loci", unit = 1000000)
+    val reader = SamSource(this.input)
+    val dict = reader.dict
+
+    val (baseCounter, uniqueBaseCounter, templateCounter, uniqueTemplateCounter, overlapDetector) = {
```
jacarey, line 64: Case class for this maybe?  Is there a threshold of tuple that we like to use for creating a higher ordered case class?

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    // Templates are assigned to an interval together.
+    val iterator = reader.iterator
+    val records = iterator.filterNot(record => record.duplicate
+      || record.secondary
+      || record.supplementary
+      || !record.pf
+      || !record.isFrPair
+      || record.mapq < minMappingQuality)
```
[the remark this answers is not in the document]
nh13: can you filter out reads where `record.start == record.mateStart` for paired end reads where both ends are mapped.

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+
+    // First we filter to ensure that we only pick one read in a template and then iterate over those reads to create
+    // an interval for each template. For single ended reads the interval is just the read.
+    val templateIntervals = records.filter(record => {
+      if (record.mateMapped) {
+        val equalStartTieBreaker =
+          if (Math.min(record.start, record.end) == Math.min(record.mateStart, record.mateEnd.getOrElse(Int.MaxValue))) {
```
nh13, line 90: Won't `Math.min(record.start, record.end)` always be `record.start`?   Also, what if this record is unmapped?

I think what you want is this:
1. record is unmapped and (not paired or mate is unmapped) => ignore
2. record is single-end (fragment) => `start=record.start` and `end=record.end`
3. record is mapped and mate is unmapped => same as (2)
4. record is unmapped and mate is mapped => use `record.mateStart` and `record.mateEnd.getOrElse(throw an exception saying you need a mate cigar)`
5. record is mapped and mate is mapped and `record.start == record.mateStart` and `record.firstOfPair` => start is `record.start` and end is `max(record.end, record.mateEnd.getOrElse(exception))`.
6. this should never happen since you filtered them out

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+      } else {
+        //If we are single ended then just take the record
+        true
+      }
+    }).map(record =>
+      // Interval should be end of read one to the beginning of read two
+      new Interval(
+        record.refName,
+        Math.max(record.start, record.end) + 1,
+        Math.min(record.mateStart, record.mateEnd.getOrElse(Int.MaxValue)) - 1
```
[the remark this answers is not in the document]
nh13: I would go back to my comment above about restructuring this section.

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+      )
+    )
+
+    var numLociMultiMatched: Long = 0
```
nh13, line 112: blasphemy.  can use `templateIntervals.count`

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+
+    templateIntervals.foreach { templateInterval =>
+      // Overlaps should only be flagged when they enclose the template bases.
+      val overlaps = overlapDetector.getOverlaps(templateInterval).toSeq
+        .filter(interval => CoordMath.encloses(interval.getStart, interval.getEnd,
+          templateInterval.getStart, templateInterval.getEnd))
```
nh13, line 118: My preference is a single line for a function call, or named parameters on multiple lines

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+
+      if (overlaps.isEmpty) {
+        baseCounter.count(this.unmatched, templateInterval.getLengthOnReference)
+        templateCounter.count(this.unmatched, 1)
+      } else {
+        val names: Seq[String] = overlaps.map(_.getName).distinct
+
+        def countForNames(namesToCount: Seq[String]): Unit = {
```
nh13, line 126: why not make this a private method?

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+
+    var numLociMultiMatched: Long = 0
+
+    templateIntervals.foreach { templateInterval =>
+      // Overlaps should only be flagged when they enclose the template bases.
+      val overlaps = overlapDetector.getOverlaps(templateInterval).toSeq
+        .filter(interval => CoordMath.encloses(interval.getStart, interval.getEnd,
+          templateInterval.getStart, templateInterval.getEnd))
+
+      if (overlaps.isEmpty) {
```
nh13, line 120: match statement instead?

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+          }
+          numLociMultiMatched += 1
+        } else {
+          // When there is only one interval matched we associate all intersected bases with the single interval
+          countForNames(names)
+        }
+      }
+      progress.record(templateInterval.getContig, templateInterval.getStart)
+    }
+    reader.close()
```
nh13, line 154: You'll need to import `SafelyCloseable` from `FgbioDef` for the second line to work
```suggestion
    progress.logLast()
    reader.safelyClose()
```

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+      )
+    }.toSeq.sortBy(_.name)
+    Metric.write(this.output, metrics)
+  }
+
+  private def buildIntervals(dict: SAMSequenceDictionary): List[Interval] = {
+    val inputIntervalList = IntervalList.fromPath(this.intervals)
+    this.nameRegex.map(_.r) match {
+      case None => inputIntervalList.uniqued(true).toList
+      case Some(regex) =>
```
nh13, line 193:
```suggestion
      case None        => inputIntervalList.uniqued(true).toList
      case Some(regex) =>
```

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+        val intervals = inputIntervalList.map { i =>
+          val name = i.getName match {
+            case regex(n) => n
+            case _ => throw new IllegalArgumentException(s"Could not match name '${i.getName}' with regex '${this.nameRegex}'")
```
nh13, line 197:
```suggestion
            case regex(n) => n
            case _        => throw new IllegalArgumentException(s"Could not match name '${i.getName}' with regex '${this.nameRegex}'")
```

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+        intervalList.toList
+    }
+  }
+}
+
+/** Metric for [[AnnotateByIntervalList]].
+  *
+  * @param name             the name of the intervals
+  * @param bases            the number of bases mapping to intervals with the given name.
+  * @param unique_bases     the number of bases mapping exclusively to intervals with the given name,
```
nh13, line 212: do you end the comment with a period, comma, or nothing?

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+  mean_coverage: Double,
+  templates: Long,
+  unique_templates: Long
+) extends Metric
+
+/**
+  * Enum for determining how to associate a read that is enclosed in more than a single interval.
+  * - Duplicate counting associates the template base with all enclosing intervals.
+  * - MostSpecific counting associates the template bases with the smallest enclosing interval.
+  * - LeastSpecific counting associates the template bases with the largest enclosing interval.
```
nh13, line 236: move the doc for each enum to where they're defined.

nh13 (changes requested):
Looking really good so far.  Besides by nitpicks (sorry).  There are two major issues:

1. how we filter reads prior to checking overlaps with intervals.
2. how we compute the "span" of a template

Happy to discuss at your leisure.

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
```diff
@@ -0,0 +1,246 @@
...
+      val intervals               = buildIntervals(dict.asSam)
+      intervals.foreach { interval => _baseCounter.count(interval.getName, 0) }
+
+      val _overlapDetector = OverlapDetector.create[Interval](intervals.iterator.toJavaList)
+      (_baseCounter, _uniqueBaseCounter, _templateCounter, _uniqueTemplateCounter, _overlapDetector)
+    }
+
+    // Templates are assigned to an interval together.
+    val iterator = reader.iterator
+    val records = iterator.filterNot(record => record.duplicate
```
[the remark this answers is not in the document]
jacarey: Increased compile time checking. It is an old habit that I keep forgetting to break.

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
jacarey: It isn't template intervals though. It is the number of templates that were enclosed by multiple intervals.  I should be able to calculate this from sum(templates) - sum(unique_templates) for all metrics now that I'm collecting them

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
jacarey: It uses local variables. Moving it out of this function scope means I need to pass them around (not a big deal I guess)

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
jacarey: not sure I understand. what is the match pattern on a Seq to check if it is empty?  is that just

`      overlaps match {
        case ol if ol.isEmpty =>
        case _ =>
      }
`

or did you mean

`      overlaps.isEmpty match {
        case true =>
        case false => 
      }
`
In either case I think a simple if else is clearer no?

src/main/scala/com/fulcrumgenomics/rnaseq/AnnotateByIntervalList.scala
jacarey: I think this handles 2, 3, 4, 5
```
        // Single end read mateStart will be set to 0
        Math.min(record.start, record.mateStart),
        // Set mate end to math min if it isn't defined (single end read)
        Math.max(record.end, record.mateEnd.getOrElse(Int.MinValue))
```
I don't throw an exception on `record.mateEnd.getOrElse` since that will mess up single ended reads.

Also for #4 there is already a require() in the accessor method to make sure mate cigar is set.

PR #637, opened 2020-12-09
Update EstimatePoolingFractions to be able to use GT.AF for per-sample allele frequencies

tfenne:
EstimatePoolingFractions is used when we do mixture experiments (i.e. mixing some number of samples), using known genotypes to accurately estimate the pooling of the samples (i.e. what fraction of the pool does each sample represent).  Up until this PR the assumption was that input samples were simple diploid germline samples, and so when performing the regression we estimate the fraction of each sample that is ALT at any SNP by a simple: homref -> 0, het -> 0.5, homvar -> 1.

What this change does is switches to use the per-sample `AF` field _if_ it is available.  If `AF` is not present for any given sample it will fall back to using the 0/0.5/1 system.  The motivation for this is to allow multi-level pooling using `EstimatePoolingFractions` and `MakeMixtureVcf`.  E.g. imagine a case where you mixed 10 samples equally into _Pool A_ and then 2 samples equally into _Pool B_, and then made a final pool of 95% _Pool B_ with only 5% _Pool A_.  In theory you could run `EstimatePoolingFractions` and feed it all twelve input samples and if you have deep enough coverage of enough SNPs it would do the right thing.  However given the very uneven target pooling, with less than ideal sequencing, it is very hard to accurately estimate the constituents of _Pool A_ in the final pool.  What this allows for is an experiment where you also sequence _Pool A_ and _Pool B_ in addition to the final pool, you can now:

1. EstimatePoolingFractions on _Pool A_ and _Pool B_
2. Use the resulting estimates to make mixture VCFs using `MakeMixtureVcf` which populates the `GT.AF` field
3. Run EstimatePoolingFractions on the final pool treating it as a mixture of the _Pool A_ and _Pool B_ mixtures.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
@@ -33,12 +33,9 @@ import com.fulcrumgenomics.commons.util.LazyLogging
 import com.fulcrumgenomics.sopt.{arg, clp}
 import com.fulcrumgenomics.util.Metric.{Count, Proportion}
 import com.fulcrumgenomics.util.{Io, Metric, Sequences}
-import com.fulcrumgenomics.vcf.ByIntervalListVariantContextIterator
 import com.fulcrumgenomics.vcf.api.{Variant, VcfSource}
 import htsjdk.samtools.util.SamLocusIterator.LocusInfo
 import htsjdk.samtools.util._
-import htsjdk.variant.variantcontext.VariantContext
-import htsjdk.variant.vcf.VCFFileReader
 import org.apache.commons.math3.stat.regression.OLSMultipleLinearRegression
 
 @clp(group=ClpGroups.SamOrBam, description=
@@ -48,6 +45,11 @@ import org.apache.commons.math3.stat.regression.OLSMultipleLinearRegression
     |mixture along with a BAM of sequencing data derived from the pool.  Performs a multiple regression
     |for the alternative allele fractions at each SNP locus, using as inputs the individual sample's genotypes.
     |Only SNPs that are bi-allelic within the pooled samples are used.
+    |
+    |Each sample's contribution of REF vs. ALT alleles at each site is derived in one of two ways. If
+    |the sample's genotype in the VCF has the `AF` attribute then the value from that field will be used.  If the
+    |genotype has no AF attribute then the contribution is estimated based on the genotype (e.g. 0/0 will be 100%
+    |ref, 0/1 will be 50% ref and 50% alt, etc.).
     |
     |Various filtering parameters can be used to control which loci are used:
     |
@@ -90,7 +92,10 @@ class EstimatePoolingFractions
       alt   = v.alleles.alts.head.value.charAt(0),
       expectedSampleFractions = sampleNames.map { s =>
         val gt = v.gt(s)
-        if (gt.isHomRef) 0 else if (gt.isHet) 0.5 else 1.0
+        gt.get[IndexedSeq[Float]]("AF") match {
+          case None      => if (gt.isHomRef) 0 else if (gt.isHet) 0.5 else 1.0
+          case Some(afs) => afs(0)
+        }
       }
     )}.toArray
 
--- a/src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
@@ -34,6 +34,7 @@ import com.fulcrumgenomics.sopt.{arg, clp}
 import com.fulcrumgenomics.util.Metric.{Count, Proportion}
 import com.fulcrumgenomics.util.{Io, Metric, Sequences}
 import com.fulcrumgenomics.vcf.ByIntervalListVariantContextIterator
+import com.fulcrumgenomics.vcf.api.{Variant, VcfSource}
 import htsjdk.samtools.util.SamLocusIterator.LocusInfo
 import htsjdk.samtools.util._
 import htsjdk.variant.variantcontext.VariantContext
@@ -76,18 +77,21 @@ class EstimatePoolingFractions
   case class Locus(chrom: String, pos: Int, ref: Char, alt: Char, expectedSampleFractions: Array[Double], var observedFraction: Option[Double] = None)
 
   override def execute(): Unit = {
-    val vcfReader   = new VCFFileReader(vcf.toFile)
+    val vcfReader   = VcfSource(vcf)
     val sampleNames = pickSamplesToUse(vcfReader)
     val intervals   = loadIntervals
 
     // Get the expected fractions from the VCF
     val vcfIterator = constructVcfIterator(vcfReader, intervals, sampleNames)
-    val loci = vcfIterator.filterNot(v => this.nonAutosomes.contains(v.getContig)).map { v => Locus(
-      chrom = v.getContig,
-      pos = v.getStart,
-      ref = v.getReference.getBaseString.charAt(0),
-      alt = v.getAlternateAllele(0).getBaseString.charAt(0),
-      expectedSampleFractions = sampleNames.map { s => val gt = v.getGenotype(s); if (gt.isHomRef) 0 else if (gt.isHet) 0.5 else 1.0 }
+    val loci = vcfIterator.filterNot(v => this.nonAutosomes.contains(v.chrom)).map { v => Locus(
+      chrom = v.chrom,
+      pos   = v.pos,
+      ref   = v.alleles.ref.bases.charAt(0),
+      alt   = v.alleles.alts.head.value.charAt(0),
+      expectedSampleFractions = sampleNames.map { s =>
+        val gt = v.gt(s)
+        if (gt.isHomRef) 0 else if (gt.isHet) 0.5 else 1.0
+      }
     )}.toArray
 
     logger.info(s"Loaded ${loci.length} bi-allelic SNPs from VCF.")
@@ -134,15 +138,12 @@ class EstimatePoolingFractions
   }
 
   /** Verify a provided sample list, or if none provided retrieve the set of samples from the VCF. */
-  private def pickSamplesToUse(vcfReader: VCFFileReader): Array[String] = {
-    if (samples.nonEmpty) {
-      val samplesInVcf = vcfReader.getFileHeader.getSampleNamesInOrder.iterator.toSet
-      val missingSamples = samples.filterNot(samplesInVcf.contains)
+  private def pickSamplesToUse(vcfIn: VcfSource): Array[String] = {
+    if (this.samples.isEmpty) vcfIn.header.samples.toArray else {
+      val samplesInVcf   = vcfIn.header.samples
+      val missingSamples = samples.toSet.diff(samplesInVcf.toSet)
       if (missingSamples.nonEmpty) fail(s"Samples not present in VCF: ${missingSamples.mkString(", ")}")
       else samples.toArray.sorted
-    }
-    else {
-      vcfReader.getFileHeader.getSampleNamesInOrder.iterator.toSeq.toArray.sorted // toSeq.toArray is necessary cos util.ArrayList.toArray() exists
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -48,6 +46,11 @@ import org.apache.commons.math3.stat.regression.OLSMultipleLinearRegression
     |for the alternative allele fractions at each SNP locus, using as inputs the individual sample's genotypes.
     |Only SNPs that are bi-allelic within the pooled samples are used.
     |
+    |Each sample's contribution of REF vs. ALT alleles at each site is derived in one of two ways. If
+    |the sample's genotype in the VCF has the `AF` attribute then the value from that field will be used.  If the
+    |genotype has no AF attribute then the contribution is estimated based on the genotype (e.g. 0/0 will be 100%
+    |ref, 0/1 will be 50% ref and 50% alt, etc.).
+    |
```
nh13, line 53:
```suggestion
    |Each sample's contribution of REF vs. ALT alleles at each site is derived in one of two ways: (1) if
    |the sample's genotype in the VCF has the `AF` attribute then the value from that field will be used, (2) if the
    |genotype has no `AF` attribute then the contribution is estimated based on the genotype (e.g. 0/0 will be 100%
    |ref, 0/1 will be 50% ref and 50% alt, etc.).
    |
```

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -76,18 +79,24 @@ class EstimatePoolingFractions
   case class Locus(chrom: String, pos: Int, ref: Char, alt: Char, expectedSampleFractions: Array[Double], var observedFraction: Option[Double] = None)
 
   override def execute(): Unit = {
-    val vcfReader   = new VCFFileReader(vcf.toFile)
+    val vcfReader   = VcfSource(vcf)
```
nh13, line 82: does this ever get closed?  Perhaps by the iterator below?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
nh13: Also, can't comment below, but it may be useful to say that by default common non-autosomal chromosomes are skipped.  Why doesn't the SAM header have a field for this?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -76,18 +79,24 @@ class EstimatePoolingFractions
...
-      expectedSampleFractions = sampleNames.map { s => val gt = v.getGenotype(s); if (gt.isHomRef) 0 else if (gt.isHet) 0.5 else 1.0 }
+    val loci = vcfIterator.filterNot(v => this.nonAutosomes.contains(v.chrom)).map { v => Locus(
+      chrom = v.chrom,
+      pos   = v.pos,
+      ref   = v.alleles.ref.bases.charAt(0),
+      alt   = v.alleles.alts.head.value.charAt(0),
+      expectedSampleFractions = sampleNames.map { s =>
+        val gt = v.gt(s)
+        gt.get[IndexedSeq[Float]]("AF") match {
+          case None      => if (gt.isHomRef) 0 else if (gt.isHet) 0.5 else 1.0
```
nh13, line 96:
```suggestion
          // Developer note: it is important we only consider bi-allelic sites, since a call could be heterozygous alt1-alt2
          case None      => if (gt.isHomRef) 0 else if (gt.isHet) 0.5 else 1.0
```

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -134,16 +143,13 @@ class EstimatePoolingFractions
...
 
   /** Verify a provided sample list, or if none provided retrieve the set of samples from the VCF. */
-  private def pickSamplesToUse(vcfReader: VCFFileReader): Array[String] = {
-    if (samples.nonEmpty) {
-      val samplesInVcf = vcfReader.getFileHeader.getSampleNamesInOrder.iterator.toSet
-      val missingSamples = samples.filterNot(samplesInVcf.contains)
+  private def pickSamplesToUse(vcfIn: VcfSource): Array[String] = {
+    if (this.samples.isEmpty) vcfIn.header.samples.toArray else {
+      val samplesInVcf   = vcfIn.header.samples
+      val missingSamples = samples.toSet.diff(samplesInVcf.toSet)
```
nh13, line 149: I prefer the old implementation as it's clearer that it's samples specified on the command line that are missing in the VCF header.  `diff` in `Set` has always bugged me that it's directional.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -163,20 +169,18 @@ class EstimatePoolingFractions
   }
 
   /** Generates an iterator over non-filtered bi-allelic SNPs where all the required samples are genotyped. */
-  def constructVcfIterator(in: VCFFileReader, intervals: Option[IntervalList], samples: Array[String]): Iterator[VariantContext] = {
-    val vcfIterator: Iterator[VariantContext] = intervals match {
+  def constructVcfIterator(in: VcfSource, intervals: Option[IntervalList], samples: Seq[String]): Iterator[Variant] = {
+    val iterator: Iterator[Variant] = intervals match {
       case None     => in.iterator
-      case Some(is) => ByIntervalListVariantContextIterator(in, is)
+      case Some(is) => is.flatMap(i => in.query(i.getContig, i.getStart, i.getEnd))
```
nh13, line 175: Probably better to use `ByIntervalListVariantContextIterator` so we handle the case if a variant overlaps multiple intervals, so the variant is only returned once.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -163,20 +169,18 @@ class EstimatePoolingFractions
...
-
-    vcfIterator
-      .filterNot(_.isFiltered)
-      .map(_.subContextFromSamples(samplesAsUtilSet, true))
-      .filter(v => v.isSNP && v.isBiallelic && !v.isMonomorphicInSamples)
-      .filter(_.getNoCallCount == 0)
-      .filter(v => v.getGenotypesOrderedByName.iterator.forall(gt => gt.getGQ >= this.minGenotypeQuality))
+    iterator
+      .filter(v => v.filters.isEmpty || v.filters == Variant.PassingFilters)
+      .filter(v => v.alleles.size == 2 && v.alleles.forall(a => a.value.length == 1))  // Just biallelic SNPs
```
nh13, line 180:
```suggestion
      .filter(v => v.alleles.size == 2 && v.alleles.forall(a => a.value.length == 1))  // Just biallelic SNVs
```

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -163,20 +169,18 @@ class EstimatePoolingFractions
...
+      .filter(v => samples.map(v.gt).forall(gt => gt.isFullyCalled && (this.minGenotypeQuality <= 0 || gt.get[Int]("GQ").exists(_ >= this.minGenotypeQuality))))
+      .map   (v => v.copy(genotypes=v.genotypes.filter { case (s, _) => samples.contains(s)} ))
+      .filter(v => v.gts.flatMap(_.calls).toSet.size > 1)  // Not monomorphic
```
nh13, line 183:
```suggestion
      .filter(v => v.gts.flatMap(_.calls).distinct.size > 1)  // Not monomorphic
```

nh13: I have a core belief that `distinct` is faster

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
```diff
@@ -163,20 +169,18 @@ class EstimatePoolingFractions
...
-      .filterNot(_.isFiltered)
-      .map(_.subContextFromSamples(samplesAsUtilSet, true))
-      .filter(v => v.isSNP && v.isBiallelic && !v.isMonomorphicInSamples)
-      .filter(_.getNoCallCount == 0)
-      .filter(v => v.getGenotypesOrderedByName.iterator.forall(gt => gt.getGQ >= this.minGenotypeQuality))
+    iterator
+      .filter(v => v.filters.isEmpty || v.filters == Variant.PassingFilters)
+      .filter(v => v.alleles.size == 2 && v.alleles.forall(a => a.value.length == 1))  // Just biallelic SNPs
+      .filter(v => samples.map(v.gt).forall(gt => gt.isFullyCalled && (this.minGenotypeQuality <= 0 || gt.get[Int]("GQ").exists(_ >= this.minGenotypeQuality))))
+      .map   (v => v.copy(genotypes=v.genotypes.filter { case (s, _) => samples.contains(s)} ))
```
nh13, line 182: Should we subset to the samples we care about before we apply the genotype-quality and fully called filters above?

nh13 (changes requested):
I had a few questions about some implementation changes, for example how you iterate over variants if intervals are provided.  Also, why we apply some genotype-specific filters prior to subsetting the samples in question.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: Added a different comment.  Also it makes me laugh when you start comments with "Developer note" - like what else are comments in code?  😛

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: I mean, set difference is always a directional, I feel like this is like complaining that `minus` is directional for integers.  Still, maybe it would be clearer as `samples -- samplesInVcf`?

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: Except that that is all written using the old HTSJDK API and I'm not up for taking the time to convert it.  I'm not overly concerned about that for now.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: It would make the code a little shorter, but I was trying to avoid a `copy()` operation on variants that we're not going to use.  Hence above I check only the genotypes for samples that we care about, and then do the copy.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: It is _at least_ as slow.  On an iterator they're about equivalent - the underlying method builds up a `HashSet` as it traverses and then skips any elements that are already present in the set. On e.g. `Seq` it's much worse as it basically builds up a `Set` and once it's done, then generates a new `Seq` with all the elements from the `Set`.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
nh13: Ha!

But why did you change it? Is more performant?

tfenne: `x.diff(y)` == `x -- y`.  I just figured the latter would be more to your liking.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
nh13: https://gist.github.com/Mortimerp9/365fd4f306b70b2570e6

Tried googling but no luck.

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: Added a close().

src/main/scala/com/fulcrumgenomics/bam/EstimatePoolingFractions.scala
tfenne: I have no faith in that benchmark.  If you look at `Iterator.distinct` it vectors to `Iterator.distinctBy(identity)`, which looks like this:

```scala
  def distinctBy[B](f: A => B): Iterator[A] = new AbstractIterator[A] {

    private[this] val traversedValues = mutable.HashSet.empty[B]
    private[this] var nextElementDefined: Boolean = false
    private[this] var nextElement: A = _

    def hasNext: Boolean = nextElementDefined || (self.hasNext && {
      val a = self.next()
      if (traversedValues.add(f(a))) {
        nextElement = a
        nextElementDefined = true
        true
      }
      else hasNext
    })

    def next(): A =
      if (hasNext) {
        nextElementDefined = false
        nextElement
      } else {
        Iterator.empty.next()
      }
  }
```

So you can see it's adding every element to a HashSet as it iterates.  Which is what `toSet` does after you dig through 16 layers of builder nonsense.

nh13 (approved):

PR #641, opened 2020-12-17
Allow PickIlluminaIndices to choose from an existing set of candidates

nh13:


src/main/java/com/fulcrumgenomics/util/PickIlluminaIndicesCommand.java
```diff
@@ -318,6 +319,7 @@ protected int execute() {
             if (first.minEditDistance() > lastMinEdit) {
                 lastMinEdit = first.minEditDistance();
                 log.info("There are " + count + " indices with a minimum edit distance of " + lastMinEdit);
+                if (count == N_INDICES) break;
```
nh13, line 322: This is a corner case bug.  If we have N candidate indices and also want N candidate indices, then without this we remove one and so return `N-1` indices.  This only mattered when the # of candidate indices was the same as the target # of indices, which is non-sensical without providing a list of indices from which to pick

tfenne (approved):
My only request is for better usage on the CLP.  I think the main usage should mention this option since it significnatly changes how the tool works.  And in the param usage I think you should mention it expects a file with one index per line and nothing else.

Issue #579, opened 2020-04-10
Scala API for accessing FASTA reference sequences


tfenne:
Anywhere in fgbio we need to access FASTA for reference genomes or similar we end up using HTSJDK's `ReferenceSequenceFile` and related classes (`ReferenceSequenceFileFactory`, `ReferenceSequence`).  These classes get the job done, and offer a fair amount of functionality, but they are old and their design is a little clunky, especially coming from scala.

We have previously generated scala APIs for concepts in HTSJDK for both SAM/BAM and for VCF.  We'd now like to do that for reference sequences.  I'm 99.9% sure the right way to do this is to wrap the HTSJDK classes and try to _never_ expose them in the scala API (similar to the VCF wrapper, and less like the SAM/BAM wrapper).

Some things that would be nice in the design:

- A more scala-esque API in general (use of `apply()` where it makes sense)
- A unified API that makes sense for a) accessing an indexed file, b) accessing a non-index file, c) pulling ranges out of a sequence already loaded into memory
- The ability to perform transforms on loaded sequences (e.g. force everything to be upper case, mask all lower case bases to Ns, etc.)
- Clarity in the API about when you're accessing metadata about a sequence vs. when you're pulling sequence data into memory

It may also be that it's time for us to think about having a scala implementation/wrapper of `SequenceDictionary` as that's often used heavily in conjunction with reference sequences, and is also super clunky.

FWIW I also have code kicking around somewhere for various activities that might be useful to think about and/or pull into fgbio as use cases for this, including:
- Invoking HTSJDK functions to generate `.fai` files
- "Normalizing" fasta files to a standard line length
- Building up a sequence dictionary from a FASTA and set of per-chromosome aliases

I would suggest starting this ticket by:

- Reviewing existing implementations including HTSJDK's, pyfaidx and any others you can think of
- Reviewing how we use the existing HTSJDK implementations through fgbio and possibly elsewhere
- Sketching out a _high level_ design of the _public API_ for review by myself and @nh13 

@nh13 Anything to add?

sstadick:
## Analysis

### Overview of existing implementations:

**HTSJDK**

- factory based object creation
- anything using `getSequenceDict` / `getSequence` will error if no index is present
- method based for access

**rust-bio**

- Nothing particularly interesting here other than Indexed readers are totally seperate and explicit

**pyfaidx**

- dict based access to sequences by name
- sequences are treated like Lists for indexing etc
- index file auto generated if it does not exist

[...]

PR #644, opened 2021-01-04
Document that sequence dictionary converts are computationally expensive, and add a cache

nh13:
CC: @mjhipp https://github.com/fulcrumgenomics/fgbio/issues/579#issuecomment-754228920

This PR has two commits for consideration:
1. document that repeatedly converting sequence dictionaries using the provided implicits is expensive
2. add a cache so that "it just works" for anybody using the implicits.  This is so that anybody that converts over to the new `SequenceDictionary` class and friends don't experience the same pain as @mjhipp 

100 iterations on hs38DH

```
[2021/01/04 15:40:10 | FgBioMain | Info] Executing DictLoop from fgbio version 1.4.0-22b0efc-SNAPSHOT as nhomer@nils-fg.local on JRE 1.8.0_212-b10 with snappy, JdkInflater, and JdkDeflater
[2021/01/04 15:40:11 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:40:11 | DictLoop | Info] Accessing SAMSequenceDictionary 100 times WITHOUT conversion step.
[2021/01/04 15:40:11 | DictLoop | Info] Elapsed time: 1.690521 ms.
[2021/01/04 15:40:11 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:40:11 | DictLoop | Info] Accessing fasta.SequenceDictionary 100 times WITHOUT conversion step.
[2021/01/04 15:40:11 | DictLoop | Info] Elapsed time: 0.413097 ms.
[2021/01/04 15:40:11 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:40:11 | DictLoop | Info] Accessing SAMSequenceDictionary 100 times with asSam conversion step.
[2021/01/04 15:40:11 | DictLoop | Info] Elapsed time: 39.607902 ms.
[2021/01/04 15:40:11 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:40:11 | DictLoop | Info] Accessing fasta.SequenceDictionary 100 times with fromSam conversion step.
[2021/01/04 15:40:11 | DictLoop | Info] Elapsed time: 12.715213 ms.
[2021/01/04 15:40:11 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:40:11 | FgBioMain | Info] DictLoop completed. Elapsed time: 0.03 minutes.
```

100,000 iterations on hs38DH:

```
[2021/01/04 15:37:14 | FgBioMain | Info] Executing DictLoop from fgbio version 1.4.0-22b0efc-SNAPSHOT as nhomer@nils-fg.local on JRE 1.8.0_212-b10 with snappy, JdkInflater, and JdkDeflater
[2021/01/04 15:37:14 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:37:14 | DictLoop | Info] Accessing SAMSequenceDictionary 100000 times WITHOUT conversion step.
[2021/01/04 15:37:14 | DictLoop | Info] Elapsed time: 5.879807 ms.
[2021/01/04 15:37:14 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:37:14 | DictLoop | Info] Accessing fasta.SequenceDictionary 100000 times WITHOUT conversion step.
[2021/01/04 15:37:14 | DictLoop | Info] Elapsed time: 3.194493 ms.
[2021/01/04 15:37:14 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:37:14 | DictLoop | Info] Accessing SAMSequenceDictionary 100000 times with asSam conversion step.
[2021/01/04 15:37:14 | DictLoop | Info] Elapsed time: 44.157055 ms.
[2021/01/04 15:37:14 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:37:14 | DictLoop | Info] Accessing fasta.SequenceDictionary 100000 times with fromSam conversion step.
[2021/01/04 15:37:16 | DictLoop | Info] Elapsed time: 1841.079818 ms.
[2021/01/04 15:37:16 | DictLoop | Info] -----------------------------------------------------------------------------
[2021/01/04 15:37:16 | FgBioMain | Info] DictLoop completed. Elapsed time: 0.06 minutes.
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
@@ -272,10 +272,16 @@ case class SequenceDictionary(infos: IndexedSeq[SequenceMetadata]) extends Itera
 }
 
 
-/** Contains useful converters to and from HTSJDK objects. */
+/** Contains useful converters to and from HTSJDK objects.
+  *
+  * These converters are computationally expensive, so it is best to only use them once when converting a sequence
+  * dictionary or sequence record.
+  *
+  *
+  * */
 object Converters {
 
-  /** Converter from a [[SequenceMetadata]] to a [[SAMSequenceRecord]] */
+  /** Converter from a [[SequenceMetadata]] to a [[SAMSequenceRecord]].  Warning: this is computationally expensive. */
   implicit class ToSAMSequenceRecord(info: SequenceMetadata) {
     def asSam: SAMSequenceRecord = {
       val rec = new SAMSequenceRecord(info.name, info.length)
@@ -285,7 +291,7 @@ object Converters {
     }
   }
 
-  /** Converter from a [[SAMSequenceRecord]] to a [[SequenceMetadata]] */
+  /** Converter from a [[SAMSequenceRecord]] to a [[SequenceMetadata]].  Warning: this is computationally expensive. */
   implicit class FromSAMSequenceRecord(rec: SAMSequenceRecord) {
     def fromSam: SequenceMetadata = {
       val attributes: Map[String, String] = rec.getAttributes.map { entry =>
@@ -300,7 +306,7 @@ object Converters {
     }
   }
 
-  /** Converter from a [[SequenceDictionary]] to a [[SAMSequenceDictionary]] */
+  /** Converter from a [[SequenceDictionary]] to a [[SAMSequenceDictionary]].  Warning: this is computationally expensive. */
   implicit class ToSAMSequenceDictionary(infos: SequenceDictionary) {
     def asSam: SAMSequenceDictionary = {
       val recs = infos.iterator.zipWithIndex.map { case (info, index) =>
@@ -310,7 +316,7 @@ object Converters {
     }
   }
 
-  /** Converter from a [[SAMSequenceDictionary]] to a [[SequenceDictionary]] */
+  /** Converter from a [[SAMSequenceDictionary]] to a [[SequenceDictionary]].  Warning: this is computationally expensive. */
   implicit class FromSAMSequenceDictionary(dict: SAMSequenceDictionary) {
     def fromSam: SequenceDictionary = SequenceDictionary(dict.getSequences.map(_.fromSam).toIndexedSeq)
   }
--- a/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
@@ -156,7 +156,7 @@ case class SequenceMetadata private[fasta]
     this.get(Keys.AlternateLocus).flatMap {
       case "*"   => None
       case range =>
-        val locus = FgBioDef.parseRange(range) match {
+        val locus = FgBioDef.GenomicRange(range) match {
           case GenomicRange("=", start, end) => AlternateLocus(refName=this.name, start=start.toInt, end=end.toInt)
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -156,7 +156,7 @@ case class SequenceMetadata private[fasta]
     this.get(Keys.AlternateLocus).flatMap {
       case "*"   => None
       case range =>
-        val locus = FgBioDef.parseRange(range) match {
+        val locus = FgBioDef.GenomicRange(range) match {
```
nh13, line 159: This is unrelated

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -190,6 +190,14 @@ case class SequenceMetadata private[fasta]
     import com.fulcrumgenomics.fasta.Converters.ToSAMSequenceRecord
     this.asSam.getSAMString
   }
+
+  override lazy val hashCode: Int = {
```
nh13, line 194: caching the hashcode is a big win

mjhipp:
This is great. Thanks @nh13!

src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
```diff
@@ -269,25 +277,42 @@ case class SequenceDictionary(infos: IndexedSeq[SequenceMetadata]) extends Itera
...
+  * These converters are computationally expensive, so it is best to only use them once when converting a sequence
+  * dictionary or sequence record.
+  *
+  *
+  * */
 object Converters {
 
-  /** Converter from a [[SequenceMetadata]] to a [[SAMSequenceRecord]] */
+  /** True to enable caching of previous conversions, false otherwise. */
+  var EnableCache: Boolean = true
```
nh13, line 295: I'd be open to making this `false` by default

tfenne (changes requested):
I think the caching here is really problematic on a couple of fronts.

1. Synchronization - the caches are all mutable HashMaps which are not internally synchronized, and they is no protection stopping them from being written to from multiple threads simultaneously (nor read simultaneously with writes).  So if e.g. someone were to do something (like read multiple BAMs) on multiple threads and cause conversions to take place they could well experience exceptions due to concurrent updates of the cache.

2. It's basically a memory leak.  If I have a large sequence dictionary that I use briefly, or many sequence dictionaries that I use briefly, this cache will cause them all to be retained in memory and that's quite counter-intuitive.  My best suggestion for this would be to use a `WeakHashMap` instead which will allow the GC to remove things from the cache when they are not referenced/held elsewhere.  The downside of that though is that the it uses weak references on the keys, and so if you read 1 BAM and populate the cache, then let go of that BAM's header, the GC is free to reclaim those entries from cache, even if another BAM header with the logically same sequence dictionary is in memory.  I think the ideal thing here would be have a `SoftHashMap` that leverages `SoftReference`s which are somewhat similar to `WeakReferences` but have semantics that better match our need - specifically `WeakReferences` are always cleared once there are only `WeakReferences` to an object (i.e. it's greedy) whereas `SoftReferences` are only cleared when a) there are only `SoftReferences` to an object _and_ the GC needs to reclaim that memory to avoid an out of memory error.

PR #646, opened 2021-01-19
Update r-packages-source in .travis.yml

mjhipp:
I noticed Travis builds were failing with a similar error that I have encountered in the past. This worked for me previously.


nh13:
Thanks @mjhipp !

PR #648, opened 2021-01-28
GroupReadsByUmi optionally allows inter-contig pairs

### LINKED ISSUE #634: Accepting discordant pairs in GroupReadsByUmi
Hello, 
This is a new feature request. 
In the note of `GroupBreadsByUmi`, it says

> Reads are aggressively filtered out so that only high quality reads/mappings are taken forward. Paired-end reads must have both reads mapped to the same chromosome with both reads having mapping quality >= min-mapq. (Note: the MQ tag is required on reads with mapped mates).

However, we are particularly interested in the discordant pairs as we are looking for SVs. Is it possible to add a switch to loosen this restriction? 

I hope this new feature also propagate to `CallMolecularConsensusReads` and `CallDuplexConsensusReads`

mjhipp:
Closes #634 

In order for pairs from different chromosomes to be grouped together, `ReadInfo` was changed to prioritize the chromosome, and take the lower chromosome of the two reads.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -83,23 +83,23 @@ object GroupReadsByUmi {
         tmp
       }
       else {
-        val lib     = library(rec)
-        val chrom   = rec.refIndex
-        val recNeg  = rec.negativeStrand
-        val recPos  = if (recNeg) rec.unclippedEnd else rec.unclippedStart
+        val lib       = library(rec)
+        val chrom     = rec.refIndex
+        val mateChrom = rec.mateRefIndex
+        val recNeg    = rec.negativeStrand
+        val recPos    = if (recNeg) rec.unclippedEnd else rec.unclippedStart
 
         val (mateNeg, matePos) = if (!rec.paired) (false, Int.MaxValue) else {
-          require(rec.refIndex == rec.mateRefIndex, s"Mate on different chrom for ${rec.name}.")
           val neg = rec.mateNegativeStrand
           val pos = if (neg) SAMUtils.getMateUnclippedEnd(rec.asSam) else SAMUtils.getMateUnclippedStart(rec.asSam)
           (neg, pos)
         }
 
-        val result = if (recPos < matePos || (recPos == matePos && !recNeg)) {
+        val result = if (chrom < mateChrom || chrom == mateChrom && (recPos < matePos || (recPos == matePos && !recNeg))) {
           new ReadInfo(chrom, recPos, matePos, recNeg, mateNeg, lib)
         }
         else {
-          new ReadInfo(chrom, matePos, recPos, mateNeg, recNeg, lib)
+          new ReadInfo(mateChrom, matePos, recPos, mateNeg, recNeg, lib)
         }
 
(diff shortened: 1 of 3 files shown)


mjhipp:
@nh13 would you rather me make this change optional/experimental?

nh13:
I think an opt-in model would be great, `--allow-inter-contig-grouping` or some better name than that, but the rest looks great to me!

mjhipp:
Thanks @nh13! Updated to make the feature opt-in.

nh13 (approved):

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -83,23 +83,23 @@ object GroupReadsByUmi {
...
 
         val (mateNeg, matePos) = if (!rec.paired) (false, Int.MaxValue) else {
-          require(rec.refIndex == rec.mateRefIndex, s"Mate on different chrom for ${rec.name}.")
           val neg = rec.mateNegativeStrand
           val pos = if (neg) SAMUtils.getMateUnclippedEnd(rec.asSam) else SAMUtils.getMateUnclippedStart(rec.asSam)
           (neg, pos)
         }
 
-        val result = if (recPos < matePos || (recPos == matePos && !recNeg)) {
+        val result = if (chrom < mateChrom || chrom == mateChrom && (recPos < matePos || (recPos == matePos && !recNeg))) {
```
nh13, line 98: Just adding one extra bracket to make 100% clear the precedence of `||` and `&&`:
```suggestion
        val result = if (chrom < mateChrom || (chrom == mateChrom && (recPos < matePos || (recPos == matePos && !recNeg)))) {
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -583,7 +586,7 @@ class GroupReadsByUmi
 
     (t.r1, t.r2, this.assigner) match {
       case (Some(r1), Some(r2), paired: PairedUmiAssigner) =>
-        require(r1.refIndex == r2.refIndex, s"Mates on different references not supported: ${r1.name}")
+        if (!this.allowInterContig) require(r1.refIndex == r2.refIndex, s"Mates on different references not supported: ${r1.name}")
         val pos1 = if (r1.positiveStrand) r1.unclippedStart else r1.unclippedEnd
         val pos2 = if (r2.positiveStrand) r2.unclippedStart else r2.unclippedEnd
         val r1Lower = pos1 < pos2 || pos1 == pos2 && r1.positiveStrand
```
nh13, line 592: I think you want to now incorporate contig coordinates here.
```suggestion
        val r1Lower = r1.refIndex < r2.refIndex || (r1.refIndex == r2.refIndex && (pos1 < pos2 || (pos1 == pos2 && r1.positiveStrand)))
```

nh13: Bonus points for a unit test that checks this.

nh13 (changes requested):
I think there's one place where the `contig` or `refIndex` wasn't considered, but should be.  If you agree, the rest looks good to me.

nh13:
Thank-you @mjhipp this is great!

Issue #656, opened 2021-02-21
How to create duplex UMIs with AnnotateBamWithUmis


birnbera:
I am working with data that uses two UMIs for paired end reads. One UMI was included as part of index 1 and the other as part of index 2. I'd like to annotate the RX field in my BAM file with both UMIs with a dash between, as in `NNNNNNNN-NNNNNNNN`. I see that `CorrectUmis` can handle duplex UMIs, such that it looks for the consensus sequence independently for each half. What would be the best way annotate a BAM with duplex UMIs in this case? It seems with `AnnotateBamWithUmis` you end up replacing the contents of the RX field when you run it twice.

nh13:
@birnbera 

You're right that `AnnotateBamWithUmis` takes only one FASTQ.  Can you try the proposed update in PR#657 where I added support for _multiple_ FASTQs?

birnbera:
Thanks for looking into this and coming up with a solution! I'm happy to run my dataset through and report back. It looks like the PR is still building in Travis CI. Do you know if there's anywhere that I can grab the built jar from the test build when it's done? I don't have a scala build environment set up. It would be preferable if there's a way to get a jar to test with.

nh13:
@birnbera https://github.com/fulcrumgenomics/fgbio/releases/tag/nh_annotate_multiple_fastqs_tag.  Download the JAR there and I can then delete the release.

birnbera:
Perfect. Just ran it and the result is as advertised. I used the resulting annotated bam (following sorting and addition of read groups) with `picard UmiAwareMarkDuplicatesWithMateCigar` and encountered no issues. Thanks again!

nh13:
@birnbera you can also try using `GroupReadsWithUmi -> CallDuplexConsensusReads -> FilterConsensusReads` if you have duplex data versus just using the UMIs to mark duplicates.  I'll close this issue once the PR is merged.

birnbera:
No issues detected after running through the suggested pipeline of additional tools.

birnbera:
[...]

PR #657, opened 2021-02-24
Add multi-FASTQ input support to AnnotateBamWithUmis

nh13:
See: https://github.com/fulcrumgenomics/fgbio/issues/656


tfenne (approved):
I think my only request would be to make the documentation clearer that the multi-fastq support is for multiple UMIs per template.  I had initially assumed (incorrectly) it was just for if your UMIs were spread over multiple input fastqs because your BAM was built from multiple FASTQs, rather than having multiple UMIs.

Issue #659, opened 2021-02-25
Calling duplex molecules using assembled paired end reads


tahuh:
Hi team.

My paired end fastq data consist of UMIs in both reads (R1 and R2) and I have assembled paired reads using several softwares
([pear](https://cme.h-its.org/exelixis/web/software/pear/) , [FLASH](https://ccb.jhu.edu/software/FLASH/) ).

When I try CallDuplexConsensusReads, the fgbio complains that GroupReadsByUMIs using `paired` strategy.

But the paired strategy does not work when I tried GroupReadsByUMIs.

Is there ways that I can make duplex sequences using assembled paired end reads?

nh13:
@tahuh `GroupReadsByUmi` currently expects the reads to paired end and not merged.  So merging the paired end reads (presumably they overlap) wont work.  The other option that wouldn't require any modification is to merge the paired end reads after consensus calling.  

That being said, I don't think it would be too hard support it.  I assume you've already extracted the UMIs and they are stored in the RX tag like `AGTC-GTCA`, where `AGTC` is from R1 and `GTCA` is from R2?  If so, I think this PR should work: https://github.com/fulcrumgenomics/fgbio/pull/660.  Can you give it a try?  Or can you send me some data that I test it out on?

tahuh:
@nh13 

Thanks for fast reply.

I will try that and let you know what happened.

nh13:
@tahuh can you let us know if this worked?

tahuh:
@nh13 Sorry for late reply.

I tried to use the method you have suggested for my research project but unfortunately me and colleagues decided not to use the read pair assembly option.

I really appreciate your kind help and feel sorry for not using the option.

I think you can close this issue.

nh13:
No worries. Thank you for reporting back!

PR #660, opened 2021-02-25
Support calling duplex molexules from merged paired end reads

nh13:
See https://github.com/fulcrumgenomics/fgbio/issues/659

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
@@ -98,6 +98,9 @@ case class Template(r1: Option[SamRecord],
     Template(x1, x2)
   }
 
+  /** True if the template has only fragment (R1) reads, false if paired (has both R1 and R2 reads). */
+  def unpaired: Boolean = r2.isEmpty
+
   /** The total count of records for the template. */
   lazy val readCount: Int = r1.size + r2.size + r1Supplementals.size + r2Supplementals.size + r1Secondaries.size + r2Secondaries.size
 }
--- a/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
@@ -569,8 +569,8 @@ class CollectDuplexSeqMetrics
     */
   private def takeNextGroup(iterator: BetterBufferedIterator[SamRecord]): Seq[SamRecord] = {
     val rec = iterator.head
-    val info = GroupReadsByUmi.ReadInfo(rec)
-    iterator.takeWhile(rec => GroupReadsByUmi.ReadInfo(rec) == info).toIndexedSeq
+    val info = GroupReadsByUmi.ReadInfo(rec, ignoreStrand=rec.unpaired)
+    iterator.takeWhile(rec => GroupReadsByUmi.ReadInfo(rec, ignoreStrand=rec.unpaired) == info).toIndexedSeq
   }
 
   /** Writes out all the metrics and plots. */
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -665,8 +665,10 @@ class GroupReadsByUmi
 
         if (r1Lower) paired.lowerReadUmiPrefix  + ":" + umis(0) + "-" + paired.higherReadUmiPrefix + ":" + umis(1)
         else         paired.higherReadUmiPrefix + ":" + umis(0) + "-" + paired.lowerReadUmiPrefix  + ":" + umis(1)
-      case (_,        _,        paired: PairedUmiAssigner) =>
-        fail(s"Template ${t.name} has only one read, paired-reads required for paired strategy.")
+      case (Some(r1), None,     paired: PairedUmiAssigner) =>
+        val umis = umi.split('-')
+        require(umis.length == 2, s"Paired strategy used but umi did not contain 2 segments: $umi")
+        paired.lowerReadUmiPrefix  + ":" + umis(0) + "-" + paired.higherReadUmiPrefix + ":" + umis(1)
       case (Some(r1), _, _) =>
         r1[String](this.rawTag)
     }
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -77,7 +77,7 @@ object GroupReadsByUmi {
     }
 
     /** Creates/retrieves a ReadEnds object from a SamRecord and stores it in a temporary attribute for later user. */
-    def apply(rec: SamRecord) : ReadInfo = {
+    def apply(rec: SamRecord, ignoreStrand: Boolean = false) : ReadInfo = {
       val tmp = rec.transientAttrs[ReadInfo](GroupReadsByUmi.ReadInfoTempAttributeName)
       if (tmp != null) {
         tmp
@@ -86,7 +86,7 @@ object GroupReadsByUmi {
         val lib       = library(rec)
         val chrom     = rec.refIndex
         val mateChrom = rec.mateRefIndex
-        val recNeg    = rec.negativeStrand
+        val recNeg    = !ignoreStrand && rec.negativeStrand
         val recPos    = if (recNeg) rec.unclippedEnd else rec.unclippedStart
 
         val (mateNeg, matePos) = if (!rec.paired) (false, Int.MaxValue) else {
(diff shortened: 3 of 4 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -592,8 +592,10 @@ class GroupReadsByUmi
 
         if (r1Lower) paired.lowerReadUmiPrefix  + ":" + umis(0) + "-" + paired.higherReadUmiPrefix + ":" + umis(1)
         else         paired.higherReadUmiPrefix + ":" + umis(0) + "-" + paired.lowerReadUmiPrefix  + ":" + umis(1)
-      case (_,        _,        paired: PairedUmiAssigner) =>
-        fail(s"Template ${t.name} has only one read, paired-reads required for paired strategy.")
+      case (Some(r1), None,     paired: PairedUmiAssigner) =>
+        val umis = umi.split('-')
+        require(umis.length == 2, s"Paired strategy used but umi did not contain 2 segments: $umi")
+        paired.lowerReadUmiPrefix  + ":" + umis(0) + "-" + paired.higherReadUmiPrefix + ":" + umis(1)
```
nh13, line 598: @tfenne if you agree, I'll add a test

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -77,7 +77,7 @@ object GroupReadsByUmi {
     }
 
     /** Creates/retrieves a ReadEnds object from a SamRecord and stores it in a temporary attribute for later user. */
-    def apply(rec: SamRecord) : ReadInfo = {
+    def apply(rec: SamRecord, ignoreStrand: Boolean = false) : ReadInfo = {
```
nh13, line 80: For "merged" paired end reads, we assume that the strand of the fragment read comes from R1:
1. if we have R1+/R2-, then the merged fragment read is `R1+ => revcomp(R2-)`
2. if we have R2+/R1-, then the merged fragment read is `R1- => revcomp(R2+)`

So when we are _retrieving_ reads to be compared, we should ignore the strand when slurping in templates, but then condition on strand for which read is `/A` and which read is `/B`.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -433,8 +433,10 @@ object Strategy extends FgBioEnum[Strategy] {
     |                  such that a read with A-B is related to but not identical to a read with B-A.
     |                  Expects the pair of UMIs to be stored in a single tag, separated by a hyphen
     |                  (e.g. `ACGT-CCGG`).  The molecular IDs produced have more structure than for single
-    |                  UMI strategies, and are of the form `{base}/{AB|BA}`. E.g. two UMI pairs would be
```
nh13, line 436: I don't think we output /AB or /BA, so these docs are very out-dated.

src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
```diff
@@ -279,6 +278,37 @@ class GroupReadsByUmiTest extends UnitSpec with OptionValues with PrivateMethodT
     aIds.head should not equal bIds.head
   }
 
+  it should "correctly group fragment reads with the paired assigner when the two UMIs are the same" in {
+    // NB: the "a*" reads are grouped, and the "b*" reads are a second group
```
nh13, line 282:
```suggestion
```

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -98,6 +98,9 @@ case class Template(r1: Option[SamRecord],
     Template(x1, x2)
   }
 
+  /** True if the template has only fragment (R1) reads, false if paired (has both R1 and R2 reads). */
+  def unpaired: Boolean = r2.isEmpty
```
tfenne, line 102: This seems dangerous.  What if `r1.paired` is true, but `r2` is missing?  What if somehow we have secondary or supplementary alignments for r2 but the primary is absent?  Perhaps `r1.exists(_.unpaired) && r2.isEmpty`?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -591,14 +593,14 @@ class GroupReadsByUmi
   /** Consumes the next group of templates with all matching end positions and returns them. */
   def takeNextGroup(iterator: BufferedIterator[Template]) : Seq[Template] = {
     val first     = iterator.next()
-    val firstEnds = ReadInfo(first.r1.getOrElse(fail(s"R1 missing for template ${first.name}")))
+    val firstEnds = ReadInfo(first.r1.getOrElse(fail(s"R1 missing for template ${first.name}")), ignoreStrand=first.unpaired)
```
tfenne, line 596: I don't get why we would need to ignore strand in the case of merged-PE reads.  Reads on opposite strands would still be different original molecules.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -591,14 +593,14 @@ class GroupReadsByUmi
...
     val firstUmi  = first.r1.get.apply[String](this.assignTag)
     val builder   = IndexedSeq.newBuilder[Template]
     builder    += first
 
     while (
       iterator.hasNext &&
-      firstEnds == ReadInfo(iterator.head.r1.get) &&
+      firstEnds == ReadInfo(iterator.head.r1.get, ignoreStrand=iterator.head.unpaired) &&
```
tfenne, line 603: This is also unsafe ... for performance reasons `ReadInfo` stores the value in a transient attribute on the record.  If you call it twice with different values for `ignoreStrand` you'll get the wrong answer the second time.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -665,8 +667,11 @@ class GroupReadsByUmi
 
         if (r1Lower) paired.lowerReadUmiPrefix  + ":" + umis(0) + "-" + paired.higherReadUmiPrefix + ":" + umis(1)
         else         paired.higherReadUmiPrefix + ":" + umis(0) + "-" + paired.lowerReadUmiPrefix  + ":" + umis(1)
-      case (_,        _,        paired: PairedUmiAssigner) =>
-        fail(s"Template ${t.name} has only one read, paired-reads required for paired strategy.")
+      case (Some(r1), None,     paired: PairedUmiAssigner) =>
+        val umis = umi.split('-')
+        require(umis.length == 2, s"Paired strategy used but umi did not contain 2 segments: $umi")
+        if (r1.positiveStrand) paired.lowerReadUmiPrefix  + ":" + umis(0) + "-" + paired.higherReadUmiPrefix + ":" + umis(1)
+        else paired.higherReadUmiPrefix + ":" + umis(0) + "-" + paired.lowerReadUmiPrefix  + ":" + umis(1)
```
tfenne, line 674:
```suggestion
        if (r1.positiveStrand) paired.lowerReadUmiPrefix  + ":" + umis(0) + "-" + paired.higherReadUmiPrefix + ":" + umis(1)
        else                   paired.higherReadUmiPrefix + ":" + umis(0) + "-" + paired.lowerReadUmiPrefix  + ":" + umis(1)
```

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -569,8 +569,8 @@ class CollectDuplexSeqMetrics
     */
   private def takeNextGroup(iterator: BetterBufferedIterator[SamRecord]): Seq[SamRecord] = {
     val rec = iterator.head
-    val info = GroupReadsByUmi.ReadInfo(rec)
-    iterator.takeWhile(rec => GroupReadsByUmi.ReadInfo(rec) == info).toIndexedSeq
+    val info = GroupReadsByUmi.ReadInfo(rec, ignoreStrand=rec.unpaired)
+    iterator.takeWhile(rec => GroupReadsByUmi.ReadInfo(rec, ignoreStrand=rec.unpaired) == info).toIndexedSeq
```
tfenne, line 573: Won't this now do the wrong thing for single UMI data?  I.e. if you have two reads who happen to have the same unclipped 5' end, but are on opposite strands, they'll now show up as having the same ReadInfo, and get bundled together for grouping?

src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
```diff
@@ -279,6 +278,37 @@ class GroupReadsByUmiTest extends UnitSpec with OptionValues with PrivateMethodT
...
+    val builder = new SamBuilder(readLength=100, sort=Some(SamOrder.Coordinate))
+    builder.addFrag(name="a01", start=100, strand=Plus, attrs=Map("RX" -> "ACT-ACT"))
+    builder.addFrag(name="a02", start=100, strand=Plus, attrs=Map("RX" -> "ACT-ACT"))
+    builder.addFrag(name="a03", start=100, strand=Plus, attrs=Map("RX" -> "ACT-ACT"))
+    builder.addFrag(name="a04", start=100, strand=Plus, attrs=Map("RX" -> "ACT-ACT"))
+    builder.addFrag(name="b01", start=100, strand=Minus, attrs=Map("RX" -> "ACT-ACT"))
+    builder.addFrag(name="b02", start=100, strand=Minus, attrs=Map("RX" -> "ACT-ACT"))
+    builder.addFrag(name="b03", start=100, strand=Minus, attrs=Map("RX" -> "ACT-ACT"))
+    builder.addFrag(name="b04", start=100, strand=Minus, attrs=Map("RX" -> "ACT-ACT"))
```
tfenne, line 291: I think this test _only_ works because there are no other reads present.  When sorting these into TemplateCoordinate the `a` reads have a 5' end position of `100` but the `b`s have a 5' end position of 199.  If there were any other reads present with 5' ends anywhere between 100-199, then they would not get sorted together and your change to ReadInfo would have no effect.

tfenne: And I really don't think we want to change the TemplateCoordinate sort order.

tfenne (changes requested):
I think there are some hard to resolve problems with this PR.  I'm leaning towards saying we should just drop it especially since the one person who opened the related issue never actually went forward with using the pe-merged data.

nh13:
Closing

PR #665, opened 2021-02-28
Add source classes for BED and generic Interval types

clintval:
I need helpers for reading generic interval data, so I added these classes!

`BedSource`: A source for BED input (similar to `IntervalListSource`).
`IntervalSource`: A source for interval data that wraps either BED or Interval List input.

In both classes, you can optionally provide a sequence dictionary that will be used to validate the intervals. I did this because I find myself having a BED and reference FASTA, more often than I have an Interval List and reference FASTA, but I would still like the safety of knowing my intervals are well-formed.

I expect to use these classes like the following, (so even BED data is safe!):

```scala
@clp(description="Does the work for you")
class DoWork(
  @arg(flag='i', doc="The path to the input intervals.") val input: PathToIntervals,
  @arg(flag='r', doc="The reference FASTA.") val reference: PathToFasta,
) extends Tool {
    val fasta     = ReferenceSequenceFileFactory.getReferenceSequenceFile(reference)
    val intervals = IntervalSource(input, Some(fasta.getSequenceDictionary.fromSam))

    intervals.foreach { interval => println(s"Found interval: $interval") }

    fasta.close()
    intervals.close()
  }
}
```

This should be OK from a type-alias perspective too, since `PathToIntervals` is defined as:

```scala
/** Represents a path to an intervals file (IntervalList or BED). */
type PathToIntervals = java.nio.file.Path
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ JeanLuc.iml
 target
 project/project
 .DS_Store
+out/
+*.iml
--- a/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
@@ -26,13 +26,12 @@
 package com.fulcrumgenomics.fasta
 
 import java.io.StringWriter
-
 import com.fulcrumgenomics.FgBioDef
 import com.fulcrumgenomics.FgBioDef._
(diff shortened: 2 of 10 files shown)


tfenne:
Conceptuall on board and thanks for doing this.  Might take a while to get to a review though.

clintval:
No rush! Take what you like from the submission.

PR #669, opened 2021-03-13
Add the --annotate-all option to AssignPrimers

nh13:
In cases where a given read end (eg. R2) has both a primary and
supplementary alignment, where only of the two is assigned a primer,
the current behavior is to annotate the mate primer/amplicon for both,
but the primer/amplicon just for the alignment for the read end that
is assigned a primer.

The --annotate-all optiona will annotate both alignments of the same end
with the primer.  This may be important for downstream translocation
detection.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/AssignPrimers.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/AssignPrimers.scala
@@ -1,7 +1,8 @@
 package com.fulcrumgenomics.util
 
 import com.fulcrumgenomics.FgBioDef.{FilePath, PathToBam, SafelyClosable}
-import com.fulcrumgenomics.bam.api.{SamRecord, SamSource, SamWriter}
+import com.fulcrumgenomics.bam.Bams
+import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamSource, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.commons.util.LazyLogging
 import com.fulcrumgenomics.sopt.{arg, clp}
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/util/AssignPrimers.scala
```diff
@@ -86,17 +93,46 @@ class AssignPrimers
...
-        rec = rec,
-        recAmplicon = recAmplicon,
-        mateAmplicon = if (rec.unpaired) None else detector.findMatePrimer(rec=rec)
-      )
-      writer.write(rec)
-      progress.record(rec)
+    if (annotateAll) {
+      // All alignments will be assigned a primer independently, and then grouped by read end. An primer assignment will
+      // be made on a given read end only if there is one and only one unique primer assignment.
+      val sorter = Bams.sorter(SamOrder.Coordinate, reader.header)
```
tfenne, line 99: I think this is a problem on a couple of fronts.  1) Since the `writer` is created using the `reader.header` above it's possible that the SO in the header and the actual ordering of the reads won't match.  2) I think the expectation is probably to produce output in the same order as the input since that's what the tool generally does.

I'd prefer to move the declaration of `writer` inside the if/else then on the `if` branch we could extract the `SamOrder` from the input header and use that to drive the `sorter` and perhaps even skip re-sorting if the input file was already query sorted or grouped?

src/main/scala/com/fulcrumgenomics/util/AssignPrimers.scala
```diff
@@ -86,17 +93,46 @@ class AssignPrimers
...
+      Bams.templateIterator(reader).foreach { template =>
+        // split into R1s and R2s
+        val (r1s, r2s) = template.allReads.toSeq.partition(r => r.unpaired || r.firstOfPair)
+        // detect the primers for each of the R1s and R2s
+        val r1Amplicons = r1s.flatMap(detector.findPrimer).distinct
+        val r1Amplicon = if (r1Amplicons.length == 1) r1Amplicons.headOption else None
```
tfenne, line 105:
```suggestion
        val r1Amplicon = r1s.flatMap(detector.findPrimer).distinct match {
            case Seq(x) => Some(x)
            case _      => None
        }
        val r1Amplicon = if (r1Amplicons.length == 1) r1Amplicons.headOption else None
```
And again below?

src/main/scala/com/fulcrumgenomics/util/AssignPrimers.scala
```diff
@@ -86,17 +93,46 @@ class AssignPrimers
...
+        template.allReads.foreach { r =>
+          sorter += r
+          progress.record(r)
+        }
+      }
+      progress.logLast()
+      // Write them out after sorting
+      sorter.foreach { rec =>
+        writer += rec
+      }
```
tfenne, line 121:
```suggestion
      writer ++= sorter
```

tfenne (changes requested):
I think there's one issue with the sorting that needs to be addressed, then you're good to go.

tfenne (approved):

PR #670, opened 2021-03-17
Support a read structure for the FASTQ in AnnotateBamWithUmis

nh13:


src/main/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmis.scala
```diff
@@ -59,6 +62,8 @@ class AnnotateBamWithUmis(
   @arg(flag='f', doc="Input FASTQ file with UMI reads.")       val fastq: PathToFastq,
   @arg(flag='o', doc="Output BAM file to write.")              val output: PathToBam,
   @arg(flag='t', doc="The BAM attribute to store UMIs in.")    val attribute: String = "RX",
+  @arg(flag='r', doc="The read structure for the FASTQ, otherwise all bases will be used.")
```
tfenne, line 65: Instead of making this optional I'd be tempted to just specify it as `+M` by default.

tfenne (approved):
Looks good to me.  I would tend to make the read structure non-optional with a default of `+M` so that there's a single code path that all the tests hit.

PR #677, opened 2021-05-19
Add End Repair Artifact Filter to FilterSomaticVcf

ThomasHSmith:
This PR covers addition of a new filter to `FilterSomaticVcf`. There is a bit of refactoring of the existing filter to avoid duplicated code. Also I took the existing filter's name.

## Changes to FilterSomaticVcf/CLI

- Renames existing filter
	- "End Repair Artifact Filter" is renamed to "A-tailing Artifact Filter" to more specifically describe the type of artifact-generating process it targets.
	- CLI options are renamed accordingly, to `--a-tailing-distance` and `--a-tailing-p-value`.
	- The `INFO` field associated with this filter is changed from `ERAP` to `ATP`.
	- The `FILTER` tag applied by this filter is now `ATailingArtifact`.
 	- The behavior of the filter was not modified.

- Added a new filter that addresses a broader category of end repair artifacts.
	- Took the name of the previous filter, "End Repair Artifact Filter".
	- Applies to any point mismatch regardless of alt base.
	- Test for artifact congruency is simply based on distance from nearest read end.
	- Uses the same method for calculating p-values as the previous filter.
	- Can be tuned using the CLI parameters `--end-repair-distance` and `--end-repair-p-value`.
	- The `INFO` field associated with this filter is `ERP`.
	- The `FILTER` tag applied by this filter is `EndRepairArtifact`.

- By default, both filters will be applied with their default distance values.
- If desired, a user can disable a specific filter by setting its distance option to `:none:`, for example `--end-repair-distance :none:` would disable the end repair artifact filter.

## Changes to classes/inheritance:

- Added `ReadEndSomaticVariantFilter.scala`:
	- Contains the `ReadEndSomaticVariantFilter` trait, which defines:
	- Functions that are shared across both filters, such as `filters()`, `annotations()`, and, in the companion object, `priors()`. Note: all three of these were lifted from EndRepairArtifactLikelihoodFilter.scala.
	- Values and functions that are required to be defined by each individual filter, such as the values `Info` and `Filter` or the functions `appliesTo()` and `isArtifactCongruent()`, which will vary depending on the filter.
- Refactored original filter:
	- Originally the `EndRepairArtifactLikelihoodFilter` class was defined in EndRepairArtifactLikelihoodFilter.scala (which is now removed).
	- New implementation has this filter represented by the class `ATailingArtifactLikelihoodFilter`, which includes the original, unaltered methods `appliesTo()` and `isArtifactCongruent()`.
- Added new end repair artifact filter:
	- The filter is represented by the class `EndRepairArtifactLikelihoodFilter`.
	- The `appliesTo()` method simply tests whether the variant is a point mismatch.
	- The `isArtifactCongruent()` method tests whether the pileup base falls within a specified distance from the nearest read end.


src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -27,46 +27,49 @@ package com.fulcrumgenomics.vcf.filtration
...
+    */
+  def isArtifactCongruent(refAllele: Byte, altAllele: Byte, entry : BaseEntry, pileupPosition: Int): Boolean
+
+  /** Applies a single filter if a) a threshold was provided at construction, b) the computed
+    * pvalue is present in the <annotations>, and c) the pvalue <= <pValueThreshold>.
+    */
+  def filters(annotations: Map[String, Any]): Iterable[String] = {
+    (this.pValueThreshold, annotations.get(Info.id).flatMap(_.toString.toDoubleOption)) match {
+      case (Some(threshold), Some(pvalue)) if pvalue <= threshold => List(Filter.id)
+      case _ => Nil
```
tfenne, line 67: Since it's only a single value you could use `Some(Filter.id)` and `None` here.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -27,46 +27,49 @@ package com.fulcrumgenomics.vcf.filtration
 import java.lang.Math.{min, pow}
 
 import com.fulcrumgenomics.bam.{Bams, BaseEntry, Pileup, PileupEntry}
-import com.fulcrumgenomics.commons.util.LazyLogging
+import com.fulcrumgenomics.commons.util.{LazyLogging, Logger}
 import com.fulcrumgenomics.util.NumericTypes.{LogProbability => LnProb}
 import com.fulcrumgenomics.util.Sequences
+import com.fulcrumgenomics.vcf.api.Allele.SimpleAllele
 import com.fulcrumgenomics.vcf.api.{VcfCount, VcfFieldType, VcfInfoHeader, _}
+import com.fulcrumgenomics.vcf.filtration.ReadEndSomaticVariantFilter.{isPointMismatch, priors}
```
tfenne, line 35: Would prefer not to import these like this.  If called from within something implementing that trait you don't need this.  And if from outside the trait, better to call with `ReadEndSomaticVariantFilter.priors(...)` to make it clearer.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -123,27 +124,117 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+    *   <priorMutation> is the prior that the genotype is the result of a true somatic mutation.
+    *   <priorArtifact> is the prior that the genotype is an artifact resulting from the mechanism specific to the filter
+    *
     *
     * Default implementation gives a mutation prior of {{{(maf * 2)^2}}} capped at 0.9999 and an
     * artifact prior of 1 - mutation prior.  This has the effect of creating priors that are heavily
     * skewed towards the artifact at very low MAFs, and towards mutations as the MAF increases towards 0.5.
     */
-  private[filtration] def priors(pileup: Pileup[PileupEntry], maf: Double) : (Double, Double) = {
+  private[filtration] def priors(pileup: Pileup[PileupEntry], maf: Double, logger : Option[Logger] = None) : (Double, Double) = {
```
tfenne, line 155: I don't love passing `logger` objects around in function calls.  Any reason not to just have the `ReadEndSomaticVariantFilter` object extend `LazyLogging` and use it's own logger?

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -123,27 +124,117 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+  * @param distance The distance from the end of the template that defines the region where this
+  *                 artifact may occur
+  * @param pValueThreshold the pvalue at and below which mutations are filtered as being likely
+  *                        caused by the artifact
+  */
+class EndRepairArtifactLikelihoodFilter(override val distance: Int = 15, override val pValueThreshold: Option[Double] = None)
+  extends ReadEndSomaticVariantFilter {
+
+  val Info = VcfInfoHeader("ERP", VcfCount.Fixed(1), VcfFieldType.Float, "P-Value for the End Repair Artifact Filter test.")
+  val Filter = VcfFilterHeader("EndRepairArtifact", "Variant is likely an artifact caused by end repair.")
```
tfenne, line 185: How about `EndRepairFillInArtifact`?

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -123,27 +124,117 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+    if (ref.length != 1) false else { // No interest in deletions/MNVs, so only consider if ref length is 1
+      gt
+        .alleles
+        .alts
+        .collect { // Collect all the alt alleles
+          case alt: SimpleAllele // Only interested in simple alleles, no symbolics or other strange stuff
+            if (alt.length == 1 & alt.bases != ref.bases) => alt // No interest in insertions, so only retain alts with length 1
+          }.nonEmpty // If anything remains, at least one of our alleles is a point mismatch
+    }
+  }
```
tfenne, line 144: A couple of things:
1. Can we call this `isHetSnv` because `pointMismatch` doesn't actually imply het.
2. This doesn't actually do what you want.  It's only checking the alleles on the VCF record and doesn't inspect the genotype at all.  So it will pass all cases where the REF allele is single-base and _at least one_ single base ALT allele exists.

E.g. it would return `true` for:
```
chr1  1234  . G  A,GAGAGAGA   . PASS  . 1/2
```

It needs to either use the methods on `gt` like `gt.isHet` and/or refer to `gt.calls` which are the called alleles in the genotype.  `gt.alleles` refers to the set of alleles on the enclosing VCF record.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -123,27 +124,117 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+
+  override val VcfInfoLines: Iterable[VcfInfoHeader]     = Seq(Info)
+  override val VcfFilterLines: Iterable[VcfFilterHeader] = Seq(Filter)
+
+  /** Applies to all het SNVs. */
+  def appliesTo(gt: Genotype): Boolean = isPointMismatch(gt)
+
+  /** Returns true if the position of the base from the closest read end does not exceed the filter's defined distance. */
+  def isArtifactCongruent(refAllele: Byte, altAllele: Byte, entry : BaseEntry, pileupPosition: Int): Boolean = {
+
```
tfenne, line 195: Remove blank line please.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -123,27 +124,117 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+  val Filter = VcfFilterHeader("ATailingArtifact", "Variant is likely an artifact caused by A-tailing.")
+
+  override val VcfInfoLines: Iterable[VcfInfoHeader]     = Seq(Info)
+  override val VcfFilterLines: Iterable[VcfFilterHeader] = Seq(Filter)
+
+  /** Only applies to het SNVs where the alt allele is A or T. */
+  def appliesTo(gt: Genotype): Boolean = {
+    isPointMismatch(gt) &&
+      gt.calls.iterator.filter(_ != gt.alleles.ref).exists(a => a.value.equalsIgnoreCase("A") || a.value.equalsIgnoreCase("T"))
+  }
```
tfenne, line 237: Even if `isPointMismatch` were correct, this changes behavior.  The old code required `!gt.isHetNonRef` which requires that the genotype is a) heterozygous and b) one of the alleles in the genotype is the reference allele.  This now only checks that the genotype is heterozygous and one of the called alleles is a non-ref is A or T, but not that other is REF.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -184,7 +275,7 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
     * the read or the A-paired allele at the end of the read - i.e. they are observations that if converted to
     * alt observations would be congruent.
     */
-  private[filtration] def isArtifactCongruent(refAllele: Byte, altAllele: Byte, entry : BaseEntry, pileupPosition: Int): Boolean = {
+   def isArtifactCongruent(refAllele: Byte, altAllele: Byte, entry : BaseEntry, pileupPosition: Int): Boolean = {
```
tfenne, line 278: Indentation got an erroneous extra space.

src/test/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilterTest.scala
```diff
@@ -29,8 +29,9 @@ import com.fulcrumgenomics.bam.{BaseEntry, Pileup, PileupBuilder, PileupEntry}
 import com.fulcrumgenomics.testing.SamBuilder.{Minus, Plus}
 import com.fulcrumgenomics.testing.{SamBuilder, UnitSpec}
 import com.fulcrumgenomics.vcf.api.{AlleleSet, Genotype}
+import com.fulcrumgenomics.vcf.filtration.ReadEndSomaticVariantFilter.{isPointMismatch, priors}
```
tfenne, line 32: Please just use Object.method() at call site instead of importing.

src/test/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilterTest.scala
```diff
@@ -42,15 +43,143 @@ class EndRepairArtifactLikelihoodFilterTest extends UnitSpec {
...
+  "isPointMismatch" should "return true for any SNV" in {
+    val bases = Seq("A", "C", "G", "T")
+    for (ref <- bases; alt <- bases; if ref != alt) yield isPointMismatch(singleGenotype(ref, alt)) shouldBe true
+  }
+
+  it should "return false for any event that is not a SNP" in {
+    ReadEndSomaticVariantFilter.isPointMismatch(singleGenotype("A",  "AT")) shouldBe false // Insertion
+    ReadEndSomaticVariantFilter.isPointMismatch(singleGenotype("TA", "A"))  shouldBe false // Deletion
+    ReadEndSomaticVariantFilter.isPointMismatch(singleGenotype("AT", "GC")) shouldBe false // MNP
+  }
```
tfenne, line 55: I think you need a more comprehensive test set here, and to use something other than the existing `singleGenotype` method to generate them.  As noted above the method isn't inspecting the genotype, but `singleGenotype` always sets the variant alleles and the genotype's alleles to the same values, so the tests didn't catch that.  
Please add some tests with > 1 ALT allele and a variety of het and hom genotypes.

tfenne (changes requested):
A few problems, but mostly easily solvable.  The main thing for discussion is that I don't like the re-use of the old names/parameters for the new filter because it could cause silent incorrect behaviour.  

I'm happy to take better suggestions on names.  I will note that I'm aware of two ways folks do "end repair".  The more common one is to fill in.  The less common one is to use an exonuclease to digest away the single-stranded overhang.  Since the artifact you're adding is specific to the fill-in method, my suggest is to add `fillIn` or something like that to the name(s).

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
mjhipp: @tfenne I share the confusion with @ThomasHSmith between `gt.calls` and `gt.alleles`. I can see in the code that `alleles` is from the `VariantContext.getAlleles()`, and `calls` is from `Genotype.getAlleles()`. I get the utility of having access to the parent Variant's alleles inside the `Genotype`, but from an api stand point I would have expected fgbio's `Genotype.alleles` to relate to htsjdk's `Genotype.getAlleles()` in the same way that `Variant.alleles` relates to `VariantContext.getAlleles()`, with some other member for the `Variant` alleles, like `Genotype.variantAlleles`

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
```diff
@@ -52,17 +52,40 @@ import scala.collection.immutable.ListMap
...
+    |### A-tailing Artifact Filter
+    |
+    |The A-tailing artifact filter attempts to measure the probability that a single-nucleotide mismatch is the product of
+    |errors in the template generated during the A-base addition steps that are common to many Illumina
+    |library preparation protocols.  The artifacts occur if/when a recessed 3' end is incorrectly filled
+    |in with one or more adenines during A-base addition. Incorrect adenine incorporation presents specifically as errors to T at
+    |the beginning of reads (and in very short templates, as matching errors to A at the ends of reads).
+    |
+    |The filter adds the `INFO` field `ATP` to SNVs with an A or T alternate allele. This field records
+    |the p-value for rejecting the possibility that the variant is an A-tailing artifact. If a threshold
```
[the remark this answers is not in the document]
ThomasHSmith: Rephrased this to be more explicit about what p-values represent, but let me know if it's still unclear.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
tfenne: @mjhipp Feedback on the API in fgbio is welcome.  That said, the HTSJDK API is a horror show and I would generally suggest we not reference that in discussion of the fgbio VCF API!

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
clintval: resolved

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
```diff
@@ -103,7 +128,12 @@ class FilterSomaticVcf
         else invalid("Must supply --sample when VCF contains more than one sample.")
     }
 
-    val filters = Seq(new EndRepairArtifactLikelihoodFilter(endRepairDistance, endRepairPValue))
+    if (aTailingDistance.isEmpty && endRepairDistance.isEmpty) logger.warning("All filters are turned off and tool will no-op!")
```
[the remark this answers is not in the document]
clintval: I confirm that the tool will No-op because it is now possible to selectively turn on/off each of the two filters so a user could presumably turn them all to "off":

```bash
❯ java -jar fgbio-1.4.0-fe54cf7-SNAPSHOT.jar FilterSomaticVcf \
    --input input.vcf.gz \
    --output output.vcf.gz \
    --bam input.bam \
    --a-tailing-distance :none: \
    --end-repair-fill-in-distance :none:
```

```bash
❯ gzip -dc input.vcf.gz | grep -v '#' | wc -l
22260
```

```diff
❯ diff <( gzip -dc input.vcf.gz ) <( gzip -dc output.vcf.gz )
<empty>
```

I'll amend the documentation to make this functionality clear!

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
```diff
@@ -74,17 +97,19 @@ import scala.collection.immutable.ListMap
...
+  @arg(flag='q', doc="Minimum base quality.")                       val minBaseQuality: Int    = 20,
+  @arg(flag='p', doc="Use only paired reads mapped in pairs.")      val pairedReadsOnly: Boolean = false,
   // Developer Note: filter-specific attributes should NOT be given flags as we will likely run
   //                 out of flags that way and end up with a mix of +flag and -flag options.
-  @arg(doc="Distance from end of read to implicate end repair.")  val endRepairDistance: Int = 2,
-  @arg(doc="Minimum acceptable p-value for end repair test.")     val endRepairPValue: Option[Double] = None
+  @arg(doc="Distance from end of read to implicate A-base addition artifacts.") val aTailingDistance: Option[Int] = Some(2),
+  @arg(doc="Minimum acceptable p-value for A-base addition artifact test.")     val aTailingPValue: Option[Double] = None,
+  @arg(doc="Distance from end of read to implicate end repair artifacts.")      val endRepairDistance: Option[Int] = Some(15),
+  @arg(doc="Minimum acceptable p-value for end repair artifact test.")          val endRepairPValue: Option[Double] = None
```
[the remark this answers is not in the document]
clintval: resolved

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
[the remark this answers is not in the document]
clintval: resolved

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
clintval: resolved

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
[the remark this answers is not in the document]
clintval: resolved

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
clintval: resolved

src/test/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilterTest.scala
clintval: resolved

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
```diff
@@ -37,10 +37,10 @@ import scala.collection.immutable.ListMap
...
 @clp(group=ClpGroups.VcfOrBcf, description=
   """
-    |Applies one or more filters to a VCF of somatic variants. The VCF must contain genotype information
-    |for the tumor sample.  If the VCF also contains genotypes for one or more other samples, the
-    |`--sample` option must be provided to specify the sample whose genotypes to examine and whose reads
-    |are present in the BAM file.
+    |Applies one or more filters to a VCF of somatic variants.
+    |
+    |If the VCF also contains genotypes for more than one sample, then the `--sample` option must be provided to specify
+    |which genotypes to examine as they relate to the reads in the BAM file (`--bam`).
```
tfenne, line 43:
```suggestion
    |Applies one or more filters to a VCF of somatic variants. The VCF must contain genotype information
    |for the tumor sample.  If the VCF also contains genotypes for one or more other samples, the
    |`--sample` option must be provided to specify the sample whose genotypes to examine and whose reads
    |are present in the BAM file.```

Comment (tfenne) [reply src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala:43]: Reverting the leading paragraph because:

a) It's an unrelated change
b) There's no need to break it into multiple paragraphs - I think it actually will lead to people not reading the second couple of sentences
c) The rewording implies you can run this on a VCF without genotypes when that is incorrect

File: src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala [hunk 15]
```diff
@@ -50,41 +50,73 @@ import scala.collection.immutable.ListMap
     |Each available filter may generate annotations in the `INFO` field of the output VCF and optionally,
     |if a threshold is specified, may apply one or more `FILTER`s to applicable variants.
     |
+    |Note: In the initial version of this tool, the only available filter was specific to A-base addition artifacts and
```
tfenne, line 53:
```suggestion
    |In previous versions of this tool, the only available filter was specific to A-base addition artifacts and
```

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
```diff
@@ -50,41 +50,73 @@ import scala.collection.immutable.ListMap
...
     |
-    |The end repair artifact filter attempts to measure the probability that a variant is caused by
-    |errors in the template generated during the end-repair and A-base addition steps that are common
-    |to many Illumina library preparation protocols.  The artifacts occur if/when the end repair creates
-    |a recessed 3' end which is subsequently and incorrectly filled in with As during A-base addition.
-    |This presents specifically as errors to T at the beginning of reads (and in very short templates,
-    |as matching errors to A at the ends of reads).
+    |The end repair fill-in artifact filter attempts to measure the probability that a single-nucleotide mismatch is the
+    |product of an error in the template generated during the end repair fill-in step that is common to many Illumina
+    |library preparation protocols, in which single-stranded 3' overhands are filled in to create a blunt end. These
```
tfenne, line 87:
```suggestion
    |library preparation protocols, in which single-stranded 3' overhangs are filled in to create a blunt end. These
```

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -24,49 +24,49 @@
...
 
-  /** The INFO header line needed by this filter. */
-  val Info = VcfInfoHeader("ERAP", VcfCount.Fixed(1), VcfFieldType.Float, "P-Value for the End-Repair Artifact test.")
+/** Trait for classes that apply filters to somatic variants calls at read ends. */
+trait ReadEndSomaticVariantFilter extends SomaticVariantFilter with LazyLogging {
 
-  /** The FILTER header line needed by this filter. */
-  val Filter = VcfFilterHeader("EndRepairArtifact", "Variant is likely an artifact caused by end-repair and A-base addition.")
+  /** The VCF header line that describes the INFO key that the filter may reference. */
+  def Info: VcfInfoHeader
```
tfenne, line 39: How about `def infoHeadLine` to make it clear in usage below that it's the entire header line and also lower case first letter to indicate def.  Or make it `val InfoHeaderLine`?

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -24,49 +24,49 @@
...
 
-  override val VcfInfoLines: Iterable[VcfInfoHeader]     = Seq(Info)
-  override val VcfFilterLines: Iterable[VcfFilterHeader] = Seq(Filter)
+  /** The VCF header line that describes the FILTER key that the filter may reference. */
+  def Filter: VcfFilterHeader
```
tfenne, line 42: Same here about naming.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -24,49 +24,49 @@
...
+    * a threshold distance from the read end. Other filters employ more elaborate tests that consider
+    * multiple criteria, such as the presence of certain allele pairs at specific read ends.
+    */
+  def isArtifactCongruent(refAllele: Byte, altAllele: Byte, entry : BaseEntry, pileupPosition: Int): Boolean
+
+  /** Applies a single filter if a) a threshold was provided at construction, b) the computed
+    * pvalue is present in the <annotations>, and c) the pvalue <= <pValueThreshold>.
+    */
+  def filters(annotations: Map[String, Any]): Iterable[String] = {
+    (this.pValueThreshold, annotations.get(Info.id).flatMap(_.toString.toDoubleOption)) match {
```
tfenne, line 62: What's with the `.toString.toDoubleOption`?  Isn't it required that subclasses produce a `pvalue: Double`?

Perhaps it's even worth changing things so that instead of having sub-classes define `VcfInfoHeader` they just return a tuple of `ID` and `description` and then have _this_ trait dictate type and cardinality?

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -113,37 +111,110 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+}
+
+/** The companion object for [[ReadEndSomaticVariantFilter]]. */
+object ReadEndSomaticVariantFilter extends LazyLogging {
 
   /** Calculates a pair of priors:
-    *   The first is the prior that the genotype is the result of a true somatic mutations
-    *   The second is the prior that the genotype is the result of the end-repair artifact process
+    *   <priorMutation> is the prior that the genotype is the result of a true somatic mutation.
+    *   <priorArtifact> is the prior that the genotype is an artifact resulting from the mechanism specific to the filter
```
tfenne, line 129:
```suggestion
    *   `priorMutation` is the prior that the genotype is the result of a true somatic mutation.
    *   `priorArtifact` is the prior that the genotype is an artifact resulting from the mechanism specific to the filter
```

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -113,37 +111,110 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+    *
     *
     * Default implementation gives a mutation prior of {{{(maf * 2)^2}}} capped at 0.9999 and an
     * artifact prior of 1 - mutation prior.  This has the effect of creating priors that are heavily
     * skewed towards the artifact at very low MAFs, and towards mutations as the MAF increases towards 0.5.
     */
   private[filtration] def priors(pileup: Pileup[PileupEntry], maf: Double) : (Double, Double) = {
```
tfenne, line 136: I don't think this should be in the companion as implementations of the trait may want to override it (as suggested by `Default implementation...` in the docs.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -113,37 +111,110 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+  *                        caused by the artifact
+  */
+class EndRepairFillInArtifactLikelihoodFilter(override val distance: Int = 15, override val pValueThreshold: Option[Double] = None)
+  extends ReadEndSomaticVariantFilter {
+
+  val Info: VcfInfoHeader = VcfInfoHeader("ERFAP", VcfCount.Fixed(1), VcfFieldType.Float, "P-Value for the End Repair Fill-in Artifact Filter test.")
+  val Filter: VcfFilterHeader = VcfFilterHeader("EndRepairFillInArtifact", "Variant is likely an artifact caused by end repair fill-in.")
+
+  override val VcfInfoLines: Iterable[VcfInfoHeader]     = Seq(Info)
+  override val VcfFilterLines: Iterable[VcfFilterHeader] = Seq(Filter)
```
tfenne, line 169: Any reason not to have these be the default impl in the trait above?

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -113,37 +111,110 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+  * @param distance The distance from the end of the template that defines the region where this
+  *                 artifact may occur
+  * @param pValueThreshold the pvalue at and below which mutations are filtered as being likely
+  *                        caused by the artifact
+  */
+class EndRepairFillInArtifactLikelihoodFilter(override val distance: Int = 15, override val pValueThreshold: Option[Double] = None)
+  extends ReadEndSomaticVariantFilter {
+
+  val Info: VcfInfoHeader = VcfInfoHeader("ERFAP", VcfCount.Fixed(1), VcfFieldType.Float, "P-Value for the End Repair Fill-in Artifact Filter test.")
+  val Filter: VcfFilterHeader = VcfFilterHeader("EndRepairFillInArtifact", "Variant is likely an artifact caused by end repair fill-in.")
```
tfenne, line 166: `@override`?

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -113,37 +111,110 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+  /** Returns true if the position of the base from the closest read end does not exceed the filter's defined distance. */
+  def isArtifactCongruent(refAllele: Byte, altAllele: Byte, entry : BaseEntry, pileupPosition: Int): Boolean = {
+    val positionInRead       = entry.positionInReadInReadOrder
+    val positionFromOtherEnd = Bams.positionFromOtherEndOfTemplate(entry.rec, pileupPosition)
+
+    // Pick the position that's closest to an end, and test for congruency based on that
+    val pos = {
+      if (positionFromOtherEnd.exists(_ < positionInRead)) positionFromOtherEnd.get
+      else                                                 positionInRead
+    }
```
tfenne, line 183: Could also do `(Some(positionInRead) ++ positionFromOtherEnd).min` I think

tfenne (changes requested):
Minor changes requested.  Thanks for sticking with this!

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
clintval: No need to override when you're making a class concrete for the first time.

Do you prefer explicitly overriding what a trait promises? I can make the update if so!

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
clintval: I agree it's a little awkward. I reverted to how it was before this PR which isn't much better.

There were some awkward design choices made before we started refactoring.

If you want me to continue refactoring forward I can! I'm trying not to let this PR drift into a bigger re-write.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
clintval: Yes that works! FYI a good amount of this code was copy-pasted from before the PR and I am trying to balance creating too much churn during this refactor for safety sake.

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
clintval: My effort was to make it clearer you don't need a tumor sample before running this tool.

There are many somatic samples out there that aren't tumor.

I'll revert... it is an unrelated change!

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
```diff
@@ -37,10 +37,11 @@ import scala.collection.immutable.ListMap
 
 @clp(group=ClpGroups.VcfOrBcf, description=
   """
-    |Applies one or more filters to a VCF of somatic variants. The VCF must contain genotype information
-    |for the tumor sample.  If the VCF also contains genotypes for one or more other samples, the
-    |`--sample` option must be provided to specify the sample whose genotypes to examine and whose reads
-    |are present in the BAM file.
+    |Applies one or more filters to a VCF of somatic variants.
```
clintval, line 40: Split off the first sentence so it pretty prints better when you do `fgbio --help`

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
```diff
@@ -113,37 +117,117 @@ class EndRepairArtifactLikelihoodFilter(val distance: Int = 2,
...
+  */
+class EndRepairFillInArtifactLikelihoodFilter(override val distance: Int = 15, override val pValueThreshold: Option[Double] = None)
+  extends ReadEndSomaticVariantFilter {
+
+  /** The VCF header line that describes the INFO key that the read-end filter references. */
+  val readEndInfoLine: VcfInfoHeader = VcfInfoHeader(
+    id          = "ERFAP",
+    count       = VcfCount.Fixed(1),
+    kind        = VcfFieldType.Float,
+    description = s"P-Value for the End Repair Fill-in Artifact Filter test, parameterized with a 5-prime distance of ${distance}bp."
```
clintval, line 172: I changed the names of the single read-end INFO/FILTER header records.

The trait minimally requires a `def` but since we only need a static I drop down to `val` here.

I also parameterized the description for all read-end filters to include the distance parameter.

I realize now I could have also parameterized the p-value threshold too!

clintval:
Back to you! Thank you Tim 🙏

src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
tfenne: I think this is a @nh13-ism, where @nh13 has a habit of doing this, and now others think there is some implied functional reason for doing so.  The current version, which has a full paragraph, not a single sentence, as the first paragraph produces this output:

```
...
VCF/BCF:                              Tools for manipulating VCF, BCF, or related data.
    AssessPhasing                      Assess the accuracy of phasing for a set of variants.
    FilterSomaticVcf                   Applies one or more filters to a VCF of somatic variants.
    FixVcfPhaseSet                     Adds/fixes the phase set (PS) genotype field.
...
```

`sopt` scans the first paragraph to identify the first _sentence_ and uses that as the summary.

tfenne (approved):

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
tfenne: Fair enough.

src/main/scala/com/fulcrumgenomics/vcf/filtration/ReadEndSomaticVariantFilter.scala
tfenne: I tend to like it mostly for GH code reviews because it makes it clear it's coming from a parent trait - in IntelliJ or similar the IDE let's you know but in GH it's not immediately obvious.

PR #681, opened 2021-06-18
[feature] TrimPrimers can trim only R1s

nh13:
Alternative to #679.  The advantage of this PR is it's a bit smaller of a modification.  The negative is using `-1` as a proxy for `None`.  In >90% of cases I'd prefer to use `Some/None`, but I think this is simple enough of a change to make it work with `TrimPrimers`

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
@@ -132,7 +132,7 @@ class TrimPrimers
       }
       else {
         require(amplicon.leftStart > 0 && amplicon.rightStart > 0,
-          f"Either the left or right amplicon coordinates must be > 0. Found ${amplicon.mkString("\t")}"
+          f"Both the left and right amplicon coordinates must be > 0. Did you forget to set '--first-of-pair'? Found ${amplicon.mkString("\t")}"
         )
       }
     }
--- a/src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
@@ -59,6 +59,19 @@ import scala.collection.BufferedIterator
     |
     |If the input BAM is not `queryname` sorted it will be sorted internally so that mate
     |information between paired-end reads can be corrected before writing the output file.
+    |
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -107,6 +122,21 @@ class TrimPrimers
...
+    // right primer.  Otherwise, coordinates must be > 0.
+    detector.detector.getAll.iterator().foreach { amplicon =>
+      if (firstOfPair) {
+        require(amplicon.leftStart == -1 || amplicon.rightStart == -1,
+          f"Either the left or right amplicon coordinates must be -1 when using --first-of-pair. Found ${amplicon.mkString("\t")}"
+        )
+      }
+      else {
+        require(amplicon.leftStart > 0 && amplicon.rightStart > 0,
+          f"Either the left or right amplicon coordinates must be > 0. Found ${amplicon.mkString("\t")}"
```
FelixMoelder, line 135: This should read:
```suggestion
          f"Both the left and right amplicon coordinates must be > 0. Did you forget to set '--first-of-pair'? Found ${amplicon.mkString("\t")}"
```
```suggestion
          f"Either the left or right amplicon coordinates must be > 0. Found ${amplicon.mkString("\t")}"
```

FelixMoelder:
Thanks for offering an alternative variant of single primer trimming.
@dlaehnemann has created a small set of test data which we applied on your and our implementation.
Performing a diff on both results showed that reads where an amplicon was found are trimmed identical by both implementations.
In case no amplicon can be found our implementation does some weired stuff trimming both ends of a read by the maximum primer length.
We did not investigate this further as it is obviously a bug which does not occur in your implementation.

Regardless on comparing both results we also checked if the reads are trimmed as we would except (and double checked the more complex cases).
It looks like everything is fine so far and the more complex cases (leading softclips, extremely short reads, reads w/o amplicon) are also handled correctly.

Therefore, we suggest merging your implementation as it is also much cleaner and we don't see a reason trying to fix ours.

[Test_primers.zip](https://github.com/fulcrumgenomics/fgbio/files/6762731/Test_primers.zip)

nh13:
@FelixMoelder thank-you for testing this implementation, we'll get it code reviewed and merged!

FelixMoelder:
@nh13 Can you estimate when this will be merged?

PR #686, opened 2021-07-14
Update unittests.yaml

nh13:


mjhipp:
Git LFS injecting lines? Looks like all failures are like this:

```
[info]   htsjdk.samtools.SAMFormatException: Error parsing text SAM file. Not enough fields; File src/test/resources/com/fulcrumgenomics/bam/estimate_pooling_fractions/HG01879.bam; Line 1
[info] Line: version git-lfs.github.com/spec/v1
```

PR #687, opened 2021-07-21
[feature] skip sort check fixmates

sstadick:
This adds a flag to optionally skip the check of the header in `SetMateInformation` when we promise we know the input reads are query grouped, such as piping from `bwa mem` into `fgbio SetMateInformation`. If the `--skip-order-check` flag is set, we take the liberty of setting the output writer ordering to the least stringent-yet-acceptable output ordering, otherwise it will assume a stricter ordering and possibly fail.


src/main/scala/com/fulcrumgenomics/bam/SetMateInformation.scala
```diff
@@ -50,20 +50,25 @@ class SetMateInformation
   @arg(flag='i', doc="Input SAM/BAM/CRAM file.")                      val input: PathToBam = Io.StdIn,
   @arg(flag='o', doc="Output SAM/BAM/CRAM file.")                     val output: PathToBam = Io.StdOut,
   @arg(flag='r', doc="Reference fasta, only needed if writing CRAM.") val ref: Option[PathToFasta] = None,
+  @arg(flag='s', doc="Skip checking the header for sort order of alignments.")
```
nh13, line 53: assume sorted?

src/main/scala/com/fulcrumgenomics/bam/SetMateInformation.scala
```diff
@@ -50,20 +50,25 @@ class SetMateInformation
...
+  if (!skipSortOrderCheck && in.header.getSortOrder != SortOrder.queryname && in.header.getGroupOrder != GroupOrder.query) {
     in.safelyClose()
     throw new ValidationException("Input is not queryname sorted or grouped.")
   }
 
   override def execute(): Unit = {
-    val out = SamWriter(output, in.header, ref=ref)
+    // if we are skipping the header order check, there is a good chance that the headers sort order is not set and the writer
+    // will expect coordinate sorted output. This lowers the expectation of the writer to the least sorted possible input.
+    val sortOrder = if (skipSortOrderCheck) Some(SamOrder.RandomQuery) else None
```
nh13, line 70:
```suggestion
    val sortOrder = if (skipSortOrderCheck) Some(SamOrder.Unknown) else None
```

nh13 (changes requested):
Can you please add a unit test where the header says `unsorted` and it fails when the new arg isn't used and passes when it is used?

nh13 (approved):

PR #689, opened 2021-07-27
[feature] Add Feature GeneAnnotations.Transcript

sstadick:
Keep all features under the _transcript_ level in the hierarchy and add annotation on the _transcript_ to differentiate biotypes at the hierarchical level.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
@@ -78,7 +78,9 @@ object GeneAnnotations {
                         cdsStart: Option[Int] = None,
                         cdsEnd: Option[Int]   = None,
                         negativeStrand: Boolean,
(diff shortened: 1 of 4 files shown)


src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -78,7 +78,9 @@ object GeneAnnotations {
                         cdsStart: Option[Int] = None,
                         cdsEnd: Option[Int]   = None,
                         negativeStrand: Boolean,
-                        exons: Seq[Exon]) extends Locatable {
+                        exons: Seq[Exon],
+                        features: Seq[Feature] = Seq(),
+                        kind: Option[String] = None) extends Locatable {
```
sstadick, line 83: Keep `kind` optional as the RefFlatSource parser is kind-unaware at the moment.

tfenne:
Replaced by a different PR.

PR #693, opened 2021-08-12
[feature] write untrimmed reads to separate output file

FelixMoelder:
We would like to propose an optional parameter for writing reads without primers to a separate file.
When processing reads in our pipeline we want to separate off-target reads from targeted ones.
In case of primer pairs this can be achieved by filtering reads that do not map to the regions spanned of primer pairs.
If only one primer exists the target regions can not be derived from the primer positions as we have an open frame.
Nevertheless, we can separate targeted reads from off-target reads by filtering reads without primers during trimming.
The proposed changes would allow to write those reads for which no amplicon can be detected to a separate file.
This approach would work for both, single and paired, primers.


src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -142,12 +160,16 @@ class TrimPrimers
...
+      val skip = trimReadsForTemplate(detector, reads)
+      (write_primerless_opt, skip) match {
+        case (Some(write_primerless), true) => reads.foreach(write_primerless)
+        case _ => reads.foreach(write)
+      }
       reads.foreach(trimProgress.record)
     }
 
     // If we had a reference and re-sorted above, reset the NM/UQ/MD tags as we push to the final output
+    //TODO How to implement resorting for primerless reads?
```
FelixMoelder, line 172: As I am new to Scala I was not sure how exactly sorting works which is why primerless reads currently remain unsorted when written to the output file.

nh13:
I've take a quick scan, and this will take more than few minutes to review, so I'll add it to my backlog.  Thank-you for the contribution (so far)!

johanneskoester:
HI @nh13, any chance to get this merged? Can we help you somehow?

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -106,14 +112,26 @@ class TrimPrimers
     //         i) No reference was given (so no coordinate sort) and the output sort order is queryname, or
     //        ii) A reference was given (so the last sort is coordinate) and the output sort order is coordinate
     //      else, we'll have to sort for potentially a third time in the SAMFileWriter
-    val (sorter, write: (SamRecord => Any), out: SamWriter) = ref match {
+    val (sorter, write: (SamRecord => Any), out: SamWriter, write_primerless_opt: Option[(SamRecord => Any)], primerless_out_opt: Option[SamWriter]) = ref match {
```
nh13, line 115: Please use lower camel case for variables and methods throughout.  See: https://docs.scala-lang.org/style/naming-conventions.html

nh13: This is getting quite difficult to read (not your fault), let me make PR to clean this up

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -164,6 +186,10 @@ class TrimPrimers
     }
 
     out.close()
+    primerless_out_opt match {
+      case Some(primerless_out) => primerless_out.close()
+      case _ => ()
+    }
```
nh13, line 192: Use a collection method (i.e. `foreach`) versus a match statement
```suggestion
    primerless_out_opt.foreach(_.close())
```

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -178,12 +204,12 @@ class TrimPrimers
...
 
   /** Trims all the reads for a given template. */
-  def trimReadsForTemplate(detector: AmpliconDetector, reads: Seq[SamRecord]): Unit = {
+  def trimReadsForTemplate(detector: AmpliconDetector, reads: Seq[SamRecord]): Boolean = {
     val rec1 = reads.find(r => r.paired && r.firstOfPair  && !r.secondary && !r.supplementary)
     val rec2 = reads.find(r => r.paired && r.secondOfPair && !r.secondary && !r.supplementary)
 
     val readsToClip = if (firstOfPair) reads.filter(_.firstOfPair) else reads
-
+    var skip = false
```
nh13, line 212: Avoid using `var`s in scala as much as possible, unless it's very tightly scoped and needed for performance (e.g. faster for loops). I'll update with my PR

nh13 (changes requested):
@FelixMoelder: I will make a PR to help expedite this PR.

nh13:
@FelixMoelder this PR is getting quite unwieldy after I tried fixing it up.  I'd rather separate the concerns about determining the reads that hand no primers/amplicons assigned.

Why not:
1. Assign primers wit h`AssignPrimers`
2. Split the output BAM from (1) into two, those who were assigned and those who were not
3. Run `TrimPrimers` on the BAM with assigned primers from (2).

I think this works out of the box

The only optimization could be to have `TrimPrimers` use the assigned primers from `AssignPrimers` versus doing the lookup a second time.  

What are your thoughts?

FelixMoelder:
Thanks a lot for your notes and suggestions.
I was not aware of fgbio being capable assigning primers to reads which is probably a feasible solution for separating trimmed from untrimmed reads.
However, in its currently implementation assigning primers to reads is only working for primer pairs.
In our case we have, once again, only single primers (see https://github.com/fulcrumgenomics/fgbio/pull/681).
To get this running it would be necessary that `AssignPrimers` is capable of handle single primers by marking read pairs in case one of both reads contains a primer.
Still, I agree that this would be a more consistent implementation of this functionality than adding an extra parameter.

nh13:
@FelixMoelder I think `AssignPrimers` will work out of the box.  From the usage:
> Each read is assigned independently of its mate (for paired end reads).
Can you try it on your data and report back?

FelixMoelder:
I have already tried on our testdata and it looks like no primers get assigned at all.
From what I see in the metrics file the issue appears to be that the amplicon region spans both primers which is `-1` in case no second primer exists.
[metrics.txt](https://github.com/fulcrumgenomics/fgbio/files/7126357/metrics.txt)

nh13:
@FelixMoelder would you be willing to share a small test example (BAM and primers)?

FelixMoelder:
@nh13 I am currently not allowed to share the data I'm working with.
But I'll generate some artificial data to share them here asap.

FelixMoelder:
@nh13 I have created some test data consisting of some reads and primers.
When testing the data applying `AssignPrimers` I realized that this indeed seems to work for single primers.
The reason I did not find any reads before was that I greped for the tags defined in the subcommand description:

```
The output will have the following tags added:

  * ap: the assigned primer coordinates (ex. 'chr1:1010873-1010894')
  * am: the mate's assigned primer coordinates (ex. 'chr1:1011118-1011137')
  * ip: the assigned amplicon id
  * im: the mate's assigned amplicon id (or '=' if the same as the assigned amplicon)
  ```
  
  Greping for any of these tags did not return a record.
When looking at the test results I realized that the reads got different tags attached:  `rp`, `mp`, `ra` and `ma`.
These tags are acutely defined as default in the parameter description. So there seems to be some confusion between the tags defined in the subcommand description and the real ones defined in the parameter descriptions.
[TestAssignPrimers.zip](https://github.com/fulcrumgenomics/fgbio/files/7159800/TestAssignPrimers.zip)

nh13:
@FelixMoelder thanks for sending along a test case.  Thank-you for your patience as I am 100% booked at work and have little time with kids at home.

FelixMoelder:
I'll close this PR as this functionality can be achieved, as suggested, by `fgio AssignPrimers` and a separat filtering step.

PR #696, opened 2021-08-28
Make Variant locatable by mixing in GenomicSpan

clintval:
This will make it easy to perform coordinate math on variants.

Curious why `Variant` has the `chrom`/`pos` field by design instead of the more common `contig`/`start`?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala
+++ b/src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala
@@ -24,6 +24,8 @@
 
 package com.fulcrumgenomics.vcf.api
 
+import com.fulcrumgenomics.util.GenomicSpan
+
 import scala.collection.immutable.ListMap
 import scala.reflect.ClassTag
 
@@ -99,7 +101,13 @@ final case class Variant(chrom: String,
                          filters: Set[String] = Variant.EmptyFilters,
                          attrs: ListMap[String,Any] = Variant.EmptyInfo,
                          genotypes: Map[String, Genotype] = Variant.EmptyGenotypes
-                        ) {
+                        ) extends GenomicSpan {
+
+  /** The chromosome on which a variant resides. */
+  override val contig: String = chrom
+
+  /** The start position of the variant. */
+  override val start: Int = pos
 
   /** The end position of the variant based on either the `END` INFO field _or_ the length of the reference allele. */
   val end: Int = get[Int]("END").getOrElse(pos + alleles.ref.length - 1)
--- a/src/test/scala/com/fulcrumgenomics/vcf/api/VariantTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/vcf/api/VariantTest.scala
@@ -24,11 +24,15 @@
 
 package com.fulcrumgenomics.vcf.api
 
-import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.testing.UnitSpec
+import htsjdk.samtools.util.Locatable
 import org.scalatest.OptionValues
+import org.scalatest.matchers.must.Matchers.convertToAnyMustWrapper
+
+import scala.collection.immutable.ListMap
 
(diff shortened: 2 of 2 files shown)


tfenne:
I'll take a look at this soon-ish.  To answer you question about why "chrom" and "pos" instead of "contig" and "start" ... I have never really approved of the use of "contig" in this context and think it is just wrong as used in HTSJDK's interval/SAMRecord etc.  At least for human and other "finished" genomes the chromosomes are generally scaffolds and not contigs, so not only is it less idiomatic that just using "chromosome", it's just plain _wrong_.  For VCF I think it's also useful to have the API names resemble the file format, so "chrom" and "pos" since they match the VCF columns.

clintval:
I agree with you on word definitions!

I also feel like the ship has sailed on what we should call such abstractions (contig vs chrom vs refName vs _other_) in upstream toolkits like htsjdk.

In code I have control over, I think I would try to pick one and stick with it so interfaces/traits can be built. And since `Locatable` is one of the most commonly used mixins in code I write, I lean towards that interface's field names (`contig`/`start`).

I did a search here and found the following:

https://github.com/fulcrumgenomics/fgbio/blob/468a843d19da4f67106dc148a1dba1f2c3390eab/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala#L139

https://github.com/fulcrumgenomics/fgbio/blob/468a843d19da4f67106dc148a1dba1f2c3390eab/src/main/scala/com/fulcrumgenomics/util/GenomicSpan.scala#L34

https://github.com/fulcrumgenomics/fgbio/blob/468a843d19da4f67106dc148a1dba1f2c3390eab/src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala#L489

https://github.com/fulcrumgenomics/fgbio/blob/e91698c6e1a77079593e60cad207da7c499ad798/src/main/scala/com/fulcrumgenomics/util/Amplicon.scala#L58

https://github.com/fulcrumgenomics/fgbio/blob/468a843d19da4f67106dc148a1dba1f2c3390eab/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala#L76

https://github.com/fulcrumgenomics/fgbio/blob/468a843d19da4f67106dc148a1dba1f2c3390eab/src/main/scala/com/fulcrumgenomics/FgBioDef.scala#L72

https://github.com/fulcrumgenomics/fgbio/blob/468a843d19da4f67106dc148a1dba1f2c3390eab/src/main/scala/com/fulcrumgenomics/vcf/api/VcfHeader.scala#L84

https://github.com/fulcrumgenomics/fgbio/blob/468a843d19da4f67106dc148a1dba1f2c3390eab/src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala#L74

https://github.com/fulcrumgenomics/fgbio/blob/468a843d19da4f67106dc148a1dba1f2c3390eab/src/main/scala/com/fulcrumgenomics/bam/Pileup.scala#L39

clintval:
FYI I'm not requesting a backwards incompatible change. The PR as-is doesn't break anything, but does let me use a `Variant` wherever a `Locatable` might be desired which is useful :D

clintval:
@tfenne do you think this PR is OK?

I have some user code that will be simplified if variants become locatable.

nh13:
@clintval I am going to try something this weekend that @tfenne and I discussed to see if we can meet both your and @tfenne 's goals

nh13:
See: https://github.com/fulcrumgenomics/fgbio/pull/699

clintval:
Thanks! If that's the way you want to go, we can do that.

It's what I'm already doing in user code, which has the side-effect of creating extra class instances.

I'd really like these features to be locatable but I also understand you don't like the getter names htsjdk chose.

It's up to you which to merge!

nh13:
@clintval will favor #699 to this, just going to wait on @tfenne's review.  

I think the tidiness of `Variant` is really nice, so I think the side-effect of extra class instances is acceptable to keep the # of methods on a `Variant` reduced.

PR #702, opened 2021-09-08
[feature] add a cache into CorrectUmis

nh13:
This is useful when we have a large set (>1K) of Umis

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
@@ -27,6 +27,7 @@ package com.fulcrumgenomics.umi
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.bam.api.{SamSource, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.commons.collection.LeastRecentlyUsedCache
 import com.fulcrumgenomics.commons.util.LazyLogging
 import com.fulcrumgenomics.sopt.{arg, clp}
 import com.fulcrumgenomics.umi.CorrectUmis._
@@ -34,6 +35,7 @@ import com.fulcrumgenomics.util.Metric.{Count, Proportion}
 import com.fulcrumgenomics.util.{Metric, _}
 
 import scala.collection.mutable
+
 
 object CorrectUmis {
   /**
@@ -117,6 +119,9 @@ object CorrectUmis {
     |Records which have their UMIs corrected (i.e. the UMI is not identical to one of the expected UMIs but is
     |close enough to be corrected) will by default have their original UMI stored in the `OX` tag. This can be
     |disabled with the `--dont-store-original-umis` option.
+    |
+    |For a large number of input UMIs, the `--cache-size` option may used to speed up the tool.  To disable
+    |using a cache, set the value to `0`.
   """)
 class CorrectUmis
 ( @arg(flag='i', doc="Input SAM or BAM file.")  val input: PathToBam,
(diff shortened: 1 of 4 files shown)


src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -128,7 +134,8 @@ class CorrectUmis
   @arg(flag='u', doc="Expected UMI sequences.", minElements=0) val umis: Seq[String] = Seq.empty,
   @arg(flag='U', doc="File of UMI sequences, one per line.", minElements=0) val umiFiles: Seq[FilePath] = Seq.empty,
   @arg(flag='t', doc="Tag in which UMIs are stored.") val umiTag: String = ConsensusTags.UmiBases,
-  @arg(flag='x', doc="Don't store original UMIs upon correction.") val dontStoreOriginalUmis: Boolean = false
+  @arg(flag='x', doc="Don't store original UMIs upon correction.") val dontStoreOriginalUmis: Boolean = false,
+  @arg(doc="The number of results to cache; zero will not cache results.") val cacheSize: Int = 100000
```
pamelarussell, line 138:
```suggestion
  @arg(doc="The number of uncorrected UMIs to cache; zero will disable the cache.") val cacheSize: Int = 100000
```

pamelarussell (approved):
One small suggestion to clarify documentation.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -252,12 +262,26 @@ class CorrectUmis
 
   /** Given a UMI sequence and a set of fixed UMIs, report the best match. */
   private[umi] def findBestMatch(bases: String, umis: Array[String]): UmiMatch = {
-    val mismatchLimit = this.maxMismatches + minDistance
-    val mismatches    = umis.map(umi => Sequences.countMismatches(bases, umi))
-    val min           = mismatches.min
-    val matched       = (min <= maxMismatches) && (mismatches.count(m => m < min + this.minDistance) == 1)
-
-    UmiMatch(matched, umis(mismatches.indexOf(min)), min)
+    if (cacheSize > 0 && cache.containsKey(bases)) cache.get(bases)
```
tfenne, line 265: You're doing double work here by using `containsKey` and then `get`.  I'd rewrite to do something like 

```scala
val previous = if (cacheSize > 0) cache.get(bases) else None
previous match {
...
}
```

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -252,12 +262,26 @@ class CorrectUmis
...
+private class LRUCache[K, V](maxEntries: Int)
+  extends java.util.LinkedHashMap[K, V](100, .75f, true) {
+  override def removeEldestEntry(eldest: java.util.Map.Entry[K, V]): Boolean = size > maxEntries
+}
+
+private object LRUCache {
+  def apply[K, V](maxEntries: Int): mutable.Map[K, V] = {
+    synchronizedMap(new LRUCache[K, V](maxEntries)).asScala
+  }
+}
```
tfenne, line 287: I think:

1. This should use scala.collection.mutable.LinkedHashMap
2. It should not extend, but have the `map` as a `private` val
3. Only public methods should be `get()` and `put()`, both synchronized, with `put()` doing the eviction as needed.

tfenne (changes requested):
I suspect this is "good enough" for what you're doing, but it's not actually an LRU cache.  To be LRU you need to update the cache on get to keep the "hot" items in cache and not have them evicted as time goes by.  I'm not sure what the overhead is though on removing and re-adding on get.  Either way, makes a good case for encapsulating the LRUCache better and not just extending Map.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
tfenne: You could also add a `getOrElseUpdate(key: Key, value: => Value)`

tfenne (approved):

PR #703, opened 2021-09-09
Validate that a Variant and its Genotypes have the same alleles

clintval:
It's my understanding that the `alleles` field of Genotype should be identical to the `alleles` field of its parent `Variant`. 

Would you be OK adding this as a construction-time validation?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala
+++ b/src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala
@@ -100,7 +100,7 @@ final case class Variant(chrom: String,
                          attrs: ListMap[String,Any] = Variant.EmptyInfo,
                          genotypes: Map[String, Genotype] = Variant.EmptyGenotypes
                         ) {
-  require(gts.forall(_.alleles == alleles), "Genotypes must have the same alleles as their Variant.")
+  require(gts.forall(gt => gt.alleles eq alleles || gt.alleles == alleles), "Genotypes must have the same alleles as their Variant.")
 
   /** The end position of the variant based on either the `END` INFO field _or_ the length of the reference allele. */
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/vcf/api/Variant.scala
```diff
@@ -100,7 +100,7 @@ final case class Variant(chrom: String,
                          attrs: ListMap[String,Any] = Variant.EmptyInfo,
                          genotypes: Map[String, Genotype] = Variant.EmptyGenotypes
                         ) {
-  require(genotypes.values.forall(_.alleles == alleles), "Genotypes must have the same alleles as their Variant.")
+  require(gts.forall(_.alleles == alleles), "Genotypes must have the same alleles as their Variant.")
```
tfenne, line 103: How about:
```suggestion
  require(gts.forall(gt => gt.alleles eq alleles || gt.alleles == alleles), "Genotypes must have the same alleles as their Variant.")
```

tfenne: I worry a bit about parsing e.g. large VCFs with thousands of samples/genotypes and the overhead.  But when we read/construct from the Java versions, we construct the allele set once and re-use it, so checking for identity first and then equality should short-circuit fairly effectively.

tfenne (changes requested):

clintval:
I think that should work well!

tfenne (approved):
Happy to merge once tests pass.  I think that's a CI issue not an issue with this PR though.  @nh13 ?

clintval:
I saw... I think the compiler doesn't like the suggestion I auto-merged but I can fix later today!

clintval:
Fixed!

PR #706, opened 2021-09-11
Add "biotype" to Gene and update NcbiRefSeqParser to support more gene biotypes

tfenne:
This is a combination of work done by @sstadick and myself, though I've backed out some of the things @sstadick did that turned out to be unecessary.

The main goals here are a) to add biotype to the `Gene` class via the `gene_biotype` field in the NCBI RefSeq GFFs and b) to make sure that we are actually able to parse all the `gene` entries in the GFF.  We were getting most of them before, but there are a subset (tRNA, *_Segment, etc.) that on the record below the `gene` records, do not define a `transcript_id` (sigh), so I've had to update how we determine what's a transcript (basically defined by having exons now).

Will squash this one we're happy, and remove the `SummarizeGff` personal tool but have left that there for the minute since it's useful until the PR is merged.


src/main/scala/com/fulcrumgenomics/personal/tfenne/SummarizeGff.scala
```diff
@@ -0,0 +1,62 @@
+/*
```
tfenne, line 1: IGNORE THIS CLASS - it will be removed before the PR is merged.

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
```diff
@@ -259,7 +262,15 @@ class NcbiRefSeqGffSource private(lines: Iterator[String],
       case xs    => (Some(xs.minBy(_.start).start), Some(xs.maxBy(_.end).end))
     }
 
-    Transcript(txRec("transcript_id"), chrom, txRec.start, txRec.end, cdsStart, cdsEnd, txRec.negative, exons)
+    // Figure out the ID of the transcript-like row
+    val id = if (txRec.has("transcript_id")) txRec("transcript_id") else txRec("ID").split('-').drop(1).mkString("-")
```
tfenne, line 266: This is done because the `ID` field is forced to be file-wide unique, and so names get prefixed with silly things. E.g. for the tRNA I've added to the test GFF, the ID field is `ID=rna-TRN-GTT5-1`.  Other types get `id-` prepended or other not-useful strings.

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -114,4 +115,84 @@ object GeneAnnotations {
...
+    def isCoding: Boolean
+  }
+
+  /** Enum to represent gene biotypes in a GFF */
+  object GeneBiotype extends FgBioEnum[GeneBiotype] {
+    /** All values of the GeneBiotype Enum */
+    def values: IndexedSeq[GeneBiotype] = findValues
+
+    /** Allows GeneBiotypes to be built from the Enum name of the biotype name from the GFF */
+    override def apply(str: String): GeneBiotype = values.find(_.key == str).getOrElse(super.apply(str))
```
nh13, line 133: The key is so close to being able to be inferred from the class name!

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -114,4 +115,84 @@ object GeneAnnotations {
...
+    case object JSegment      extends GeneBiotype { val key: String = "J_segment";      val isCoding = true }
+
+    // lnc_RNA -> exon
+    case object LncRna        extends GeneBiotype { val key: String = "lncRNA";         val isCoding = false }
+
+    // primary_transcript -> {exon, miRNA}  <-- odd
+    case object MiRna         extends GeneBiotype { val key: String = "miRNA";          val isCoding = false }
+
+    // transcript -> exon                   <-- odd
+    case object MiscRna       extends GeneBiotype { val key: String = "misc_RNA";       val isCoding = false } // TODO: check this
```
nh13, line 157: TODO

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -114,4 +115,84 @@ object GeneAnnotations {
...
+
+    // *_feature or nothing                 <-- odd
+    case object Other         extends GeneBiotype { val key: String = "other";          val isCoding = false } // TODO: check this
```
nh13, line 160: TODO

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -114,4 +115,84 @@ object GeneAnnotations {
...
+  /** Enum to represent gene biotypes in a GFF */
+  object GeneBiotype extends FgBioEnum[GeneBiotype] {
+    /** All values of the GeneBiotype Enum */
+    def values: IndexedSeq[GeneBiotype] = findValues
+
+    /** Allows GeneBiotypes to be built from the Enum name of the biotype name from the GFF */
+    override def apply(str: String): GeneBiotype = values.find(_.key == str).getOrElse(super.apply(str))
+
+    // antisense_RNA -> exon
+    case object AntisenseRna  extends GeneBiotype { val key: String = "antisense_RNA";  val isCoding = false } // TODO: check this
```
nh13, line 136: TODO

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -114,4 +115,84 @@ object GeneAnnotations {
...
+
+  /** Enum to represent gene biotypes in a GFF */
+  object GeneBiotype extends FgBioEnum[GeneBiotype] {
+    /** All values of the GeneBiotype Enum */
+    def values: IndexedSeq[GeneBiotype] = findValues
+
+    /** Allows GeneBiotypes to be built from the Enum name of the biotype name from the GFF */
+    override def apply(str: String): GeneBiotype = values.find(_.key == str).getOrElse(super.apply(str))
+
+    // antisense_RNA -> exon
```
nh13, line 135: What do these comments mean?

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
```diff
@@ -114,4 +115,84 @@ object GeneAnnotations {
...
+  /** Trait that gene biotypes will extends */
+  sealed trait GeneBiotype extends EnumEntry {
+    /** The biotype name in the GFF */
+    def key: String
+    /** True if this biotype represents a protein coding feature */
+    def isCoding: Boolean
+  }
+
+  /** Enum to represent gene biotypes in a GFF */
+  object GeneBiotype extends FgBioEnum[GeneBiotype] {
```
nh13, line 128: 1. Is there a URL you can link to that describes the set of possible biotypes in a GFF from NCBI's RefSeq?  
2. Here are the [biotypes from GENCODE and Ensembl](https://www.gencodegenes.org/pages/biotypes.html).  I'd put a note in that we only support biotypes from (1), but future biotypes could be added.

I am concerned this is too NCBI-specific now, any thoughts on how to make that clearer?

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
```diff
@@ -226,27 +226,30 @@ class NcbiRefSeqGffSource private(lines: Iterator[String],
     * */
   private def parseGene(chrom: String, geneRec: GffRecord, iter: BetterBufferedIterator[GffRecord]): Option[Gene] = {
     val txsBuilder = Seq.newBuilder[Transcript]
+    val biotype = GeneBiotype(geneRec("gene_biotype"))
```
nh13, line 229:
```suggestion
    val txsBuilder = Seq.newBuilder[Transcript]
    val biotype    = GeneBiotype(geneRec("gene_biotype"))
```

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
```diff
@@ -226,27 +226,30 @@ class NcbiRefSeqGffSource private(lines: Iterator[String],
...
 
     // Loop while the next record is a child of this gene and is "transcript-like", which includes any
     // direct children that define the attribute "transcript_id", not just records with kind=transcript.
-    while (iter.hasNext && iter.head.parentId.contains(geneRec.id) && iter.head.has("transcript_id")) {
+    while (iter.hasNext && iter.head.parentId.contains(geneRec.id)) {
```
nh13, line 233: 1. Why not use `headOption`?
```suggestion
    while (iter.headOption.exists(_.contains(geneRec.id))) {
```
2. I think the comment about "transcript-like" is out of date

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
```diff
@@ -226,27 +226,30 @@ class NcbiRefSeqGffSource private(lines: Iterator[String],
...
       val tx = parseTranscript(chrom, txRec, iter)
-      if (this.includeXs || !tx.name.startsWith("X")) txsBuilder += tx
+      tx.filter(t => this.includeXs || !t.name.startsWith("X")).foreach(txsBuilder += _)
     }
 
     val txs = txsBuilder.result()
     if (txs.isEmpty) None else {
       val locus = GeneLocus(txs)
-      Some(Gene(geneRec.name, Seq(locus)))
+      Some(Gene(geneRec.name, Seq(locus), Some(biotype)))
```
nh13, line 242:
```suggestion
      Some(Gene(name=geneRec.name, locia=Seq(locus), biotype=Some(biotype)))
```

nh13: Also, why not inline `GeneLocus(txs)`?

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
nh13: I am a bit confused.  Where does _fgbio_ prepend a field to the transcript ID besides the tests?  Or are you now requiring the user to have a some field before the "-" to make it work when `transcript_id` is not not found?

nh13 (changes requested):
A bunch of minor nitpicks, with two issues:
1. how do we deal with biotypes that aren't NCBI specific and make that clearer?
2. it makes sense that the ID must be file-wide unique, but I am confused about some of the assumptions when _parsing_ the GFF, since I don't see where we prepend a leading field in the ID elsewhere.

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
tfenne: fgbio doesn't prepend anything ... RefSeq does.  Perhaps a couple of examples will be helpful?  Apologies that this is a bit hard to read in github:

```
NC_000001.11  Gnomon       gene        89006     267354    .  -  .  ID=gene-LOC100996442;Dbxref=GeneID:100996442;Name=LOC100996442;gbkey=Gene;gene=LOC100996442;gene_biotype=misc_RNA
NC_000001.11  Gnomon       transcript  89006     267354    .  -  .  ID=rna-XR_001737582.2;Parent=gene-LOC100996442;Dbxref=GeneID:100996442,Genbank:XR_001737582.2;Name=XR_001737582.2;gene=LOC100996442;transcript_id=XR_001737582.2
...
NC_000001.11  tRNAscan-SE  gene        16520585  16520658  .  -  .  ID=gene-TRN-GTT5-1;Dbxref=GeneID:107985757,HGNC:HGNC:38582;Name=TRN-GTT5-1;gbkey=Gene;gene=TRN-GTT5-1;gene_biotype=tRNA
NC_000001.11  tRNAscan-SE  tRNA        16520585  16520658  .  -  .  ID=rna-TRN-GTT5-1;Parent=gene-TRN-GTT5-1;Dbxref=GeneID:107985757,HGNC:HGNC:38582;gene=TRN-GTT5-1;inference=COORDINATES: profile:tRNAscan-SE:1.23;product=tRNA-Asn
```

This is two genes and their "transcript" records that come next in the file.  You can see that the first "transcript" has `transcript_id=XR_001737582.2` and also `ID=rna-XR_001737582.2`.  That's just how it is in the file - I think they do this so that e.g. if a gene and a transcript have the same ID, they still end up file-wide unique because each level/type gets a different prefix like `rna-`.
The second "transcript" doesn't have `transcript_id` defined, so I need to do something else.  In this case stripping off the `rna-` and yielding `TRN-GTT5-1` from `ID=rna-TRN-GTT5-1` is about the most sensible thing I could see to do. 

If that doesn't make sense perhaps we should discuss live.

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
tfenne: I'm not sure what to suggest here.  I can't find any obvious source of information from NCBI on how they came to this set of Biotypes.  Certainly we can add more if/when we need to if we build parsers for Ensembl/Gencode GFFs.

tfenne: Have added a comment as suggested.

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
tfenne: I know.  I'm almost tempted to rename the members to match the values in the GFF.  Any thought on whether it is better to have scala-style names or mirror the values in the GFF?

src/main/scala/com/fulcrumgenomics/util/GeneAnnotations.scala
tfenne: I think these were Seth's notes from the structure of entries below the gene in the RefSeq GFF.  They don't make sense here so I'll pull them out.

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
tfenne: Good catch on the comment.  I tend to prefer `iter.hasNext` perhaps just due to familiarity.

nh13: I always though that scala folks wanted to avoid `.head`, `.get` and so on, preferring to rely on options

src/main/scala/com/fulcrumgenomics/util/NcbiRefSeqGffSource.scala
nh13: That makes sense, can you add a note in the docs that the id is transformed?

nh13 (approved):

PR #708, opened 2021-09-14
[feature] GenomicRange to support contig names with colons

nh13:


src/main/scala/com/fulcrumgenomics/FgBioDef.scala
```diff
@@ -81,7 +81,7 @@ object FgBioDef extends CommonsDef {
       * @param range the string to parse.
       */
     def apply(range: String): GenomicRange = {
-      val (refName: String, rest: String) = range.indexOf(':') match {
+      val (refName: String, rest: String) = range.lastIndexOf(':') match {
```
kstromhaug, line 84: is it possible that there could be a stray colon at the end?

kstromhaug (approved):

tfenne (approved):

PR #711, opened 2021-09-16
Add an option to DemuxFastqs to insert sample barcodes in the FASTQ header

kstromhaug:
1. Deprecates the `--illumina-standards` option in favor of more explicit fine-grained options
2. `--fastq-include-sample-barcodes` will insert the sample barcode(s) into the FASTQ comment in the FASTQ header
3. `--fastq-skip-read-numbers` will _not_ insert trailing `/1` or `/2/ into the read names in the FASTQs
4. `--illumina-file-names` will name the output files according to bcl2fastq output file names

The above (2-3) are named the way they are with defaults to not change the current default behavior.

Also added the `--platform` option as not all reads are Illumina these days.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -810,10 +810,10 @@ object ReadInfo {
   private def reject(name: String) = throw new IllegalArgumentException(s"Cannot extract ReadInfo due to missing comment: $recname")
 
   /** Builds the [[ReadInfo]] by parsing a [[FastqRecord]]. */
-  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name))
+  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name)))
 
   /** Builds the [[ReadInfo]] by parsing a [[DemuxRecord]]. */
-  def apply(rec: DemuxRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name))
+  def apply(rec: DemuxRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name)))
 
   /** Builds the [[ReadInfo]] by parsing a standard input FASTQ. */
   def apply(name: String, comment: String): ReadInfo = try { // <- comment is an Option[String] in FastqRecord and DemuxRecord
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -807,11 +807,13 @@ case class ReadInfo(readNumber: Int, passQc: Boolean, internalControl: Boolean, 
 }
 
 object ReadInfo {
+  private def reject(name: String) = throw new IllegalArgumentException(s"Cannot extract ReadInfo due to missing comment: $recname")
+
   /** Builds the [[ReadInfo]] by parsing a [[FastqRecord]]. */
-  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.get)
+  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name))
 
   /** Builds the [[ReadInfo]] by parsing a [[DemuxRecord]]. */
-  def apply(rec: DemuxRecord): ReadInfo = this(rec.name, rec.comment.get)
+  def apply(rec: DemuxRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name))
 
   /** Builds the [[ReadInfo]] by parsing a standard input FASTQ. */
   def apply(name: String, comment: String): ReadInfo = try { // <- comment is an Option[String] in FastqRecord and DemuxRecord
@@ -823,10 +825,10 @@ object ReadInfo {
       s"Expected the comment format 'ReadNum:FilterFlag:0:SampleNumber', found '$readInfo'")
     val Seq(readNumber, keep, internalControl, sampleInfo) = commentInfo
 
-    val keepBoolean: Boolean = {
-      if (keep == "Y") true
-      else if (keep == "N") false
-      else throw new IllegalStateException(s"Cannot parse filter/keep flag: $keep")
+    val keepBoolean = keep match {
+      case "Y" => true
+      case "N" => false
+      case   x => throw new IllegalStateException(s"Cannot parse filter/keep flag: $x")
     }
 
     ReadInfo(
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -807,7 +807,7 @@ case class ReadInfo(readNumber: Int, passQc: Boolean, internalControl: Boolean, 
 }
 
 object ReadInfo {
-  private def reject(name: String) = throw new IllegalArgumentException(s"Cannot extract ReadInfo due to missing comment: $recname")
+  private def reject(name: String) = throw new IllegalArgumentException(s"Cannot extract ReadInfo due to missing comment: $name")
 
   /** Builds the [[ReadInfo]] by parsing a [[FastqRecord]]. */
   def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name)))
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -483,9 +483,11 @@ private class SamRecordWriter(prefix: PathPrefix,
       record.mateUnmapped = true
       if (rec.readNumber == 1) record.firstOfPair = true else record.secondOfPair = true
     }
-    rec.sampleBarcode.foreach(bc => record("BC") = bc)
+//    rec.sampleBarcode.foreach(bc => record("BC") = bc)
+    if (rec.sampleBarcode.nonEmpty) record("BC") = rec.sampleBarcode.mkString("-")
     record(ReservedTagConstants.READ_GROUP_ID) =  rgId
-    rec.molecularBarcode.foreach(mb => record(umiTag) = mb)
+//    rec.molecularBarcode.foreach(mb => record(umiTag) = mb)
+    if (rec.molecularBarcode.nonEmpty) record(umiTag) = rec.molecularBarcode.mkString("-")
     writer += record
   }
 
@@ -516,9 +518,15 @@ private[fastq] class FastqRecordWriter(prefix: PathPrefix, val pairedEnd: Boolea
       require(rec.name.count(_ == ':') == 6,
         s"Expected the read name format 'Instrument:RunID:FlowCellID:Lane:Tile:X:Y', found '${rec.name}'")
       require(comment.count(_ == ':') == 3,
-        s"Expected the comment format 'ReadNum:FilterFlag:0:SampleNumber', found '${rec.name}'")
+        s"Expected the comment format 'ReadNum:FilterFlag:0:SampleNumber', found '${comment}'")
 
-      s"${rec.name} $comment"
+      val fields = comment.split(' ').toSeq
+      val Seq(readNum, qc, internalControl, sampleBarcode) = fields.head.split(':').toSeq
+      val newSampleBarcode = rec.sampleBarcode.mkString("+")
+      val newFirstTerm = Seq(readNum, qc, internalControl, newSampleBarcode).mkString(":")
+      val newComment = (newFirstTerm +: fields.tail).mkString(" ")
+
+      s"${rec.name} $newComment"
     }
     else {
       Seq(Some(rec.name), rec.sampleBarcode, rec.molecularBarcode).flatten.mkString(":")
@@ -556,8 +564,8 @@ private[fastq] case class SampleInfo(sample: Sample, isUnmatched: Boolean = fals
 private[fastq] object FastqDemultiplexer {
 
   /** Stores the minimal information for a single template read. */
-  case class DemuxRecord(name: String, bases: String, quals: String, molecularBarcode: Option[String],
-                         sampleBarcode: Option[String], readNumber: Int, pairedEnd: Boolean, comment: Option[String],
+  case class DemuxRecord(name: String, bases: String, quals: String, molecularBarcode: Seq[String],
+                         sampleBarcode: Seq[String], readNumber: Int, pairedEnd: Boolean, comment: Option[String],
                          originalBases: Option[String] = None, originalQuals: Option[String] = None)
 
   /** A class to store the [[SampleInfo]] and associated demultiplexed [[DemuxRecord]]s.
@@ -684,9 +692,9 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
     val (sampleInfo, numMismatches) = matchSampleBarcode(subReads)
 
     // Method to get all the bases of a given type
-    def bases(segmentType: SegmentType): Option[String] = {
-      val b = subReads.filter(_.kind == segmentType).map(_.bases).mkString("-")
-      if (b.isEmpty) None else Some(b)
+    def bases(segmentType: SegmentType): Seq[String] = {
+      subReads.filter(_.kind == segmentType).map(_.bases)
+//       if (b.isEmpty) None else Some(b) // <-- ask about this part, not quite sure what Some does.
     }
 
     // Get the molecular and sample barcodes
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -483,10 +483,8 @@ private class SamRecordWriter(prefix: PathPrefix,
       record.mateUnmapped = true
       if (rec.readNumber == 1) record.firstOfPair = true else record.secondOfPair = true
     }
-//    rec.sampleBarcode.foreach(bc => record("BC") = bc)
     if (rec.sampleBarcode.nonEmpty) record("BC") = rec.sampleBarcode.mkString("-")
     record(ReservedTagConstants.READ_GROUP_ID) =  rgId
-//    rec.molecularBarcode.foreach(mb => record(umiTag) = mb)
     if (rec.molecularBarcode.nonEmpty) record(umiTag) = rec.molecularBarcode.mkString("-")
     writer += record
   }
@@ -694,7 +692,6 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
     // Method to get all the bases of a given type
     def bases(segmentType: SegmentType): Seq[String] = {
       subReads.filter(_.kind == segmentType).map(_.bases)
-//       if (b.isEmpty) None else Some(b) // <-- ask about this part, not quite sure what Some does.
     }
 
     // Get the molecular and sample barcodes
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -264,10 +264,10 @@ object DemuxFastqs {
       |
       |The following options affect the output format:
       |
-      |1. If `--fastq-skip-read-numbers` is specified, then trailing /1 and /2 for R1 and R2 respectively, will not be
+      |1. If `--omit-fastq-read-numbers` is specified, then trailing /1 and /2 for R1 and R2 respectively, will not be
       |appended to e FASTQ read name.  By default they will be appended.
-      |2. If `--fastq-include-sample-barcodes` is specified, then sample barcode will replace the last field in the
-      |first comment in the FASTQ header.
+      |2. If `--include-sample-barcodes-in-fastq` is specified, then sample barcode will replace the last field in the
+      |first comment in the FASTQ header, e.g. replace 'NNNNNN' in the header `@Instrument:RunID:FlowCellID:Lane:Tile:X:Y 1:N:0:NNNNNN`
       |3. If `--illumina-file-names` is specified, the output files will be named according to the Illumina FASTQ file
       |naming conventions:
       |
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -260,15 +260,36 @@ object DemuxFastqs {
       |    --metadata SampleSheet.csv --metrics metrics.txt --output output_folder
       |```
       |
-      |## Illumina FASTQ File Naming
+      |## Output Standards
```
nh13, line 263: @tfenne I originally authored this, so could you take a look to see if makes sense?

Some context: the `--illumina-standards` option was added a long time ago so that the output FASTQs would follow the format so that they can be uploaded to BaseSpace.  This included the file names AND the header format.  But in some docs for BaseSpace now, the last field in the comment is the sample barcode (index) while previously it was sample number.  

With the three new options, we've deprecated the above option in favor of three explicit options for clearer behavior.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -298,19 +319,46 @@ class DemuxFastqs
  @arg(doc="The sequencing center from which the data originated") val sequencingCenter: Option[String] = None,
  @arg(doc="Predicted median insert size, to insert into the read group header") val predictedInsertSize: Option[Integer] = None,
  @arg(doc="Platform model to insert into the group header (ex. miseq, hiseq2500, hiseqX)") val platformModel: Option[String] = None,
+ @arg(doc="Platform to insert into the group header (e.g Illumina)") val platform: String = "Illumina",
```
nh13, line 322: easter egg!

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -330,11 +378,6 @@ class DemuxFastqs
     case n => invalid(s"Found $n read structures with template bases but expected 1 or 2.")
   }
 
-  if (illuminaStandards) {
```
nh13, line 333: not sure why this was here originally, so removed it

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -457,7 +501,7 @@ class DemuxFastqs
 
 /** A writer than writes [[DemuxRecord]]s */
 private trait DemuxWriter extends Closeable {
-  def add(rec: DemuxRecord): Unit
+  def add(rec: DemuxRecord): Any
```
nh13, line 504: so that it returns the record output (either `SamRecord` or `FastqRecord`) for testing.  I suppose in the future we can make this a type parameter.

src/test/scala/com/fulcrumgenomics/fastq/DemuxFastqsTest.scala
```diff
@@ -516,70 +516,111 @@ class DemuxFastqsTest extends UnitSpec with OptionValues with ErrorLogLevel {
...
+        observedAssignments should contain theSameElementsInOrderAs assignments
+      }
+      else {
+        observedAssignments should contain theSameElementsInOrderAs barcodes
+      }
+   }
+  }
 
-        headersR1 should contain theSameElementsInOrderAs headersR2
+  it should "demultiplex with --fastqs-include-read-numbers=false --fastqs-includ-sample-barcodes=false" in {
```
nh13, line 610: Fix with search/replace here and below.
```suggestion
  it should "demultiplex with --fastqs-include-read-numbers=false --fastqs-include-sample-barcodes=false" in {
```

nh13 (approved):
@kstromhaug please do not merge yet, will get @tfenne's review

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -298,19 +319,46 @@ class DemuxFastqs
...
+ @deprecated(message="Use outputStandards instead", since="1.3.0")
+ @arg(doc="Output FASTQs according to Illumina BaseSpace Sequence Hub naming standards.  This is differfent than Illumina naming standards.",
+   mutex=Array("fastqSkipReadNumbers", "fastqIncludeSampleBarcodes", "illuminaFileNames"))
+ val illuminaStandards: Boolean = false,
+ @arg(doc="Include a trailing /1 or /2 for R1 and R2 in the FASTQ read name.", mutex=Array("illuminaStandards"))
+ var fastqSkipReadNumbers: Boolean = false,
+ @arg(doc="Update the sample barcode in the FASTQ header.", mutex=Array("illuminaStandards"))
+ var fastqIncludeSampleBarcodes: Boolean = false,
+ @arg(doc="Name the output files according to the Illumina file name standards.", mutex=Array("illuminaStandards"))
+ var illuminaFileNames: Boolean = false
```
nh13, line 342: Can be `val`s now that we have a `FastqStandards` container class
```suggestion
 val fastqSkipReadNumbers: Boolean = false,
 @arg(doc="Update the sample barcode in the FASTQ header.", mutex=Array("illuminaStandards"))
 val fastqIncludeSampleBarcodes: Boolean = false,
 @arg(doc="Name the output files according to the Illumina file name standards.", mutex=Array("illuminaStandards"))
 val illuminaFileNames: Boolean = false
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -260,15 +260,36 @@ object DemuxFastqs {
...
       |
-      |The output FASTQs will be written according to Illumina naming standards when the `--illumina-standards` option
-      |is specified.  This requires that the input fastq files contain Illumina format read names with comments containing
-      |`<ReadNum>:<FilterFlag>:<0>:<SampleNumber>` (no angle brackets).  The output create output files with Illumina
-      |names:
+      |The following options affect the output format:
       |
-      |1. The file extension will be `_R1_001.fastq.gz` for read one, and `_R2_001.fastq.gz` for read two (if paired end).
-      |2. The per-sample output prefix will be `<SampleName>_S<SampleOrdinal>_L<LaneNumber>` (without angle brackets).
+      |1. If `--fastq-skip-read-numbers` is specified, then trailing /1 and /2 for R1 and R2 respectively, will not be
```
tfenne, line 267: I think I would rename this put `skip` at the front, e.g. `--skip-fastq-read-numbers` and actually reading that I think `--omit-fastq-read-numbers` would make more sense.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -260,15 +260,36 @@ object DemuxFastqs {
...
+      |appended to e FASTQ read name.  By default they will be appended.
+      |2. If `--fastq-include-sample-barcodes` is specified, then sample barcode will replace the last field in the
```
tfenne, line 269: Similarly, I think it reads better as `--include-sample-barcodes-in-fastq`

tfenne: Might be nice to include an example here too, to make it clearer - e.g. `@foo.bar.splat beep:boop:ACGTGTA`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -260,15 +260,36 @@ object DemuxFastqs {
...
+      |first comment in the FASTQ header.
+      |3. If `--illumina-file-names` is specified, the output files will be named according to the Illumina FASTQ file
+      |naming conventions:
+      |
+      |  a. The file extension will be `_R1_001.fastq.gz` for read one, and `_R2_001.fastq.gz` for read two (if paired end).
+      |  b. The per-sample output prefix will be `<SampleName>_S<SampleOrdinal>_L<LaneNumber>` (without angle brackets).
+      |
+      |Options (1) and (2) require the FASTQ read names to contain the following elements:
```
tfenne, line 277:
```suggestion
      |Options (1) and (2) require the input FASTQ read names to contain the following elements:
```

tfenne: Why do they?  Not outputting fastq read numbers shouldn't depend on the header lines in the fastq file.  Similarly, does the rest of the header comment matter if you're just replacing the last field?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -260,15 +260,36 @@ object DemuxFastqs {
...
+      |[See the Illumina FASTQ conventions for more details.](https://support.illumina.com/help/BaseSpace_OLH_009008/Content/Source/Informatics/BS/FASTQFiles_Intro_swBS.htm)
+      |
+      |The `--illumina-standards` option may not be specified with the three options above.  Use this option if you
+      |intend to upload to Illumina BaseSpace.  This option implies:
+      |
+      |`--fastq-skip-read-numbers=true --fastq-include-sample-barcodes=false --illumina-file-names=true`
+      |
+      |[See the Illumina Basespace standards described here](https://help.basespace.illumina.com/articles/tutorials/upload-data-using-web-uploader/).
+      |
+      |To output with Illumina conventions that match `bcl2fastq` and `BCLconvert`, use:
```
tfenne, line 290:
```suggestion
      |To output with recent Illumina conventions (circa 2021) that match `bcl2fastq` and `BCLconvert`, use:
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
tfenne, line 322:
```suggestion
 @arg(doc="Platform to insert into the read group header of BAMs (e.g Illumina)") val platform: String = "Illumina",
```

Also aren't the platforms supposed to be all upper case?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -298,19 +319,46 @@ class DemuxFastqs
...
         and read two) as defined by the corresponding read structure(s).
      """)
- val includeAllBasesInFastqs: Boolean = false
+ val includeAllBasesInFastqs: Boolean = false,
+ @deprecated(message="Use outputStandards instead", since="1.3.0")
+ @arg(doc="Output FASTQs according to Illumina BaseSpace Sequence Hub naming standards.  This is differfent than Illumina naming standards.",
+   mutex=Array("fastqSkipReadNumbers", "fastqIncludeSampleBarcodes", "illuminaFileNames"))
+ val illuminaStandards: Boolean = false,
+ @arg(doc="Include a trailing /1 or /2 for R1 and R2 in the FASTQ read name.", mutex=Array("illuminaStandards"))
+ var fastqSkipReadNumbers: Boolean = false,
```
tfenne, line 338:
```suggestion
 @arg(doc="Do not include trailing /1 or /2 for R1 and R2 in the FASTQ read name.", mutex=Array("illuminaStandards"))
 val fastqSkipReadNumbers: Boolean = false,
```

tfenne: Why `var`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -298,19 +319,46 @@ class DemuxFastqs
...
+ @arg(doc="Update the sample barcode in the FASTQ header.", mutex=Array("illuminaStandards"))
+ var fastqIncludeSampleBarcodes: Boolean = false,
```
tfenne, line 340:
```suggestion
 @arg(doc="Insert the sample barcode into the FASTQ header.", mutex=Array("illuminaStandards"))
 var fastqIncludeSampleBarcodes: Boolean = false,
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
tfenne: No reason not to do it now right?  Would take 1 minute to do.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -466,13 +510,13 @@ private class SamRecordWriter(prefix: PathPrefix,
                               val umiTag: String,
                               val numSamples: Int) extends DemuxWriter {
   val order: Option[SamOrder] = if (header.getSortOrder == SortOrder.unsorted) None else SamOrder(header)
-  private val writer = SamWriter(PathUtil.pathTo(s"${prefix}.bam"), header, sort=order,
+  private val writer = SamWriter(PathUtil.pathTo(s"$prefix.bam"), header, sort=order,
```
tfenne, line 513: Any reason to make this change?  I tend to think it's clearer to keep the braces unless the next character is a space.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -483,61 +527,61 @@ private class SamRecordWriter(prefix: PathPrefix,
...
     }
-    else {
-      Seq(Some(rec.name), rec.sampleBarcode, rec.molecularBarcode).flatten.mkString(":")
+    val comment = rec.readInfo match {
+      case None       => rec.comment
+      case Some(info) =>
+        if (fastqStandards.includeSampleBarcodes && rec.sampleBarcode.nonEmpty) {
+          Some(info.copy(sampleInfo=rec.sampleBarcode.mkString("+")).toString)
+        }
+        else Some(info.toString)
```
tfenne, line 570: I'm not really sure I'm following what's going on here.  A couple of comments to clarify intent would be helpful.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -556,9 +600,10 @@ private[fastq] case class SampleInfo(sample: Sample, isUnmatched: Boolean = fals
 private[fastq] object FastqDemultiplexer {
 
   /** Stores the minimal information for a single template read. */
-  case class DemuxRecord(name: String, bases: String, quals: String, molecularBarcode: Option[String],
-                         sampleBarcode: Option[String], readNumber: Int, pairedEnd: Boolean, comment: Option[String],
-                         originalBases: Option[String] = None, originalQuals: Option[String] = None)
+  case class DemuxRecord(name: String, bases: String, quals: String, molecularBarcode: Seq[String],
+                         sampleBarcode: Seq[String], readNumber: Int, pairedEnd: Boolean, comment: Option[String],
+                         originalBases: Option[String] = None, originalQuals: Option[String] = None,
+                         readInfo: Option[ReadInfo] = None)
```
tfenne, line 606: Ideally case casses should either a) fit all on one line, or b) have a member per line.  E.g. 
```suggestion
  case class DemuxRecord(name: String,
                         bases: String,
                         quals: String,
                         molecularBarcode: Seq[String],
                         sampleBarcode: Seq[String],
                         readNumber: Int,
                         pairedEnd: Boolean,
                         comment: Option[String],
                         originalBases: Option[String] = None,
                         originalQuals: Option[String] = None,
                         readInfo: Option[ReadInfo] = None)
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -729,3 +776,76 @@ object OutputType extends FgBioEnum[OutputType] {
...
+
+  /** Builds the [[ReadInfo]] by parsing a standard input FASTQ. */
+  def apply(name: String, comment: Option[String]): ReadInfo = try {
+    val commentFields = comment.getOrElse {
+      throw new IllegalArgumentException(s"Comment not found.")
+    }.split(' ')
+    val readInfo = commentFields.head
+
+    require(name.count(_ == ':') == 6,
+      s"Expected the read name format 'Instrument:RunID:FlowCellID:Lane:Tile:X:Y', found '${name}'")
```
tfenne, line 815: Why require this if we don't actually rely on it?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -729,3 +776,76 @@ object OutputType extends FgBioEnum[OutputType] {
...
+    require(readInfo.count(_ == ':') == 3,
+      s"Expected the comment format 'ReadNum:FilterFlag:0:SampleNumber', found '$readInfo'")
```
tfenne, line 817: Also, why not do the `split` on 819 and then just check the length of the resulting seq instead of scanning the string twice?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -729,3 +776,76 @@ object OutputType extends FgBioEnum[OutputType] {
...
+  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment)
+
+  /** Builds the [[ReadInfo]] by parsing a [[DemuxRecord]]. */
+  def apply(rec: DemuxRecord): ReadInfo = this(rec.name, rec.comment)
+
+  /** Builds the [[ReadInfo]] by parsing a standard input FASTQ. */
+  def apply(name: String, comment: Option[String]): ReadInfo = try {
+    val commentFields = comment.getOrElse {
+      throw new IllegalArgumentException(s"Comment not found.")
+    }.split(' ')
```
tfenne, line 811: If the comment is required, the type should not be `Option[String]` but just `String`.

tfenne: I also don't really understand what's going on here.  Do people really generate fastqs in the wild with multiple space-separated comments?  If so I think it would make more sense to just call this `comments` since we then break down the first comment into _it's_ constituent fields.

tfenne (changes requested):
I didn't really read through the tests - @nh13 I'm assuming you're on that?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
nh13: Yes, there are space separated comments in the wild.  I like you're suggestion too.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
nh13: 1. We had this check before, so I'd like to keep it in. 
2. Even if we don't explicitly rely on it, it's just a way to check that the FASTQ header is approximately in the format we expect for parsing the read info.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -729,3 +785,73 @@ object OutputType extends FgBioEnum[OutputType] {
...
+case class ReadInfo(readNumber: Int, passQc: Boolean, internalControl: Boolean, sampleInfo: String, rest: Seq[String]) {
+  override def toString: String = {
+    val leading = f"$readNumber:${if (passQc) "Y" else "N"}:${if (internalControl) 1 else 0}:$sampleInfo"
+    if (rest.isEmpty) leading else leading + " " + rest.mkString(" ")
+  }
+}
+
+object ReadInfo {
+  /** Builds the [[ReadInfo]] by parsing a [[FastqRecord]]. */
+  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.get)
```
tfenne, line 811:
```suggestion
  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(s"Can't extra read info when comment not present: ${rec.name}"))
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -729,3 +785,73 @@ object OutputType extends FgBioEnum[OutputType] {
...
+
+  /** Builds the [[ReadInfo]] by parsing a [[DemuxRecord]]. */
+  def apply(rec: DemuxRecord): ReadInfo = this(rec.name, rec.comment.get)
```
tfenne, line 814:
```suggestion
  private def reject(name: String) = throw new IllegalArgumentException(s"Cannot extract ReadInfo due to missing comment: $recname")

  /** Builds the [[ReadInfo]] by parsing a [[FastqRecord]]. */
  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name))

  /** Builds the [[ReadInfo]] by parsing a [[DemuxRecord]]. */
  def apply(rec: DemuxRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name))
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -729,3 +785,73 @@ object OutputType extends FgBioEnum[OutputType] {
...
+    val commentInfo = readInfo.split(':').toSeq
+    require(commentInfo.length == 4,
+      s"Expected the comment format 'ReadNum:FilterFlag:0:SampleNumber', found '$readInfo'")
+    val Seq(readNumber, keep, internalControl, sampleInfo) = commentInfo
+
+    val keepBoolean: Boolean = {
+      if (keep == "Y") true
+      else if (keep == "N") false
+      else throw new IllegalStateException(s"Cannot parse filter/keep flag: $keep")
+    }
```
tfenne, line 830:
```suggestion
    val keepBoolean = keep match {
      case "Y" => true
      case "N" => false
      case   x => throw new IllegalStateException(s"Cannot parse filter/keep flag: $x")
    }
```

tfenne (approved):
A couple of minor suggestions that can be commited and then merge away.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -729,3 +785,75 @@ object OutputType extends FgBioEnum[OutputType] {
...
+}
+
+object ReadInfo {
+  private def reject(name: String) = throw new IllegalArgumentException(s"Cannot extract ReadInfo due to missing comment: $recname")
+
+  /** Builds the [[ReadInfo]] by parsing a [[FastqRecord]]. */
+  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name))
+
+  /** Builds the [[ReadInfo]] by parsing a [[DemuxRecord]]. */
+  def apply(rec: DemuxRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name))
```
nh13, line 816:
```suggestion
  def apply(rec: FastqRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name)))

  /** Builds the [[ReadInfo]] by parsing a [[DemuxRecord]]. */
  def apply(rec: DemuxRecord): ReadInfo = this(rec.name, rec.comment.getOrElse(reject(rec.name)))
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -729,3 +785,75 @@ object OutputType extends FgBioEnum[OutputType] {
...
+  */
+case class ReadInfo(readNumber: Int, passQc: Boolean, internalControl: Boolean, sampleInfo: String, rest: Seq[String]) {
+  override def toString: String = {
+    val leading = f"$readNumber:${if (passQc) "Y" else "N"}:${if (internalControl) 1 else 0}:$sampleInfo"
+    if (rest.isEmpty) leading else leading + " " + rest.mkString(" ")
+  }
+}
+
+object ReadInfo {
+  private def reject(name: String) = throw new IllegalArgumentException(s"Cannot extract ReadInfo due to missing comment: $recname")
```
nh13, line 810:
```suggestion
  private def reject(name: String) = throw new IllegalArgumentException(s"Cannot extract ReadInfo due to missing comment: $name")
```

PR #713, opened 2021-09-20
Add an option to filter reads on the header filter flag

kstromhaug:
Added a --keep-only-passing flag to keep only reads annotated with Y in the header comment.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -142,13 +142,14 @@ object DemuxFastqs {
   def demultiplexingIterator(sources: Seq[FastqSource],
                              demultiplexer: FastqDemultiplexer,
                              threads: Int,
-                             batchSize: Int = DemuxBatchRecordsSize): Iterator[DemuxResult] = {
+                             batchSize: Int = DemuxBatchRecordsSize,
+                             omitFailingReads: Boolean): Iterator[DemuxResult] = {
 
     require(demultiplexer.expectedNumberOfReads == sources.length,
       s"The demultiplexer expects ${demultiplexer.expectedNumberOfReads} reads but ${sources.length} FASTQ sources given.")
 
     val zippedIterator = FastqSource.zipped(sources)
-    if (threads > 1) {
+    val resultIterator = if (threads > 1) {
       // Developer Note: Iterator does not support parallel operations, so we need to group together records into a
       // [[List]] or [[Seq]].  A fixed number of records are grouped to reduce memory overhead.
       val pool = new ForkJoinPool(threads)
@@ -164,6 +165,7 @@ object DemuxFastqs {
     else {
       zippedIterator.map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
     }
+    resultIterator.filter(r => r.keep(omitFailingReads))
   }
 }
 
@@ -290,6 +292,8 @@ object DemuxFastqs {
       |To output with recent Illumina conventions (circa 2021) that match `bcl2fastq` and `BCLconvert`, use:
       |
       |`--omit-fastq-read-numbers=true --include-sample-barcodes-in-fastq=true --illumina-file-names=true`
+      |
+      |By default all input reads are output.  If your input FASTQs contain reads that do not pass filter (as defined by the Y/N filter flag in the FASTQ comment) these can be filtered out during demultiplexing using the `--omit-failing-reads` option.
     """,
   group=ClpGroups.Fastq
 )
@@ -339,7 +343,9 @@ class DemuxFastqs
  @arg(doc="Insert the sample barcode into the FASTQ header.", mutex=Array("illuminaStandards"))
  val includeSampleBarcodesInFastq: Boolean = false,
  @arg(doc="Name the output files according to the Illumina file name standards.", mutex=Array("illuminaStandards"))
- val illuminaFileNames: Boolean = false
+ val illuminaFileNames: Boolean = false,
+ @arg(doc="Keep only passing filter reads if true, otherwise keep all reads. Passing filter reads are determined from the comment in the FASTQ header.")
+ val omitFailingReads: Boolean = false
 ) extends FgBioTool with LazyLogging {
 
   // Support the deprecated --illumina-standards option
@@ -427,8 +433,8 @@ class DemuxFastqs
       minMismatchDelta = minMismatchDelta,
       maxNoCalls       = maxNoCalls,
       includeOriginal  = this.includeAllBasesInFastqs,
-      fastqStandards   = this.fastqStandards
-    )
+      fastqStandards   = this.fastqStandards,
+      omitFailingReads = this.omitFailingReads)
 
     val progress = ProgressLogger(this.logger, unit=1e6.toInt)
 
@@ -438,7 +444,8 @@ class DemuxFastqs
     val iterator  = demultiplexingIterator(
       sources       = sources,
       demultiplexer = demultiplexer,
-      threads       = threads
+      threads       = threads,
+      omitFailingReads = this.omitFailingReads
     )
 
     // Write the records out in its own thread
@@ -510,7 +517,7 @@ private class SamRecordWriter(prefix: PathPrefix,
                               val umiTag: String,
                               val numSamples: Int) extends DemuxWriter[SamRecord] {
   val order: Option[SamOrder] = if (header.getSortOrder == SortOrder.unsorted) None else SamOrder(header)
-  private val writer = SamWriter(PathUtil.pathTo(s"${prefix}.bam"), header, sort=order,
+  private val writer = SamWriter(PathUtil.pathTo(s"$prefix.bam"), header, sort=order,
     async = DemuxFastqs.UseAsyncIo,
     maxRecordsInRam = Math.max(10000,  DemuxFastqs.MaxRecordsInRam / numSamples))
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -148,7 +148,8 @@ object DemuxFastqs {
       s"The demultiplexer expects ${demultiplexer.expectedNumberOfReads} reads but ${sources.length} FASTQ sources given.")
 
     val zippedIterator = FastqSource.zipped(sources)
-    if (threads > 1) {
+
```
nh13, line 151: bingo!
```suggestion
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -290,6 +292,10 @@ object DemuxFastqs {
       |To output with Illumina conventions that match `bcl2fastq` and `BCLconvert`, use:
       |
       |`--fastq-skip-read-numbers=true --fastq-include-sample-barcodes=true --illumina-file-names=true`
+      |
+      |To output reads with a passing filter flag, use:
+      |
+      |`--keep-only-passing=true`
```
nh13, line 298: needs more description of what a "passing filter flag" is and where it comes from (i.e. the `<is filtered>` field in the FASTQ comment above).

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -331,15 +337,17 @@ class DemuxFastqs
...
    mutex=Array("fastqSkipReadNumbers", "fastqIncludeSampleBarcodes", "illuminaFileNames"))
  val illuminaStandards: Boolean = false,
  @arg(doc="Include a trailing /1 or /2 for R1 and R2 in the FASTQ read name.", mutex=Array("illuminaStandards"))
  var fastqSkipReadNumbers: Boolean = false,
  @arg(doc="Update the sample barcode in the FASTQ header.", mutex=Array("illuminaStandards"))
  var fastqIncludeSampleBarcodes: Boolean = false,
  @arg(doc="Name the output files according to the Illumina file name standards.", mutex=Array("illuminaStandards"))
- var illuminaFileNames: Boolean = false
+ var illuminaFileNames: Boolean = false,
+ @arg(doc="Keep only passing filter reads if true, otherwise keep all reads. Passing filter reads are determined from the comment in the FASTQ header")
```
nh13, line 349:
```suggestion
 @arg(doc="Keep only passing filter reads if true, otherwise keep all reads. Passing filter reads are determined from the comment in the FASTQ header.")
```

nh13 (changes requested):
Can you add a bit more in the usage about the new option?

Once the #711 PR is merged, we'll rebase and retarget this PR to the `master` branch before merging.  Approvals needed of course.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -762,7 +772,8 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
         )
       }
 
-    DemuxResult(sampleInfo=sampleInfo, numMismatches=numMismatches, records=demuxRecords)
+    val passQc = if (!keepOnlyPassing) true else demuxRecords.forall(d => d.readInfo.forall(_.passQc))
+    DemuxResult(sampleInfo=sampleInfo, numMismatches=numMismatches, records=demuxRecords, passQc=passQc)
```
nh13, line 776: @kstromhaug @tfenne part of me wonders if we should check `keepOnlyPassing` here when setting `passQc` on the `DemuxResult`, but instead check it when we filter the `resultIterator`.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -762,7 +772,8 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
         )
       }
 
-    DemuxResult(sampleInfo=sampleInfo, numMismatches=numMismatches, records=demuxRecords)
+    val passQc = if (!keepOnlyPassing) true else demuxRecords.forall(d => d.readInfo.forall(_.passQc))
```
nh13, line 775: Then update `resultsIterator.filter`
```suggestion
    val passQc = demuxRecords.forall(d => d.readInfo.forall(_.passQc))
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -164,6 +165,7 @@ object DemuxFastqs {
     else {
       zippedIterator.map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
     }
+    resultIterator.filter(r => r.passQc)
```
nh13, line 168:
```suggestion
    resultIterator.filter(r => r.keep(keepOnlyPassing))
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -611,7 +622,14 @@ private[fastq] object FastqDemultiplexer {
     *                      for the unmatched sample.
     * @param records the records, one for each read that has template bases.
     */
-  case class DemuxResult(sampleInfo: SampleInfo, numMismatches: Int, records: Seq[DemuxRecord])
+  case class DemuxResult(sampleInfo: SampleInfo, numMismatches: Int, records: Seq[DemuxRecord], passQc: Boolean = true) {
+
+    def keep(keepOnlyPassing: Boolean = false): Boolean = {
```
nh13, line 627: Add a doc-string :)
```suggestion
    /** Returns true if this [[DemuxResult]] should be kept for output, false otherwise.
     *
     * Returns true if `keepOnlyPassing` is false or if `passQc` is true
     *
     * @param keepOnlyPassing true to keep only passing reads, false to keep all reads
    */
    def keep(keepOnlyPassing: Boolean = false): Boolean = {
```

nh13 (approved):
One minor docstring that I'd like you to add, but LGTM.  Please wait for @tfenne's review, and #711 to merge into main before proceeding.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -331,15 +337,17 @@ class DemuxFastqs
...
  val illuminaStandards: Boolean = false,
  @arg(doc="Include a trailing /1 or /2 for R1 and R2 in the FASTQ read name.", mutex=Array("illuminaStandards"))
  var fastqSkipReadNumbers: Boolean = false,
  @arg(doc="Update the sample barcode in the FASTQ header.", mutex=Array("illuminaStandards"))
  var fastqIncludeSampleBarcodes: Boolean = false,
  @arg(doc="Name the output files according to the Illumina file name standards.", mutex=Array("illuminaStandards"))
- var illuminaFileNames: Boolean = false
+ var illuminaFileNames: Boolean = false,
+ @arg(doc="Keep only passing filter reads if true, otherwise keep all reads. Passing filter reads are determined from the comment in the FASTQ header.")
+ val keepOnlyPassing: Boolean = false
```
nh13, line 350: @kstromhaug @tfenne  do you think it's clearer to have this option named `--omit-failing-reads` with default `false` (outputs all reads), and when `true` only emits passing reads determined from the FASTQ header/comment?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -290,6 +292,10 @@ object DemuxFastqs {
       |To output with Illumina conventions that match `bcl2fastq` and `BCLconvert`, use:
       |
       |`--fastq-skip-read-numbers=true --fastq-include-sample-barcodes=true --illumina-file-names=true`
+      |
+      |To output reads with a passing filter flag, encoded as 0 (not control) or >0 (an internal control) in the <is filter> flag of the header comment, use:
+      |
+      |`--keep-only-passing=true`
```
tfenne, line 298: This sounds to me like it's mixing two things - the `passing filter` flag and the `is control` flag/code.  How about

```suggestion
      |By default all input reads are output.  If your input FASTQs contain reads that do not pass filter (as defined by the Y/N filter flag in the FASTQ comment) these can be filtered out during demultiplexing using the `--skip-failing-reads` option.
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
tfenne: I would definitely rename this to be something like `--skip-failing-reads` or `--omit-failing-reads` as I think that is clearer.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -762,7 +787,8 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
         )
       }
 
-    DemuxResult(sampleInfo=sampleInfo, numMismatches=numMismatches, records=demuxRecords)
+    val passQc = demuxRecords.forall(d => d.readInfo.forall(_.passQc))
+    DemuxResult(sampleInfo=sampleInfo, numMismatches=numMismatches, records=demuxRecords, passQc=passQc)
```
tfenne, line 791: Wouldn't it be a lot more efficient to do this up front?  What's the point of demuxing reads that we're just going to throw on the floor?  I would suggest doing this wherever `demultiplex()` is called from and then just skipping over the input fastq records.

tfenne (changes requested):
A couple of minor things, but one more major comment/suggestion about moving the logic to skip failing reads up out of `demultiplex()` so that we don't spend a lot of effort demuxing reads we don't output.

I guess the other question is if you say to only do passing reads, do you want stats for non-passing reads?  I would think not?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
kstromhaug: Leaving as is for now

tfenne (approved):

PR #714, opened 2021-09-20
Add option to DemuxFastqs to filter on the internal control flag in the header comment

kstromhaug:


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -293,9 +293,13 @@ object DemuxFastqs {
       |
       |`--fastq-skip-read-numbers=true --fastq-include-sample-barcodes=true --illumina-file-names=true`
       |
-      |To output reads with a passing filter flag, use:
+      |To output reads with a passing filter flag, as encoded in the <is filtered> field in the header comment, use:
```
nh13, line 296: Adding backticks, since the usage is parsed as markdown, which is really nice.
```suggestion
      |To output reads with a passing filter flag, as encoded in the `<is filtered>` field in the header comment, use:
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -293,9 +293,13 @@ object DemuxFastqs {
...
       |
       |`--keep-only-passing=true`
+      |
+      |To output only reads that are not control reads, as encoded in the <control number> field in the header comment, use:
```
nh13, line 300:
```suggestion
      |To output only reads that are not control reads, as encoded in the `<control number>` field in the header comment, use:
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
nh13: Also, I think you're mixing up PRs, as I think I asked for this in the other PR: https://github.com/fulcrumgenomics/fgbio/pull/713#discussion_r712400396

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -347,7 +351,9 @@ class DemuxFastqs
  @arg(doc="Name the output files according to the Illumina file name standards.", mutex=Array("illuminaStandards"))
  var illuminaFileNames: Boolean = false,
  @arg(doc="Keep only passing filter reads if true, otherwise keep all reads. Passing filter reads are determined from the comment in the FASTQ header")
- val keepOnlyPassing: Boolean = false
+ val keepOnlyPassing: Boolean = false,
+ @arg(doc="Keep reads marked as control if true, otherwise removes them. Control reads are determined from the comment in the FASTQ header")
+ val keepControlReads: Boolean = true
```
nh13, line 356: 1. I'd default this to `false`, to keep the current default behavior of the tool, which is to output all reads. 
2. I word-smithed this to be the same as `--keep-only-passing` for consistency.
```suggestion
 @arg(doc="Keep only reads identified as control if true, otherwise keep all reads. Control reads are determined from the comment in the FASTQ header")
 val keepControlReads: Boolean = false
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -773,7 +785,8 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
       }
 
     val passQc = if (!keepOnlyPassing) true else demuxRecords.forall(d => d.readInfo.forall(_.passQc))
-    DemuxResult(sampleInfo=sampleInfo, numMismatches=numMismatches, records=demuxRecords, passQc=passQc)
+    val isControl = if (keepControlReads) false else demuxRecords.forall(d => d.readInfo.forall(_.internalControl))
```
nh13, line 788: ditto to this comment: https://github.com/fulcrumgenomics/fgbio/pull/713/files#r712552311.  I think we want to set the value regardless of `keepControlReads`, then filter later (in the `resultsIterator.filter`)
```suggestion
    val isControl = demuxRecords.forall(d => d.readInfo.forall(_.internalControl))
```

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -448,7 +457,8 @@ class DemuxFastqs
       sources       = sources,
       demultiplexer = demultiplexer,
       threads       = threads,
-      keepOnlyPassing = this.keepOnlyPassing
+      keepOnlyPassing = this.keepOnlyPassing,
+      keepControlReads = this.keepControlReads
```
nh13, line 461: please align the `=`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -621,22 +631,24 @@ private[fastq] object FastqDemultiplexer {
...
 
     /** Returns true if this [[DemuxResult]] should be kept for output, false otherwise.
      *
      * Returns true if `keepOnlyPassing` is false or if `passQc` is true
      *
      * @param keepOnlyPassing true to keep only passing reads, false to keep all reads
     */
-    def keep(keepOnlyPassing: Boolean = false): Boolean = {
-      if (!keepOnlyPassing) true else passQc
+    def keep(keepOnlyPassing: Boolean = false, keepControlReads: Boolean = false): Boolean = {
```
nh13, line 645: Please add unit tests that test this method directly.

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -346,8 +351,11 @@ class DemuxFastqs
  var fastqIncludeSampleBarcodes: Boolean = false,
  @arg(doc="Name the output files according to the Illumina file name standards.", mutex=Array("illuminaStandards"))
  var illuminaFileNames: Boolean = false,
- @arg(doc="Keep only passing filter reads if true, otherwise keep all reads. Passing filter reads are determined from the comment in the FASTQ header.")
- val keepOnlyPassing: Boolean = false
+
+ @arg(doc="Keep only passing filter reads if true, otherwise keep all reads. Passing filter reads are determined from the comment in the FASTQ header")
+ val keepOnlyPassing: Boolean = false,
+ @arg(doc="Keep only reads identified as control if true, otherwise keep all reads. Control reads are determined from the comment in the FASTQ header")
+ val keepControlReads: Boolean = false
```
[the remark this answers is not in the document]
tfenne: What's the expected default here?  Since we didn't have this option previously do you want to have the default be to retain control reads so by default we output everything that's input?
Either way, I think boolean options are best when they default to false, because then the user can just use `--thing` instead of `--thing=false`.  So here I think I would suggest we have:
`val [omit|skip|discard]ControlReads: Boolean = false`

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -621,22 +631,24 @@ private[fastq] object FastqDemultiplexer {
...
      * Returns true if `keepOnlyPassing` is false or if `passQc` is true
      *
      * @param keepOnlyPassing true to keep only passing reads, false to keep all reads
     */
-    def keep(keepOnlyPassing: Boolean = false): Boolean = {
-      if (!keepOnlyPassing) true else passQc
+    def keep(keepOnlyPassing: Boolean = false, keepControlReads: Boolean = false): Boolean = {
+      val keepByQc = if (!keepOnlyPassing) true else passQc
+      val keepByControl = if (keepControlReads) true else !isControl
+      keepByQc & keepByControl
```
tfenne, line 648: For booleans this isn't terrible, but `&` is bitwise and, and this should be using `&&` for logical and.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -787,8 +802,10 @@ private class FastqDemultiplexer(val sampleInfos: Seq[SampleInfo],
         )
       }
 
+
     val passQc = demuxRecords.forall(d => d.readInfo.forall(_.passQc))
-    DemuxResult(sampleInfo=sampleInfo, numMismatches=numMismatches, records=demuxRecords, passQc=passQc)
+    val isControl = demuxRecords.forall(d => d.readInfo.forall(_.internalControl))
+    DemuxResult(sampleInfo=sampleInfo, numMismatches=numMismatches, records=demuxRecords, passQc=passQc, isControl=isControl)
```
tfenne, line 808: Same point here as in #713 - I think this should get moved up so as not to do all the demuxing work when we don't need to.

tfenne (changes requested):

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -345,7 +349,9 @@ class DemuxFastqs
  @arg(doc="Name the output files according to the Illumina file name standards.", mutex=Array("illuminaStandards"))
  var illuminaFileNames: Boolean = false,
  @arg(doc="Keep only passing filter reads if true, otherwise keep all reads. Passing filter reads are determined from the comment in the FASTQ header.")
- val skipFailingReads: Boolean = false
+ val skipFailingReads: Boolean = false,
+ @arg(doc="Keep only reads identified as control if true, otherwise keep all reads. Control reads are determined from the comment in the FASTQ header.")
```
nh13, line 353:
```suggestion
 @arg(doc="Do not keep reads identified as control if true, otherwise keep all reads. Control reads are determined from the comment in the FASTQ header.")
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -434,7 +440,8 @@ class DemuxFastqs
       maxNoCalls       = maxNoCalls,
       includeOriginal  = this.includeAllBasesInFastqs,
       fastqStandards   = this.fastqStandards,
-      skipFailingReads  = this.skipFailingReads
+      skipFailingReads  = this.skipFailingReads,
+      omitControlReads = this.omitControlReads
```
nh13, line 444: align equals

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -443,10 +450,11 @@ class DemuxFastqs
...
-      sources       = sources,
-      demultiplexer = demultiplexer,
-      threads       = threads,
-      skipFailingReads = this.skipFailingReads
+      sources          = sources,
+      demultiplexer    = demultiplexer,
+      threads          = threads,
+      skipFailingReads  = this.skipFailingReads,
+      omitControlReads = this.omitControlReads
     )
```
nh13, line 458:
```suggestion
      sources           = sources,
      demultiplexer     = demultiplexer,
      threads           = threads,
      skipFailingReads  = this.skipFailingReads,
      omitControlReads  = this.omitControlReads
    )
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -619,22 +627,25 @@ private[fastq] object FastqDemultiplexer {
...
      * Returns true if `skipFailingReads` is false or if `passQc` is true
      *
      * @param skipFailingReads true to keep only passing reads, false to keep all reads
     */
-    def keep(skipFailingReads: Boolean = false): Boolean = {
-      if (!skipFailingReads) true else passQc
+
+    def keep(skipFailingReads: Boolean, omitControlReads: Boolean): Boolean = {
+      val keepByQc = if (!skipFailingReads) true else passQc
+      val keepByControl = if (!omitControlReads) true else !isControl
```
nh13, line 644:
```suggestion
      val keepByQc      = if (!skipFailingReads) true else passQc
      val keepByControl = if (!omitControlReads) true else !isControl
```

nh13 (changes requested):
Minor changes left!

nh13 (approved):
Nice work! LGTM

tfenne (approved):
My only concern is that I feel like we should pick one of `skip` or `omit` and use it in both options rather than mix and match.

PR #716, opened 2021-09-23
Add an option to mask bases given a quality threshold in DemuxFastqs.

kstromhaug:


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -308,7 +317,8 @@ class DemuxFastqs
         and template bases) for every read with template bases (ex. read one
         and read two) as defined by the corresponding read structure(s).
      """)
- val includeAllBasesInFastqs: Boolean = false
+ val includeAllBasesInFastqs: Boolean = false,
+ @arg(doc="Mask bases with a quality score below a specified threshold") val qualityThreshold: Int = 0,
```
tfenne, line 321:
```suggestion
 @arg(doc="Mask bases with a quality score below the specified threshold to N") val qualityThreshold: Int = 0,
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -375,6 +385,10 @@ class DemuxFastqs
       }
     }
 
+    // validate the quality threshold provided
+    require(qualityThreshold >= 0 && qualityThreshold <= 40,
+      s"The quality threshold ($qualityThreshold) is not within the expected range (0, 40)")
```
tfenne, line 390: Why 40?  I would argue there's no need to restrict this on the upper end, or even at all - what harm does it do if someone specifies `-10` or `60`?

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -556,17 +572,40 @@ private[fastq] case class SampleInfo(sample: Sample, isUnmatched: Boolean = fals
 private[fastq] object FastqDemultiplexer {
 
   /** Stores the minimal information for a single template read. */
-  case class DemuxRecord(name: String, bases: String, quals: String, molecularBarcode: Option[String],
+  final case class DemuxRecord(name: String, bases: String, quals: String, molecularBarcode: Option[String],
```
tfenne, line 575: I think you can pull this `final` back out.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -556,17 +572,40 @@ private[fastq] case class SampleInfo(sample: Sample, isUnmatched: Boolean = fals
...
     */
-  case class DemuxResult(sampleInfo: SampleInfo, numMismatches: Int, records: Seq[DemuxRecord])
+  final case class DemuxResult(sampleInfo: SampleInfo, numMismatches: Int, records: Seq[DemuxRecord]) {
+
+    /** A function to mask bases that have a quality score less than or equal to a specified threshold
+      * @param threshold The threshold for masking bases, inclusive. Bases with a quality score <= threshold will be masked
+      * @return a new DemuxResult with updated bases
+      */
+    def maskLowQualityBases(threshold: Int, qualityEncoding: QualityEncoding): DemuxResult = { // using this.type here causes a mismatch error
+      if (threshold == 0) this
```
tfenne, line 605:
```suggestion
      if (threshold <= 0) this
```
Just to be defensive.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -556,17 +572,40 @@ private[fastq] case class SampleInfo(sample: Sample, isUnmatched: Boolean = fals
...
+    /** A function to mask bases that have a quality score less than or equal to a specified threshold
+      * @param threshold The threshold for masking bases, inclusive. Bases with a quality score <= threshold will be masked
+      * @param qualityEncoding The encoding used for quality scores in the Fastq file.
+      * @return a new DemuxRecord with updated bases
+      */
+    def maskLowQualityBases(threshold: Int, qualityEncoding: QualityEncoding): DemuxRecord = {
+      val bases = this.bases.zip(quals).map { case (base, qual) =>
+        if (qualityEncoding.toStandardNumeric(ch = qual) <= threshold) "N" else base
+      }
+    this.copy(bases = bases.mkString)
```
tfenne, line 588: It would be more efficient go translate the other way and ask "what character is `threshold` in the given encoding, then compare to that.

tfenne: I think too if the assumption is that the majority of reads would not have bases below threshold it would be faster to first check, e.g. 

```scala
val thresh = qualityEncoding.toAlpha(threshold)
if (this.quals.exists(_ < = thresh)) {
  // do the masking and return copy(bases=newBases)
}
else {
  this
}
```

tfenne: Also, when masking it would likely be much more efficient to e.g.:

```scala
val bases = this.bases.getBytes
forloop (from=0, until=quals.length) { i => 
  if (quals.charAt(i) <= thresh) bases(i) = 'N'
}
new String(bases)
```

src/test/scala/com/fulcrumgenomics/fastq/DemuxFastqsTest.scala
```diff
@@ -320,6 +320,97 @@ class DemuxFastqsTest extends UnitSpec with OptionValues with ErrorLogLevel {
...
   }
 
+  def makeDemuxRecord(bases: String, quals: String): DemuxResult = {
+    val structures = Seq(ReadStructure("5B40T"))
+    val sampleInfos = toSampleInfos(structures)
+    val demuxRecord = DemuxRecord(name = "name", bases = bases, quals = quals, molecularBarcode = Option("MB"), sampleBarcode = Option("SB"), readNumber = 1, pairedEnd = false, comment = None)
+    DemuxResult(sampleInfo = sampleInfos(0), numMismatches = 0, records = Seq(demuxRecord))
+  }
+
+  "DemuxResult.countMismatches" should "mask bases that are less than or equal to the quality threshold" in {
```
tfenne, line 330:
```suggestion
  "DemuxResult.maskLowQualityBases" should "mask bases that are less than or equal to the quality threshold" in {
```

src/test/scala/com/fulcrumgenomics/fastq/DemuxFastqsTest.scala
```diff
@@ -320,6 +320,97 @@ class DemuxFastqsTest extends UnitSpec with OptionValues with ErrorLogLevel {
...
+}
+
+  "DemuxFastqs" should "run end to end and return the same bases when the threshold is 0 for 1 thread" in {
+    val records = testEndToEndWithQualityThreshold(threshold = 0, threads = 1)
+    records.length shouldBe 1
+    records(0).bases.length shouldBe 39
+    records(0).bases shouldEqual "A"*39
+  }
+
+  "DemuxFastqs" should "run end to end and replace any bases below a specified quality threshold for 1 thread" in {
```
tfenne, line 390: For the first test you put the label (e.g. "DemuxFastqs") but for subsequent ones you want to use just `it`

tfenne (changes requested):
A bunch of feedback on how to make it significantly more efficient.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -269,6 +273,11 @@ object DemuxFastqs {
       |
       |1. The file extension will be `_R1_001.fastq.gz` for read one, and `_R2_001.fastq.gz` for read two (if paired end).
       |2. The per-sample output prefix will be `<SampleName>_S<SampleOrdinal>_L<LaneNumber>` (without angle brackets).
+      |
+      |To mask bases that are below a specified quality threshold, use the below command, where X is the desired threshold.
+      |Defaults to 0 to keep all bases
+      | --quality-threshold X
```
nh13, line 279: 1. Remember that the usage is in markdown, so you can use back-ticks.  I think the is much clearer.
```suggestion
      |Use the `--quality-threshold` option to mask bases that are below a specified quality threshold.
```
2. More importantly, I don't think you need to describe this option to the usage at all.  And if so, it doesn't belong in the Illumina FASTQ File Naming section.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
nh13, line 321:
```suggestion
 @arg(doc="Mask bases as Ns with a quality score below the given threshold") val qualityThreshold: Int = 0,
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
nh13: 1. Agree with @tfenne .  
2. `qualityEncoding.toAlpha(threshold)` could be computed once and stored somewhere so we don't have to do it each time.  Perhaps in the main class and pass the ascii value.
3.  This should just return `this` if the quality threshold is not defined or zero or below.  That would be much faster when using the default value.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
nh13: Ah, here's where we check if `threshold` is leq zero.  Not a bad idea to also add the check in `DemuxRecord`

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -142,12 +142,17 @@ object DemuxFastqs {
...
+                             threshold: Int,
+                             qualityEncoding: QualityEncoding,
                              batchSize: Int = DemuxBatchRecordsSize): Iterator[DemuxResult] = {
 
     require(demultiplexer.expectedNumberOfReads == sources.length,
       s"The demultiplexer expects ${demultiplexer.expectedNumberOfReads} reads but ${sources.length} FASTQ sources given.")
 
     val zippedIterator = FastqSource.zipped(sources)
+
+    val thresh = qualityEncoding.toStandardAscii((threshold + qualityEncoding.asciiOffset).toChar)
```
nh13, line 154: I like that you haven't pulled this out all the way to the `demuxFastqs` class, as this method is only called once anyhow.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -158,11 +163,13 @@ object DemuxFastqs {
           batch
             .parWith(pool=pool)
             .map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
+            .map { result => result.maskLowQualityBases(threshold=thresh, qualityEncoding=qualityEncoding) }
```
nh13, line 166: what do you think the benefits of this are?
```suggestion
            .map { readRecords => 
                demultiplexer.demultiplex(readRecords: _*)
                    .maskLowQualityBases(threshold=thresh, qualityEncoding=qualityEncoding) 
            }
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -158,11 +163,13 @@ object DemuxFastqs {
...
             .seq
         }
     }
     else {
       zippedIterator.map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
+        .map { result => result.maskLowQualityBases(threshold=thresh, qualityEncoding=qualityEncoding)}
```
nh13, line 172: Fine to keep them as two maps since there's no parallelism.  Also just a bit of a reformat to see the two `map`s
```suggestion
      zippedIterator
        .map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
        .map { result => result.maskLowQualityBases(threshold=thresh, qualityEncoding=qualityEncoding)}
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -269,6 +276,7 @@ object DemuxFastqs {
       |
       |1. The file extension will be `_R1_001.fastq.gz` for read one, and `_R2_001.fastq.gz` for read two (if paired end).
       |2. The per-sample output prefix will be `<SampleName>_S<SampleOrdinal>_L<LaneNumber>` (without angle brackets).
+      |
```
nh13, line 279:
```suggestion
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -558,15 +569,40 @@ private[fastq] object FastqDemultiplexer {
   /** Stores the minimal information for a single template read. */
   case class DemuxRecord(name: String, bases: String, quals: String, molecularBarcode: Option[String],
                          sampleBarcode: Option[String], readNumber: Int, pairedEnd: Boolean, comment: Option[String],
-                         originalBases: Option[String] = None, originalQuals: Option[String] = None)
+                         originalBases: Option[String] = None, originalQuals: Option[String] = None) {
+
+    /** A function to mask bases that have a quality score less than or equal to a specified threshold
```
nh13, line 574: No need to say "A function"
```suggestion
    /** Masks bases that have a quality score less than or equal to a specified threshold.
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -558,15 +569,40 @@ private[fastq] object FastqDemultiplexer {
...
+      * @param qualityEncoding The encoding used for quality scores in the Fastq file.
+      * @return a new DemuxRecord with updated bases
+      */
+    def maskLowQualityBases(threshold: Char, qualityEncoding: QualityEncoding): DemuxRecord = {
+      if (this.quals.exists(_ <= threshold)) {
+        val bases = this.bases.toCharArray
+        for (i <- Range(0, this.quals.length)) if (quals.charAt(i) <= threshold) bases(i) = 'N'
+        this.copy(bases = bases.mkString)
+      }
+      else this
```
nh13, line 585: 1.  The usage for DemuxFastqs says "Mask bases with a quality score below the specified threshold as Ns".  But this implements `<=`.  Please update the function usages please as I prefer `<` versus `<=`.
2. Can you add a short-circuit so that if `threshold` is zero or below, you just return `this`
3. The suggestion below incorporates (1) & (2), and I prefer to have the shorter conditional branch first, as then it's much easer to scan.
```suggestion
      if (threshold <= 0 || this.quals.forall(_ > threshold) this else {
        val bases = this.bases.toCharArray
        for (i <- Range(0, this.quals.length)) if (quals.charAt(i) <= threshold) bases(i) = 'N'
        this.copy(bases = bases.mkString)
      }
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -558,15 +569,40 @@ private[fastq] object FastqDemultiplexer {
...
     * @param records the records, one for each read that has template bases.
     */
-  case class DemuxResult(sampleInfo: SampleInfo, numMismatches: Int, records: Seq[DemuxRecord])
+  case class DemuxResult(sampleInfo: SampleInfo, numMismatches: Int, records: Seq[DemuxRecord]) {
+
+    /** A function to mask bases that have a quality score less than or equal to a specified threshold
+      * @param threshold The threshold for masking bases, inclusive. Bases with a quality score <= threshold will be masked
+      * @return a new DemuxResult with updated bases
+      */
+    def maskLowQualityBases(threshold: Char, qualityEncoding: QualityEncoding): DemuxResult = { // using this.type here causes a mismatch error
```
nh13, line 601: I think we have a type alias you can use wherever you have `Char` for a PhredScore.  Can you fix here and everywhere else?
```suggestion
    def maskLowQualityBases(threshold: PhredScore, qualityEncoding: QualityEncoding): DemuxResult = { // using this.type here causes a mismatch error
```

nh13 (changes requested):
A few more things

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -142,12 +142,17 @@ object DemuxFastqs {
...
+                             qualityEncoding: QualityEncoding,
                              batchSize: Int = DemuxBatchRecordsSize): Iterator[DemuxResult] = {
 
     require(demultiplexer.expectedNumberOfReads == sources.length,
       s"The demultiplexer expects ${demultiplexer.expectedNumberOfReads} reads but ${sources.length} FASTQ sources given.")
 
     val zippedIterator = FastqSource.zipped(sources)
+
+    val thresh = qualityEncoding.toStandardAscii(PhredScore.cap(threshold + qualityEncoding.asciiOffset).toChar).toByte
+
```
nh13, line 155:
```suggestion
    val zippedIterator = FastqSource.zipped(sources)
    val thresh         = qualityEncoding.toStandardAscii(PhredScore.cap(threshold + qualityEncoding.asciiOffset).toChar).toByte

```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -157,12 +162,17 @@ object DemuxFastqs {
         .flatMap { batch =>
           batch
             .parWith(pool=pool)
-            .map { readRecords => demultiplexer.demultiplex(readRecords: _*) }
+            .map { readRecords => 
+                demultiplexer.demultiplex(readRecords: _*)
+                    .maskLowQualityBases(threshold=thresh, qualityEncoding=qualityEncoding)
```
nh13, line 167: why is this better than having two `maps`?  I think it is

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -558,15 +571,39 @@ private[fastq] object FastqDemultiplexer {
...
-                         originalBases: Option[String] = None, originalQuals: Option[String] = None)
+                         originalBases: Option[String] = None, originalQuals: Option[String] = None) {
+
+    /** Masks bases that have a quality score less than or equal to a specified threshold.
+      * @param threshold The threshold for masking bases, exlusive. Bases with a quality score < threshold will be masked
+      * @param qualityEncoding The encoding used for quality scores in the Fastq file.
+      * @return a new DemuxRecord with updated bases
+      */
+    def maskLowQualityBases(threshold: PhredScore, qualityEncoding: QualityEncoding): DemuxRecord = {
+      if (threshold <= 0 || this.quals.forall(_ > threshold)) this else {
```
nh13, line 582: bug in my previous suggestion
```suggestion
      if (threshold <= 0 || this.quals.forall(_ >= threshold)) this else {
```

nh13 (changes requested):
Minor comments, so close!

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
kstromhaug: To avoid iterating through all the records again, and instead doing both steps at once?

nh13: @kstromhaug I believe this will create two "parallel" collections.  The order in which the `map` are applied is non-deterministic, so we could end up having `demultiplex` get called before any `maskLowQualityBases` gets applied, so it could cause a lot more memory usage.

nh13 (approved):
LGTM, good job!

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -142,12 +142,16 @@ object DemuxFastqs {
...
                              threads: Int,
+                             threshold: Int,
+                             qualityEncoding: QualityEncoding,
                              batchSize: Int = DemuxBatchRecordsSize): Iterator[DemuxResult] = {
 
     require(demultiplexer.expectedNumberOfReads == sources.length,
       s"The demultiplexer expects ${demultiplexer.expectedNumberOfReads} reads but ${sources.length} FASTQ sources given.")
 
     val zippedIterator = FastqSource.zipped(sources)
+    val thresh         = qualityEncoding.toStandardAscii(PhredScore.cap(threshold + qualityEncoding.asciiOffset).toChar).toByte
```
tfenne, line 153: I think there are several things wrong with this:
1. `PhredScore.cap()` is intended to be applied to a numeric phred score, not a phred score plus some offet.  I don't think we need this call at all.
2. qualityEncoding.toStandardAscii returns the byte/char _as encoded by QualityEncoding.Standard_, which will do the wrong thing if another quality encoding is in use.

I think all this really needs to be is `(threshold + qualityEncoding.asciiOffset).to[Byte|Char].  Sorry if I led you astray, it doesn't look like there's a method on QualityEncoding to do this.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -558,15 +570,39 @@ private[fastq] object FastqDemultiplexer {
...
                          sampleBarcode: Option[String], readNumber: Int, pairedEnd: Boolean, comment: Option[String],
-                         originalBases: Option[String] = None, originalQuals: Option[String] = None)
+                         originalBases: Option[String] = None, originalQuals: Option[String] = None) {
+
+    /** Masks bases that have a quality score less than or equal to a specified threshold.
+      * @param threshold The threshold for masking bases, exlusive. Bases with a quality score < threshold will be masked
+      * @param qualityEncoding The encoding used for quality scores in the Fastq file.
+      * @return a new DemuxRecord with updated bases
+      */
+    def maskLowQualityBases(threshold: PhredScore, qualityEncoding: QualityEncoding): DemuxRecord = {
```
tfenne, line 580: Changing the type to `PhredScore` here is also confusing.  The value is _not_  simple PhredScore, but is a PhredScore plus an offset.  I think this should be `threshold: Byte` and the documentation needs to be very very clear that the value is comparable to the characters directly, not the numeric value of the scores.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -558,15 +570,39 @@ private[fastq] object FastqDemultiplexer {
...
+      if (threshold <= 0 || this.quals.forall(_ >= threshold)) this else {
+        val bases = this.bases.toCharArray
+        for (i <- Range(0, this.quals.length)) if (quals.charAt(i) < threshold) bases(i) = 'N'
+        this.copy(bases = bases.mkString)
```
tfenne, line 584: This is still less efficient that it could be.  IIRC I suggested using `new String(bases)` instead of `bases.mkString`.  The former does a very efficient array copy, while the latter assembles the string one character at a time through a StringBuilder which is going to be much slower.

tfenne: Also the `forloop` construct in `FgbioDef` is more efficient that doing `for (i <- Range())`

tfenne (changes requested):

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -142,12 +142,18 @@ object DemuxFastqs {
...
+                             qualityEncoding: QualityEncoding,
                              batchSize: Int = DemuxBatchRecordsSize): Iterator[DemuxResult] = {
 
     require(demultiplexer.expectedNumberOfReads == sources.length,
       s"The demultiplexer expects ${demultiplexer.expectedNumberOfReads} reads but ${sources.length} FASTQ sources given.")
 
+    if (threshold+qualityEncoding.asciiOffset > Byte.MaxValue) throw new IllegalArgumentException(f"the threshold $threshold exceeds max possible value")
+
     val zippedIterator = FastqSource.zipped(sources)
+    val thresh         = (threshold + qualityEncoding.asciiOffset).toByte
```
tfenne, line 155:
```suggestion
    val zippedIterator = FastqSource.zipped(sources)

    if (threshold + qualityEncoding.asciiOffset > Byte.MaxValue)
      throw new IllegalArgumentException(f"Quality threshold $threshold exceeds max possible value")
    val thresh         = (threshold + qualityEncoding.asciiOffset).toByte
```

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -558,15 +572,40 @@ private[fastq] object FastqDemultiplexer {
...
     * @param records the records, one for each read that has template bases.
     */
-  case class DemuxResult(sampleInfo: SampleInfo, numMismatches: Int, records: Seq[DemuxRecord])
+  case class DemuxResult(sampleInfo: SampleInfo, numMismatches: Int, records: Seq[DemuxRecord]) {
+
+    /** A function to mask bases that have a quality score less than or equal to a specified threshold
+      * @param threshold The threshold for masking bases, exclusive. Bases with a quality score < threshold will be masked
+      * @return a new DemuxResult with updated bases
+      */
+    def maskLowQualityBases(threshold: Byte, qualityEncoding: QualityEncoding): DemuxResult = { // using this.type here causes a mismatch error
```
tfenne, line 604:
```suggestion
    def maskLowQualityBases(threshold: Byte, qualityEncoding: QualityEncoding): DemuxResult = {
```

tfenne (approved):
Some minor quibbles on formatting/wording suggested.

PR #717, opened 2021-09-24
Add helpers for mateCigar and matesOverlap on SamRecord

clintval:


src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -169,22 +172,30 @@ trait SamRecord {
...
   @inline final def mateUnclippedEnd: Option[Int] = {
     require(paired && mateMapped, "Cannot get mate unclipped end position on read without a mapped mate.")
-    get[String]("MC").map { cig =>
-      val cigar = Cigar(cig)
+    mateCigar.map { cigar =>
       mateStart + cigar.lengthOnTarget - 1 + cigar.reverseIterator.takeWhile(_.operator.isClipping).map(_.length).sum
     }
   }
+  @inline final def mateOverlaps: Option[Boolean] = {
+    Option.when(mapped && paired && mateMapped) {
```
clintval, line 195: I ordered the booleans intentionally to short-circuit fast in the common cases.

I return `None` when all the prerequisites are not satisfied.

However, I noticed all the other mate functions above this one raise exceptions!

I try to avoid try-catch/exceptions when possible, are you open to making the other functions pure functions?

nh13: @tfenne likely has a more articulate opinion, and while I also really like pure functions, I also don't want to have `mateOverlaps.getOrElse(throw new IllegalStateException("I have to write this every time"))` everywhere, so I think some of the methods should not be pure and do the existence check for you.  I am open to suggestions for this.

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -169,22 +172,30 @@ trait SamRecord {
 
   @inline final def mateStart: Int = getMateAlignmentStart
   @inline final def mateStart_=(s: Int):Unit = setMateAlignmentStart(s)
+
+  @inline final def mateCigar: Option[Cigar] = {
+    require(paired && mateMapped, s"Cannot get a mate cigar with a record missing the $McTag tag.")
```
nh13, line 177:
```suggestion
    require(paired && mateMapped, s"Cannot get a mate cigar on read without a mapped mate.")
```

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -289,6 +300,9 @@ object SamRecord {
   /** The maximum insert size that can be stored in a [[SamRecord]]. */
   val MaximumInsertSize: Int = SAMRecord.MAX_INSERT_SIZE
 
+  /** The SAM tag for the mate's cigar. */
+  val McTag: String = MC.toString
```
nh13, line 304: This is a slippery slope.  I we _want_ to include them, I'd create `object SamTags` to store them in.  Honestly, I don't see the full utility of having such a facade and not just using either "MC" hardcoded as we've done, or `htsjdk.samtools.SAMTag.MC` directly

nh13 (changes requested):
One minor change.

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
clintval: I don't think any performance boost is meaningful so I typically do it because I'm allergic to string literals. I can rip it out!

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
clintval: My vote would be to take all the exception raising codepaths and route them to `None`.

All of the functions already return `Option[T]` so more `None` cases seem like a logical thing to do when there is no way to derive the value the caller is looking for. The current implementation throws exceptions in some cases, and returns `None` in others.

I don't think the idea needs to get in the way of this PR moving forward though although `mateOverlaps` in this PR does the behavior I propose.

clintval:
Back to you!

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: Should you even be asking about a "mate" when the read is not paired?  If not, then returning `None` and not handling that as an error is dangerous.  Perhaps the better return type is `Try[T]` versus `Option[T]`?

clintval: Yeah, it's certainly philosophical. If you want the mate's end but there is a mate but no way to get it's end versus if there is no mate at all, I would say both of those are cases where there are no mate ends (`None').

I reserve exceptions for exceptional scenarios.

nh13: > I reserve exceptions for exceptional scenarios

I remember reviewing the original API and thinking about this too, and @tfenne and I landed on the current compromise.

tfenne: This is tricky because there are several cases where mateOverlaps cannot function:

1. The read has no mate - why are you calling mateOverlaps?  Perhaps you should get an exception?
2. The mate exists but is not mapped, this seems like a gray zone
3. The mate exists, is mapped, but the MC tag is absent, `None` seems like a sensible result since it could overlap but we cannot tell (though arguably we _can_ tell in some instances when `start <= mateStart <= end`).

So I think as implemented, the returns don't make a lot of sense to me.  I think:

1. Should either throw an exception (this would be consistent with how other mate related methods work when there is no mate) _or_ return Some(false) because it is knowable that this read does not have a mate overlapping it if a) there is no mate or b) the mate isn't mapped or c) this read isn't mapped
2. I'd argue this should likely return `Some(false)` because again, we can assert that the mate is not mapped in a way that overlaps this read
3. Returning None makes sense here because we can't definitely tell whether the mate overlaps or not.

I.e. true = they overlap, false = the don't overlap (possibly because they or their alignments don't exist), None = they _might_ overlap and we cannot tell.

I think I'd rewrite as (and note the quibble on the name ;) ):

```scala
@inline final def matesOverlap: Option[Boolean] = {
  if (!mapped || !paired || !mateMapped || refIndex != mateRefIndex) Some(false) else {
    if (mateStart >= start && mateStart <= end) Some(true)
    else mateEnd.map(mateEnd => CoordMath.overlaps(start, end, mateStart, mateEnd))
  }
}
```

clintval:
Thanks Tim I think that makes a lot of sense!

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -169,22 +170,31 @@ trait SamRecord {
...
   }
   @inline final def mateUnclippedEnd: Option[Int] = {
     require(paired && mateMapped, "Cannot get mate unclipped end position on read without a mapped mate.")
-    get[String]("MC").map { cig =>
-      val cigar = Cigar(cig)
+    mateCigar.map { cigar =>
       mateStart + cigar.lengthOnTarget - 1 + cigar.reverseIterator.takeWhile(_.operator.isClipping).map(_.length).sum
     }
   }
+  @inline final def matesOverlap: Option[Boolean] = {
```
nh13, line 192: 1. mateOverlaps? I don’t see how this works on multiple mates :0
2. I prefer exception if the read pair doesn’t have a mate to be consistent with the current implementation. Happy to have another PR address the exception vs pure function issue to being the discussion there. If you and @tfenne object, let’s take this offline.

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
clintval: Both these were caused by suggestions of Tim: https://github.com/fulcrumgenomics/fgbio/pull/717#discussion_r716027927

I think "mates overlap" was Tim's choice because the read pair's are each others mates.

As in, "do mates overlap? Yes, mates overlap."

Let me know what you'd like me to do!

nh13: I disagree with @tfenne so let's take this offline.  Since I don't like the new name ( E.g. "does I overlap my mate", "yes, my mate overlaps") and I think we should keep the implementation consistent.

nh13 (approved):

clintval:
Wait to merge, I am redundantly using `require` twice for all these methods since they delegate back to `mateCigar` but I can revert that change so we don't have the overhead of checking something twice.

clintval (commented):
@nh13 This is ready for final review!

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -185,6 +190,12 @@ trait SamRecord {
...
+    require(mapped && paired && mateMapped, "Cannot determine if mates overlap without paired mates that are both mapped.")
+    if (refIndex != mateRefIndex) Some(false)
+    else if (mateStart >= start && mateStart <= end) Some(true)
```
nh13, line 196: if `end < mateStart` then I think we should return `Some(false)` right?

nh13 (approved):

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
clintval: I think you are right, we could code another fast path.

clintval: I added the case and luckily I already have a unit test in place to trigger the branch.

I'll confirm all branches are actually tested once codecov tells me.

PR #729, opened 2021-10-18
Option to indicated sorted FASTQ to add UMIs more quickly

kstromhaug:
…lookup

### DIFF
--- a/src/test/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmisTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmisTest.scala
@@ -51,7 +51,7 @@ class AnnotateBamWithUmisTest extends UnitSpec {
 
   it should "successfully add UMIs to a BAM in when the fastq is sorted" in {
     val out       = makeTempFile("with_umis.", ".bam")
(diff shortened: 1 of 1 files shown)


Poshi:
Given that #709 has been closed in favor of this one, I'd like to note that here I'm missing a small functionality that was present at the closed one: the ability to keep the UMI qualities along with their bases (if desired).
UMI qualities could be needed to better assess the mismatched bases, or to be able to recover the original files from the final bam. Usually they are stored in the QX tag, although I took the approach of letting the user say which tag name to use. If no tag name given, then the current behavior is expected.
It would be nice to add that functionality to this "almost equivalent" PR.

nh13:
@Poshi good catch! We prefer that the two distinct features, though on the same tool, be made in separate PRs. Would you like to make a PR, or we can?

Poshi:
PR created (https://github.com/fulcrumgenomics/fgbio/pull/733).
BTW, the merged PR (#729) contains some semantic error that can also make difficult to work with partially split data sets. I'll create an issue with a better explanation.
Of course we all have personal lives and sponsored work must be prioritized, but given the fact that I've done the sponsored work without the sponsoring, it would be nice to take those fruits before they are too ripe (and a newly created PR takes their place) :-)

nh13:
@Poshi sorry if my tone came off poorly, it's hard in written form.  If folks either report issues/bugs with reproducible examples, or contribute, we attempt to prioritize them, but also try to set expectations low!

Issue #732, opened 2021-10-20
Support outputting the UMI base qualities when extracting/annotating UMIs


nh13:
Optionally extract/annotate the UMI qualities in addition to the UMI bases in the following tools:

- [x] `AnnotateBamWithUmis` (see #733)
- [ ] `ExtractUmisFromBam`
- [x] `FastqToBam` (see #740)
- [ ] ~`DemuxFastqs`~

PR #733, opened 2021-10-21
New: store UMI qualities if requested (AnnotateBamWithUmis)

Poshi:
Update of issue #732. This commit only considers AnnotateBamWithUmis tool.


nh13:
@Poshi I added them for you so you don't have to!

Poshi:
@nh13 Too late. I already have them. Do you want me to push them? Or better keep yours?
I added two tests that only checks that qualities are added (bases added are checked in other tests). To mimmick hot the bases added test works I had to modify the input SAM file, but this didn't break any other test, so it should be OK.

Poshi:
Done. I think it was all. Any other idea or suggestion will be welcomed, but tackled after sleeping :-)
Thanks for your patience and promptly reply.

PR #737, opened 2021-11-02
Skip empty lines at end of sample sheet when parsing sample data

mjhipp:
It is not uncommon for someone to accidentally add an extra blank line to the end, or for Microsoft Excel to add rows with just commas at the bottom of a sample sheet after editing. 

A blank row will hit this error (even at the bottom of the file):

https://github.com/fulcrumgenomics/fgbio/blob/b01fc04e46ab5ddf18076136f5ac8c1ebe9ad53f/src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala#L106-L108

A row with just commas hits this error:

https://github.com/fulcrumgenomics/fgbio/blob/b01fc04e46ab5ddf18076136f5ac8c1ebe9ad53f/src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala#L119

This is a suggested change to make it easier to parse some sample sheets with these types of data-less rows.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
+++ b/src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
@@ -98,16 +98,27 @@ object SampleSheet {
       case None => throw new IllegalArgumentException("Could not find the header for sample data.")
     }).map(_.toUpperCase.trim)
 
-    // get the rows (values), so skip "[Data]" and the header
+    // get the rows (values), so skip "[Data]" and header line
     val lineNumber = preData.size + 2 // 0-based
-    val sampleDatums = postData.drop(2).zipWithIndex.map { case (line, rowNumber) =>
-        val values = line.split(SplitRegex, -1)
-        // check we have the correct # of columns
-        if (values.size != header.length) {
-          throw new IllegalArgumentException(s"Found a line with a mismatching number of columns on line #$lineNumber: " + line)
+    val sampleDatums = postData
+      .drop(2) // skip the "[Data]" and header line
+      .zipWithIndex // add the row #
+      .map { case (line, rowNumber) => (line.split(SplitRegex, -1), rowNumber) } // just split the values
+      .reverse // to skip empty lines, and lines with all empty values, at the end of the sample
+      .dropWhile { case (values, rowNumber) => values.isEmpty || values.forall(_.trim.isEmpty) } // no values, regardless of number, so drop/skip it
+      .reverse // put back in the correct order
+      .map { case (values, rowNumber) =>
+        if (values.isEmpty || values.forall(_.trim.isEmpty)) { // empty line, or all values are empty
+          throw new IllegalArgumentException(s"Found empty line on row ${rowNumber + 1} of the sample data section.") // 1-indexed sample data row
+        }
+        else if (values.size != header.length) { // there aren't the required # of values
+          throw new IllegalArgumentException(
+            s"Found a line with a mismatching number of columns on line #${lineNumber + rowNumber + 1}: " + values.mkString(",")
+          )
         }
         header.zip(values.map(_.trim)).toMap
       }
+
     lane match {
       case Some(l: Int) => sampleDatums.filter { SampleSheet.getIntField(_, Lane).exists(_ == l) }
       case None         => sampleDatums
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -100,13 +100,16 @@ object SampleSheet {
 
     // get the rows (values), so skip "[Data]" and the header
     val lineNumber = preData.size + 2 // 0-based
-    val sampleDatums = postData.drop(2).zipWithIndex.map { case (line, rowNumber) =>
+    val sampleDatums = postData.drop(2).zipWithIndex.flatMap { case (line, rowNumber) =>
         val values = line.split(SplitRegex, -1)
+        // If this is an empty line, skip it
+        if (values.forall(_.trim.isEmpty)) None
```
nh13, line 106: this skips any empty lines, even in the middle of the data section, which isn't what you wanted (lines at the end only).

nh13:
1. Can you discard empty lines at the end as well as "all comma lines", but not allow empty lines in the middle of the data section?  Can use something like `.reverseIterator.dropWhile(_ => ...).reverseIterator`
2. I don't see an improvement for the second error.  
3. Could you also add unit tests?

mjhipp:
> 1. Can you discard empty lines at the end as well as "all comma lines", but not allow empty lines in the middle of the data section?

Yes, this was main issue with the PR as written, I can update to account for this.

> 2\. I don't see an improvement for the second error.

Do you mean you don't think my changes fix the error? My thinking was `",,,"` -> `["", "", "", ""]` and `["", "", "", ""].forall(_.trim.isEmpty) == true` so it would be skipped and avoid the error. I will add unit tests to make sure.

nh13:
> 2. I don't see an improvement for the second error.

You're right, sorry, end of day skim and I missed that.  Unit tests for documentation :p

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -100,14 +100,25 @@ object SampleSheet {
...
+          Some(header.zip(values.map(_.trim)).toMap)
         }
-        header.zip(values.map(_.trim)).toMap
       }
+    // skip empty lines that come after the last sample data line
+    // throw an error if empty lines are mixed in with sample data
+    val sampleDatums = maybeSampleDatums.take(maybeSampleDatums.lastIndexWhere(_.isDefined) + 1) match {
+      case datums if datums.forall(_.isDefined) => datums.flatten
+      case datums =>
+        val emptyRows = datums.zipWithIndex.filter(_._1.isEmpty).map(_._2).map(_ + 1) // 1 indexed
```
mjhipp, line 119: did 1-indexed row number thinking it would be easier for user (what is row 0?)

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -116,8 +127,8 @@ object SampleSheet {
 
   /** Creates a sample from the given row data. */
   protected def makeSample(sampleOrdinal: Int, sampleDatum: Map[String, String]): Sample = {
-    val sampleName: String          = SampleSheet.getStringField(sampleDatum, SampleName)    getOrElse (throw new IllegalArgumentException(s"Missing: $SampleName"))
-    val sampleId: String            = SampleSheet.getStringField(sampleDatum, SampleId)      getOrElse (throw new IllegalArgumentException(s"Missing: $SampleId"))
+    val sampleName: String          = SampleSheet.getStringField(sampleDatum, SampleName)    getOrElse (throw new IllegalArgumentException(s"Sample at ordinal #$sampleOrdinal is missing: $SampleName"))
+    val sampleId: String            = SampleSheet.getStringField(sampleDatum, SampleId)      getOrElse (throw new IllegalArgumentException(s"Sample at ordinal #$sampleOrdinal is missing: $SampleId"))
```
mjhipp, line 131: Similar to above. Before, the "comma-only" lines just gave the error message "Missing: Sample_Name", which was a bit confusing.

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -100,14 +100,25 @@ object SampleSheet {
 
     // get the rows (values), so skip "[Data]" and the header
     val lineNumber = preData.size + 2 // 0-based
-    val sampleDatums = postData.drop(2).zipWithIndex.map { case (line, rowNumber) =>
+    val maybeSampleDatums = postData.drop(2).zipWithIndex.map { case (line, rowNumber) =>
         val values = line.split(SplitRegex, -1)
+        // treat as empty line iff all columns are empty (empty line or comma-only line, often added by spreadsheet editors)
```
nh13, line 105:
```suggestion
        // treat as empty line if all columns are empty (empty line or comma-only line, often added by spreadsheet editors)
```

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
```diff
@@ -100,14 +100,25 @@ object SampleSheet {
...
+        if (values.forall(_.trim.isEmpty)) None
```
nh13, line 106: This will allow rows where the # of fields is wrong _in the middle_ of the data section, not just the end.

nh13 (changes requested):
@mjhipp thank-you for this effort. I think this PR is getting unwildly, as is the case for sample sheet parsing.

How about we:
1. Read in all the sample lines as-is: 

```scala
val valuesAndRowNumber = postData.drop(2).zipWithIndex.map { case (line, rowNumber) =>
   (line.split(SplitRegex, -1), rowNumber)
}
```

2. Remove those from the end while they are empty or have all empty values.  In the latter case, the # of columns should still match the header!

```scala
valuesAndRowNumber.reverseIterator.dropWhile { case (values, rowNumber) =>
  if (values.isEmpty) true
  else if (values.size != header.length) {
    throw new IllegalArgumentException(s"Found a line with a mismatching number of columns on line #$lineNumber: " + values.mkString(","))
  }
  else {
    values.forall(_.trim.isEmpty)
  }
}.reverseIterator
```

src/main/scala/com/fulcrumgenomics/illumina/SampleSheet.scala
mjhipp: It sets the row `datum` to `None`, which is then checked below when creating `sampleDatums`. 

If there are any `None`s mixed in the middle of the data section it throws the error:

```
"Found empty rows in sample data section: ${emptyRows.mkString(", ")}"
```

nh13:
@mjhipp for the patience and perseverance!

nh13 (approved):

PR #740, opened 2021-11-13
New: added UMI qualities keeping in FastqToBam

Poshi:
Trying to complete another task of issue #732. This PR adds a command line parameter to give a tag for the UMI qualities and, in case it is present, add them to the output BAM.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
@@ -85,6 +85,7 @@ class FastqToBam
   @arg(flag='r', doc="Read structures, one for each of the FASTQs.", minElements=0)            val readStructures: Seq[ReadStructure] = Seq(),
   @arg(flag='s', doc="If true, queryname sort the BAM file, otherwise preserve input order.")  val sort: Boolean = false,
   @arg(flag='u', doc="Tag in which to store molecular barcodes/UMIs.")                         val umiTag: String = ConsensusTags.UmiBases,
+  @arg(flag='q', doc="Tag in which to store molecular barcode/UMI qualities.")                 val umiQualTag: Option[String] = None,
   @arg(          doc="Read group ID to use in the file header.")                               val readGroupId: String = "A",
   @arg(          doc="The name of the sequenced sample.")                                      val sample: String,
   @arg(          doc="The name/ID of the sequenced library.")                                  val library: String,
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -149,6 +150,7 @@ class FastqToBam
       val subs = fqs.iterator.zip(structures.iterator).flatMap { case(fq, rs) => rs.extract(fq.bases, fq.quals) }.toIndexedSeq
       val sampleBarcode = subs.iterator.filter(_.kind == SampleBarcode).map(_.bases).mkString("-")
       val umi           = subs.iterator.filter(_.kind == MolecularBarcode).map(_.bases).mkString("-")
+      val umiQual       = subs.iterator.filter(_.kind == MolecularBarcode).map(_.quals).mkString(" ")
```
nh13, line 153: Any reason not to use the same separator as the bases?
```suggestion
      val umiQual       = subs.iterator.filter(_.kind == MolecularBarcode).map(_.quals).mkString("-")
```

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
Poshi: Yes, there's a reason: because the [official specification](https://samtools.github.io/hts-specs/SAMtags.pdf) says it.
The tag that should contain the qualities of the bases in the _RX_ tag is the _QX_ tag. And the description of that tag (page 6) literally says:
> In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the recommended implementation concatenates all the quality strings with a space (‘ ’) between the different strings.

It is just a recommendation, but unless we offer the user a way to customize the separator, I would go for the recommended one (like we are following the specs by using an hyphen as a separator in the _RX_ tag).

nh13: Got it.  Fixed `AnnotateBamWithUmis` to have the same behavior, as well as fixed a bug: https://github.com/fulcrumgenomics/fgbio/pull/742.

Poshi: I didn't remember having tweaked this behavior in `AnnotateBamWithUmis`, I just assumed the code was OK and added a very similar code for the qualities.
Either case, now that the issue is identified, it should be easy to solve :-)

Poshi:
Sorry, I'm still kind of new in some workflows. I explained why I want to reject the suggested change and you agreed with me. I marked the conversation as resolved without commiting the suggested change. But the summary still says that a change is requested. And I cannot find the proper way to reject the change, so it looks like I'm just ignoring it.
Is there anything else I should do for a proper ending of the review process? I cannot find anything in the github documentation, only how to accept and commit the proposed change :-/

PR #742, opened 2021-11-15
AnnotateBamWithUmis Bug Fix

nh13:
This [commit](https://github.com/Poshi/fgbio/commit/468a843d19da4f67106dc148a1dba1f2c3390eab#diff-6af11c9e15cd56e53dde749b569558bdbb3755c359842f43abef11c6ba04d07e) incorrectly joined the UMI bases with no separator when there were multiple segments.   Then #733 was introduced which only returned the first set of UMI bases ([commit](https://github.com/Poshi/fgbio/blob/daa0254beaab498f66de13bfb5c02f7e2e2b041c/src/main/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmis.scala#L94)).  The latter would not properly annotate `5M3S5M` for example, as it would only take the first 5 bp.

- [x] needs tests

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmis.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmis.scala
@@ -91,11 +91,10 @@ class AnnotateBamWithUmis(
   }
 
   /** Extracts the UMI bases and qualities given the read structure */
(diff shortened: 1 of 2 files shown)


Poshi:
Looks good :-)

nh13:
@tfenne can you take a quick look?

tfenne (approved):

PR #747, opened 2021-12-15
Forcing log4j transitive dependency (through GKL) to version that doesn't have zero day exploit.

tfenne:
The changes to the plugins and sbt version were just to get a working run of `dependencyTree` so I could track down _why_ we had a dependency on log4j in the first place.

We should probably do a release once this is merged.  Sigh.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ JeanLuc.iml
 target
 project/project
 .DS_Store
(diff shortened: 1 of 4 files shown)


jacarey (approved):
LGTM

nh13 (approved):

PR #754, opened 2022-01-08
Speed up property access on Cigar case class

clintval:
I have an algorithm that repeatedly and unpredictably accesses the properties on Cigar classes. 

This algorithm could be significantly faster if the Cigar properties could be calculated once.

Would you be willing to drop these values down to set-once lazy values?

This PR should not cause any issues with users of this code since `case class` sub-classing should be avoided anyway.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
+++ b/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
@@ -57,7 +57,7 @@ object Cigar {
 
   /** Constructs a Cigar object from a Cigar string. */
   def apply(cigar: String): Cigar = {
-    require(cigar != null && cigar.length > 0, "Can't parse a null or empty cigar string.")
+    require(cigar != null && cigar.nonEmpty, "Can't parse a null or empty cigar string.")
     val elements = new ArrayBuffer[CigarElem]
     val chars    = cigar.toCharArray
 
@@ -96,7 +96,7 @@ object Cigar {
   */
 case class Cigar(elems: IndexedSeq[CigarElem]) extends Iterable[CigarElem] {
   // Cache whether or not the Cigar is coalesced already (i.e. has no pair of adjacent elements with the same operator)
-  private val isCoalesced: Boolean = {
+  private lazy val isCoalesced: Boolean = {
     var itIs = true
     var index = 0
     while (index < elems.length-1 && itIs) {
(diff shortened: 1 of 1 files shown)


clintval:
I'm now seeing some really significant speedups (5%-35%) in many of my tools with this PR!

This is because `SamRecord.end()` uses `cigar.lengthOnTarget()` which must recalculate every time it is called.

https://github.com/fulcrumgenomics/fgbio/blob/ff43d10d5443874504632f6af2b27f6c7e0f23c2/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala#L149

For tools where either of the above methods are called more than once, this PR will make a positive difference.

nh13:
@clintval do you have a sense of the memory increase this causes?

clintval:
TLDR: How would you like me to benchmark? I'm personally not worried about the memory overhead.

**My first assumption** is that most programs will not call every single one of these properties for the lifetime of the cigar class instance. This means that only a few of the integers will need to be stored for any given cigar instance in an average case. In the worst case, a program will force all of the integers to be computed and they will be stored in the heap.

**My second assumption** is that tools often need to re-access cigar properties multiple times, especially across library function boundaries. If we could cache that computation, then repeated access would be amortized (faster). Coding the properties to be lazy also ensures the properties are only computed and stored if needed. Given the existing behavior, when tools need to re-access cigar properties, I can imagine there will be multiple references to different integers (with the same value) which would actually increase heap size.

I believe the speedup of not having to `.foldLeft()` more than once over all cigar operators is worth any extra weight of caching a single integer per property. Fold left is not the fastest way to aggregate across a collection (you need to box integers and create tuples once per element) and some cigar sequences can also be very long!

This is a difficult test to benchmark in a GCd language like Scala. Changing the dynamics of the heap will change when garbage collection will be called to cleanup stale references to cigar's and their integers. However, I have access to JProfiler, so if there is a test you would like me to run, I can code it up!

tfenne:
I don't think there's a need to benchmark, but it is worth pointing out that this will increase the size of Cigar instances even when they are just being ferried around and not looked at.  Depending on how it actually works, Cigar either inflates by ~17 bytes or ~35 bytes to hold a) 5 booleans to check for initialization and either b) 4 Ints and 1 boolean, or 5 64bit references.

If you had 1m Cigar objects kicking around in memory that'd suck up another 16MB assuming primitives.  I would _guess_ that programs are either not holding that many in memory at one time, or if they are that +/-16MB will be un-noticeable.

My only other thought would be that if you are finding this makes such a big difference, it may well make sense to have _another_ case class `CigarStats(coalesced: Boolean, lenOfQ: Int, lenOnT: Int, alignedBases: Int, clippedBases: Int)` which is not exposed in the API, but would allow you to `private lazy val stats = CigarStats(this)` and compute all five stats in a single loop for improved efficiency.

clintval:
I liked your idea @tfenne, and theoretically it should be more time-efficient in most of my code.

There may be a slightly higher memory overhead for the case class wrapper of the statistics but I think that is OK for the reason you state: if you have millions of cigars in memory, you probably have a very memory-consuming application anyway, and a few dozen additional megabytes aren't likely to make a practical difference. Reverting the properties on `Cigar` back down to methods means we no longer need to track if they are initialized.

The `CigarStats` class takes up 32 bytes with padding:

```scala
@ import $ivy.`org.apache.spark::spark-sql:3.2.0`
@ import org.apache.spark.util.SizeEstimator._
@ case class CigarStats(lengthOnQuery: Int, lengthOnTarget: Int, alignedBases: Int, clippedBases: Int)
@ estimate(stats)
res5: Long = 32L
```

What do you think?

clintval:
I should clarify that the library function execution speedups I am seeing are very good.

But many of my applications are IO-bound so this may not be dramatic when actually running a full IO test on a CLI tool.

Regardless, I think it is still a worthwhile performance enhancement and shaves minutes off a few of my CLI tools.

nh13 (approved):
Let's wait for @tfenne's approval too, but I like this compromise.

tfenne (approved):

PR #757, opened 2022-01-11
Retry the Intel deflate test as for some levels it's close to equal time

nh13:
@tfenne this is why after the updated to `0.8.8` we are seeing spurious CI failures.  This is a short term fix.

### DIFF
--- a/src/test/scala/com/fulcrumgenomics/cmdline/IntelCompressionTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/cmdline/IntelCompressionTest.scala
@@ -25,7 +25,6 @@
 package com.fulcrumgenomics.cmdline
 
 import java.nio.file.Paths
-
 import com.fulcrumgenomics.FgBioDef.SafelyClosable
 import com.fulcrumgenomics.bam.api.SamSource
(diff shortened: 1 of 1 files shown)


tfenne (approved):
Or perhaps it should test that it's faster or at least not more than ~5% slower?

nh13:
> Or perhaps it should test that it's faster or at least not more than ~5% slower?

Why not both!

src/test/scala/com/fulcrumgenomics/cmdline/IntelCompressionTest.scala
```diff
@@ -76,7 +84,7 @@ class IntelCompressionTest extends UnitSpec {
 
       info(f"Intel: ${intelTime}ms JDK: ${jdkTime}ms speedup: ${jdkTime/intelTime.toFloat}%.2fx")
 
-      intelTime should be <= jdkTime
+      intelTime should be <= (jdkTime * 1.05)
```
nh13, line 87:
```suggestion
      intelTime.toDouble should be <= (jdkTime * 1.05)
```

PR #758, opened 2022-01-11
Make it easier to extend FgMetricsDoclet

nh13:
For other scala toolkits, we'd need to copy over the class wholesale or at least copy the private methods.  This makes it easier to extend `FgMetricsDoclet`.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/internal/FgMetricsDoclet.scala
+++ b/src/main/scala/com/fulcrumgenomics/internal/FgMetricsDoclet.scala
@@ -122,7 +122,7 @@ class FgMetricsDoclet extends Doclet(reporter = new ConsoleReporter(new Settings
   }
 
   /** Finds the [[scala.tools.nsc.doc.model.DocTemplateEntity]] instances that correspond to subclasses of [[Metric]] */
-  protected def findMetricsClasses: List[DocTemplateEntity] = {
+  private def findMetricsClasses: List[DocTemplateEntity] = {
     def find(template: DocTemplateEntity): List[DocTemplateEntity] = {
       template :: template.templates.collect { case d: DocTemplateEntity => find(d) }.flatten
     }
--- a/src/main/scala/com/fulcrumgenomics/internal/FgMetricsDoclet.scala
+++ b/src/main/scala/com/fulcrumgenomics/internal/FgMetricsDoclet.scala
@@ -97,11 +97,6 @@ class FgMetricsDoclet extends Doclet(reporter = new ConsoleReporter(new Settings
        |
        |* `Count` is an integer representing the count of some item
        |* `Proportion` is a real number with a value between 0 and 1 representing a proportion or fraction
-       |
-       |## Table of Contents
-       |
-       ||Metric Type|Description|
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/internal/FgMetricsDoclet.scala
```diff
@@ -92,8 +86,26 @@ class FgMetricsDoclet extends Doclet(reporter = new ConsoleReporter(new Settings
...
+       |This page contains descriptions of all metrics produced by all $toolkitName tools.  Within the descriptions
+       |the type of each field/column is given, including two commonly used types:
+       |
+       |* `Count` is an integer representing the count of some item
+       |* `Proportion` is a real number with a value between 0 and 1 representing a proportion or fraction
+       |
+       |## Table of Contents
+       |
+       ||Metric Type|Description|
+       ||-----------|-----------|""".stripMargin
```
nh13, line 104:
```suggestion
```

src/main/scala/com/fulcrumgenomics/internal/FgMetricsDoclet.scala
```diff
@@ -115,7 +122,7 @@ class FgMetricsDoclet extends Doclet(reporter = new ConsoleReporter(new Settings
   }
 
   /** Finds the [[scala.tools.nsc.doc.model.DocTemplateEntity]] instances that correspond to subclasses of [[Metric]] */
-  private def findMetricsClasses: List[DocTemplateEntity] = {
+  protected def findMetricsClasses: List[DocTemplateEntity] = {
```
tfenne, line 125: I'm curious if you're overriding this one function?  Perhaps there should be another `def` that returns a set of packages within which to extract Metric classes, and then filter here?

tfenne (approved):

src/main/scala/com/fulcrumgenomics/internal/FgMetricsDoclet.scala
nh13, line 125:
```suggestion
  private def findMetricsClasses: List[DocTemplateEntity] = {
```

src/main/scala/com/fulcrumgenomics/internal/FgMetricsDoclet.scala
```diff
@@ -92,8 +86,21 @@ class FgMetricsDoclet extends Doclet(reporter = new ConsoleReporter(new Settings
...
+
+  protected def preamble: String = {
+    s"""
+       |# $toolkitName Metrics Descriptions
+       |
+       |This page contains descriptions of all metrics produced by all $toolkitName tools.  Within the descriptions
+       |the type of each field/column is given, including two commonly used types:
+       |
+       |* `Count` is an integer representing the count of some item
+       |* `Proportion` is a real number with a value between 0 and 1 representing a proportion or fraction
```
nh13, line 99:
```suggestion
       |* `Proportion` is a real number with a value between 0 and 1 representing a proportion or fraction""".stripMargin
```

tfenne:
@nh13 I think this can be merged, right?

PR #760, opened 2022-01-13
Add an entrypoint to SamSource for bytes of SAM/BAM/CRAM

clintval:


src/main/scala/com/fulcrumgenomics/bam/api/SamSource.scala
```diff
@@ -66,6 +65,29 @@ object SamSource {
...
+
+  /** Constructs a [[SamSource]] to read from the provided input stream.
+    *
+    * @param stream the input stream of SAM/BAM/CRAM bytes
+    * @param ref an optional reference sequencing for decoding CRAM files
+    * @param async if true use extra thread(s) to speed up reading
+    * @param stringency the validation stringency to apply when reading the data
+    * @param factory a SAMRecordFactory; MUST return classes that mix in [[SamRecord]]
+    */
+  def apply(
```
clintval, line 77: Unfortunately cannot set defaults here due to multiple `apply` overloadings.

src/main/scala/com/fulcrumgenomics/bam/api/SamSource.scala
```diff
@@ -66,6 +65,29 @@ object SamSource {
...
+    stream: InputStream,
+    ref: Option[PathToFasta],
+    async: Boolean,
+    stringency: ValidationStringency,
+    factory: SAMRecordFactory,
+  ): SamSource = {
+    val fac = SamReaderFactory.make()
+    fac.samRecordFactory(factory)
+    fac.setUseAsyncIo(async)
+    fac.validationStringency(stringency)
```
nh13, line 87: do you want to refactor and have common code in the apply methods in one place?

nh13 (changes requested):
One suggestion for you

src/main/scala/com/fulcrumgenomics/bam/api/SamSource.scala
clintval: Yeah that could be nice, it would help future users from forgetting to set a parameter if a new entrypoint was added.

clintval:
Would you be willing to do a re-review?

src/main/scala/com/fulcrumgenomics/bam/api/SamSource.scala
nh13: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
What if you renamed the function to say `fromInputStream`?

nh13 (approved):

Issue #759, opened 2022-01-12
CallMolecularConsensusReads truncate reads


ruolin:
Hi fgbio community, 

I have a single read-pair family (cD=1) which after call consensus was truncated. This is indeed weird because it should just do nothing and output the same read. To reproduce this run `java fgbio-1.5.0.jar CallMolecularConsensusReads -i a.bam -o a.ccs.bam -M 1`. The following is the a.bam. 

```
@HD     VN:1.6  SO:unsorted     GO:query        SS:unsorted:template-coordinate
@SQ     SN:1    LN:249250621    AS:GRCh37       M5:1b22b98cdeb4a9304cb5d48026a85128     UR:http://www.broadinstitute.org/ftp/pub/seq/references/Homo_sapiens_assembly19.fasta   SP:Homo Sapiens
@SQ     SN:2    LN:243199373    AS:GRCh37       M5:a0d9851da00400dec1098a9255ac712e     UR:http://www.broadinstitute.org/ftp/pub/seq/references/Homo_sapiens_assembly19.fasta   SP:Homo Sapiens
@SQ     SN:3    LN:198022430    AS:GRCh37       M5:fdfd811849cc2fadebc929bb925902e5     UR:http://www.broadinstitute.org/ftp/pub/seq/references/Homo_sapiens_assembly19.fasta   SP:Homo Sapiens
@SQ     SN:4    LN:191154276    AS:GRCh37       M5:23dccd106897542ad87d2765d28a19a1     UR:http://www.broadinstitute.org/ftp/pub/seq/references/Homo_sapiens_assembly19.fasta   SP:Homo Sapiens
@SQ     SN:5    LN:180915260    AS:GRCh37       M5:0740173db9ffd264d728f32784845cd7     UR:http://www.broadinstitute.org/ftp/pub/seq/references/Homo_sapiens_assembly19.fasta   SP:Homo Sapiens
@SQ     SN:6    LN:171115067    AS:GRCh37       M5:1d3a93a248d92a729ee764823acbbc6b     UR:http://www.broadinstitute.org/ftp/pub/seq/references/Homo_sapiens_assembly19.fasta   SP:Homo Sapiens
@SQ     SN:7    LN:159138663    AS:GRCh37       M5:618366e953d6aaad97dbe4777c29375e     UR:http://www.broadinstitute.org/ftp/pub/seq/references/Homo_sapiens_assembly19.fasta   SP:Homo Sapiens
[...]

PR #761, opened 2022-01-14
Support for clipping reads that extend past their mate

nh13:
### Summary 

This adds:

1.  methods in `SamRecordClipper` to clip bases that map beyond the end of the mate for FR pairs.
2. Use these methods in `ClipBam` to implement the option `--clip-bases-past-mate`.
3.   Use these method in consensus calling to properly decide the number of bases to clip based on insert size for FR pairs.  This fixes a bug in the consensus caller that could improperly trim small inserts. 

### More details on the  bug in the consensus caller that could improperly trim small inserts

If the reads are longer than the insert size, but aligned with
insertions, the consensus read may have the insertion trimmed.

See issue #759.  Corrected output to include the full length read:

```
@HD	VN:1.6	SO:queryname	GO:none
@RG	ID:A	LB:lib1	SM:A	PL:ILLUMINA	PU:unit1
@CO	Read group A contains consensus reads generated from 1 input read groups.
lib1:8193187/A	77	*	0	0	*	*	0	0	ACCATTCTAATTACAGTGGGGGAAGTGTGCGCGCAAGTGTGTGCGTGTGTGCGCACGCGCAGGTGTGTGCGTGAGTGAAGGGATGTAGACAGAGGCTGCTGGCCGGG	CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC	cD:i:1	cE:f:0.0	RG:Z:A	MI:Z:8193187/A	cM:i:1	RX:Z:TTA-TCT	cd:B:s,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1	ce:B:s,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
lib1:8193187/A	141	*	0	0	*	*	0	0	CCCGGCCAGCAGCCTCTGTCTACATCCCTTCACTCACGCACACACCTGCGCGTGCGCACACACGCACACACTTGCGCGCACACTTCCCCCACTGTAATTAGAATGGT	CCCCCCCCCCCCCCCCCCCCC9CCC9CCC9CCCCCCCCC9CCCCCCCCCCCCCCCCCCCCCCCCCCCCC99CCCCCCC9CCCCCCCCCCCCCCCC9CCCCCCCCCCC	cD:i:1	cE:f:0.0	RG:Z:A	MI:Z:8193187/A	cM:i:1	RX:Z:TTA-TCT	cd:B:s,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1	ce:B:s,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
+++ b/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
@@ -95,21 +95,21 @@ private[alignment] case class CigarStats(lengthOnQuery: Int,
                                          lengthOnTarget: Int,
                                          alignedBases: Int,
                                          clippedBases: Int,
-                                         softClippedFivePrimeBases: Int,
-                                         softClippedThreePrimeBases: Int)
+                                         leadingSoftClippedBases: Int,
+                                         trailingSoftClippedBases: Int)
 
 /** Companion object for [[Cigar]]. */
 private[alignment] object CigarStats {
 
   /** Build a [[CigarStats]] from a [[Cigar]]. */
   private[alignment] def apply(cigar: Cigar): CigarStats = {
-    var lengthOnQuery              = 0
-    var lengthOnTarget             = 0
-    var alignedBases               = 0
-    var clippedBases               = 0
-    var softClippedFivePrimeBases  = 0
-    var softClippedThreePrimeBases = 0
-    var index                      = 0
+    var lengthOnQuery            = 0
+    var lengthOnTarget           = 0
+    var alignedBases             = 0
+    var clippedBases             = 0
+    var leadingSoftClippedBases  = 0
+    var trailingSoftClippedBases = 0
+    var index                    = 0
 
     // leading clipping
     while (index < cigar.elems.length && cigar.elems(index).operator.isClipping) {
@@ -117,7 +117,7 @@ private[alignment] object CigarStats {
       lengthOnQuery  += elem.lengthOnQuery
       lengthOnTarget += elem.lengthOnTarget
       clippedBases += elem.length
-      if (elem.operator == CigarOperator.S) softClippedFivePrimeBases += elem.length
+      if (elem.operator == CigarOperator.S) leadingSoftClippedBases += elem.length
       index += 1
     }
 
@@ -136,7 +136,7 @@ private[alignment] object CigarStats {
       lengthOnQuery  += elem.lengthOnQuery
       lengthOnTarget += elem.lengthOnTarget
       clippedBases += elem.length
-      if (elem.operator == CigarOperator.S) softClippedThreePrimeBases += elem.length
+      if (elem.operator == CigarOperator.S) trailingSoftClippedBases += elem.length
       index += 1
     }
 
@@ -145,12 +145,12 @@ private[alignment] object CigarStats {
     )
 
     CigarStats(
-      lengthOnQuery              = lengthOnQuery,
-      lengthOnTarget             = lengthOnTarget,
-      alignedBases               = alignedBases,
-      clippedBases               = clippedBases,
-      softClippedFivePrimeBases  = softClippedFivePrimeBases,
-      softClippedThreePrimeBases = softClippedThreePrimeBases
+      lengthOnQuery            = lengthOnQuery,
+      lengthOnTarget           = lengthOnTarget,
+      alignedBases             = alignedBases,
+      clippedBases             = clippedBases,
+      leadingSoftClippedBases  = leadingSoftClippedBases,
+      trailingSoftClippedBases = trailingSoftClippedBases
     )
   }
 
@@ -201,11 +201,27 @@ case class Cigar(elems: IndexedSeq[CigarElem]) extends Iterable[CigarElem] {
   /** Returns the number of bases that are clipped between the two sequences. */
   def clippedBases: Int = stats.clippedBases
 
-  /** Returns the number of bases that are soft-clipped on the 5' end of this sequence.  Ignores hard-clips. */
-  def softClippedFivePrimeBases: Int = stats.softClippedFivePrimeBases
+  /** Returns the number of bases that are soft-clipped at the start of the sequence  Ignores hard-clips. */
+  def leadingSoftClippedBases: Int = stats.leadingSoftClippedBases
 
-  /** Returns the number of bases that are soft-clipped on the 3' end of this sequence.  Ignores hard-clips. */
-  def softClippedThreePrimeBases: Int = stats.softClippedThreePrimeBases
+  /** Returns the number of bases that are soft-clipped at the end of this sequence.  Ignores hard-clips. */
+  def trailingSoftClippedBases: Int = stats.trailingSoftClippedBases
+
+  /** Returns the number of bases that are hard-clipped at the start of the sequence. */
+  def leadingHardClippedBases = this.headOption.map { elem =>
+    if (elem.operator == CigarOperator.H) elem.length else 0
+  }.getOrElse(0)
+
+  /** Returns the number of bases that are clipped at the start of the sequence. */
+  def leadingClippedBases: Int = leadingHardClippedBases + leadingSoftClippedBases
+
+  /** Returns the number of bases that are hard-clipped at the end of the sequence. */
+  def trailingHardClippedBases = this.lastOption.map { elem =>
+    if (elem.operator == CigarOperator.H) elem.length else 0
+  }.getOrElse(0)
+
+  /** Returns the number of bases that are clipped at the end of the sequence. */
+  def trailingClippedBases: Int = trailingHardClippedBases + trailingSoftClippedBases
 
   /** Truncates the cigar based on either query or target length cutoff. */
   private def truncate(len: Int, shouldCount: CigarElem => Boolean): Cigar = {
--- a/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
+++ b/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
@@ -91,33 +91,71 @@ object Cigar {
 }
 
 /** A data class for holding statistics about a [[Cigar]]. */
-private[alignment] case class CigarStats(lengthOnQuery: Int, lengthOnTarget: Int, alignedBases: Int, clippedBases: Int)
+private[alignment] case class CigarStats(lengthOnQuery: Int,
+                                         lengthOnTarget: Int,
+                                         alignedBases: Int,
+                                         clippedBases: Int,
+                                         softClippedFivePrimeBases: Int,
+                                         softClippedThreePrimeBases: Int)
 
 /** Companion object for [[Cigar]]. */
 private[alignment] object CigarStats {
 
   /** Build a [[CigarStats]] from a [[Cigar]]. */
   private[alignment] def apply(cigar: Cigar): CigarStats = {
-    var lengthOnQuery  = 0
-    var lengthOnTarget = 0
-    var alignedBases   = 0
-    var clippedBases   = 0
+    var lengthOnQuery              = 0
+    var lengthOnTarget             = 0
+    var alignedBases               = 0
+    var clippedBases               = 0
+    var softClippedFivePrimeBases  = 0
+    var softClippedThreePrimeBases = 0
+    var index                      = 0
 
-    forloop(0, cigar.elems.length) { index =>
+    // leading clipping
+    while (index < cigar.elems.length && cigar.elems(index).operator.isClipping) {
       val elem = cigar.elems(index)
       lengthOnQuery  += elem.lengthOnQuery
       lengthOnTarget += elem.lengthOnTarget
-      if (elem.operator.isAlignment)     alignedBases += elem.length
-      else if (elem.operator.isClipping) clippedBases += elem.length
+      clippedBases += elem.length
+      if (elem.operator == CigarOperator.S) softClippedFivePrimeBases += elem.length
+      index += 1
     }
 
+    // aligned bases
+    while (index < cigar.elems.length && !cigar.elems(index).operator.isClipping) {
+      val elem = cigar.elems(index)
+      lengthOnQuery  += elem.lengthOnQuery
+      lengthOnTarget += elem.lengthOnTarget
+      if (elem.operator.isAlignment) alignedBases += elem.length
+      index += 1
+    }
+
+    // trailing clipping
+    while (index < cigar.elems.length && cigar.elems(index).operator.isClipping) {
+      val elem = cigar.elems(index)
+      lengthOnQuery  += elem.lengthOnQuery
+      lengthOnTarget += elem.lengthOnTarget
+      clippedBases += elem.length
+      if (elem.operator == CigarOperator.S) softClippedThreePrimeBases += elem.length
+      index += 1
+    }
+
+    require(index == cigar.elems.length,
+      s"Invalid cigar, did not consume all the elements: ${Cigar(cigar.elems.take(index))}[${Cigar(cigar.elems.drop(index))}]"
+    )
+
     CigarStats(
-      lengthOnQuery  = lengthOnQuery,
-      lengthOnTarget = lengthOnTarget,
-      alignedBases   = alignedBases,
-      clippedBases   = clippedBases
+      lengthOnQuery              = lengthOnQuery,
+      lengthOnTarget             = lengthOnTarget,
+      alignedBases               = alignedBases,
+      clippedBases               = clippedBases,
+      softClippedFivePrimeBases  = softClippedFivePrimeBases,
+      softClippedThreePrimeBases = softClippedThreePrimeBases
     )
   }
+
+  /** Build a [[CigarStats]] from a string representation of the cigar. */
+  private[alignment] def apply(cigar: String): CigarStats = this.apply(Cigar(cigar))
 }
 
 /**
@@ -162,6 +200,12 @@ case class Cigar(elems: IndexedSeq[CigarElem]) extends Iterable[CigarElem] {
 
   /** Returns the number of bases that are clipped between the two sequences. */
   def clippedBases: Int = stats.clippedBases
+
+  /** Returns the number of bases that are soft-clipped on the 5' end of this sequence.  Ignores hard-clips. */
+  def softClippedFivePrimeBases: Int = stats.softClippedFivePrimeBases
+
+  /** Returns the number of bases that are soft-clipped on the 3' end of this sequence.  Ignores hard-clips. */
+  def softClippedThreePrimeBases: Int = stats.softClippedThreePrimeBases
 
   /** Truncates the cigar based on either query or target length cutoff. */
   private def truncate(len: Int, shouldCount: CigarElem => Boolean): Cigar = {
@@ -297,21 +341,21 @@ case class Alignment(query: Array[Byte],
 
       cig.operator match {
         case CigarOperator.INSERTION =>
-          forloop (from=0, until=len) { i =>
+          forloop (from=0, until=len) { _ =>
             queryBuffer.append(this.query(qOffset).toChar)
             alignBuffer.append(gapChar)
             targetBuffer.append(padChar)
             qOffset += 1
           }
         case CigarOperator.DELETION =>
-          forloop (from=0, until=len) { i =>
+          forloop (from=0, until=len) { _ =>
             queryBuffer.append(padChar)
             alignBuffer.append(gapChar)
             targetBuffer.append(this.target(tOffset).toChar)
             tOffset += 1
           }
         case op =>
-          forloop (from=0, until=len) { i =>
+          forloop (from=0, until=len) { _ =>
             val q = this.query(qOffset).toChar
             val t = this.target(tOffset).toChar
             queryBuffer.append(q)
--- a/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
@@ -79,7 +79,7 @@ class ClipBam
   @arg(          doc="Require at least this number of bases to be clipped on the 5' end of R2") val readTwoFivePrime: Int  = 0,
   @arg(          doc="Require at least this number of bases to be clipped on the 3' end of R2") val readTwoThreePrime: Int = 0,
   @arg(          doc="Clip overlapping reads.", mutex=Array("clipBasesPastMate")) val clipOverlappingReads: Boolean = false,
-  @arg(          doc="Clip reads who sequence past the start of their mate.", mutex=Array("clipOverlappingReads")) val clipBasesPastMate: Boolean = false
+  @arg(          doc="Clip reads in FR pairs that sequence past the far end of their mate.", mutex=Array("clipOverlappingReads")) val clipBasesPastMate: Boolean = false
 ) extends FgBioTool with LazyLogging {
   Io.assertReadable(input)
   Io.assertReadable(ref)
--- a/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
@@ -189,7 +189,11 @@ class ClipBam
     }
 
     val (numExtendingPastMateStartReadOne, numExtendingPastMateStartReadTwo) = {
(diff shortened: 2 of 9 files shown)


src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -245,11 +245,36 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
     // end of the read, or the end of the insert, whichever is shorter!
     val len = {
       var index = if (!rec.isFrPair) trimToLength - 1 else {
-        // Adjust the insert size based on clipping on the 5' end of the given read.
-        // This *does not* consider any clipping on the 5' end of its mate.
-        val fivePrimeClipping  = cigar.takeWhile(_.operator.isClipping).filter(_.operator == CigarOperator.S).sumBy(_.length)
-        val adjustedInsertSize = abs(rec.insertSize) + fivePrimeClipping
```
nh13, line 251: tl;dr: using the insert size does not account for inserted bases!

src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -532,6 +532,20 @@ class VanillaUmiConsensusCallerTest extends UnitSpec with OptionValues {
     s2.cigar.toString shouldBe "46S96M"
   }
 
+  it should "not to trim based on insert size in the presence of soft-clipping (+/-)" in {
```
nh13, line 535: this is just swapping R1 and R2 from the previous test

src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -560,4 +574,18 @@ class VanillaUmiConsensusCallerTest extends UnitSpec with OptionValues {
     source.cigar.toString() shouldBe "3M"
     source.sam shouldBe Some(rec)
   }
+
+  it should "not trim reads with an insert size shorter than the read length when there's an insertion in the middle" in {
```
nh13, line 578: this is the new test to confirm the new behavior.  Previously, the insert size would be calculated as 80bp, so the consensus read would have been 80bp.  This happened because the insert size doesn't count inserted bases in the insert.  The correct length of the consensus bases is 100bp.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -245,11 +245,36 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
...
-        min(adjustedInsertSize, trimToLength) - 1
+        // If this read reads through the template (i.e. past the start of the mate), then we need to trim bases on the
+        // 3' end.  If not, then we can use the full read. This *does not* consider any clipping on the 5' end of its mate.
+        val adjustedReadLength = {
+          if (rec.positiveStrand) {
+            val mateEnd = rec.mateEnd.getOrElse(rec.start + abs(rec.insertSize) - 1)
```
nh13, line 252: If MC isn't present, this relies on the aligner setting insert size correctly

tfenne: Use `CoordMath.getEnd` here?

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -245,11 +245,36 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
...
+            if (rec.end >= mateEnd) {
+              // the # of bases to keep is the last read base of where the mate ends
+              rec.readPosAtRefPos(pos=mateEnd, returnLastBaseIfDeleted=false)
+            } else {
+              // Find where the end of the read goes past the mate start, if at all
+              val refPosAtReadEnd  = min(mateEnd, rec.end) // last 3' base
```
tfenne, line 258: Why is this needed?  We're in an else under `rec.end >= mateEnd` therefore rec.end will _always_ be <= mateEnd and therefore the value returned.   Am I missing something?

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -245,11 +245,36 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
...
+              rec.readPosAtRefPos(pos=refPosAtReadEnd, returnLastBaseIfDeleted=false)
```
tfenne, line 259: It's a while since I've been through this code, but this will remove any trailing clipping, insertions, and maybe deletions too.  Is that intentional?  I.e. if the cigar is `80M20S`, say because the read runs off into a large insertion, this will return `80`, and cause the read to get trimmed to 80 bases, no?

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -245,11 +245,36 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
...
+            }
+          } else {
+            // Find where the start of the read that goes before the mate start, if at all
```
tfenne, line 262:
```suggestion
            // Find where the start of the read goes before the mate start, if at all
```
Just saying out loud that I find these comments hard to grok because they use start/end as then are in the SAM file, as opposed to e.g. if you're looking at reads in IGV (where directionality/strand is conveyed) I think we tend to flip start/end for reverse reads to match sequencing order.

tfenne (changes requested):
I think I might need to be walked through this PR.  I understand the bug/goal, but have questions about the implementation.

I don't understand why the `else` branches are the way they are.  It seems to be that they will now _always_ remove 3' soft-clipping and I don't think that's desirable.  I think there are three conceptual cases we need to walk through and make sure we agree on the expected outcome:

1. A read in an FR pair is aligned such that it has _aligned_ bases that extend past the "far" end of it's mate.   This is simple, we know we want to trim any bases past the mate's far end.
2. A read in an FR pair is aligned such that it _doesn't_ have aligned bases that extend past the "far" end of it's mate, but is close and has soft-clipping that if converted to an `M` cigar, would put it past it's mate end.  I'm not sure what to do in this case - the old code would have trimmed, but I'm not entirely sure that's the right thing to do.
3. Reads are far apart (e.g. 2x150 with 350bp insert size) but have soft-clipping at the 3' end.  In this case I believe we do not want to remove the clipping.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
nh13: I think this should be `min(mateStart, rec.end)` based on the comment

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -233,7 +234,7 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
   /**
     * Ensures that there are at least clipLength bases clipped at the 5' end
     * of the read, _including_ any existing soft and hard clipping.  Calculates any
-    * additional clipping and delegates to [com.fulcrumgenomics.bam.SamRecordClipper.[clipStartOfAlignment]].
+    * additional clipping and delegates to [[com.fulcrumgenomics.bam.SamRecordClipper.clipStartOfAlignment]].
```
nh13, line 237: doc fix

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -162,7 +163,7 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
       val oldElems = rec.cigar.elems.reverse
       val oldBases = rec.bases.reverse
       val oldQuals = rec.quals.reverse
-      val (newElems, readBasesClipped, refBasesClipped, bases, quals) = clip(oldElems, numberOfBasesToClip, oldBases, oldQuals)
+      val (newElems, readBasesClipped, _, bases, quals) = clip(oldElems, numberOfBasesToClip, oldBases, oldQuals)
```
nh13, line 166: intellij warning

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -289,6 +290,96 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
     }
   }
 
+  /** Clips overlapping read pairs, where both ends of the read pair are mapped to the same chromosome, and in FR orientation.
```
nh13, line 293: moved from `ClipBam`

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -289,6 +290,96 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+    *
+    * @param rec the read
+    * @param mate the mate
+    * @return the number of overlapping bases to that were clipped on the record and mate respectively (3' end in sequencing order)
+    */
+  def clipOverlappingReads(rec: SamRecord, mate: SamRecord): (Int, Int) = {
+    if (rec.matesOverlap.contains(false)) (0, 0) // do not overlap, don't clip
+    else {
+      require(rec.isFrPair, "Only know how to clip FR overlapping read pairs")
```
nh13, line 302: I can be convinced to return `(0, 0)` versus requiring here

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -289,6 +290,96 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+        (numOverlappingBasesReadOne, numOverlappingBasesReadTwo)
+      }
+    }
+  }
+
+  /** Clips a read that sequences past the start (in sequencing order) of the mate, and in FR orientation.
+    *
+    * @param rec the read
+    * @param mate the mate
+    * @param doNotClip do not clip the reads, only return the number of bases that would be clipped
```
nh13: I could be convinced to update all the methods to have a mode where they don't actually clip, but just return how many additional bases _they would clip_.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -440,7 +531,7 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
               unreachable(s"Cigar $es doesn't contain soft clipping at the start")
           }
 
-          elems = newElems.result
+          elems = newElems.result()
```
nh13, line 565: scala warning

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -168,6 +170,9 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
   /** A consensus caller used to generate consensus UMI sequences */
   private val consensusBuilder = new SimpleConsensusCaller()
 
+  /** Clipper utility used to _calculate_ clipping, but not do the actual clipping */
+  private val clipper = new SamRecordClipper(mode=ClippingMode.Soft, autoClipAttributes=true)
```
nh13, line 174: would be nice to have mode that is like "dry run", but wouldn't want that on the command line I don't think (e.g. in `ClipBam`).  Not sure how to prevent that.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -241,16 +246,27 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
...
     val len = {
       var index = if (!rec.isFrPair) trimToLength - 1 else {
-        // Adjust the insert size based on clipping on the 5' end of the given read.
-        // This *does not* consider any clipping on the 5' end of its mate.
-        val fivePrimeClipping  = cigar.takeWhile(_.operator.isClipping).filter(_.operator == CigarOperator.S).sumBy(_.length)
-        val adjustedInsertSize = abs(rec.insertSize) + fivePrimeClipping
-        min(adjustedInsertSize, trimToLength) - 1
+        // Get the number of mapped bases to clip that maps beyond the mate's end.  Soft-clipped bases are not counted.
+        val numBasesExtendingPasteMateEnd = this.clipper.clipExtendingPastMateEnd(rec=rec, doNotClip=true)
+        if (numBasesExtendingPasteMateEnd == 0) trimToLength - 1
```
nh13: this matches the previous behavior.  We don't trim if only soft-clipped bases are "beyond the end of the mate"

src/test/scala/com/fulcrumgenomics/alignment/AlignmentTest.scala
```diff
@@ -281,3 +281,23 @@ class AlignmentTest extends UnitSpec {
     alignment.subByTarget(6, 11).cigar.toString() shouldBe "5D1="
   }
 }
+
+class CigarStatsTest extends UnitSpec {
```
nh13, line 285: we didn't have tests for these, tsk tsk

src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
```diff
@@ -524,4 +524,16 @@ class ClipBamTest extends UnitSpec with ErrorLogLevel with OptionValues {
       }
     }
   }
+
+  it should "clip FR reads that extend past the mate" in {
```
nh13, line 528: `SamRecordClipper` has nice tests, so just wrote the minimal test here to ensure it works in `ClipBam`.

src/test/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmisTest.scala
```diff
@@ -211,7 +211,6 @@ class AnnotateBamWithUmisTest extends UnitSpec {
     Io.writeLines(reverseFq, Io.readLines(fq).grouped(4).toSeq.reverse.flatten)
     val annotator = new AnnotateBamWithUmis(input=sam, fastq=Seq(fq, reverseFq), readStructure=Seq(ReadStructure("2M4B+M"), ReadStructure("1B+M")), output=out, attribute=umiTag, sorted=true)
     val result = intercept[Exception] { annotator.execute() }
-    println(result)
```
nh13, line 214: there's one, set for stun

src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -513,25 +513,39 @@ class VanillaUmiConsensusCallerTest extends UnitSpec with OptionValues {
...
 
-    s1.baseString shouldBe "T"*142
+    s1.baseString shouldBe "T"*142 // all the bases remain, no clipping
     s1.cigar.toString shouldBe "23S72M47S"  // NB: cigar is reversed in toSourceRead
-    s2.baseString shouldBe "A"*142
+    s2.baseString shouldBe "A"*142 // all the bases remain, no clipping
     s2.cigar.toString shouldBe "46S96M"
   }
 
+  it should "not trim based on insert size in the presence of soft-clipping (+/-)" in {
```
nh13, line 535: tests mirrors the previous one, but swaps R1 and R2

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -91,33 +91,71 @@ object Cigar {
 }
 
 /** A data class for holding statistics about a [[Cigar]]. */
-private[alignment] case class CigarStats(lengthOnQuery: Int, lengthOnTarget: Int, alignedBases: Int, clippedBases: Int)
+private[alignment] case class CigarStats(lengthOnQuery: Int,
+                                         lengthOnTarget: Int,
+                                         alignedBases: Int,
+                                         clippedBases: Int,
+                                         softClippedFivePrimeBases: Int,
+                                         softClippedThreePrimeBases: Int)
```
tfenne, line 99: The way these are calculated, they are not 5' and 3', they are "leading" and "trailing" or "starting" and "ending".

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -78,13 +78,14 @@ class ClipBam
   @arg(          doc="Require at least this number of bases to be clipped on the 3' end of R1") val readOneThreePrime: Int = 0,
   @arg(          doc="Require at least this number of bases to be clipped on the 5' end of R2") val readTwoFivePrime: Int  = 0,
   @arg(          doc="Require at least this number of bases to be clipped on the 3' end of R2") val readTwoThreePrime: Int = 0,
-  @arg(          doc="Clip overlapping reads.") val clipOverlappingReads: Boolean = false
+  @arg(          doc="Clip overlapping reads.", mutex=Array("clipBasesPastMate")) val clipOverlappingReads: Boolean = false,
+  @arg(          doc="Clip reads who sequence past the start of their mate.", mutex=Array("clipOverlappingReads")) val clipBasesPastMate: Boolean = false
```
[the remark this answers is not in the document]
nh13: I prefer the mutex, since one is a subset of the other.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -289,6 +290,127 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+
+  /** Returns the number of bases extending past the mate end for FR pairs including any soft-clipped bases, zero otherwise.
+    *
+    * The mate end is computed via the mate-cigar (MC) SAM tag if present, otherwise the reported insert size is used.
+    *
+    * @param rec the record to clip
+    */
+  def numBasesExtendingPastMateEnd(rec: SamRecord): Int = {
+    val mateEnd = rec.mateEnd.getOrElse(rec.start + abs(rec.insertSize) - 1)
+    numBasesExtendingPastMateEnd(rec=rec, mateEnd=mateEnd)
```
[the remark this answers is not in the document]
tfenne: I think the thing that's confusing to me is that you're calling a method called `numBasesExtendingPastMateEnd` and passing it `mateEnd`, except you're using `mateEnd` in two different ways.  For the parameter you really mean "end" as in `SamRecord.end` or it's equivalent, and in the method name you mean `the end that's the start in sequencing order".

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -168,6 +170,9 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
   /** A consensus caller used to generate consensus UMI sequences */
   private val consensusBuilder = new SimpleConsensusCaller()
 
+  /** Clipper utility used to _calculate_ clipping, but not do the actual clipping */
```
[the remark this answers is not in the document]
tfenne: Yeah, I got there in the end.

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
```diff
@@ -31,13 +31,27 @@ import com.fulcrumgenomics.testing.SamBuilder.{Minus, Plus, Strand}
...
-    new SamBuilder(readLength=len).addFrag(start=start, cigar = cigar, strand=strand, attrs=attrs).get
+    new SamBuilder(readLength=len).addFrag(start=start, cigar=cigar, strand=strand, attrs=attrs).value
+  }
+
+  /** Returns a pair of SAM records (read pair) with the respective starts / cigars / strands requested. */
+  def pair(start1: Int, cigar1: String, strand1: Strand = Plus, start2: Int, cigar2: String, strand2: Strand = Minus): (SamRecord, SamRecord) = {
+    val cig1 = TextCigarCodec.decode(cigar1)
+    val len1 = cig1.getReadLength
+    val cig2 = TextCigarCodec.decode(cigar2)
+    val len2 = cig2.getReadLength
```
tfenne, line 49:
```suggestion
    val len1 = Cigar(cig1).lengthOnQuery
    val len2 = Cigar(cig2).lengthOnQuery
```

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
```diff
@@ -31,13 +31,28 @@ import com.fulcrumgenomics.testing.SamBuilder.{Minus, Plus, Strand}
...
+    len1 shouldBe len2
+    new SamBuilder(readLength=len1)
+      .addPair(start1=start1, cigar1=cigar1, strand1=strand1, start2=start2, cigar2=cigar2, strand2=strand2) match {
+      case Seq(r1, r2) => (r1, r2)
+      case _ => throw new IllegalArgumentException("Bug")
```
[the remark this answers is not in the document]
tfenne: I think you can just leave off the default case and it'll explode on it's own if untrue.  Or if you really want the default case perhaps use `unreachable()` to make it clear it's never supposed to trigger?

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
```diff
@@ -551,4 +565,184 @@ class SamRecordClipperTest extends UnitSpec {
...
+    clipper(Soft).clipOverlappingReads(rec=rec, mate=mate) shouldBe (64, 75)
+    rec.start shouldBe 50
+    rec.end shouldBe 85
+    rec.cigar.toString shouldBe "36M64S"
+    mate.start shouldBe 86
+    mate.end shouldBe 120
+    mate.cigar.toString shouldBe "75S15M10D10M"
+  }
+
+  it should "clip reads that extend past each other with bpth read having deletions" in {
```
tfenne, line 673:
```suggestion
  it should "clip reads that extend past each other with both read having deletions" in {
```

tfenne (approved):

PR #767, opened 2022-02-05
More ergonomic methods for converting between HTSJDK and fgbio SequenceDictionary objects.

tfenne:
I find it really cumbersome to convert between the fgbio and the HTSJDK sequence dictionary types.  This PR attempts to fix this.

For conversions from the fgbio types to HTSJDK types there is now just a `.toSam` function directly on `SequenceMetadata` and `SequenceDictionary`.  I don't believe there is any reason to hide these behind implicits - it makes them substantially harder to discover and use, for no real benefit.  I have named them`.toSam` to imply that they do a conversion, not `.asSam` as we use in `SamRecord` because that implies (to me) that it's a cast, which it _is_ in `SamRecord` but is not here.

For conversions from the HTSJDK types to scala types I've added `apply()`methods to the fgbio types that just take the HTSJDK types.  I've also modified the implicits to add a `toScala` as I think that is more natural to type:

```scala
val dict = intervalList.dict.toScala // get the sequence dictionary and make a scala one, vs.
val d2   = intervalList.dict.fromSam // what does this mean exactly?
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala
@@ -108,6 +108,18 @@ object SequenceMetadata {
     )
   }
 
+  /** Generates a new SequenceMetadata from a SAMSequenceRecord. */
+  def apply(rec: SAMSequenceRecord): SequenceMetadata = {
+    val attributes: Map[String, String] = rec.getAttributes.map { entry => entry.getKey -> entry.getValue }.toMap
+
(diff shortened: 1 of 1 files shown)


nh13 (approved):
I guess the message I'm getting is there's almost never a reason to use implicits.

tfenne:
I think that's true - I think implicits in `FgBioDef` are useful, where you can just pull them in with `import FgBioDef._`, but implicits you have to go looking for and import specifically are harder to find (and have IDEs auto-import) than explicit methods.

PR #768, opened 2022-02-06
Make GroupReadsByUmi more permissive in the alignments it accepts.

tfenne:
@nh13 I think this does what we want.  I fixed up one test that was failing as a result, but have not added more tests.  I'd like to verify it works as intended on real data before writing the unit tests, otherwise the implementation might change before then.

I think this change might be enough to warrant a 2.0 release as I would like the behavior to be default or even non-optionally on to pass through both inter-contig and half-mapped mate pairs.  I wonder too if we should also drop the default `minMapQ` to 1 from 30?

I also enabled (trivially) threading in `CallMolecularConsensusReads` while I was there as I was testing consensus calling and got tired of waiting.

@clintval I'd be very curious on your take on this PR.  Our goal is to enable SV calling off the consensus BAM by allowing through any reads that might usefully support an SV including cross-chromosomal mappings _and_ read pairs with only one of the two mates mapped.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -65,6 +65,13 @@ class TransientAttrs(private val rec: SamRecord) {
     case null => default
     case value => value.asInstanceOf[A]
   }
+  def getOrElseUpdate[A](key: Any, default: => A): A = rec.asSam.getTransientAttribute(key) match {
+    case null  =>
+      val value: A = default
+      update(key, value)
+      value
+    case value => value.asInstanceOf[A]
+  }
 }
 
 /**
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -227,7 +227,7 @@ trait SamRecord {
   @inline final def remove(name: String): Unit                   = setAttribute(name, null)
 
   // transient attributes
-  @inline final def transientAttrs: TransientAttrs = new TransientAttrs(this)
+  @inline final val transientAttrs: TransientAttrs = new TransientAttrs(this)
 
   // TODO long-term: replace these two methods with methods on [[Cigar]] to save creating alignment blocks in memory
   @inline final def refPosAtReadPos(pos: Int) = getReferencePositionAtReadPosition(pos)
--- a/src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
@@ -27,7 +27,7 @@ package com.fulcrumgenomics.umi
 
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.bam.api.{SamOrder, SamSource, SamWriter}
-import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioMain, FgBioTool}
 import com.fulcrumgenomics.commons.io.Io
 import com.fulcrumgenomics.commons.util.{LazyLogging, LogLevel, Logger}
 import com.fulcrumgenomics.sopt._
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -65,9 +65,22 @@ object GroupReadsByUmi {
   private val ReadInfoTempAttributeName = "__GRBU_ReadInfo"
 
   /** A case class to represent all the information we need to order reads for duplicate marking / grouping. */
-  case class ReadInfo(refIndex: Int, start1: Int, start2: Int, strand1: Boolean, strand2: Boolean, library: String)
+  case class ReadInfo(refIndex1: Int,
+                      start1: Int,
+                      strand1: Byte,
+                      refIndex2: Int,
+                      start2: Int,
+                      strand2: Byte,
+                      library: String)
 
   object ReadInfo {
+    // Use the Max possible value for ref/pos/strand when we have unmapped reads to mirror what's done
+    // in the TemplateCoordinate sort order where we sort by the _lower_ of the two mates' positions
+    // and want to use the mapped position when one mate is unmapped
+    private val UnknownRef    = Int.MaxValue
+    private val UnknownPos    = Int.MaxValue
+    private val UnknownStrand = Byte.MaxValue
+
     /** Looks in all the places the library name can be hiding. Returns the library name
       * if one is found, otherwise returns "unknown".
       */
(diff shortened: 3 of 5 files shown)


src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -227,7 +227,7 @@ trait SamRecord {
   @inline final def remove(name: String): Unit                   = setAttribute(name, null)
 
   // transient attributes
-  @inline final def transientAttrs: TransientAttrs = new TransientAttrs(this)
+  @inline final val transientAttrs: TransientAttrs = new TransientAttrs(this)
```
tfenne, line 230: This is unrelated, but IntelliJ kept getting confused and thinking uses of `rec.transientAttrs[String]("foo")` were errors, and this is likely better anyway.

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -143,7 +145,7 @@ class CallMolecularConsensusReads
 
     // The output file is unmapped, so for now let's clear out the sequence dictionary & PGs
     val outHeader = UmiConsensusCaller.outputHeader(in.header, readGroupId, sortOrder)
-    val out       = SamWriter(output, outHeader, sort=sortOrder)
+    val out       = SamWriter(output, outHeader, sort=sortOrder, async=threads > 1)
```
nh13, line 148: If someone uses `fgbio --async-io=true CallMolecularConsensusReads --threads 1` then this wont be async.  I think that's an issue.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -83,29 +93,38 @@ object GroupReadsByUmi {
...
+          new ReadInfo(ref2, start2, strand2, ref1, start1, strand1, lib)
         }
 
         rec.transientAttrs(GroupReadsByUmi.ReadInfoTempAttributeName, result)
         result
       }
     }
+
+    // Encodes a known strand into a byte value
+    @inline private def strandToByte(positive: Boolean): Byte = if (positive) 1 else 0
```
nh13, line 127: previously, we stored a boolean for strand, which was whether the read was on the _negative_ strand (true if so, false if on positive).  Since `false < true`, that means that positive would be earlier than negative, all else being equal.  I think positive being `1` and negative being `0` flips that. 
```suggestion
    @inline private def strandToByte(positive: Boolean): Byte = if (positive) 0 else 1
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -425,15 +444,16 @@ class GroupReadsByUmi
...
                        |otherwise discard reads with UMIs shorter than this length and allow for differing UMI lengths.
                        |""")
   val minUmiLength: Option[Int] = None,
-  @arg(flag='x', doc= """Allow read pairs with primary alignments on different contigs to be grouped when using the
-                       |paired assigner (otherwise filtered out).""")
-  val allowInterContig: Boolean = false
+  @arg(flag='x', doc= """
+                       |DEPRECATED: this option will be removed in future versions and inter-contig reads will be
+                       |automatically processed.""")
+  val allowInterContig: Boolean = true
```
nh13, line 456: agreed

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -425,15 +444,16 @@ class GroupReadsByUmi
   @arg(flag='T', doc="The output tag for UMI grouping.") val assignTag: String = "MI",
   @arg(flag='m', doc="Minimum mapping quality.")         val minMapQ: Int      = 30,
```
nh13, line 445: Changed both the doc and default value:
```suggestion
  @arg(flag='m', doc="Minimum mapping quality for mapped reads.")         val minMapQ: Int      = 1,
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: FWIW ReadInfo doesn't have any part in the _sorting_ of the reads, it's only used to compare reads with each other and decide if they have the same "ends" and should have their UMIs looked at together.
That said, it doesn't matter which way around it is, so I'll commit the change.

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
tfenne: Added a check to see if `SamSource.UseAsync` was on as well.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -83,29 +93,38 @@ object GroupReadsByUmi {
         tmp
```
clintval, line 93: Scala doc references "ReadEnds" but while you're here you could fix that.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -83,29 +93,38 @@ object GroupReadsByUmi {
...
+          (ref1 == ref2 && start1 < start2) ||
+          (ref1 == ref2 && start1 == start2 && strand1 < strand2)
+
+
+        val result = if (earlier) {
+          new ReadInfo(ref1, start1, strand1, ref2, start2, strand2, lib)
         }
         else {
-          new ReadInfo(mateChrom, matePos, recPos, mateNeg, recNeg, lib)
+          new ReadInfo(ref2, start2, strand2, ref1, start1, strand1, lib)
```
clintval, line 118: Since you have duplicate types (many integers) I tend to do by-name parameterization to protect from accidentally swapping arguments.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -83,29 +93,38 @@ object GroupReadsByUmi {
...
-          (neg, pos)
+        val lib = library(rec)
+
+        val (ref1, start1, strand1) = if (rec.unmapped) (UnknownRef, UnknownPos, UnknownStrand) else {
+          val start = if (rec.positiveStrand) rec.unclippedStart else rec.unclippedEnd
+          (rec.refIndex, start, strandToByte(rec.positiveStrand))
+        }
+
+        val (ref2, start2, strand2) = if (rec.unpaired || rec.mateUnmapped) (UnknownRef, UnknownPos, UnknownStrand) else {
+          val start = if (rec.matePositiveStrand) SAMUtils.getMateUnclippedStart(rec.asSam) else SAMUtils.getMateUnclippedEnd(rec.asSam)
```
clintval, line 104: Worth it to add helpers on `SamRecord`? In parity with:

https://github.com/fulcrumgenomics/fgbio/blob/7a91c97350b308c35d054d5080a42482d0eff12d/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala#L151-L152

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -423,17 +442,18 @@ class GroupReadsByUmi
...
+  @arg(flag='s', doc="The UMI assignment strategy.")     val strategy: Strategy,
   @arg(flag='e', doc="The allowable number of edits between UMIs.") val edits: Int = 1,
   @arg(flag='l', doc= """The minimum UMI length. If not specified then all UMIs must have the same length,
                        |otherwise discard reads with UMIs shorter than this length and allow for differing UMI lengths.
                        |""")
   val minUmiLength: Option[Int] = None,
-  @arg(flag='x', doc= """Allow read pairs with primary alignments on different contigs to be grouped when using the
-                       |paired assigner (otherwise filtered out).""")
-  val allowInterContig: Boolean = false
+  @arg(flag='x', doc= """
```
clintval, line 453: Would you be also willing to add a formal `@deprecated` scala tag to this param?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -423,17 +442,18 @@ class GroupReadsByUmi
   @arg(flag='f', doc="Optional output of tag family size counts.") val familySizeHistogram: Option[FilePath] = None,
   @arg(flag='t', doc="The tag containing the raw UMI.")  val rawTag: String    = "RX",
   @arg(flag='T', doc="The output tag for UMI grouping.") val assignTag: String = "MI",
-  @arg(flag='m', doc="Minimum mapping quality.")         val minMapQ: Int      = 30,
+  @arg(flag='m', doc="Minimum mapping quality for mapped reads.")         val minMapQ: Int      = 1,
```
clintval, line 445: Maybe drop the mapped reads qualifier and drop down to 0? (https://github.com/samtools/htsjdk/blob/7ac95d5f7afca846f412cc8f4e6fd59422c48a71/src/main/java/htsjdk/samtools/SAMRecord.java#L125-L128)

clintval (approved):
I like it!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -83,29 +93,38 @@ object GroupReadsByUmi {
...
         }
 
-        val result = if (chrom < mateChrom || (chrom == mateChrom && (recPos < matePos || (recPos == matePos && !recNeg)))) {
-          new ReadInfo(chrom, recPos, matePos, recNeg, mateNeg, lib)
+        val earlier =
+          (ref1 < ref2) ||
+          (ref1 == ref2 && start1 < start2) ||
+          (ref1 == ref2 && start1 == start2 && strand1 < strand2)
+
+
```
nh13, line 113:
```suggestion
```

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -163,7 +165,8 @@ class CallMolecularConsensusReads
       rejects        = rej
     )
 
-    val iterator = new ConsensusCallingIterator(in.iterator, caller, Some(ProgressLogger(logger, unit=5e5.toInt)))
+    val progress = ProgressLogger(logger, unit=5e5.toInt)
```
nh13, line 168: Add `progress.logLast` after `out` is closed.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -65,9 +65,19 @@ object GroupReadsByUmi {
...
+                      start1: Int,
+                      strand1: Byte,
+                      refIndex2: Int,
+                      start2: Int,
+                      strand2: Byte,
+                      library: String)
 
   object ReadInfo {
+    private val UnknownRef    = Int.MaxValue
+    private val UnknownPos    = Int.MaxValue
```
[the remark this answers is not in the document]
nh13: I don't think we want sentinels, we want unknown ref/pos to be sorted _at the end_, so they need to have values larger than any other possible value.  I think the sentinel values you point us to are `-1`!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -469,9 +495,10 @@ class GroupReadsByUmi
     in.iterator
       .filter(r => !r.secondary && !r.supplementary)
       .filter(r => (includeNonPfReads || r.pf)                                      || { filteredNonPf += 1; false })
-      .filter(r => (r.mapped && (r.unpaired || r.mateMapped))                       || { filteredPoorAlignment += 1; false })
+      .filter(r => (r.mapped || (r.paired || r.mateMapped))                         || { filteredPoorAlignment += 1; false })
```
nh13, line 498: Somethings wrong here
1. I am a bit confused by the remaining braces here.  Do we need them since we have only OR logical relationships?  
2. if `r.paired` is `false`, then how can `r.mateMapped` every be true? 

I think you want "either the read is mapped, or it's paired and its mate is mapped"
```suggestion
      .filter(r => (r.mapped || (r.paired && r.mateMapped))                         || { filteredPoorAlignment += 1; false })
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -469,9 +495,10 @@ class GroupReadsByUmi
...
       .filter(r => (allowInterContig || r.unpaired || r.refIndex == r.mateRefIndex) || { filteredPoorAlignment += 1; false })
       .filter(r => mapqOk(r, this.minMapQ)                                          || { filteredPoorAlignment += 1; false })
+      // TODO: should we only filter UMIs with >= n Ns ?
```
nh13, line 501: remove
```suggestion
```

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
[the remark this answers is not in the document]
tfenne: What Nils said.  I'll add a comment above here to explain that.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Huh, turns out there are already `mateUnclippedStart` and `mateUnclippedEnd` functions there!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Good catch, thanks!

tfenne:
@nh13 Want to take one more look?  The only major change I made (beyond simple fixups) was to refactor how `ReadInfo` objects are created, and make a `ReadInfo.apply(Template)` and use that instead (the `ReadInfo.apply(SamRecord)` is used by CollectDuplexSeqMetrics, so also stays).  Change includes:

- Refactor of `ReadInfo.apply()`
- Addition of `TransientAttributes.getOrElseUpdate()`
- A test to make sure that when clipping is present the two `ReadInfo.apply()` methods give the same result.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -405,20 +439,11 @@ object Strategy extends FgBioEnum[Strategy] {
...
-    |(Note: the `MQ` tag is required on reads with mapped mates). By default, paired-end reads must have both reads
-    |mapped to the same chromosome (to turn off this filter, use `--allow-inter-contig`).
-    |
-    |This is done with the expectation that the next step is building consensus reads, where
-    |it is undesirable to either:
-    |
-    |   1. Assign reads together that are really from different source molecules
-    |   2. Build two groups from reads that are really from the same molecule
+    |During grouping, reads are filtered out if a) all reads with the same queryname are unmapped, b) any primary
+    |read has mapping quality < `min-map-q` (default=1), or c) the primary mappings for R1 and R1 are on different
```
nh13, line 443:
```suggestion
    |read has mapping quality < `min-map-q` (default=1), or c) the primary mappings for R1 and R2 are on different
```

nh13 (approved):
One minor typo, the LGTM!

PR #771, opened 2022-02-17
Only create the cache when it's going to be used since the constructo…

tfenne:
…r doesn't like it when you pass 0 to it!


nh13 (approved):
Sorry, this was my fault, thank-you for fixing!

PR #774, opened 2022-02-17
Reduce memory usage by GroupReadsByUmi in a corner case

tfenne:


src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -185,7 +185,10 @@ class CorrectUmis
         case Some(umi: String) =>
           val sequences = umi.split('-')
           if (sequences.exists(_.length != umiLength)) {
-            if (wrongLengthRecords == 0) logger.warning(s"Read (${rec.name}) detected with unexpected length UMI(s): ${sequences.mkString(" ")}")
+            if (wrongLengthRecords == 0) {
+              logger.warning(s"Read (${rec.name}) detected with unexpected length UMI(s): ${sequences.mkString(" ")}.")
+              logger.warning(s"Expected UMI length: ${umiLength}")
```
tfenne, line 190: Apologies for the unrelated change here.  I had a stupid type that took me far too long to figure out because I used `-u` instead of `-U` and `CorrectUmis` happily decided my filename was the sole UMI sequence to correct to.  It the message here had told me my expcted UMI length was 30+ that would have helped!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -259,13 +265,35 @@ object GroupReadsByUmi {
     /** String that is prefixed onto the UMI from the read with that maps to a higher coordinate in the genome.. */
     private[umi] val higherReadUmiPrefix: String = ("b" * (maxMismatches+1)) + ":"
 
+
+    /** Returns true if the two UMIs are the same. */
+    final override def isSameUmi(a: Umi, b: Umi): Boolean = {
+      if (a == b) true else {
+        val (a1, a2) = split(a)
```
nh13, line 272:
```suggestion
        // same as `a == reverse(b)` but more efficient than creating new strings
        val (a1, a2) = split(a)
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -259,13 +265,35 @@ object GroupReadsByUmi {
...
+      }
+    }
+
+    /** Returns the UMI with the lexically lower half first. */
+    override def canonicalize(u: Umi): Umi = {
+      val (a, b) = split(u)
+      if (a < b) u else s"${b}-${a}"
+    }
+
+    /** Splits the paired UMI into it's two parts. */
```
nh13, line 284:
```suggestion
    /** Splits the paired UMI into its two parts. */
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -441,6 +469,17 @@ class GroupReadsByUmi
 
   private val assigner = strategy.newStrategy(this.edits)
 
+  /** True if no differences in UMIs are tolerated and the Molecular ID tag is MI, false otherwise. True here enables
+    * an optimization where, when bringing groups of reads into memory, we can _also_ group by UMI thus
+    * reducing the number in memory.  This is helpful since edits=0 is often used for data that has
```
nh13, line 474:
```suggestion
    * reducing the number of reads in memory.  This is helpful since edits=0 is often used for data that has
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -441,6 +469,17 @@ class GroupReadsByUmi
...
+    * high numbers of reads with the same start/stop coordinates.
+    * We do this be setting the MI tag to the canonicalized, (optionally truncated) UMI prior to sorting, so that
```
nh13, line 476:
```suggestion
    * We do this by setting the MI tag to the canonicalized (optionally truncated) UMI prior to sorting, so that
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -480,7 +519,24 @@ class GroupReadsByUmi
           }
         } || { filterUmisTooShort += 1; false}
       }
-      .foreach(r => { sorter += r; kept += 1; sortProgress.record(r) })
+      .foreach { r =>
+        // If we're able to also group by the UMI because edits aren't allowed, push the trimmed, canonicalized UMI
+        // into the assign tag (which must be MI if cantakeNextGroupByUmi is true), since that is used by the
+        // SamOrder to sort the reads _and_ we'll overwrite it on the way out!
```
nh13, line 525: I'd emphasize that we may trim the UMI to the minimum length here for sorting, but later we re-parse the raw UMI based on the minimum length UMI in the group we'll consider, with that minimum length UMI perhaps being longer than the minimum length specified on the command line.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -531,11 +587,22 @@ class GroupReadsByUmi
...
-    first :: buffer.toList
+    val firstUmi  = first.r1.get.apply[String](this.assignTag)
+    val builder   = IndexedSeq.newBuilder[Template]
+    builder    += first
+
+    while (
+      iterator.hasNext &&
+      firstEnds == ReadInfo(iterator.head.r1.get) &&
+      // This last condition only works because we put a canonicalized UMI into rec(assignTag) if canTakeNextGroupByUmi
+      (!canTakeNextGroupByUmi || firstUmi == iterator.head.r1.get.apply[String](this.assignTag))
```
nh13, line 600: So this does change behavior in a subtle way.  Suppose we have three templates that have all but the same assign tag.  Let's say `AAAA`, `GGGGG`, `GGGGT`, with `--min-umi-length=3`.  

Previously, all templates would be read into memory, and `truncateUmis` would truncate to the length of smallest UMI observed in the group of templates, in this case 4bp long due to the `AAAA` (not 3 as per the command line!).  So the three templates would have UMIs `AAAA->AAAA`, `GGGGG->GGGG`, and `GGGGT->GGGG`.  So we'd assign two unique molecules (`AAAA` and `GGGG`, with the last molecule containing the last two reads).

In the new implementation, we truncate the raw UMI bases based on `--min-umi-length=3` to set `MI` for sorting.  So we'd truncate to length 3 for sorting: `AAAA->AAA`, `GGG->GGG`, and `GGG->GGG`.  When we read back in after sorting, we read in the first template by itself (only read with `MI:AAA`), so no truncation is applied and it stays the same (`AAAA`).  We then read in all templates with MI having `GGG`, which gives the second two templates.  Now we go back to the raw UMIs to find the length of the shortest UMI of the two.  Both are 5bp long, so we do not truncate and keep them the same (`GGGGG` and `GGGGT`).  But now these UMIs differ, so we get three molecules!

One _could_ argue that the new implementation is an improvement, but it does change behavior subtly.

nh13 (changes requested):
I think there's a subtle change to the behavior which we could think of as an improvement, but it does change the behavior, so we should discuss.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Done

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Hrm, that is an interesting point.  So basically if you have variable length single UMIs _and_ have edits = 0 the behavior will be subtly different.  My instinct is to call this an improvement and move on, but I don't have a great sense of who (or on what kind of data) the variable length support is used, so I'm not 100% sure.

PR #775, opened 2022-02-18
Add a tool to copy the UMI from the read name

nh13:


tfenne (approved):
A couple of thoughts:

1. How about moving the core 'move from read name into tag' into a method somewhere that we could then also plumb it into e.g. `GroupReadsByUmi` directly without forcing folks to run a separate tool?
2. I feel like we should do some validation of the UMIs that are extracted.  That could be as simple as checking that they only contains `ACGTN-` and either erroring or logging a warning.  I don't love that if my read names don't have UMIs in them, but are Illumina format read names, I'll end up with e.g. the `y` coordinate of the read in the RX tag and zero warnings that something unexpected has happened.

nh13:
@tfenne made most of the changes you suggested, with the exception of plumbing this into other tools

src/main/scala/com/fulcrumgenomics/umi/package.scala
```diff
@@ -0,0 +1,70 @@
...
+    *
+    * @param rec the record to modify
+    * @param removeUmi true to remove the UMI from the read name, otherwise only copy the UMI to the tag
+    * @param umiDelimiter if not None, replaces any delimiter found in the UMI with a dash ('-') as per the SAM spec
+    * @return the modified record
+    */
+  def copyUmiFromReadName(rec: SamRecord,
+                          removeUmi: Boolean = false,
+                          nameDelimiter: Char = ':',
+                          umiDelimiter: Option[Char] = Some('+')): SamRecord = {
```
tfenne, line 46: Is there a reason for the default being `+` - i.e. is it commonly used somewhere?  I'm not familiar.

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
```diff
@@ -0,0 +1,62 @@
...
+  """
+    |Copies the UMI at the end of the BAM's read name to the RX tag.
+    |
+    |The UMI is assumed to be at the end of the read name, with colon preceding it.
+  """)
+class CopyUmiFromReadName
+( @arg(flag='i', doc="The input BAM file") input: PathToBam,
+  @arg(flag='o', doc="The output BAM file") output: PathToBam,
+  @arg(doc="Remove the UMI from the read name") removeUmi: Boolean = false
+) extends FgBioTool with LazyLogging {
```
tfenne, line 45: Should we also expose the incoming UMI delimiter option here?

tfenne (approved):
A few minor comments - feel free to merge.

src/main/scala/com/fulcrumgenomics/umi/package.scala
nh13: This is how `bcl2fastq` delimits dual umis

src/main/scala/com/fulcrumgenomics/umi/package.scala
```diff
@@ -0,0 +1,70 @@
...
+ * THE SOFTWARE.
+ *
+ */
+
+package com.fulcrumgenomics
+
+import com.fulcrumgenomics.bam.api.SamRecord
+import htsjdk.samtools.util.SequenceUtil
+
+package object umi {
```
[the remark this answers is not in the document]
nh13: true, true.  Done

PR #776, opened 2022-02-19
Multi-thread CallMolecularConsensusReads

nh13:


tfenne (approved):
FWIW I already did something similar here: https://github.com/fulcrumgenomics/fgbio/pull/768

PR #778, opened 2022-02-19
ZippperBams: A replacement for picard's MergeBamAlignment

tfenne:
@nh13 not ready for a full on code review, but it would be helpful for you to take a look at this and see if you think there is any functionality we routinely use in MergeBamAlignment that is missing from this?  

The only thing I can think of so far is the clipping of read-pairs that extend past each other's starts.  I'm hesitate to do that here as it would then require fixing up tags like NM, MD, etc. that then also require either iterating reads in coordinate order to use the reference efficiently, or holding the entire genome in memory (the latter may not be so bad these days).

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
@@ -73,6 +73,11 @@ case class Template(r1: Option[SamRecord],
   /** Returns an iterator of all reads for the template. */
   def allReads: Iterator[SamRecord] = r1.iterator ++ r2.iterator ++ allSupplementaryAndSecondary
 
+  /** Returns an iterator of all read 1s for the template. */
+  def allR1s: Iterator[SamRecord] = r1.iterator ++ r1Secondaries.iterator ++ r1Supplementals.iterator
+
+  /** Returns an iterator of all read 2s for the template. */
+  def allR2s: Iterator[SamRecord] = r2.iterator ++ r2Secondaries.iterator ++ r2Supplementals.iterator
   /**
     * Produces a copy of the template that has had mapping information removed.  Discards secondary and supplementary
     * records, and retains the primary records after un-mapping them.  Auxillary tags listed in [[Bams.AlignmentTags]]
@@ -96,6 +101,19 @@ case class Template(r1: Option[SamRecord],
     }
 
     Template(x1, x2)
+  }
+
+  /** Fixes mate information and sets mate cigar on all primary and supplementary (but not secondary) records. */
+  def fixMateInfo(): Unit = {
+    for (primary <- r1; supp <- r2Supplementals) {
+      SamPairUtil.setMateInformationOnSupplementalAlignment(supp.asSam, primary.asSam, true)
+    }
+    for (primary <- r2; supp <- r1Supplementals) {
+      SamPairUtil.setMateInformationOnSupplementalAlignment(supp.asSam, primary.asSam, true)
+    }
+    for (first <- r1; second <- r2) {
+      SamPairUtil.setMateInfo(first.asSam, second.asSam)
+    }
   }
 
   /** The total count of records for the template. */
--- a/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
@@ -36,7 +36,6 @@ import htsjdk.samtools._
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker
 import htsjdk.samtools.util.{CloserUtil, CoordMath, SequenceUtil}
 
-import scala.collection.mutable.{ArrayBuffer, ListBuffer}
 import scala.math.{max, min}
 
 /**
@@ -112,7 +111,7 @@ case class Template(r1: Option[SamRecord],
       SamPairUtil.setMateInformationOnSupplementalAlignment(supp.asSam, primary.asSam, true)
     }
     for (first <- r1; second <- r2) {
-      SamPairUtil.setMateInfo(first.asSam, second.asSam)
+      SamPairUtil.setMateInfo(first.asSam, second.asSam, true)
     }
   }
 
--- a/src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
@@ -25,7 +25,6 @@
 package com.fulcrumgenomics.bam
 
 import com.fulcrumgenomics.FgBioDef._
-import com.fulcrumgenomics.bam.ZipperBams.{buildOutputHeader, checkSort, templateIterator}
 import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamSource, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.commons.async.AsyncIterator
@@ -37,18 +36,12 @@ import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
 import htsjdk.samtools.{SAMFileHeader, SAMSequenceDictionary}
 import htsjdk.variant.utils.SAMSequenceDictionaryExtractor
 
-/*
-Feature List
-
-    Trim back reads that are FR pairs that extend past each other's starts
- */
-
-object ZipperBams extends LazyLogging {
+/** Companion object that holds implementation methods for the ZipperBams command line tool. */
+private[bam] object ZipperBams extends LazyLogging {
   /** Class to hold info about all the extended tags/attrs we want to manipulate. */
   case class TagInfo(remove: IndexedSeq[String], reverse: IndexedSeq[String], revcomp: IndexedSeq[String]) {
     val setToReverse: java.util.Set[String] = remove.iterator.toJavaSet
     val setToRevcomp: java.util.Set[String] = revcomp.iterator.toJavaSet
-
     val hasRevsOrRevcomps: Boolean = reverse.nonEmpty || revcomp.nonEmpty
   }
 
@@ -60,6 +53,13 @@ object ZipperBams extends LazyLogging {
     }
   }
 
+  /** Builds the header for the output BAM from the headers of the unmapped and mapped BAM.
+    *
+    * @param unmapped the header from the unmapped BAM
+    * @param mapped the header from the mapped BAM
+    * @param sort optionally, the sort order specified for the output BAM
+    * @param dict the sequence dictionary of the reference genome to insert into the output header
+    */
   def buildOutputHeader(unmapped: SAMFileHeader,
                         mapped: SAMFileHeader,
                         sort: Option[SamOrder],
@@ -87,7 +87,16 @@ object ZipperBams extends LazyLogging {
     val recs      = source.iterator.bufferBetter
     val templates = new Iterator[Template] {
       override def hasNext: Boolean = recs.hasNext
-      override def next(): Template = Template(recs)
+      override def next(): Template = {
+        try {
+          Template(recs)
+        }
+        catch {
+          case iae: IllegalArgumentException => throw new IllegalStateException(
+            s"Error reading from ${source.toSamReader.getResourceDescription}: ${iae.getMessage}"
+          )
+        }
+      }
     }
 
     if (bufferSize > 0) {
@@ -110,16 +119,18 @@ object ZipperBams extends LazyLogging {
     for (u <- unmapped.primaryReads) {
       // Get all the mapped records for the unmapped record
       val ms = if (u.unpaired || u.firstOfPair) mapped.allR1s.toSeq else mapped.allR1s.toSeq
+
+      // Copy over pass/fail qc flag
+      ms.foreach(_.pf = u.pf)
+
       // Copy over tags on any positive strand reads first
       ms.iterator.filter(_.positiveStrand).foreach(m => copyTags(u, m))
 
-      // Then reverse complement the unmapped read if we need to
-      if (tagInfo.hasRevsOrRevcomps && ms.exists(_.negativeStrand)) {
-        u.asSam.reverseComplement(tagInfo.setToRevcomp, tagInfo.setToReverse, true)
+      // Then reverse complement the unmapped read if we need to and do the negative strand reads
+      if (ms.exists(_.negativeStrand)) {
+        if (tagInfo.hasRevsOrRevcomps) u.asSam.reverseComplement(tagInfo.setToRevcomp, tagInfo.setToReverse, true)
+        ms.iterator.filter(_.negativeStrand).foreach(m => copyTags(u, m))
       }
-
-      // And deal with the negative strand mappings
-      ms.iterator.filter(_.negativeStrand).foreach(m => copyTags(u, m))
     }
 
     mapped
@@ -153,9 +164,12 @@ class ZipperBams
   @arg(flag='u', doc="Unmapped SAM or BAM.") val unmapped: PathToBam,
   @arg(flag='r', doc="Path to the reference used in alignment.") val ref: PathToFasta,
   @arg(flag='o', doc="Output SAM or BAM file.") val output: PathToBam = Io.StdOut,
-  @arg(doc="Tags to remove from the mapped BAM records.", minElements=0) val tagsToRemove: IndexedSeq[String] = IndexedSeq.empty,
-  @arg(doc="Set of optional tags to reverse on reads mapped to the negative strand.", minElements=0) val tagsToReverse: IndexedSeq[String] = IndexedSeq.empty,
-  @arg(doc="Set of optional tags to reverse complement on reads mapped to the negative strand.", minElements=0) val tagsToRevcomp: IndexedSeq[String] = IndexedSeq.empty,
+  @arg(doc="Tags to remove from the mapped BAM records.", minElements=0)
+  val tagsToRemove: IndexedSeq[String] = IndexedSeq.empty,
+  @arg(doc="Set of optional tags to reverse on reads mapped to the negative strand.", minElements=0)
+  val tagsToReverse: IndexedSeq[String] = IndexedSeq.empty,
+  @arg(doc="Set of optional tags to reverse complement on reads mapped to the negative strand.", minElements=0)
+  val tagsToRevcomp: IndexedSeq[String] = IndexedSeq.empty,
   @arg(flag='s', doc="Sort the output BAM into the given order.") val sort: Option[SamOrder] = None,
   @arg(flag='b', doc="Buffer this many read-pairs while reading the input BAMs.") val buffer: Int = 5000
 ) extends FgBioTool {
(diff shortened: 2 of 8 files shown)


src/main/scala/com/fulcrumgenomics/cmdline/FgBioTool.scala
```diff
@@ -24,11 +24,11 @@
 package com.fulcrumgenomics.cmdline
 
 import com.fulcrumgenomics.cmdline.FgBioMain.FailureException
+import com.fulcrumgenomics.commons.util.LazyLogging
 import com.fulcrumgenomics.sopt.cmdline.ValidationException
 
 /** All fgbio tools should extend this. */
-// Todo: extend Lazy Logging?
-trait FgBioTool {
+trait FgBioTool extends LazyLogging {
```
nh13, line 31: ty

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+
+@clp(group=ClpGroups.SamOrBam, description=
+  """
+    |Zips together an unmapped and mapped BAM to transfer metadata into the output BAM.
+    |
+    |Both the unmapped and mapped BAMs _must_ be a) queryname sorted or grouped (i.e. all records with the same
+    |name are grouped together in the file), and b) have the same ordering of querynames.  If either of these are
+    |violated the output is undefined!
+    |
+    |By default the mapped BAM is read from stdin and the output BAM is written to stdout. This can be changed
```
nh13, line 141:
```suggestion
    |By default the mapped BAM is read from standard input (stdin) and the output BAM is written to standard output (stdout). This can be changed
```

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+    |using the `--input/-i` and `--output/-o` options.
+    |
+    |By default the output BAM file is emitted in the same order as the input BAMs.  This can be overridden
+    |using the `--sort` option, though in practice it may be faster to do the following:
+    |
+    |  `fgbio --compression 0 ZipperBams -i mapped.bam -u unmapped.bam -r ref.fa | samtools sort -@ 16`
```
nh13, line 147: Also, why not use a triple-back-tick code block?

```suggestion
    |  `fgbio --compression 0 ZipperBams -i mapped.bam -u unmapped.bam -r ref.fa | samtools sort -@ $(nproc)`
```

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+    |
+    |
```
nh13, line 149:
```suggestion
```

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+  """)
+class ZipperBams
```
nh13, line 151: This feels like it should interleave BAMs, like if I had a bam of just R1s and a bam of just R2s, it would give me R1 then R2 then R1 then R2...

The main purpose of this tool is for alignment, so how about a name with that in it? Perhaps we ask the twitterverse?

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+( @arg(flag='i', doc="Mapped SAM or BAM.") val input: PathToBam = Io.StdIn,
+  @arg(flag='u', doc="Unmapped SAM or BAM.") val unmapped: PathToBam,
+  @arg(flag='r', doc="Path to the reference used in alignment.") val ref: PathToFasta,
+  @arg(flag='o', doc="Output SAM or BAM file.") val output: PathToBam = Io.StdOut,
+  @arg(doc="Tags to remove from the mapped BAM records.", minElements=0) val tagsToRemove: IndexedSeq[String] = IndexedSeq.empty,
+  @arg(doc="Set of optional tags to reverse on reads mapped to the negative strand.", minElements=0) val tagsToReverse: IndexedSeq[String] = IndexedSeq.empty,
+  @arg(doc="Set of optional tags to reverse complement on reads mapped to the negative strand.", minElements=0) val tagsToRevcomp: IndexedSeq[String] = IndexedSeq.empty,
```
nh13, line 158: Any harm in putting in the usage our recommended values for duplex sequencing?

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+    val setToReverse: java.util.Set[String] = remove.iterator.toJavaSet
+    val setToRevcomp: java.util.Set[String] = revcomp.iterator.toJavaSet
+
+    val hasRevsOrRevcomps: Boolean = reverse.nonEmpty || revcomp.nonEmpty
+  }
+
+  /** Checks the source's header declares it is queryname sorted or grouped and logs a warning if not. */
+  def checkSort(source: SamSource, path: FilePath, name: String): Unit = {
+    if (source.header.getSortOrder != SortOrder.queryname && source.header.getGroupOrder != GroupOrder.query) {
+      logger.warning(s"${name} file ${path} does not appear to be queryname sorted or grouped.")
```
nh13, line 58:
```suggestion
      logger.warning(s"${name} file ${path} does not appear to be queryname sorted or grouped.")
      logger.warning(s"Continuing, but your output may be incorrect.")
```

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+    }
+  }
+
+  def buildOutputHeader(unmapped: SAMFileHeader,
```
nh13, line 62: One line doc string since this is a public method

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+                        mapped: SAMFileHeader,
+                        sort: Option[SamOrder],
```
nh13, line 64:
```suggestion
                        sort: Option[SamOrder] = None,
```

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+
+  override def execute(): Unit = {
+    val unmappedSource = SamSource(unmapped)
+    val mappedSource   = SamSource(input)
+    checkSort(unmappedSource, unmapped, "unmapped")
+    checkSort(mappedSource,  input, "input")
+
+    val dict    = SAMSequenceDictionaryExtractor.extractDictionary(this.ref)
+    val header  = buildOutputHeader(unmappedSource.header, mappedSource.header, this.sort, dict)
+    val out     = SamWriter(output, header, sort=this.sort, async=true)
```
nh13, line 172: Why explicitly set `async=true` here? I am seeing a trend of having this always on, so why not just set it as always on at the tool level?

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+
+    val unmappedIter = templateIterator(unmappedSource, bufferSize=buffer)
+    val mappedIter   = templateIterator(mappedSource, bufferSize=buffer)
+
+    unmappedIter.foreach { template =>
+      if (mappedIter.hasNext && mappedIter.head.name == template.name) {
+        out ++= merge(unmapped=template, mapped=mappedIter.next(), tagInfo=tagInfo).allReads
+      }
+      else {
+        out ++= template.allReads
```
nh13, line 183:
```suggestion
        logger.debug("Found an unmapped read with no corresponding mapped read: ${template.name}")
        out ++= template.allReads
```

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+import htsjdk.samtools.{SAMFileHeader, SAMSequenceDictionary}
+import htsjdk.variant.utils.SAMSequenceDictionaryExtractor
+
+/*
+Feature List
+
+    Trim back reads that are FR pairs that extend past each other's starts
+ */
+
+object ZipperBams extends LazyLogging {
```
nh13, line 46: How much do you want to public?  Consider making more methods and classes private

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+    header.setGroupOrder(unmapped.getGroupOrder)
+    sort.foreach(s => s.applyTo(header))
+
+    header
+  }
+
+  /** Constructs a template iterator that assumes the input is sorted or grouped, and turns it into an async iter. */
+  def templateIterator(source: SamSource, bufferSize: Int): BetterBufferedIterator[Template] = {
+    val recs = source.iterator.bufferBetter
+    val templates = new Iterator[Template] {
```
nh13, line 87: blasphemy
```suggestion
    val recs      = source.iterator.bufferBetter
    val templates = new Iterator[Template] {
```

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+  /** Merge information from the unmapped template into the mapped template. */
+  def merge(unmapped: Template, mapped: Template, tagInfo: TagInfo): Template = {
+    // Fix mate info first
+    mapped.fixMateInfo()
+
+    // Remove tags from the mapped records
+    for (rec <- mapped.allReads; tag <- tagInfo.remove) rec.remove(tag)
+
+    // Copy tags over from the unmapped record
+    for (u <- unmapped.primaryReads) {
```
nh13, line 109: do we check any where that either/and (1) the unmapped BAM contains only unmapepd reads, (2) there are only primary reads?

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+    if (bufferSize > 0) {
+      AsyncIterator(templates, Some(bufferSize)).bufferBetter
+    }
+    else {
+      templates.bufferBetter
+    }
+  }
+
+  /** Merge information from the unmapped template into the mapped template. */
+  def merge(unmapped: Template, mapped: Template, tagInfo: TagInfo): Template = {
```
nh13, line 101: 1. do we want to copy over any flags from the _unmapped_ reads (e.g. pass/fail vendor QC?)
2. do we want to _keep_ the proper pair flag from the _mapped_ reads?

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,191 @@
...
+
+    header
+  }
+
+  /** Constructs a template iterator that assumes the input is sorted or grouped, and turns it into an async iter. */
+  def templateIterator(source: SamSource, bufferSize: Int): BetterBufferedIterator[Template] = {
+    val recs = source.iterator.bufferBetter
+    val templates = new Iterator[Template] {
+      override def hasNext: Boolean = recs.hasNext
+      override def next(): Template = Template(recs)
```
nh13, line 89: So I think this will catch the case where the aligner produces multiple primaries for a given read end.  Do we want to have a better error message than what's in `Template`?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -98,6 +103,19 @@ case class Template(r1: Option[SamRecord],
...
+  /** Fixes mate information and sets mate cigar on all primary and supplementary (but not secondary) records. */
+  def fixMateInfo(): Unit = {
+    for (primary <- r1; supp <- r2Supplementals) {
+      SamPairUtil.setMateInformationOnSupplementalAlignment(supp.asSam, primary.asSam, true)
+    }
+    for (primary <- r2; supp <- r1Supplementals) {
+      SamPairUtil.setMateInformationOnSupplementalAlignment(supp.asSam, primary.asSam, true)
+    }
+    for (first <- r1; second <- r2) {
+      SamPairUtil.setMateInfo(first.asSam, second.asSam)
```
nh13, line 115: Do you want to set the mate cigar?  I think if not given, the default is `false`
```suggestion
      SamPairUtil.setMateInfo(first.asSam, second.asSam, true)
```

nh13 (changes requested):
1. I do like the idea of opting in to store the whole reference in memory to update the NM/MD tags and clip overlapping read pairs. 
2. Some very minor comments.

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: Indeed - I just don't want to name it anything close to `MergeBamAlignment` and that was the best I could come up with personally.  Perhaps we start with our slack and if no good ideas are forthcoming, ask twitter?

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: I was thinking actually that since tags have to be two characters we could also allow named sets, so you could do --reverse Duplex and --revcop Duplex and it would put in all the relevant tags?

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: Should probably not do this .. there's no easy way for a single tool to override the default though as it's off in another class that the tool doesn't see directly.

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: Agreed.

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: Oh, that's a really good question.  Probably we should copy some non-alignment related flags over.  

I'm in favor of, at least at first, having the tool not try to override anything the aligner does but _just_ import information from the unmapped BAM.  Most of the time MergeBamAlignment tries to be smarter than modern aligners it makes things worse.

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: Yeah, probably.

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: Made the companion itself private!

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: Honestly I'm not sure.

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
nh13: I don't have a better name, ship it.

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,242 @@
...
+  val tagsToRemove: IndexedSeq[String] = IndexedSeq.empty,
+  @arg(doc="Set of optional tags to reverse on reads mapped to the negative strand.", minElements=0)
+  val tagsToReverse: IndexedSeq[String] = IndexedSeq.empty,
+  @arg(doc="Set of optional tags to reverse complement on reads mapped to the negative strand.", minElements=0)
+  val tagsToRevcomp: IndexedSeq[String] = IndexedSeq.empty,
+  @arg(flag='s', doc="Sort the output BAM into the given order.") val sort: Option[SamOrder] = None,
+  @arg(flag='b', doc="Buffer this many read-pairs while reading the input BAMs.") val buffer: Int = 5000
+) extends FgBioTool {
+  import ZipperBams._
+
```
nh13, line 209: add `Io.assert*` here

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,242 @@
...
+    if (tagInfo.remove.nonEmpty)  logger.info(s"Tags for removal from mapped BAM: ${tagInfo.remove.mkString(", ")}")
+    if (tagInfo.reverse.nonEmpty) logger.info(s"Tags being reversed: ${tagInfo.reverse.mkString(", ")}")
+    if (tagInfo.revcomp.nonEmpty) logger.info(s"Tags being reverse complemented: ${tagInfo.revcomp.mkString(", ")}")
+
+    val unmappedIter = templateIterator(unmappedSource, bufferSize=buffer)
+    val mappedIter   = templateIterator(mappedSource, bufferSize=buffer)
+
+    unmappedIter.foreach { template =>
+      if (mappedIter.hasNext && mappedIter.head.name == template.name) {
```
nh13, line 229: Happy to keep your version for speed, but this also works:
```suggestion
      if (mappedIter.headOption.exists(_.name == template.name)) {
```

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -0,0 +1,242 @@
...
+    val tagInfo = TagInfo(remove=tagsToRemove, reverse=tagsToReverse, revcomp=tagsToRevcomp)
+
+    if (tagInfo.remove.nonEmpty)  logger.info(s"Tags for removal from mapped BAM: ${tagInfo.remove.mkString(", ")}")
+    if (tagInfo.reverse.nonEmpty) logger.info(s"Tags being reversed: ${tagInfo.reverse.mkString(", ")}")
+    if (tagInfo.revcomp.nonEmpty) logger.info(s"Tags being reverse complemented: ${tagInfo.revcomp.mkString(", ")}")
+
+    val unmappedIter = templateIterator(unmappedSource, bufferSize=buffer)
+    val mappedIter   = templateIterator(mappedSource, bufferSize=buffer)
+
+    unmappedIter.foreach { template =>
```
nh13, line 228: progress logging please, with `progress.logLast` too please.

nh13 (approved):
Minor comments, no need to re-review.

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: I'll keep mine, thanks!

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
tfenne: The writer handles the progress logging for both sorting (if it's happening) and writing, and does `logLast` appropriately.  I think that's sufficient.

PR #779, opened 2022-02-20
ConsensusCallingIterator could fail when no consensus reads are called

nh13:
This occurs when `consensusReadsFromSamRecords` returns `Seq.empty` (due to filters) for ALL input reads/templates, so we hit the "Bug: input is not empty" exception


nh13:
Closing in favor of #780

PR #780, opened 2022-02-20
Alternative bugfix for "ConsensusCallingIterator could fail when no consensus reads are called"

tfenne:
An alternative approach to #779 that eliminates the `Iterator.continually` for a regular iterator that is (IMHO) a little easier to reason about.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
@@ -25,6 +25,7 @@
 package com.fulcrumgenomics.umi
 
 import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.bam.api
 import com.fulcrumgenomics.bam.api.SamRecord
 import com.fulcrumgenomics.commons.async.AsyncIterator
 import com.fulcrumgenomics.commons.util.LazyLogging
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -63,41 +64,43 @@ class ConsensusCallingIterator[ConsensusRead <: SimpleRead](sourceIterator: Iter
     }
     else {
       val halfMaxRecords   = maxRecordsInRam / 2
-      val groupingIterator = new SamRecordGroupedIterator(new AsyncIterator(progressIterator, Some(halfMaxRecords)).start(), caller.sourceMoleculeId)
       val pool             = new ForkJoinPool(threads - 1, ForkJoinPool.defaultForkJoinWorkerThreadFactory, null, true)
-      val bufferedIter     = groupingIterator.bufferBetter
       val callers          = new IterableThreadLocal[UmiConsensusCaller[ConsensusRead]](() => caller.emptyClone())
+      val groupingIterator = {
+        val async   = AsyncIterator(progressIterator, Some(halfMaxRecords))
```
nh13, line 70: I run into this so frequently with our async classes, namely forgetting to start them.  Perhaps the behavior should be changed to always start, but have a way to say "don't start"
```suggestion
        val async   = AsyncIterator(progressIterator, Some(halfMaxRecords)).start()
```

nh13 (approved):
One bug found, but the rest looks better and tidier.  It may be worth confirming there's no performance regression.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
nh13: ffs, and I got it wrong too, since you changed this from the primary constructor to a secondary constructor, with the latter automatically starting!

tfenne: Yeah, for better or worse, we decided the apply() methods should start them, while the constructors shouldn't.  Perhaps we should just change them to have the constructors take a non-defaulted `start: Boolean` so it's super clear.

PR #781, opened 2022-02-20
Change default sort orders of consensus callers

nh13:
This affects CallDuplexConsensusReads and CallMolecularConsensusReads.

When no sort order is given, the output is unsorted but query-grouped,
which is sufficient for either downstream re-alignment or consensus
filtering.  We can thus avoid sorting the output consensus reads when
calling.

I _think_ `picard MergeBamAlignment` will not like this, as it wants `queryname`, not `unsorted:querygrouped`, so leaving this as a draft for now.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
@@ -121,7 +121,7 @@ class CallMolecularConsensusReads
           """)
  val maxReads: Option[Int] = None,
  @arg(flag='B', doc="If true produce tags on consensus reads that contain per-base information.") val outputPerBaseTags: Boolean = DefaultProducePerBaseTags,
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -121,7 +121,7 @@ class CallMolecularConsensusReads
           """)
  val maxReads: Option[Int] = None,
  @arg(flag='B', doc="If true produce tags on consensus reads that contain per-base information.") val outputPerBaseTags: Boolean = DefaultProducePerBaseTags,
- @arg(flag='S', doc="The sort order of the output, if `:none:` then the same as the input.") val sortOrder: Option[SamOrder] = Some(SamOrder.Queryname),
+ @arg(flag='S', doc="The sort order of the output, if `:none:` then the same as the input.") val sortOrder: Option[SamOrder] = None,
```
nh13, line 124:
```suggestion
 @arg(flag='S', doc="The sort order of the output, the same as the input if not given.") val sortOrder: Option[SamOrder] = None,
```

tfenne (approved):

PR #782, opened 2022-02-21
Specify an output sort order in FilterConsensusReads

nh13:
Also ReferenceSequenceIterator closes the underlying reference file.

@tfenne not sure if all this work is worth it.  But if the input order is queryname and the output order is queryname, then this will skip sorting when `--load-full-reference=true`.  This could be nice if we want to rely on multi-threaded sort tools or don't care about the order after filtering.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
@@ -24,19 +24,23 @@
 
 package com.fulcrumgenomics.umi
 
-import java.lang.Math.{max, min}
-
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.bam.Bams
 import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamSource, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.commons.io.Writer
 import com.fulcrumgenomics.commons.util.LazyLogging
+import com.fulcrumgenomics.fasta.ReferenceSequenceIterator
 import com.fulcrumgenomics.sopt.{arg, clp}
 import com.fulcrumgenomics.util.NumericTypes.PhredScore
 import com.fulcrumgenomics.util.{Io, ProgressLogger}
-import htsjdk.samtools.SAMFileHeader.SortOrder
+import htsjdk.samtools.SAMFileHeader
+import htsjdk.samtools.SAMFileHeader.GroupOrder
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker
 import htsjdk.samtools.util.SequenceUtil
+
+import java.io.Closeable
+import java.lang.Math.{max, min}
 
 /** Filter values for filtering consensus reads */
 private[umi] case class ConsensusReadFilter(minReads: Int, maxReadErrorRate: Double, maxBaseErrorRate: Double)
@@ -114,7 +118,9 @@ class FilterConsensusReads
   @arg(flag='q', doc="The minimum mean base quality across the consensus read.")
   val minMeanBaseQuality: Option[PhredScore] = None,
   @arg(flag='s', doc="Mask (make `N`) consensus bases where the AB and BA consensus reads disagree (for duplex-sequencing only).")
-  val requireSingleStrandAgreement: Boolean = false
+  val requireSingleStrandAgreement: Boolean = false,
+  @arg(flag='S', doc="The sort order of the output, if `:none:` then the same as the input.") val sortOrder: Option[SamOrder] = Some(SamOrder.Coordinate),
+  @arg(flag='l', doc="Load the full reference sequence in memory") val loadFullReference: Boolean = false
 ) extends FgBioTool with LazyLogging {
   // Baseline input validation
   Io.assertReadable(input)
@@ -169,12 +175,9 @@ class FilterConsensusReads
   private val EmptyFilterResult = FilterResult(keepRead=true, maskedBases=0)
 
   override def execute(): Unit = {
-    val in        = SamSource(input)
-    val header    = in.header.clone()
-    header.setSortOrder(SortOrder.coordinate)
-    val sorter    = Bams.sorter(SamOrder.Coordinate, header, maxRecordsInRam=MaxRecordsInMemoryWhenSorting)
-    val out       = SamWriter(output, header)
-    val progress1 = ProgressLogger(logger, verb="Filtered and masked")
+    val progress = ProgressLogger(logger, verb="Filtered and masked")
+    val in       = SamSource(input)
+    val out      = buildOutputWriter(in.header)
 
     // Go through the reads by template and do the filtering
     val templateIterator = Bams.templateIterator(in, maxInMemory=MaxRecordsInMemoryWhenSorting)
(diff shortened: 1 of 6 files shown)


src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -201,34 +204,126 @@ class FilterConsensusReads
...
+      new Writer[SamRecord] with Closeable {
+        private val _sorter = Bams.sorter(order=SamOrder.Coordinate, header=header, maxRecordsInRam=MaxRecordsInMemoryWhenSorting)
+        override def write(rec: SamRecord): Unit = {
+          progress.record(rec)
+          this._sorter += rec
+        }
+        def close(): Unit = {
+          this._sorter.foreach { rec => writer += regenerateNmUqMdTags(rec) }
+          writer.close()
+          refCloseMethod()
```
nh13, line 323:
```suggestion
          refCloseMethod()
          progress.logLast()
```

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -201,34 +208,63 @@ class FilterConsensusReads
...
-    logger.info(f"Output ${keptReads}%,d of ${totalReads}%,d primary consensus reads.")
-    logger.info(f"Masked ${maskedBases}%,d of ${totalBases}%,d bases in retained primary consensus reads.")
+    logger.info(f"Output $keptReads%,d of $totalReads%,d primary consensus reads.")
+    logger.info(f"Masked $maskedBases%,d of $totalBases%,d bases in retained primary consensus reads.")
+  }
+
+  /** Builds the writer to which filtered records should be written.
+    *
+    *  If the input order is [[SamOrder.Queryname]] or query grouped, then the filtered records will also be in the same
+    *  order.  So if the output order specified AND does not match the the input order, sorting will occur.
```
nh13, line 236:
```suggestion
    *  order.  So if the output order is specified AND does not match the the input order, sorting will occur.
```

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -201,34 +208,63 @@ class FilterConsensusReads
...
+    *
+    *  If the input order is not [[SamOrder.Queryname]] or query grouped, then the input records will be resorted into
+    *  [[SamOrder.Queryname]].  So if the output order specified AND is not [[SamOrder.Queryname]], sorting will occur.
```
nh13, line 239:
```suggestion
    *  [[SamOrder.Queryname]].  So if the output order is specified AND is not [[SamOrder.Queryname]], sorting will occur.
```

src/test/scala/com/fulcrumgenomics/umi/FilterConsensusReadsTest.scala
```diff
@@ -290,6 +290,67 @@ class FilterConsensusReadsTest extends UnitSpec {
     }
   }
 
+  private case class ReadNames(in: Seq[String], out: Seq[String])
+
+  private def sortOrderTest(name1: String, start1R1: Int, start1R2: Int, name2: String, start2R1: Int, start2R2: Int,
+                            inOrder: SamOrder, outOrder: Option[SamOrder] = None): ReadNames = {
```
tfenne, line 296: I think that in order to test more combinations it would be better to take `inSort` and `inGroup` separately as `SamOrder` has a limited set of combinations.

tfenne (changes requested):
Top level usage on FilterConsensusReads needs updating.  The second to last paragraph talks about sorting, which is now incorrect.  And it should mention the need to load the whole genome into memory in addition to potentially needing memory for sorting.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -169,12 +175,13 @@ class FilterConsensusReads
...
   override def execute(): Unit = {
-    val in        = SamSource(input)
-    val header    = in.header.clone()
-    header.setSortOrder(SortOrder.coordinate)
-    val sorter    = Bams.sorter(SamOrder.Coordinate, header, maxRecordsInRam=MaxRecordsInMemoryWhenSorting)
-    val out       = SamWriter(output, header)
-    val progress1 = ProgressLogger(logger, verb="Filtered and masked")
+    logger.info("Reading the reference into memory")
+    val refMap = ReferenceSequenceIterator(ref, stripComments=true).map { ref => ref.getContigIndex -> ref}.toMap
+    logger.info(f"Read ${refMap.size}%,d contigs.")
```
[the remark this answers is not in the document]
nh13: Leaving it here for now

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -169,12 +175,13 @@ class FilterConsensusReads
...
+
+    val progress = ProgressLogger(logger, verb="Filtered and masked")
+    val in       = SamSource(input)
+    val out      = buildOutputWriter(in.header, refMap)
```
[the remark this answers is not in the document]
nh13: Leaving it as is.  See https://github.com/fulcrumgenomics/fgbio/issues/799

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -201,34 +208,63 @@ class FilterConsensusReads
...
-      progress2.record(rec)
-    }
-
+    logger.info("Finalizing the output")
     in.safelyClose()
     out.close()
-    logger.info(f"Output ${keptReads}%,d of ${totalReads}%,d primary consensus reads.")
-    logger.info(f"Masked ${maskedBases}%,d of ${totalBases}%,d bases in retained primary consensus reads.")
+    logger.info(f"Output $keptReads%,d of $totalReads%,d primary consensus reads.")
+    logger.info(f"Masked $maskedBases%,d of $totalBases%,d bases in retained primary consensus reads.")
```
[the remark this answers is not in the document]
nh13: don't need it, and IntelliJ is making it a game to get rid of the warnings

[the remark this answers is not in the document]
tfenne: So ... IntelliJ is not the boss of us.  Put cursor into warning, option-enter, disable inspection.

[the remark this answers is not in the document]
tfenne: I really do think that Intellij is wrong on this.  Since braces become necessary the moment the expression becomes more than a simple variable reference (e.g. `${masedBases / totalBases}`) _and_ the fact that braces make it much clearer where the expression ends, I try to use them 100% of the time, and would certainly prefer they not get removed.

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -201,34 +207,87 @@ class FilterConsensusReads
...
+        SamOrder.values.find { order =>
+          order.sortOrder == outputSortOrder && order.groupOrder == outputGroupOrder && order.subSort == outputSubSort
+        }.orElse {
+          // this can only happen if the input order is unrecognized
+          throw new IllegalArgumentException(
+            s"The input BAM had an unrecognized sort order (SO:$inSortOrder GO:$inGroupOrder SS: $inSubSort)" +
+            s"\nTry re-running with --sort-order for a supported output order." in $input
+          )
+        }
+      }
```
tfenne, line 278: Can't this whole `else` block now just be `this.sortOrder`?  I think your intent is:
1. If `this.sortOrder` is None, and input is query sorted or grouped, no sorting occurs anywhere and output == input
2. Else if `this.sortOrder` is None, and input is _not_ query sorted or grouped, sorting to queryname occurs, and output is in SamOrder.Queryname
3. Else output will be in `this.sortOrder` and sorting will occur _only_ if the `afterFiltering` ordering is different than `this.sortOrder`.

In this `else` branch we've confirmed that afterFiltering and output orderings are different, and that can _only_ occur if `this.sortOrder` is defined, so we can just use it?

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -114,7 +118,9 @@ class FilterConsensusReads
   @arg(flag='q', doc="The minimum mean base quality across the consensus read.")
   val minMeanBaseQuality: Option[PhredScore] = None,
   @arg(flag='s', doc="Mask (make `N`) consensus bases where the AB and BA consensus reads disagree (for duplex-sequencing only).")
-  val requireSingleStrandAgreement: Boolean = false
+  val requireSingleStrandAgreement: Boolean = false,
```
tfenne, line 121: I can't comment on the tool usage, but you seemed to miss my top-level review comment last time, so I'm replicating it here:

> Top level usage on FilterConsensusReads needs updating. The second to last paragraph talks about sorting, which is now incorrect. And it should mention the need to load the whole genome into memory in addition to potentially needing memory for sorting.

tfenne (approved):
One simplification suggested, and usage still needs updating.  Merge when addressed.

tfenne (approved):

PR #790, opened 2022-02-25
Adds an optimized path when creating a consensus from a single input read.

tfenne:
This probably needs a very careful review, and we may want to add more tests to make sure it is doing the right thing in all cases, but in my hands this cuts consensus calling time in half for `CallMolecularConsensusReads` when there are a lot of input molecules with only a single supporting read.

On an example BAM file these two commits get me from a runtime on an example BAM of  10.80 minutes on master to 5.00 after the first commit and 4.45 minutes after the second commit.

Third commit adds a new ParIterator based on the iterata one though cleaned up a bit and extended a bit.  Using that instead of the prior parallelization cuts runtime to ~3.4 minutes with the same number of threads.

And with the fourth commit (adding an async iterator on the input side too) we're down to 2.14 minutes!


src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -196,39 +205,57 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
...
-          if (depth < this.options.minReads)       (NoCall, NotEnoughReadsQual)
-          else if (rawQual < this.options.minConsensusBaseQuality) (NoCall, TooLowQualityQual)
-          else (rawBase, rawQual)
+          consensusBases(i)  = base
+          consensusQuals(i)  = qual
+          consensusDepths(i) = if (isNoCall) 0 else 1
+          consensusErrors(i) = 0
         }
+      }
+      else {
```
tfenne, line 224: Everything from here on down is identical, just indented under an else, and github has gotten confused.

src/main/scala/com/fulcrumgenomics/util/ParIterator.scala
```diff
@@ -0,0 +1,130 @@
...
+import com.fulcrumgenomics.commons.async.AsyncIterator
+
+import java.util.concurrent.ForkJoinPool
+import scala.collection.AbstractIterator
+import scala.collection.parallel.{ForkJoinTaskSupport, TaskSupport}
+
+/**
+  * Companion object to help with manufacturing ParIterators.
+  */
+object ParIterator {
```
tfenne, line 37: This whole file will go down into commons (along with some implicits in CommonsDef) once we're happy with it.

src/main/scala/com/fulcrumgenomics/util/ParIterator.scala
```diff
@@ -0,0 +1,130 @@
...
+    apply(iter, chunkSize, support)
+  }
+
+  /**
+    * Constructs a ParIterator from the underlying iterator, that will perform operations in parallel on chunks
+    * of `chunkSize` using the provided `TaskSupport`.
+    */
+  def apply[A](iter: Iterator[A], chunkSize: Int, support: TaskSupport): ParIterator[A] = {
+    val grouped = iter.grouped(chunkSize)
+    val async = AsyncIterator(grouped, Some(2))
```
tfenne, line 54: It's possible that this should be optional, but it makes a huge different to have the next chunk ready to go when the parallel operation on the prior chunk wraps up.  Cut runtime by ~1/3 and finally got utilization up to what you'd hope with the number of threads.

src/main/scala/com/fulcrumgenomics/util/ParIterator.scala
tfenne: Actually, don't review it here!  Review it here instead: https://github.com/fulcrumgenomics/commons/pull/76

src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -196,39 +204,57 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
...
-      while (positionInRead < consensusLength) {
-        // Add the evidence from all reads that are long enough to cover this base
-        capped.foreach { read =>
-          if (read.length > positionInRead) {
-            val base = read.bases(positionInRead)
-            val qual = read.quals(positionInRead)
-            if (base != NoCall) builder.add(base=base, qual=qual)
-          }
-        }
+      if (capped.length == 1) {
```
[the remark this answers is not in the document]
nh13: I suppose my optimization assumes their all the same length!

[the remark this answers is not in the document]
tfenne: Agreed, though somewhere higher up the chain we avoid calling consensus if we have fewer than minReads already, and if you set maxReads < minReads that's on you ;)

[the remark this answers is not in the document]
tfenne: Added a validation in CallMolecularConsensusReads to make sure max-reads >= min-reads.

PR #793, opened 2022-02-27
Change default validation stringency to SILENT and make common option…

tfenne:
…s more easily accessible at runtime.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
+++ b/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
@@ -56,12 +56,25 @@ object FgBioMain {
   case class FailureException private[cmdline] (exit: Int = 1, message: Option[String] = None) extends RuntimeException
 }
 
+/** Public object that holds a singleton instance of the common args so that it can be accessed elsewhere. */
+object FgBioCommonArgs {
+  private var _args: FgBioCommonArgs = new FgBioCommonArgs()
+
+  /** Setter that is private to the cmdline package. */
+  private[cmdline] def args_=(args: FgBioCommonArgs): Unit = {
+    this._args = args
+  }
+
+  /** Access the common args for a command line program. */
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
```diff
@@ -56,12 +56,25 @@ object FgBioMain {
...
+  def args: FgBioCommonArgs = this._args
+}
+
 class FgBioCommonArgs
 ( @arg(doc="Use asynchronous I/O where possible, e.g. for SAM and BAM files.") val asyncIo: Boolean = false,
   @arg(doc="Default GZIP compression level, BAM compression level.")           val compression: Int = 5,
   @arg(doc="Directory to use for temporary files.")                            val tmpDir: DirPath  = Paths.get(System.getProperty("java.io.tmpdir")),
   @arg(doc="Minimum severity log-level to emit.")                              val logLevel: LogLevel = LogLevel.Info,
-  @arg(doc="Validation stringency for SAM/BAM reading.")                       val samValidationStringency: Option[ValidationStringency] = None
+  @arg(doc="Validation stringency for SAM/BAM reading.")                       val samValidationStringency: ValidationStringency = ValidationStringency.STRICT
```
tfenne, line 77: I'm tired of paying an extra 50% for HTSJDK to run decoding and validation when it's not needed.  Thoughts?

nh13: I agree. So do we want silent instead of strict?

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
tfenne: Whoopsie, silly autocomplete.  Yes we do.

nh13 (approved):

PR #794, opened 2022-02-27
GroupReadsByUmi only sort input if it is not TemplateCoordinate sorted

nh13:
This extracts some functionality from #791, where we skip sorting the input if the input is already `TemplateCoordinate` sorted.  This allows other tools to perform the sort.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -27,8 +27,8 @@
 package com.fulcrumgenomics.umi
 
 import java.util.concurrent.atomic.AtomicLong
-
 import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.bam.api.SamOrder.TemplateCoordinate
 import com.fulcrumgenomics.bam.{Bams, Template}
 import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamSource, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
@@ -102,8 +102,8 @@ object GroupReadsByUmi {
 
         val r1Earlier =
           (ref1 < ref2) ||
-          (ref1 == ref2 && start1 < start2) ||
-          (ref1 == ref2 && start1 == start2 && strand1 < strand2)
+            (ref1 == ref2 && start1 < start2) ||
+            (ref1 == ref2 && start1 == start2 && strand1 < strand2)
 
         if (r1Earlier) new ReadInfo(ref1, start1, strand1, ref2, start2, strand2, lib)
         else           new ReadInfo(ref2, start2, strand2, ref1, start1, strand1, lib)
@@ -381,7 +381,7 @@ object GroupReadsByUmi {
       * the main assignment method to filter out the ones that shouldn't be in the Map.
       */
     override def assign(rawUmis: Seq[Umi]): Map[Umi, MoleculeId] = {
-      super.assign(rawUmis).filter { case (umi, id) => rawUmis.contains(umi) }
+      super.assign(rawUmis).filter { case (umi, _) => rawUmis.contains(umi) }
     }
   }
 }
@@ -469,6 +469,9 @@ object Strategy extends FgBioEnum[Strategy] {
     |compared, where `len` is the length of the shortest UMI. The UMI length is the number of [ACGT] bases in the UMI
     |(i.e. does not count dashes and other non-ACGT characters). This option is not implemented for reads with UMI pairs
     |(i.e. using the paired assigner).
+    |
+    |If the input is not template-coordinate sorted (i.e. `SO:unsorted GO:query SS:unsorted:template-coordinate`), then
+    |this tool will re-sort the input. The ouitput will be written in template-coordinate order.
   """
 )
 class GroupReadsByUmi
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -279,7 +279,7 @@ object GroupReadsByUmi {
     /** Returns the UMI with the lexically lower half first. */
     override def canonicalize(u: Umi): Umi = {
       val (a, b) = split(u)
-      if (a < b) u else s"${b}-${a}"
+      if (a < b) u else s"$b-$a"
```
tfenne, line 282: Stop stealing my braces!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -520,7 +518,7 @@ class GroupReadsByUmi
           }
         } || { filterUmisTooShort += 1; false}
       }
-      .foreach { r =>
+      .map { r =>
```
tfenne, line 521:
```suggestion
      .tapEach { r =>
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -536,33 +534,40 @@ class GroupReadsByUmi
...
-    this.minUmiLength.foreach { _ => logger.info(f"Filtered out $filterUmisTooShort%,d reads that contained UMIs that were too short.") }
-
-    // Output the reads in the new ordering
-    logger.info("Assigning reads to UMIs and outputting.")
     val outHeader = header.clone()
-    SamOrder.TemplateCoordinate.applyTo(outHeader)
-    val out = SamWriter(output, outHeader)
-
-    val iterator = Bams.templateIterator(sorter.iterator, out.header, Bams.MaxInMemory, Io.tmpDir)
     val tagFamilySizeCounter = new NumericCounter[Int]()
```
tfenne, line 542: It's weird that after your edit this is sandwiched between creation of the outHeader and setting the SamOrder on the outHeader.  Maybe move it up a line?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -536,33 +534,40 @@ class GroupReadsByUmi
...
+    SamOrder.TemplateCoordinate.applyTo(outHeader)
+    val out                        = SamWriter(output, outHeader)
+    val templateCoordinateIterator = {
+      // Skip sorting if the input is already template coordinate!
+      val iter = if (SamOrder(in.header).contains(TemplateCoordinate)) filteredIterator else {
+        logger.info("Sorting the input to TeplateCoordinate order.")
```
tfenne, line 548:
```suggestion
        logger.info("Sorting the input to TemplateCoordinate order.")
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -573,9 +578,14 @@ class GroupReadsByUmi
...
       templatesByMi.values.foreach(ps => tagFamilySizeCounter.count(ps.size))
     }
-
     out.close()
 
+    logger.info(f"Accepted $kept%,d reads for grouping.")
+    if (filteredNonPf > 0) logger.info(f"Filtered out $filteredNonPf%,d non-PF reads.")
+    logger.info(f"Filtered out $filteredPoorAlignment%,d reads due to mapping issues.")
+    logger.info(f"Filtered out $filteredNsInUmi%,d reads that contained one or more Ns in their UMIs.")
+    this.minUmiLength.foreach { _ => logger.info(f"Filtered out $filterUmisTooShort%,d reads that contained UMIs that were too short.") }
```
tfenne, line 587: It makes me a little sad that now if GroupReads is doing the sorting I have to wait until the output is written to get the stats.

tfenne (changes requested):
`canTakeNextGroupByUmi` on line 480 will, I think, need updating.  If the sort was done externally then I don't think we can rely on the fact that they records are also sorted by the same UMI sequence we're going to look at, and so it's not safe to take the next group by UMI.

Usage needs updating to clearly explain when sorting will and won't happen.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: I have found a way, I think

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne, line 282:
```suggestion
      if (a < b) u else s"${b}-${a}"
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -491,22 +491,21 @@ class GroupReadsByUmi
     rec.mapq >= minMapQ && mateMqOk
   }
 
+  // A handful of counters for tracking reads
```
tfenne, line 494: Reproducing top-level comments from my last review here since you seem to have missed them and I can't comment on the appropriate lines since they are unmodified:

> canTakeNextGroupByUmi on line 480 will, I think, need updating. If the sort was done externally then I don't think we can rely on the fact that they records are also sorted by the same UMI sequence we're going to look at, and so it's not safe to take the next group by UMI.

> Usage needs updating to clearly explain when sorting will and won't happen.

tfenne (commented):
Looks good except for updating usage and fixing up the `canTakeNextGroupByUmi`.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -294,7 +294,7 @@ object GroupReadsByUmi {
     /** Takes a UMI of the form "A-B" and returns "B-A". */
     def reverse(umi: Umi): Umi = {
       val (first, second) = split(umi)
-      s"${second}-${first}"
+      s"$second-$first"
```
nh13, line 297:
```suggestion
      s"${second}-${first}"
```

nh13:
Done.

tfenne (approved):

PR #800, opened 2022-03-02
Added ability for FastqToBam to also extract UMIs from read names.

tfenne:
This ended up being more complicated than I thought, but I think it's still worthwhile.  I refactored `Umis.copyUmiFromReadName()` to enable me to apply it just to read names, and also to add a strict mode.  I got rid of the parameter for setting the conversion of `+` to `-` and just made that happen all the time, since I hope we really only see `+` or `-` in the wild.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
@@ -25,14 +25,13 @@
 package com.fulcrumgenomics.fastq
 
 import java.util
-
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamWriter}
 import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
 import com.fulcrumgenomics.commons.CommonsDef.PathToFastq
 import com.fulcrumgenomics.commons.util.LazyLogging
 import com.fulcrumgenomics.sopt.{arg, clp}
-import com.fulcrumgenomics.umi.ConsensusTags
+import com.fulcrumgenomics.umi.{ConsensusTags, Umis}
 import com.fulcrumgenomics.util.SegmentType._
 import com.fulcrumgenomics.util.{Io, ProgressLogger, ReadStructure}
 import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
@@ -72,6 +71,13 @@ import htsjdk.samtools.{ReservedTagConstants, SAMFileHeader, SAMReadGroupRecord}
     |For more information on read structures see the
     |[Read Structure Wiki Page](https://github.com/fulcrumgenomics/fgbio/wiki/Read-Structures)
     |
+    |UMIs may be extracted from the read sequences, the read names, or both.  If `--extract-umis-from-read-names` is
+    |specified, any UMIs present in the read names are extracted; read names are expected to be `:`-separated with
+    |any UMIs present in the 8th field.  If this option is specified, the `--umi-qual-tag` option may not be used as
+    |qualities are not available for UMIs in the read name. If UMI segments are present in the read structures those
+    |will also be extracted.  If UMIs are present in both, the final UMIs be constructed by first taking the UMIs
+    |from the read names, then adding a hyphen, then the UMIs extracted from the reads.
+    |
     |The same number of input files and read structures must be provided, with one exception: if supplying exactly
     |1 or 2 fastq files, both of which are solely template reads, no read structures need be provided.
     |
@@ -86,6 +92,7 @@ class FastqToBam
   @arg(flag='s', doc="If true, queryname sort the BAM file, otherwise preserve input order.")  val sort: Boolean = false,
   @arg(flag='u', doc="Tag in which to store molecular barcodes/UMIs.")                         val umiTag: String = ConsensusTags.UmiBases,
   @arg(flag='q', doc="Tag in which to store molecular barcode/UMI qualities.")                 val umiQualTag: Option[String] = None,
+  @arg(flag='n', doc="Extract UMI(s) from read names and prepend to UMIs from reads.")         val extractUmisFromReadNames: Boolean = false,
   @arg(          doc="Read group ID to use in the file header.")                               val readGroupId: String = "A",
   @arg(          doc="The name of the sequenced sample.")                                      val sample: String,
   @arg(          doc="The name/ID of the sequenced library.")                                  val library: String,
@@ -107,6 +114,7 @@ class FastqToBam
   Io.assertCanWriteFile(output)
   validate(input.length == actualReadStructures.length, "input and read-structure must be supplied the same number of times.")
   validate(1 to 2 contains actualReadStructures.flatMap(_.templateSegments).size, "read structures must contain 1-2 template reads total.")
+  validate(!extractUmisFromReadNames || umiQualTag.isEmpty, "Cannot extract UMI qualities when also extracting UMI from read names.")
 
   override def execute(): Unit = {
     val encoding = qualityEncoding
@@ -153,6 +161,9 @@ class FastqToBam
       val umiQual       = subs.iterator.filter(_.kind == MolecularBarcode).map(_.quals).mkString(" ")
       val templates     = subs.iterator.filter(_.kind == Template).toList
 
+      // If requested, pull out the UMI(s) from the read name
+      val umiFromReadName = if (extractUmisFromReadNames) Umis.extractUmisFromReadName(fqs.head.name, strict=true) else None
+
       templates.zipWithIndex.map { case (read, index) =>
         // If the template read had no bases, we'll substitute in a single N @ Q2 below to keep htsjdk happy
         val empty = read.bases.length == 0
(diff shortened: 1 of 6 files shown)


src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -72,6 +71,13 @@ import htsjdk.samtools.{ReservedTagConstants, SAMFileHeader, SAMReadGroupRecord}
     |For more information on read structures see the
     |[Read Structure Wiki Page](https://github.com/fulcrumgenomics/fgbio/wiki/Read-Structures)
     |
+    |UMIs may be extracted from the read sequences, the read names, or both.  If `--extract-umis-from-read-names` is
+    |specified, any UMIs present in the read names are extracted; read names are expected to be `:`-separated with
+    |any UMIs present in the 8th field.  If this option is specified, the `--umi-qual-tag` option may not be used as
+    |qualities are not available for UMIs in the read name. If UMI segments are present in the read structures those
+    |will also be extracted.  If UMIs are present in both, the final UMIs are constructed by first taking the UMIs
+    |from the read names, then adding a hyphen, then the UMIs extracted from the reads.
```
nh13, line 79: I really hope we don't see UMIs in the read name and the reads themselves.  The reason they're in names to begin with is someone extracted them to begin with.  So I don't think we should support UMIs in both read names and in the reads themselves, even though it's easy enough to do.

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
```diff
@@ -42,9 +42,7 @@ import com.fulcrumgenomics.util.{Io, ProgressLogger}
 class CopyUmiFromReadName
 ( @arg(flag='i', doc="The input BAM file") input: PathToBam,
   @arg(flag='o', doc="The output BAM file") output: PathToBam,
-  @arg(doc="Remove the UMI from the read name") removeUmi: Boolean = false,
-  @arg(doc="Replaces any occurrences of this delimiter found in the UMI with a dash ('-') as per the SAM specification")
-  umiDelimiter: Option[Char] = None
```
nh13, line 47: The usage needs to be updated to say that UMIs can be either dash or plus delimited.

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -35,37 +35,62 @@ object Umis {
     * The read name is split by the given name delimiter, and the last field is assumed to be the UMI sequence.  The UMI
     * will be copied to the `RX` tag as per the SAM specification.
     *
+    * An exception will be thrown if the read name does not contain a valid UMI in the last delimited segment.
+    *
     * @param rec the record to modify
     * @param removeUmi true to remove the UMI from the read name, otherwise only copy the UMI to the tag
-    * @param umiDelimiter if not None, replaces any occurrences of this delimiter found in the UMI with a dash ('-')
-    *                     as per the SAM specification
+    * @param delimiter the delimiter of fields within the readname
```
nh13, line 42:
```suggestion
    * @param delimiter the delimiter of fields within the read name
```

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -35,37 +35,62 @@ object Umis {
...
+    *
+    *  If `strict` is true the name _must_ contain 8 colon-separated segments, with the UMI being the last.  If
+    *  `strict` is false the last segment is returned so long as it appears to be a valid UMI.
+    */
+  def extractUmisFromReadName(header: String, delimiter: Char = ':', strict: Boolean): Option[String] = {
+    // Support full FASTQ headers by first removing any spaces and trailing text
+    val name = {
+      val idx = header.indexOf(' ')
+      if (idx >= 0) header.substring(0, idx) else header
+    }
```
nh13, line 71: I have a feeling we debated this in picard at some point, whether or not to support headers.

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -35,37 +35,62 @@ object Umis {
...
+  /**
+    * Extracts the UMI from an Illumina fastq style read name.  Illumina documents their FASTQ read names as:
+    *   @<instrument>:<run number>:<flowcell ID>:<lane>:<tile>:<x-pos>:<y-pos>:<UMI> <read>:<is filtered>:<control number>:<index>
+    *
+    *  See https://support.illumina.com/help/BaseSpace_OLH_009008/Content/Source/Informatics/BS/FileFormat_FASTQ-files_swBS.htm
+    *
+    *  If `strict` is true the name _must_ contain 8 colon-separated segments, with the UMI being the last.  If
+    *  `strict` is false the last segment is returned so long as it appears to be a valid UMI.
+    */
+  def extractUmisFromReadName(header: String, delimiter: Char = ':', strict: Boolean): Option[String] = {
```
nh13, line 66: this is extracting not from a read name (which has no spaces) but from a FASTQ header (which may have spaces).  Shall we have two methods?  `extractUmisFromReadName` and `extractUmisFromFastqHeader` to make this clearer?  If the read names come from a BAM, then we don't need to search for a space.

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -35,37 +35,62 @@ object Umis {
...
+  }
+
+  /**
+    * Extracts the UMI from an Illumina fastq style read name.  Illumina documents their FASTQ read names as:
+    *   @<instrument>:<run number>:<flowcell ID>:<lane>:<tile>:<x-pos>:<y-pos>:<UMI> <read>:<is filtered>:<control number>:<index>
+    *
+    *  See https://support.illumina.com/help/BaseSpace_OLH_009008/Content/Source/Informatics/BS/FileFormat_FASTQ-files_swBS.htm
+    *
+    *  If `strict` is true the name _must_ contain 8 colon-separated segments, with the UMI being the last.  If
+    *  `strict` is false the last segment is returned so long as it appears to be a valid UMI.
```
nh13, line 64:
```suggestion
    *  If `strict` is true the name _must_ contain either 7 or 8 colon-separated segments, 
    with the UMI being the last in the case of 8 and `None` in the case of 7.
    * 
    * If `strict` is false the last segment is returned so long as it appears to be a valid UMI.
```

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -35,37 +35,62 @@ object Umis {
...
-    // validate the UMI
-    require(umi.nonEmpty && umi.forall { char => SequenceUtil.isValidBase(char.toByte) || SequenceUtil.isNoCall(char.toByte) || char == '-' },
-      f"UMI '$umi' is not valid (not `[ACGTNacgtn-]+`) for read: ${rec.name}"
-    )
+    rec
+  }
+
+  /**
+    * Extracts the UMI from an Illumina fastq style read name.  Illumina documents their FASTQ read names as:
+    *   @<instrument>:<run number>:<flowcell ID>:<lane>:<tile>:<x-pos>:<y-pos>:<UMI> <read>:<is filtered>:<control number>:<index>
```
nh13, line 59: UMI is optional no?  

Also document that dual-UMIs should be plus delimited

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -35,37 +35,62 @@ object Umis {
...
+      else if (colons == 7) Some(name.substring(name.lastIndexOf(delimiter) + 1, name.length))
+      else throw new IllegalArgumentException(s"Trying to extract UMI from read with ${colons + 1} parts (7-8 expected): ${name}")
+    } else {
+      val idx = name.lastIndexOf(delimiter)
+      require(idx != -1, s"Read did not have multiple '${delimiter}'-separated fields: ${name}")
+      Some(name.substring(idx + 1, name.length))
     }
-    rec
+
+    val umi   = rawUmi.map(raw => (if (raw.indexOf('+') > 0) raw.replace('+', '-') else raw).toUpperCase)
```
nh13, line 85: why upper case?  Is that strictly necessary?

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
tfenne: That's a direct copy/paste from the Illumina doc.  I'll add context but prefer not to edit the actual line.

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
tfenne: I lack faith in people/software that jam more things into read names.  What's the downside?

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
tfenne: My thought process here was 1) it's easy enough to do and 2) if it's possible then sooner or later someone will do it.  Illumina being illumina only supports UMIs in ways _their_ library preps put UMIs onto molecules.  What if someone uses two UMIs with one being in a place Illumina supports (e.g. inline) and the other being in a place they don't (e.g. inline after a skip, or as the second half of an index read)?

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
nh13: If I encode things into my UMI that are case sensitive.  No bother

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
tfenne: Realized that after I did this, the FastqRecord class splits the name and rest for me, so didn't need the split on space.  Also didn't write tests for that(!).  So I've removed that and made it just for read names again.

PR #802, opened 2022-03-03
Remove mutex for clipping options in ClipBam

nh13:
Now --clip-overlapping-reads will supersede --clip-bases-past-mate

Also, see branch name


tfenne (approved):
Lol.

PR #804, opened 2022-03-04
[docs] Fix doc comment to match actual output-type options

sstadick:
And also remove a trailing whitspace :)


nh13 (approved):
Thank-you!

PR #805, opened 2022-03-04
Personal tool to correct overlapping bases

nh13:


src/main/scala/com/fulcrumgenomics/personal/nhomer/ReadAndRefPosIterator.scala
```diff
@@ -0,0 +1,203 @@
...
+
+
+/** An iterator for each _mapped_ read base (i.e. has a corresponding reference base) that overlaps the alignment of
+  * its mate, with each value being the 1-based position in the read and reference respectively.
+  *
+  * Insertions and deletions will not be returned.  For example, an insertion consumes a read base, but has no
+  * corresponding reference base.  A deletion consumes a reference base, but has no corresponding read base.
+  *
+  * @param rec the record to iterate
+  * @param mate the mate of the record
```
kstromhaug, line 188: add @param tags for the other two parameters as well?

src/main/scala/com/fulcrumgenomics/personal/nhomer/ReadAndRefPosIterator.scala
```diff
@@ -0,0 +1,203 @@
...
+            endReadPos: Option[Int] = None
+           ): MateOverlappingReadAndRefPosIterator = new MateOverlappingReadAndRefPosIterator(
+    rec          = rec,
+    mate         = mate,
+    startReadPos = startReadPos.getOrElse(1),
+    endReadPos   = endReadPos.getOrElse(rec.length),
+  )
+}
+
+
```
kstromhaug, line 180:
```suggestion
```

src/main/scala/com/fulcrumgenomics/personal/nhomer/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,189 @@
...
+    var r1OverlappingBases = 0
+    var r2OverlappingBases = 0
+
+    // add all the bases prior to the overlapping bases
+    r1BasesBuilder.addAll(r1Bases.take(r1LastReadPos))
+    r1QualsBuilder.addAll(r1Quals.take(r1LastReadPos))
+    r2BasesBuilder.addAll(r2Bases.take(r2LastReadPos))
+    r2QualsBuilder.addAll(r2Quals.take(r2LastReadPos))
+
+    // Walk through the iterators by reference
```
kstromhaug, line 94:
```suggestion
    // Walk through the iterators by reference position
```

src/main/scala/com/fulcrumgenomics/personal/nhomer/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,189 @@
...
+  private val r2QualsBuilder = Array.newBuilder[PhredScore]
+  private val r1Builders     = Seq(r1BasesBuilder, r1QualsBuilder)
+  private val r2Builders     = Seq(r2BasesBuilder, r2QualsBuilder)
+
+  def correct(template: Template): CorrectionStats =  (template.r1, template.r2) match {
+    case (Some(r1), Some(r2)) if r1.mapped && r2.mapped && r1.matesOverlap.contains(true) => correct(r1, r2)
+    case _ => CorrectionStats(0, 0, 0, 0)
+  }
+
+  def correct(r1: SamRecord, r2: SamRecord): CorrectionStats = {
```
kstromhaug, line 64: Add docs

src/main/scala/com/fulcrumgenomics/personal/nhomer/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,189 @@
...
+    val r2Corrected = r2Bases.zip(r2.bases).count { case (left, right) => left != right }
+
+    CorrectionStats(r1OverlappingBases, r2OverlappingBases, r1Corrected, r2Corrected)
+  }
+
+  private def consensusCall(base1: Byte, qual1: PhredScore, base2: Byte, qual2: PhredScore): (Byte, PhredScore) = {
+    // Capture the raw consensus base prior to masking it to N, so that we can compute
+    // errors vs. the actually called base.
+    val (rawBase: Byte, rawQual: PhredScore) = {
+      if      (base1 == base2 && maxQualOnAgreement)   (base1,  Math.max(qual1, qual2).toByte)        // use the maximum base quality
```
kstromhaug, line 175:
```suggestion
      if      (base1 == base2 && maxQualOnAgreement)   (base1,  Math.max(qual1, qual2).toByte) // use the maximum base quality
```

src/main/scala/com/fulcrumgenomics/personal/nhomer/CorrectOverlappingBases.scala
```diff
@@ -0,0 +1,174 @@
...
+      templatesMetric += datum.templateMetric
+      basesMetric     += datum.basesMetric
+    }
+
+    Metric.write(metrics, templatesMetric, basesMetric)
+  }
+
+}
+
+case class CorrectOverlappingBasesMetric
```
kstromhaug, line 151: Perhaps a docstring here and below

kstromhaug (approved):

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
```diff
@@ -0,0 +1,182 @@
...
+    def apply(): ThreadData = ThreadData(
+      caller         = new OverlappingBasesConsensusCaller(onlyMaskDisagreements=onlyMaskDisagreements, maxQualOnAgreement=maxQualOnAgreement),
+      templateMetric = CallOverlappingConsensusBasesMetric(tpe=CountType.Templates),
+      basesMetric    = CallOverlappingConsensusBasesMetric(tpe=CountType.Bases)
+    )
+  }
+
+  override def execute(): Unit = {
+    val source           = SamSource(input)
+    val writer           = SamWriter(output, source.header)
```
tfenne, line 104: What about if `sortOrder` is specified?

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
```diff
@@ -0,0 +1,182 @@
...
+    val progress         = new ProgressLogger(logger)
+    val templateIterator = Bams.templateIterator(source)
+    val threadData       = new IterableThreadLocal(() => ThreadData())
+
+    ParIterator(templateIterator, threads=threads)
+      .map { template =>
+        val threadDatum = threadData.get()
+        threadDatum.synchronized {
```
tfenne, line 112: Why do you need synchronization here?  Aren't you only operating on a per-thread value?

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
```diff
@@ -0,0 +1,182 @@
...
+          // update metrics
+          threadDatum.templateMetric.total += 1
+          threadDatum.basesMetric.total += template.primaryReads.map(_.length).sum
```
tfenne, line 115:
```suggestion
          threadDatum.basesMetric.total += template.primaryReads.sumBy(_.length)
```

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
```diff
@@ -0,0 +1,182 @@
...
+    val templatesMetric = CallOverlappingConsensusBasesMetric(tpe=CountType.Templates)
+    val basesMetric     = CallOverlappingConsensusBasesMetric(tpe=CountType.Bases)
+    threadData.foreach { datum =>
+      templatesMetric += datum.templateMetric
+      basesMetric     += datum.basesMetric
+    }
+
+    Metric.write(metrics, templatesMetric, basesMetric)
+  }
+
```
tfenne, line 147:
```suggestion
```

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
```diff
@@ -0,0 +1,182 @@
...
+  * the [[CallOverlappingConsensusBases]] tool.
+  *
+  * @param tpe template if the counts are per template, bases if counts are in units of bases.
+  * @param total the total number of templates (bases) examined
+  * @param overlapping the total number of templates (bases) that were overlapping
+  * @param corrected the total number of templates (bases) that were corrected.
+  */
+case class CallOverlappingConsensusBasesMetric
+(
+  tpe: CountType,
```
tfenne, line 160: I prefer to use `kind` here since it's not reserved and looks better than `tpe` in tsv files.

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+
+
+/** Consensus calls the overlapping mapped bases in read pairs.
+  *
+  * This will iterate through the mapped bases that overlap between the read and mate in a read pair.  If the read and
+  * mate agree at a given reference position, then read and mate base will not change and the base quality returned
+  * is controlled by `maxQualOnAgreement`. If they disagree at a given reference position, then the base and quality
+  * returned is controlled by `onlyMaskDisagreements`.
+  *
+  * @param onlyMaskDisagreements if the read and mate bases disagree at a given reference position, true to mask (make
```
tfenne, line 41: Ditto here.  `maskDisagreements`

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+  *                              'N') the read and mate bases, otherwise pick the base with the highest base quality and
+  *                              return a base quality that's the difference between the higher and lower base quality.
+  * @param maxQualOnAgreement    if the read and mate bases agree at a given reference position, true to for the
+  *                              resulting base quality to be the maximum base quality, otherwise the sum of the base
+  *                              qualities.
+  */
+class OverlappingBasesConsensusCaller(onlyMaskDisagreements: Boolean = false,
+                                      maxQualOnAgreement: Boolean = false) {
```
tfenne, line 49: Feels like we should also have the ability to cap the resulting base quality.  Or should `maxQualOnAgreement` default to true?

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+
+  /** Consensus calls the overlapping bases if and only if the template is a paired end where both ends map with at
+    * least one base overlapping.
+    *
+    * @param template the template to potentially correct.
+    * @return summary statistics about how many bases were examined and modified
+    */
+  def call(template: Template): CorrectionStats = (template.r1, template.r2) match {
+    case (Some(r1), Some(r2)) if r1.mapped && r2.mapped && r1.matesOverlap.contains(true) => call(r1, r2)
+    case _ => CorrectionStats(0, 0, 0, 0)
```
tfenne, line 68: Maybe make a `CorrectStats.AllZeros` constant and return that here so as to avoid creating lots of these?

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  /** Consensus calls the overlapping bases if and only if both ends map with at least one base overlapping.
+    *
+    * @param r1 the first read in the pair
+    * @param r2 the second read in the pair
+    * @return summary statistics about how many bases were examined and modified
+    */
+  def call(r1: SamRecord, r2: SamRecord): CorrectionStats = {
+    require(r1.mapped && r2.mapped && r1.paired && r2.paired && r1.name == r2.name && r1.matesOverlap.contains(true))
```
tfenne, line 78: Do you also want to require that `r1.firstofPair` and`r2.secondOfPair`?

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    val r1Bases = r1.bases
+    val r1Quals = r1.quals
+    val r2Bases = r2.bases
+    val r2Quals = r2.quals
+
+    // Initialize the iters
+    val r1Iter        = MateOverlappingReadAndRefPosIterator(r1, r2).buffered
+    val r2Iter        = MateOverlappingReadAndRefPosIterator(r2, r1).buffered
```
tfenne, line 97: I was surprised to see that we needed two of these

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+    var r1LastReadPos = r1Iter.head.read - 1
+    var r2LastReadPos = r2Iter.head.read - 1
```
tfenne, line 99: Seeing this before seeing I was very confused by what `.read` is.  I would rename to `readPos` and `refPos` to make it clearer.

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+
+    // Clear and resize the builders
+    r1Builders.foreach { builder =>
+      builder.clear()
+      builder.sizeHint(r1.length)
+    }
+    r2Builders.foreach { builder =>
+      builder.clear()
+      builder.sizeHint(r2.length)
+    }
```
tfenne, line 88: I'm _so_ confused.  Why are you using builders here?  Looking at the code below it becomes really complicated because you're copying slices of the bases/quals into the builder based on what you've skipped over etc.  Why not do the edits in place in the `r1.bases` and `r1.quals` etc.?  That way you can ignore/skip over bases that are not overlapped and/or not aligned to the reference, and just tweak the bases that are aligned to the reference?  

It feels like this would be both simpler, _and_ lead to way less temporary object creation?

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+        r2QualsBuilder.addAll(r2Quals.slice(from=r2LastReadPos + 1, until=r2Head.read + 1))
+        r2LastReadPos = r2Head.read
+        r2Iter.next()
+      }
+      else { // matched reference bases, so consensus call
+        // add read bases from insertions
+        r1BasesBuilder.addAll(r1Bases.slice(from = r1LastReadPos + 1, until = r1Head.read))
+        r1QualsBuilder.addAll(r1Quals.slice(from = r1LastReadPos + 1, until = r1Head.read))
+        r2BasesBuilder.addAll(r2Bases.slice(from = r2LastReadPos + 1, until = r2Head.read))
+        r2QualsBuilder.addAll(r2Quals.slice(from = r2LastReadPos + 1, until = r2Head.read))
```
tfenne, line 130: These are invoked every time as far as I can see, even when the last base was also mapped to the reference.  So many empty slices created.

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+
+        r1OverlappingBases += 1
+        r2OverlappingBases += 1
```
tfenne, line 133: Are these _ever_ incremented separately?

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
tfenne: Having gone through the code I now understand what this does, and the documentation is _really_ confusing to me.  What you're really saying, I think is:

`If true: any time bases disagree mask them to N with qual=2, and any time bases agree pass the bases and quality scores through unmodified.  If false: when bases disagree set the output base to be the higher quality base and the qual the delta between the quality scores, and when bases agree adjust the base quality based on `maxQualOnAgreement`.

I think this would make a lot more sense to rework the options as two "strategies":

1. `disagreementStrategy`: one of `MaskBoth`, `MaskLowerQual`, `Consensus`
2. `agreementStrategy`: one of `Consensus`, `MaxQual`, `PassThrough`

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -0,0 +1,250 @@
...
+
+    val r1Corrected = r1Bases.zip(r1.bases).count { case (left, right) => left != right }
+    val r2Corrected = r2Bases.zip(r2.bases).count { case (left, right) => left != right }
+
+    CorrectionStats(r1OverlappingBases, r2OverlappingBases, r1Corrected, r2Corrected)
+  }
+
+  private def consensusCall(base1: Byte, qual1: PhredScore, base2: Byte, qual2: PhredScore): (Byte, PhredScore) = {
+    // Capture the raw consensus base prior to masking it to N, so that we can compute
+    // errors vs. the actually called base.
```
tfenne, line 188: I tend to think that if _either_ base is already a no-call then we should probably do _nothing_ as we don't actually have a second base call (not convert the output base to a no-call).  That has ramifications beyond this method though, so perhaps we should only call this if both bases are [ACGT]?

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
```diff
@@ -0,0 +1,182 @@
...
+            if (correctedBases > 0) {
+              threadDatum.templateMetric.corrected += 1
+              threadDatum.basesMetric.corrected += correctedBases
+            }
+          }
+        }
+        template.allReads.foreach(progress.record)
+        template
+      }.toAsync(ParIterator.DefaultChunkSize * 8).foreach { template =>
+      writer ++= template.allReads
```
tfenne, line 132: I'm realizing that doing this invalidates the NM and MD tags on reads that we edit.  Should we remove those tags?  Recalculate?  Just document this fact?

src/main/scala/com/fulcrumgenomics/bam/ReadAndRefPosIterator.scala
```diff
@@ -0,0 +1,202 @@
...
+  * 1-based position in the read and reference respectively.
+  *
+  * Insertions and deletions will not be returned.  For example, an insertion consumes a read base, but has no
+  * corresponding reference base.  A deletion consumes a reference base, but has no corresponding read base.
+  *
+  * @param rec the record over which read and reference positions should be returned
+  * @param startReadPos the 1-based start position in the read
+  * @param endReadPos the 1-based inclusive end position in the read
+  * @param startRefPos the 1-based start position in the reference
+  * @param endRefPos the 1-based inclusive end position in the reference
```
tfenne, line 44: I think these need to be clearly documents as filters on what is returned by the iterator, i.e. it's not up to the user to calculate these correctly, just to use them to limit the set of values returned by the iterator, right?
Should they then be renamed to `[min|max][Read|Ref]pos`?  And perhaps have default values of 1 and MaxInt?

kstromhaug (approved):
LGTM

src/main/scala/com/fulcrumgenomics/bam/ReadAndRefPosIterator.scala
```diff
@@ -0,0 +1,202 @@
...
+  * */
+class ReadAndRefPosIterator (rec: SamRecord,
```
tfenne, line 46: This seems really complicated.  I'm curious why you decided to reimplement from scratch rather than doing something like:

```scala
val blocks = rec.asSam.getAlignmentBlocks.iterator.filter(b => /* block overlaps read and reference windows given */)
var currentBlock: Option[AlignmentBlock] = blocks.nextOption
var currentOffset: Int = 0
```

src/main/scala/com/fulcrumgenomics/bam/ReadAndRefPosIterator.scala
```diff
@@ -0,0 +1,202 @@
...
+  }
+}
+
+object ReadAndRefPosIterator {
+  /** Stores a 1-based position in the read and reference
+    *
+    * @param read 1-based position in the read
+    * @param ref 1-based position in the reference
+    */
+  case class ReadAndRefPos(read: Int, ref: Int)
```
tfenne, line 142: read -> readPos, ref -> refPos

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -105,7 +106,23 @@ class CallDuplexConsensusReads
...
+     |quality.
+  """)
+ val onlyMaskDisagreements: Boolean = false,
+ @arg(doc=
+  """
+    |When `--consensus-call-overlapping-bases` is being used, if the read and mate bases agree at a given reference
+    |position, true to for the resulting base quality to be the maximum base quality, otherwise the sum of the base
+    |qualities.
+    |""")
+ val maxQualOnAgreement: Boolean = false
```
tfenne, line 125: I don't think we should expose these here.  We should pick sensible ways of doing it for doing it inside the caller, and if you want full control use the standalone tool.

src/main/scala/com/fulcrumgenomics/umi/CallMolecularConsensusReads.scala
```diff
@@ -123,7 +125,23 @@ class CallMolecularConsensusReads
...
+    |quality.
+  """)
+ val onlyMaskDisagreements: Boolean = false,
+ @arg(doc=
+   """
+     |When `--consensus-call-overlapping-bases` is being used, if the read and mate bases agree at a given reference
+     |position, true to for the resulting base quality to be the maximum base quality, otherwise the sum of the base
+     |qualities.
+     |""")
+ val maxQualOnAgreement: Boolean = false
```
tfenne, line 144: Ditto.

tfenne (changes requested):

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
```diff
@@ -0,0 +1,182 @@
...
+  """
+    |Consensus calls overlapping bases in read pairs.
+    |
+    |## Inputs and Outputs
+    |
+    |In order to correctly correct reads by template, the input BAM must be either `queryname`  or `query` grouped.  The
+    |sort can be done in streaming fashion with:
+    |
+    |```
+    |samtools sort -n -u in.bam | fgbio ClipBam -i /dev/stdin ...
```
nh13:
```suggestion
    |fgbio SortBam -i in.bam -o out.bam -s queryname | fgbio ClipBam -i /dev/stdin ...
```

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
```diff
@@ -0,0 +1,182 @@
...
+  """
+    |Consensus calls overlapping bases in read pairs.
+    |
+    |## Inputs and Outputs
+    |
+    |In order to correctly correct reads by template, the input BAM must be either `queryname` sorted or `query` grouped.  The
+    |sort can be done in streaming fashion with:
+    |
+    |```
+    |samtools sort -n -u in.bam | fgbio CallOverlappingConsensusBases -i /dev/stdin ...
```
nh13, line 51:
```suggestion
    |fgbio --compression 0 SortBam -i in.bam -o out.bam -s queryname | fgbio CallOverlappingConsensusBases -i /dev/stdin ...
```

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
nh13: s

src/main/scala/com/fulcrumgenomics/bam/CallOverlappingConsensusBases.scala
nh13: Recalculate isn't too hard these days.  I'l fix it.

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
nh13: done

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
nh13: Yeah, I think simplifying it to just have a `ReadAndRefPosIterator`, since that's essentially what they are, is fine.

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
nh13: Removed.  Hopefully the defaults are sensible.

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
nh13: it is done.

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
nh13: I don't think it this is relevant anymore after implementing agreement and disagreement strategies

nh13:
@tfenne back to you.

1. There's the question about capping the base quality coming out, which I dont' think is needed now that we have agreement and disagreement strategies:
> Wait - so the default is to output Q60 if the two inputs are Q30? Should there be some cap like in the main consensus callers?

We don't have an explicit cap in the main consensus callers, except to make sure it doesn't go beyond 93 or whatever the htsjdk maximum value is.  The other callers have an effective cap due to the pre and post labeling error rates, which we aren't using in this simplified model.  I could switch over to ConsensusCaller and have the pre-error rate be effectively zero, and the post-error rate by 40 or whatever the default is, but is it worth it?

2. fixed everything else you asked for _except_ your [comment here](https://github.com/fulcrumgenomics/fgbio/pull/805/#discussion_r834714385). 

See this comment in SamRecords that you wrote on 5/19/2017:

> // TODO long-term: replace these two methods with methods on [[Cigar]] to save creating alignment blocks in memory

I took that to heart and started an PR for replacing those methods in #809, but didn't end up needing them here, since it was rather than parsing and iterating through the cigar for each overlapped read/reference position, it was more efficient to create iterators (one for a read, one for a mate) that kept track of where it was in the cigar for each read and returned the desired overlapping read/references positions.  Yes it's complex, but this booking keeping is hard!

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -55,7 +55,7 @@ import scala.collection.immutable.IndexedSeq
     |done in streaming fashion with, for example:
     |
     |```
-    |samtools sort -n -u in.bam | fgbio ClipBam -i /dev/stdin ...
+    |fgbio --compression 0 SortBam -i in.bam -o out.bam -s queryname | fgbio ClipBam -i /dev/stdin ...
```
tfenne, line 58: Can go back to samtools since beecb49f2b3472d2e8589460b06c7f9753781ef2 (here and elsewhere).

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -76,152 +70,104 @@ class OverlappingBasesConsensusCaller(onlyMaskDisagreements: Boolean = false,
...
+            base1 = base1,
+            qual1 = r1.quals(r1Head.readPos - 1),
+            base2 = base2,
+            qual2 = r2.quals(r2Head.readPos - 1)
+          )
+
+          if (newBase1 != NoCall && newBase1 != base1) r1Corrected += 1
+          if (newBase2 != NoCall && newBase2 != base2) r2Corrected += 1
+
+          // Only add the consensus call if we aren't masking disagreements or its a no-call
```
tfenne, line 115: Maybe kill this comment?  I think `consensusCall` always returns the bases/quals you want to use, and there's no conditional/guard here.

src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -241,10 +187,86 @@ object OverlappingBasesConsensusCaller {
...
+      (NoCall, NoCallQual, NoCall, NoCallQual)
+    }
+  }
+
+  /** Mask the base with the lowest base quality, with the other base unchanged.  If the base qualities are the same,
+    * mask both bases. */
+  case object MaskLowerQual extends DisagreementStrategy {
+    def baseAndQual(base1: Byte, qual1: PhredScore, base2: Byte, qual2: PhredScore): (Byte, PhredScore, Byte, PhredScore) = {
+      //                       r1-base  r1-qual                        r2-base  r2-qual
+      if (qual1 > qual2)      (base1,   PhredScore.cap(qual1 - qual2), NoCall,  NoCallQual)
```
tfenne, line 251: From the docs I was thinking this would leave the qual1 unchanged, not subtract the qual2 from it.  Thoughts?  I think either way is fine, but the documentation should be made clearer if you leave the code this way.

tfenne (approved):
Some minor comments.  Please merge when done.

PR #806, opened 2022-03-07
Fix most scala 2.13 deprecation warnings.

tfenne:
Mostly just adding and removing `()` in a bunch of places, plus a handful of other things.


nh13 (approved):
Very nice cleanup!

PR #807, opened 2022-03-09
Require template sort orders in ClipBam and FilterConsensusReads

nh13:
This requires that the input BAM files are queryname sorted or query
    grouped in both tools.  Supporting the sorting of the input is now
    removed.  Use `samtools sort -n -u in.bam | fgbio <tool> -i /dev/stdin`
    instead.  The header of the input BAM must properly specify one of these
    two sort orders.
    
    Adds a method to Bams assert the header is queryname sorted or query
    grouped.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
@@ -26,16 +26,19 @@ package com.fulcrumgenomics.bam
 
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.bam.api.SamOrder.Queryname
-import com.fulcrumgenomics.bam.api.{SamOrder, SamRecord, SamRecordCodec, SamSource}
+import com.fulcrumgenomics.bam.api._
 import com.fulcrumgenomics.commons.collection.{BetterBufferedIterator, SelfClosingIterator}
+import com.fulcrumgenomics.commons.io.Writer
 import com.fulcrumgenomics.commons.util.LazyLogging
+import com.fulcrumgenomics.fasta.ReferenceSequenceIterator
 import com.fulcrumgenomics.util.{Io, ProgressLogger, Sorter}
 import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
 import htsjdk.samtools.SamPairUtil.PairOrientation
 import htsjdk.samtools._
-import htsjdk.samtools.reference.{ReferenceSequence, ReferenceSequenceFileWalker}
+import htsjdk.samtools.reference.ReferenceSequence
 import htsjdk.samtools.util.{CloserUtil, CoordMath, SequenceUtil}
 
+import java.io.Closeable
 import scala.math.{max, min}
 
 /**
@@ -467,7 +470,6 @@ object Bams extends LazyLogging {
     (min(firstEnd,secondEnd), max(firstEnd,secondEnd))
   }
 
-
   /** If the read is mapped in an FR pair, returns the distance of the position from the other end
     * of the template, other wise returns None.
     *
@@ -488,6 +490,37 @@ object Bams extends LazyLogging {
     }
     else {
       None
+    }
+  }
+
+  /** Requires that the header is queryname sorted or query grouped. */
+  def requireTemplateGrouped(header: SAMFileHeader, toolName: String): Unit = {
+    require(header.getSortOrder == SortOrder.queryname || header.getGroupOrder == GroupOrder.query,
+      "Input was not queryname sorted or query grouped, found: " +
+        s"SO:${header.getSortOrder} GO:${header.getGroupOrder}" +
+        Option(header.getAttribute("SS")).map(ss => f" SS:$ss").getOrElse("") +
+        f". Use `samtools sort -n -u in.bam | fgbio $toolName -i /dev/stdin`"
+    )
+  }
+
+  /** Builds a [[Writer]] of [[SamRecord]]s that regenerates the NM, UQ, and MD tags using the given map of reference
+    * sequences.
+    *
+    * @param writer the writer to write to
+    * @param ref the path to the reference FASTA
+    */
+  def regenerateNmUqMdTagsWriter(writer: SamWriter, ref: PathToFasta): Writer[SamRecord] with Closeable = {
+    logger.info("Reading the reference fasta into memory")
+    val refMap = ReferenceSequenceIterator(ref, stripComments=true).map { ref => ref.getContigIndex -> ref}.toMap
+    logger.info(f"Read ${refMap.size}%,d contigs.")
+
+    // Create the final writer based on if the full reference has been loaded, or not
+    new Writer[SamRecord] with Closeable {
+      override def write(rec: SamRecord): Unit = {
+        Bams.regenerateNmUqMdTags(rec, refMap(rec.refIndex))
+        writer += rec
+      }
+      def close(): Unit = writer.close()
     }
   }
 }
--- a/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
@@ -493,13 +493,19 @@ object Bams extends LazyLogging {
     }
   }
 
+  /** Returns true if the header is queryname sorted or query grouped */
+  def isQueryGrouped(header: SAMFileHeader): Boolean = {
+    header.getSortOrder == SortOrder.queryname || header.getGroupOrder == GroupOrder.query
+  }
+
   /** Requires that the header is queryname sorted or query grouped. */
-  def requireTemplateGrouped(header: SAMFileHeader, toolName: String): Unit = {
-    require(header.getSortOrder == SortOrder.queryname || header.getGroupOrder == GroupOrder.query,
+  def requireQueryGrouped(header: SAMFileHeader, toolName: String = "<tool>", path: Option[PathToBam] = None): Unit = {
+    require(isQueryGrouped(header),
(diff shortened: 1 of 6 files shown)


src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -490,4 +492,90 @@ object Bams extends LazyLogging {
       None
     }
   }
+
+
+  /** Builds a [[SamWriter]] assuming that input reads will be passed through [[Bams.templateIterator()]] then into this
+    * writer.
+    *
+    * The [[Bams.templateIterator()]] will not resort the input reads if the input file is [[SamOrder.Queryname]] or
```
nh13, line 500:
```suggestion
    * The [[Bams.templateIterator()]] will not re-sort the input reads if the input file is [[SamOrder.Queryname]] or
```

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -490,4 +492,90 @@ object Bams extends LazyLogging {
...
+    * @param inHeader the SAM header from the input file
+    * @param outOrder the desired output sort order, if any
+    * @param ref the path to the reference FASTA, used for CRAM output
+    * @param maxRecordsInMemoryWhenSorting the maximum number of records to keep in memory when sorting
+    */
+  def writerFromTemplateIterator(output: PathToBam,
+                                 inHeader: SAMFileHeader,
+                                 outOrder: Option[SamOrder] = None,
+                                 ref: Option[PathToFasta] = None,
+                                 maxRecordsInMemoryWhenSorting: Int = 250e3.toInt
```
nh13, line 521: previously in `FilterConsensusReads`, but happy to set it to something else

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -490,4 +492,90 @@ object Bams extends LazyLogging {
...
+                                ): SamWriter = {
+    val outHeader    = inHeader.clone()
```
nh13, line 523: this is extracted from `FilterConsensusReads`

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -490,4 +492,90 @@ object Bams extends LazyLogging {
...
+    writer
+  }
+
+  /** Builds a [[Writer]] of [[SamRecord]]s that regenerates the NM, UQ, and MD tags using the given map of reference
+    * sequences.
+    *
+    * @param writer the writer to write to
+    * @param ref the path to the reference FASTA
+    */
+  def regenerateNmUqMdTagsWriter(writer: SamWriter, ref: PathToFasta): Writer[SamRecord] with Closeable = {
```
nh13, line 567: ditto, pulled out from `FilterConsensusReads` and happy to write tests

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -79,7 +77,9 @@ class ClipBam
   @arg(          doc="Require at least this number of bases to be clipped on the 5' end of R2") val readTwoFivePrime: Int  = 0,
   @arg(          doc="Require at least this number of bases to be clipped on the 3' end of R2") val readTwoThreePrime: Int = 0,
   @arg(          doc="Clip overlapping reads.") val clipOverlappingReads: Boolean = false,
-  @arg(          doc="Clip reads in FR pairs that sequence past the far end of their mate.") val clipBasesPastMate: Boolean = false
+  @arg(          doc="Clip reads in FR pairs that sequence past the far end of their mate.") val clipBasesPastMate: Boolean = false,
+  @arg(flag='S', doc="The sort order of the output. If not given, output will be in the same order as input if the input is query name sorted or query grouped, otherwise queryname order.")
+  val sortOrder: Option[SamOrder] = None
```
nh13, line 82: breaking change: previously the output was allows coordinate, now it's whatever the input was (or this)

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -228,63 +221,16 @@ class FilterConsensusReads
     logger.info(f"Masked $maskedBases%,d of $totalBases%,d bases in retained primary consensus reads.")
   }
 
-  /** Builds the writer to which filtered records should be written.
```
nh13, line 231: moved to `Bams`

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -228,65 +229,6 @@ class FilterConsensusReads
...
-    *  If the input order is [[SamOrder.Queryname]] or query grouped, then the filtered records will also be in the same
-    *  order.  So if the output order is specified AND does not match the the input order, sorting will occur.
-    *
-    *  If the input order is not [[SamOrder.Queryname]] or query grouped, then the input records will be resorted into
-    *  [[SamOrder.Queryname]].  So if the output order is specified AND is not [[SamOrder.Queryname]], sorting will occur.
-    *
-    *  Otherwise, we can skip sorting!
-    *
-    * */
-  private def buildOutputWriter(inHeader: SAMFileHeader, refMap: Map[Int, ReferenceSequence]): Writer[SamRecord] with Closeable = {
```
nh13, line 242: removed

src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -228,65 +229,6 @@ class FilterConsensusReads
...
-            Some(order)
-          }
-      }
-    }
-
-    val writer = SamWriter(output, outHeader, sort=writerSortOrder, maxRecordsInRam=MaxRecordsInMemoryWhenSorting)
-    writerSortOrder.foreach(o => logger.info(f"Output will be sorted into $o order"))
-
-    // Create the final writer based on if the full reference has been loaded, or not
-    new Writer[SamRecord] with Closeable {
```
nh13, line 281: this moved to Bams

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -51,11 +50,17 @@ import scala.collection.immutable.IndexedSeq
...
+    |sort can be done in streaming fashion with:
+    |
+    |```
+    |samtools sort -n -u in.bam | fgbio ClipBam -i /dev/stdin ...
+    |```
+    |
+    |The output sort order may be specified with `--sort-order`.  If not given, then the output will be in the same
+    |order as input.
+    |
+    |Any existing `NM`, `UQ` and `MD` tags are repaired.
```
tfenne, line 63:
```suggestion
    |Any existing `NM`, `UQ` and `MD` tags are repaired, and mate-pair information updated.
```

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -119,24 +130,12 @@ class ClipBam
       }
 
       template.allReads.foreach { r =>
-        sorter += r
+        out += r
         progress.record(r)
```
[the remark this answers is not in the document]
tfenne: I can't put it as a suggestion due to the deleted line, but I think this can now be `out ++=template.allReads`

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -490,4 +492,35 @@ object Bams extends LazyLogging {
...
+    logger.info("Reading the reference fasta into memory")
+    val refMap = ReferenceSequenceIterator(ref, stripComments=true).map { ref => ref.getContigIndex -> ref}.toMap
+    logger.info(f"Read ${refMap.size}%,d contigs.")
```
tfenne, line 515: Can we bump these to debug?  I don't think logging statements like this really belong in a utility function, so I would prefer to remove or at least make debug.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
tfenne: How about `nmUqMdTagRegeneratingWriter()`?  I think it needs to be `{adjective}Writer`.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -490,4 +492,35 @@ object Bams extends LazyLogging {
       None
     }
   }
+
+  /** Requires that the header is queryname sorted or query grouped. */
+  def requireTemplateGrouped(header: SAMFileHeader, toolName: String): Unit = {
+    require(header.getSortOrder == SortOrder.queryname || header.getGroupOrder == GroupOrder.query,
```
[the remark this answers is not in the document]
nh13: 1. I will make the tool name a `Path` optional and add them to the message.  
2. The impetus for having this method here was to NOT have to re-write this error message everywhere.  I would rather have this in a single place.  I will a `isQueryGrouped: Boolean` method and leave this here.

I think both are good compromises given I am in the opposite camp.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
nh13: Ok.  It does take a while to load some large references, so it's nice to have to not duplicate logging messages everywhere. But here it makes more sense to move them into the tools

nh13:
@tfenne disagreed with you on moving the `require` in `requireQueryGrouped`, but the rest I changed.

tfenne (approved):

PR #809, opened 2022-03-10
Add scala implementations for some SamRecord methods

nh13:
* refPosAtReadPos
* readPosAtRefPos

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -270,7 +270,7 @@ trait SamRecord {
         val end = readPos + elem.length - 1
         if (pos <= end) {
           if (!elem.operator.consumesReferenceBases()) {
-            if (returnLastBaseIfInserted) {
+            if (returnLastBaseIfInserted && elem.operator == CigarOperator.INSERTION) {
               result = Some(refPos - 1)
             }
           }
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -234,9 +234,107 @@ trait SamRecord {
   // transient attributes
   @inline final val transientAttrs: TransientAttrs = new TransientAttrs(this)
 
-  // TODO long-term: replace these two methods with methods on [[Cigar]] to save creating alignment blocks in memory
-  @inline final def refPosAtReadPos(pos: Int) = getReferencePositionAtReadPosition(pos)
-  @inline final def readPosAtRefPos(pos: Int, returnLastBaseIfDeleted: Boolean) = getReadPositionAtReferencePosition(pos, returnLastBaseIfDeleted)
+  @deprecated(message="Use referencePosAtReadPos instead", since="2.0.0")
+  @inline final def refPosAtReadPos(pos: Int): Int = referencePosAtReadPos(pos).getOrElse(0)
+  @deprecated(message="Use readPosAtReferencePos instead", since="2.0.0")
+  @inline final def readPosAtRefPos(pos: Int, returnLastBaseIfDeleted: Boolean): Int = {
+    readPosAtReferencePos(pos, returnLastBaseIfDeleted=returnLastBaseIfDeleted).getOrElse(0)
+  }
+
+  /** Returns the 1-based reference position for the 1-based position in the read, or [[scala.None]] if
+    * the given base in the read does not align to a reference base (e.g. due to soft-clipping,
+    * or an insertion).
+    *
+    * If the requested read position is less than or equal to zero, or greater than the read length [[com.fulcrumgenomics.bam.api.SamRecord.length]], then
+    * an exception will be thrown.
+    *
+    * Hard-clipping (`H`) is ignored and not counted as part of the read.
+    *
+    * @param pos the 1-based read position to query
+    * @param returnLastBaseIfInserted if the reference is an insertion, true to return the previous reference base,
+    *                                 false to return None
+    * */
+  @inline final def referencePosAtReadPos(pos: Int, returnLastBaseIfInserted: Boolean = false): Option[Int] = {
+    require(this.mapped, s"read was not mapped: ${this}")
+    require(1 <= pos, s"position given '$pos' was less than one: ${this}")
+    require(pos <= cigar.lengthOnQuery , s"position given '$pos' was longer than the # of read bases '${cigar.lengthOnQuery}': ${this}")
+
+    val elems   = this.cigar.elems.iterator
+    var readPos = 1
+    var refPos  = this.start
+    var result: Option[Int] = None
+
+    while (elems.hasNext && readPos <= pos) {
+      val elem = elems.next()
+      if (elem.operator.consumesReadBases()) {
+        val end = readPos + elem.length - 1
+        if (pos <= end) {
+          if (!elem.operator.consumesReferenceBases()) {
+            if (returnLastBaseIfInserted) {
+              result = Some(refPos - 1)
+            }
+          }
+          else {
+            result = Some(refPos + (pos - readPos))
+          }
+        }
+      }
+
+      readPos += elem.lengthOnQuery
+      refPos += elem.lengthOnTarget
+    }
+
+    result
+  }
+
+  /** Returns the 1-based read position for the 1-based position in the reference, or [[scala.None]] if
+    * the given base in the read does not align to a read base (e.g. due to soft-clipping,
+    * or a deletion).
+    *
+    * @param pos the 1-based reference position to query
+    * @param returnLastBaseIfDeleted if the reference is a deletion, true to return the previous read base, false to
+    *                                return None
+    * @param returnLastBaseIfSkipped if the reference is a skip, true to return the previous read base, false to
+    *                                return None
+    * */
+  @inline final def readPosAtReferencePos(pos: Int,
+                                          returnLastBaseIfDeleted: Boolean = false,
+                                          returnLastBaseIfSkipped: Boolean = false): Option[Int] = {
+    require(this.mapped, s"read was not mapped: ${this}")
+    require(this.start <= pos, s"position given '$pos' was before the start of the alignment '${this.start}': ${this}")
+    require(pos <= this.end , s"position given '$pos' was past the end of the alignment '${this.end}': ${this}")
+
+    val elems   = this.cigar.elems.iterator
+    var readPos = 1
+    var refPos  = this.start
+    var result: Option[Int] = None
+
+    while (elems.hasNext && refPos <= pos) {
+      val elem = elems.next()
+      if (elem.operator.consumesReferenceBases()) {
+        val end = refPos + elem.length - 1
+        if (pos <= end) {
+          if (!elem.operator.consumesReadBases()) {
+            if (returnLastBaseIfDeleted && elem.operator == CigarOperator.DELETION) {
+              result = Some(readPos - 1)
+            }
+            else if (returnLastBaseIfSkipped && elem.operator == CigarOperator.SKIPPED_REGION) {
+              result = Some(readPos - 1)
+            }
+          }
+          else {
+            result = Some(readPos + (pos - refPos))
+          }
+        }
+      }
+
+      readPos += elem.lengthOnQuery
+      refPos += elem.lengthOnTarget
+    }
+
+    result
+  }
+
 
   ////////////////////////////////////////////////////////////////////////////
   // Non-wrapper methods.
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -254,7 +254,7 @@ trait SamRecord {
     * @param returnLastBaseIfInserted if the reference is an insertion, true to return the previous reference base,
     *                                 false to return None
     * */
-  @inline final def referencePosAtReadPos(pos: Int, returnLastBaseIfInserted: Boolean = false): Option[Int] = {
+  final def referencePosAtReadPos(pos: Int, returnLastBaseIfInserted: Boolean = false): Option[Int] = {
     require(this.mapped, s"read was not mapped: ${this}")
     require(1 <= pos, s"position given '$pos' was less than one: ${this}")
     require(pos <= cigar.lengthOnQuery , s"position given '$pos' was longer than the # of read bases '${cigar.lengthOnQuery}': ${this}")
@@ -297,9 +297,9 @@ trait SamRecord {
     * @param returnLastBaseIfSkipped if the reference is a skip, true to return the previous read base, false to
     *                                return None
     * */
-  @inline final def readPosAtReferencePos(pos: Int,
-                                          returnLastBaseIfDeleted: Boolean = false,
-                                          returnLastBaseIfSkipped: Boolean = false): Option[Int] = {
+  final def readPosAtReferencePos(pos: Int,
+                                  returnLastBaseIfDeleted: Boolean = false,
+                                  returnLastBaseIfSkipped: Boolean = false): Option[Int] = {
     require(this.mapped, s"read was not mapped: ${this}")
     require(this.start <= pos, s"position given '$pos' was before the start of the alignment '${this.start}': ${this}")
     require(pos <= this.end , s"position given '$pos' was past the end of the alignment '${this.end}': ${this}")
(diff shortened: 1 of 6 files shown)


src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
   // transient attributes
   @inline final val transientAttrs: TransientAttrs = new TransientAttrs(this)
 
-  // TODO long-term: replace these two methods with methods on [[Cigar]] to save creating alignment blocks in memory
-  @inline final def refPosAtReadPos(pos: Int) = getReferencePositionAtReadPosition(pos)
-  @inline final def readPosAtRefPos(pos: Int, returnLastBaseIfDeleted: Boolean) = getReadPositionAtReferencePosition(pos, returnLastBaseIfDeleted)
```
nh13, line 241: these are breaking changes to the API.  I like returning `Option[Int]` better, so I can return `None` instead of `0`

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+  /** Returns the 1-based reference position for the 1-based position in the read, or [[None]] if the reference does
+    * not map at the read position (e.g. insertion).
+    *
+    * @param pos the 1-based read position to query
+    * @param returnLastBaseIfInserted if the reference is an insertion, true to returned the previous reference base,
+    *                                 false to return None
+    * */
+  @inline final def refPosAtReadPos(pos: Int, returnLastBaseIfInserted: Boolean = false): Option[Int] = if (pos <= 0) None else {
```
nh13, line 246: `returnLastBaseIfInserted` is a new parameter, which htsjdk doesn't even have!

src/test/scala/com/fulcrumgenomics/bam/api/SamRecordTest.scala
```diff
@@ -289,4 +289,48 @@ class SamRecordTest extends UnitSpec with OptionValues {
     rec1.matesOverlap shouldBe None // Mate's start is not enclosed by rec, and mate's end cannot be determined
     rec2.matesOverlap.value shouldBe true // Mate's start is enclosed by rec, regardless of where mate end is
   }
+
```
nh13, line 292: tests copied from htsjdk, with a few added

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+    var readPos      = 1
+    var refPos       = this.start
+    var elementIndex = 0
+    val elems        = this.cigar.elems
+
+    def continue(): Boolean = if (elementIndex >= elems.length) false else {
+      val elem = elems(elementIndex)
+      if (pos > CoordMath.getEnd(readPos, elem.lengthOnQuery)) true // current cigar element is before the desired position
+      else {
+        if (pos < readPos) readPos = 0 // past the point in the read
```
tfenne, line 256: How is this possible?  We're walking down the read, and not skipping any elements, so how could we have shot past the pos that we are trying to find?

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
-  @inline final def readPosAtRefPos(pos: Int, returnLastBaseIfDeleted: Boolean) = getReadPositionAtReferencePosition(pos, returnLastBaseIfDeleted)
+  /** Returns the 1-based reference position for the 1-based position in the read, or [[None]] if the reference does
+    * not map at the read position (e.g. insertion).
+    *
+    * @param pos the 1-based read position to query
+    * @param returnLastBaseIfInserted if the reference is an insertion, true to returned the previous reference base,
+    *                                 false to return None
+    * */
+  @inline final def refPosAtReadPos(pos: Int, returnLastBaseIfInserted: Boolean = false): Option[Int] = if (pos <= 0) None else {
+    var readPos      = 1
```
tfenne, line 247: I don't think this will handle leading soft-clipping correctly.  I believe it will return `rec.start` for any leading soft-clipped read position, and that seems wrong.

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+    def continue(): Boolean = if (elementIndex >= elems.length) false else {
+      val elem = elems(elementIndex)
+      if (pos > CoordMath.getEnd(readPos, elem.lengthOnQuery)) true // current cigar element is before the desired position
+      else {
+        if (pos < readPos) readPos = 0 // past the point in the read
+        else if (elem.operator == CigarOperator.INSERTION) {
+          if (!returnLastBaseIfInserted) refPos = 0
+          else refPos -= 1
+        } // in an insertion, no reference position
+        else refPos += pos - readPos  // get the offset
```
tfenne, line 261: I think you need to also handle soft-clipping here to return None.

tfenne: And I think before returning false you need to also check that the cigar operator consumes query bases.  E.g. I think if you have a cigar like `20M10D20M` with `rec.start == 100` and you ask for `rec.refPosAtReadPos(21)` I think this will erroneously return 121 instead of 131.

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+    val elems        = this.cigar.elems
+
+    def continue(): Boolean = if (elementIndex >= elems.length) false else {
+      val elem = elems(elementIndex)
+      if (pos > CoordMath.getEnd(readPos, elem.lengthOnQuery)) true // current cigar element is before the desired position
+      else {
+        if (pos < readPos) readPos = 0 // past the point in the read
+        else if (elem.operator == CigarOperator.INSERTION) {
+          if (!returnLastBaseIfInserted) refPos = 0
+          else refPos -= 1
```
tfenne, line 259: Could you add comments here - presumably setting refPos to 0 is your way of saying "return None"?

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+    }
+
+    while (continue()) {
+      val elem      = elems(elementIndex)
+      refPos       += elem.lengthOnTarget
+      readPos      += elem.lengthOnQuery
+      elementIndex += 1
+    }
+
+    if (this.start <= refPos && readPos <= this.end) Some(refPos) else None
```
tfenne, line 273: Do I also need to tweet to get you to stop using yoda conditionals (in a language where it can't go wrong with a typo)?

```suggestion
    if (refPos >= this.start && refPos <= this.end) Some(refPos) else None
```

I think this check was also wrong - why are you comparing `readPos` which is in the range 1..rec.length against `this.end` which is a reference position?

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+      var readPos      = 0
+      var refPos       = this.start
+      var elementIndex = 0
+      val elems        = this.cigar.elems
+
+      def continue(): Boolean = if (elementIndex >= elems.length) false else {
+        val elem = elems(elementIndex)
+        if (pos > CoordMath.getEnd(refPos, elem.lengthOnTarget)) true // current cigar element is before the desired position
+        else { // overlaps!
+          if (elem.operator == CigarOperator.DELETION) { // deletion
```
tfenne, line 296: Should we also handle CigarOperator.SKIP?  I'm honestly not sure.

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
tfenne, line 246: In the other method you check for `rec.mapped`, why not do that here also?

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+      readPos      += elem.lengthOnQuery
+      elementIndex += 1
+    }
+
+    if (this.start <= refPos && readPos <= this.end) Some(refPos) else None
+  }
+
+
+  /** Returns the 1-based position into the read's bases where the position is mapped either as a match or mismatch,
+    * [[None]] if the read does not map the position.
```
tfenne, line 278:
```suggestion
  /** Returns the 1-based position into the read's bases corresponding to the reference position. 
    * If `returnLastBaseIfDeleted` is false (default) then a position is only returned if a read base
    * is directly aligned to the reference position.  If `returnLastBaseIfDeleted` is true and the
    * reference base is deleted in the read, the read-base immediately prior to the deletion is
    * returned.
    */
```

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+    *
+    * @param pos the 1-based reference position to query
+    * @param returnLastBaseIfDeleted if the reference is a deletion, true to returned the previous read base, false to
+    *                                return None
+    * */
+  @inline final def readPosAtRefPos(pos: Int, returnLastBaseIfDeleted: Boolean = false): Option[Int] = {
+    if (this.unmapped || pos < this.start || this.end < pos) None
```
tfenne, line 285: Should this be a require() instead?
```suggestion
    if (this.unmapped || pos < this.start || pos > this.end) None
```

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+
+
+  /** Returns the 1-based position into the read's bases where the position is mapped either as a match or mismatch,
+    * [[None]] if the read does not map the position.
+    *
+    * @param pos the 1-based reference position to query
+    * @param returnLastBaseIfDeleted if the reference is a deletion, true to returned the previous read base, false to
+    *                                return None
+    * */
+  @inline final def readPosAtRefPos(pos: Int, returnLastBaseIfDeleted: Boolean = false): Option[Int] = {
```
tfenne, line 284: I think here and in the other method we should rename the parameter to `returnPriorBaseIf[Deleted]` as I _think_ `prior` is clearer that we're talking about the base prior to the one requested, vs. `last` could imply the last base in the read.

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+    if (this.unmapped || pos < this.start || this.end < pos) None
+    else { // overlaps
+      var readPos      = 0
```
tfenne, line 287: Why not 1 since we're returning 1-based?

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -236,9 +236,83 @@ trait SamRecord {
...
+        if (pos > CoordMath.getEnd(refPos, elem.lengthOnTarget)) true // current cigar element is before the desired position
+        else { // overlaps!
+          if (elem.operator == CigarOperator.DELETION) { // deletion
+            if (!returnLastBaseIfDeleted) readPos = 0 // don't return a read position, so zero out the current read position
+          } else {
+            readPos += pos - refPos + 1 // get the offset
+          }
+          false
+        }
+      }
```
tfenne, line 303: I think this implementation has the same problem with clipping.  If I have a read that has `rec.start = 100` and cigar `10S90M` and I request `readPosAtRefPos(100)` I think this will erroneously return `1` and not `11`.

src/test/scala/com/fulcrumgenomics/bam/api/SamRecordTest.scala
```diff
@@ -289,4 +289,48 @@ class SamRecordTest extends UnitSpec with OptionValues {
...
+  "SamRecord.readPosAtRefPos" should "return the correct value" in {
+    case class ReadPosAtRefPos(cigar: String, refPos: Int, readPos: Option[Int], returnLastBaseIfDeleted: Boolean)
+    val testCases = Seq(
```
tfenne, line 295: Here and below I think a lot more test cases are needed.  For soft-clipping I think you need tests where:
1. you have leading soft-clipping and request the read position at `rec.start` and `rec.start + 1`
2. you have trailing soft-clipping and you request `read.end` and `read.end - 1`

I think there should also be cases here where there is hard-clipping (should just be ignored).  And skips (e.g. `100N`) in the cigar (presumably return None.

src/test/scala/com/fulcrumgenomics/bam/api/SamRecordTest.scala
```diff
@@ -289,4 +289,48 @@ class SamRecordTest extends UnitSpec with OptionValues {
...
+      val readPos = frag.readPosAtRefPos(pos=testCase.refPos, returnLastBaseIfDeleted=testCase.returnLastBaseIfDeleted)
+      readPos shouldBe testCase.readPos
+      readPos.getOrElse(0) shouldBe frag.asSam.getReadPositionAtReferencePosition(testCase.refPos, testCase.returnLastBaseIfDeleted)
+    }
+  }
+
+  "SamRecord.refPosAtReadPos" should "return the correct value" in {
+    case class RefPosAtReadPos(cigar: String, refPos: Option[Int], readPos: Int, returnLastBaseIfInserted: Boolean)
+    val testCases = Seq(
+      RefPosAtReadPos("3S9M",   Some(7),  10, false),
```
tfenne, line 318: Likewise here I think you need tests with leading soft and hard clipping.  If you request a refPos at a base that is in the leading or trailing soft-clipping it should always return None.

tfenne (changes requested):

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: I really don't like `>=`.  It's not about yoda positionals.   I don't see how what you wrote is more readable?

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: A leading insertion

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: Also, after your fix, isn't this more readable `this.start <= refPos && refPos <= this.end`, which is very similar to `start <= refPos <= end`, which is what I really wanted to write?

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: 1. I think it handles soft-clipped bases just fine, both shown by the added tests, as well as that the `refPos` will now be past `this.end`!

nh13: added a comment

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: Fixed the bug, but I like my yoda-like positionals

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: Might as well, but with a separate boolean param

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: It wont since we check `pos` against `CoordMath.getEnd(refPos, elem.lengthOnTarget)`.  Since a soft-clip has `0` for `elem.lengthOnTarget`, then `CoordMath.getEnd` will return `refPos-1`, which is one before `pos`

nh13: Added tests

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
nh13: fixed

nh13:
@tfenne back to you

tfenne (commented):
Thanks for all the tests - that's super comforting!

PR #811, opened 2022-03-10
Add a github action for generating tool and metrics docs

nh13:


.github/workflows/unittests.yaml
```diff
@@ -56,3 +56,52 @@ jobs:
...
+          bash run_metrics_doclet.sh
+      - name: Commit generated docs
+        shell: bash -l {0}
+        run: |
+          git add docs/tools/*md
+          git add docs/*md
+          git config --global user.name "Github Actions"
+          git config --global user.email "nobody@fulcrumgenomics.com"
+          git commit --all --message "Generate docs files" --author "Nobody <nobody@fulcrumgenomics.com>" || echo "nothing to commit"
+          git push --set-upstream origin "HEAD:${{ github.head_ref }}" --follow-tags --atomic
```
clintval, line 107: This may cause race conditions.

It will also cause quite a few merge conflicts for all developers.

nh13:
In private discussions, we'll want to change this to:

1. On any PR, we test that docs and metrics compile without error, and are available (as artifacts) so folks can review them if needed.
1. On any commit to main, the docs and metrics are built and published to the gh-pages branch as the latest "develop" version.  To avoid race conditions, we should set [concurrency](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency).
2. On any tagged release, we publish the documentation and metrics as per the wiki.

PR #814, opened 2022-03-11
Draft best practices doc

tfenne:


docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+
+This document describes the best practise(s) for building pipelines to go from FASTQ files that contain a mixture of template (genomic) bases and UMI bases, through to aligned and filtered consensus reads.  
+
+The pipeline is broken into two phases:
+
+1. Getting to aligned and grouped raw reads
+2. Going from raw reads to aligned and filtered consensus reads
+
+We provide two alternatives for the second phase.  One intended for R&D environment where it is desirable to be able to branch off from the pipeline and test e.g. multiple consensus calling or filtering parameters.  The other is intended for high throughput production environments where performance and throughput take precedence over flexibility.
+
```
nh13, line 13: table of conetents

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+## Tools used
+
+The steps described below use a mixture of:
+
+- `fgbio` (version 2.0 or higher)
+- `samtools` (version
```
nh13, line 19: version?

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
+<img src="https://github.com/fulcrumgenomics/fgbio/blob/main/.github/logos/fulcrumgenomics.svg" alt="Fulcrum Genomics" height="35"/>
+
+# fgbio Best Practise FASTQ->Consensus Pipeline
```
nh13, line 3:
```suggestion
# fgbio Best Practise FASTQ -> Consensus Pipeline
```

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+
+This document describes the best practise(s) for building pipelines to go from FASTQ files that contain a mixture of template (genomic) bases and UMI bases, through to aligned and filtered consensus reads.  
+
+The pipeline is broken into two phases:
+
+1. Getting to aligned and grouped raw reads
```
nh13, line 9:
```suggestion
1. Getting to aligned and grouped raw reads to identify which raw reads belong to the same source molecule.
```

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+2. Going from raw reads to aligned and filtered consensus reads
```
nh13, line 10:
```suggestion
2. Going from grouped raw reads to aligned and filtered consensus reads
```

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+```mermaid
+graph TD;
+A["fgbio FastqToBam"]-->B["samtools fastq | bwa mem | fgbio ZipperBam"];
+B-->C["fgbio GroupReadsByUmi"];
+```
+
+In this example we'll use a single sample with paired-end fastq files and an inline 8bp UMI at the beginning of read one (see the fgbio [Read Structure](https://github.com/fulcrumgenomics/fgbio/wiki/Read-Structures) page for how to specify other UMI setups).
+
+### Step 1.1: FASTQ -> uBam
+
```
nh13, line 36: This step extracts the UMI bases and stores them in standard SAM tags.  
This is preferred over non-standard solutions that store UMIs in FASTQ read names or otherwise.

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+      --min-distance 2 \
+      --umi-files file-of-expected-umis-one-per-line.txt \
+      --dont-store-original-umis
+```
+
+Neither option above requires or performs any sorting.
+
+### Step 1.2: uBam -> Mapped BAM
+
+This step uses a combination of `samtools` (to convert the uBam back to FASTQ), `bwa mem` to do the actual alignment and `fgbio ZipperBams` to carry metaadata and UMIs forward from the unmapped BAM.  No sorting is performed in this step.
```
nh13, line 72:
```suggestion
This step uses a combination of `samtools` (to convert the uBam back to FASTQ), `bwa mem` to do the actual alignment and `fgbio ZipperBams` to carry metadata and UMIs forward from the unmapped BAM.  No sorting is performed in this step.
```

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+      --unmapped-bam s1.unmapped.bam \
+      --ref ref.fa \
+      --output s1.mapped.bam 
+```
+
+Some important `bwa mem` options used here are:
+
+- `-t 16` to set the number of threads to use to 16; this can be set to any number between 1 and the number of CPU cores available
+- `-K 150000000` to tell `bwa` to process reads/bases in fixed size chunks regardless of the number of threads; this makes `bwa` deterministic and reproducible with different thread counts
+- `-Y` to use soft-clipping (instead of hard clipping) in supplemntal alignments
```
nh13, line 88:
```suggestion
1. `-t 16` to set the number of threads to use to 16; this can be set to any number between 1 and the number of CPU cores available
1. `-K 150000000` to tell `bwa` to process reads/bases in fixed size chunks regardless of the number of threads; this makes `bwa` deterministic and reproducible with different thread counts
1. `-Y` to use soft-clipping (instead of hard clipping) in supplemental alignments; this useful for downstream structural variation detection, among other things
```

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+  --edits 1 \
+  --output s1.grouped.bam \
+  --family-size-histogram s1.tag-family-sizes.txt
+```
+
+However:
+
+1. If UMIs are significantly longer (e.g. 20bp) or have more errors, `--edits` can be increased
+2. For multiplex PCR and similar data where reads' genomic positions are fixed by the primers it is recommended to use `--strategy Identity` to reduce runtime at the expense of lower accuracy
+
```
nh13, line 111: The mapped BAM may be streamed directly into `GroupReadsByUmi`.

nh13: I am curious as to why not just stream into `GroupReadsByUmi`, since the latter is going to have to sort anyways.  Or stream the mapped reads into `fgbio SortSam` (ugh)

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+```mermaid
+graph TD;
+A["fgbio CallMolecularConsensusReads"]-->B["samtools fastq | bwa mem | fgbio ZipperBam"];
+B-->C["fgbio FilterConsensusReads | samtools sort"];
+```
+
+### Step 2(a).1: GroupedBam -> Consensus uBam
+
+This step generates unmapped consensus reads from the grouped reads.  Depending on your data you may choose to raise or lower the `--min-input-base-quality` to accept or reject more bases from the grouped raw reads.  In an R&D setting it is frequently useful to see _all_ consensus reads, but there is a performance trade-off.  Frequently a very large fraction of molecules will have only a single read[-pair]; if consensus reads with depth 1 will be discarded later setting `--min-reads 2` can cause this step to run significantly faster.
```
nh13, line 124:
```suggestion
This step generates unmapped consensus reads from the grouped reads.  Depending on your data you may choose to raise or lower the `--min-input-base-quality` to accept or reject more bases from the grouped raw reads.  In an R&D setting it is frequently useful to see _all_ consensus reads, but there is a performance trade-off.  Frequently a very large fraction of molecules will have only a single read[-pair]; if consensus reads with depth 1 will be discarded later setting `--min-reads 2` or higher can cause this step to run significantly faster.
```

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+
+```bash
+fgbio -Xmx4g --compression 0 CallMolecularConsensusReads \
```
nh13, line 127: should we use compression `1` whenever we touch disk?
```suggestion
fgbio -Xmx4g --compression 1 CallMolecularConsensusReads \
```

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+We provide two alternatives for the second phase.  One intended for R&D environment where it is desirable to be able to branch off from the pipeline and test e.g. multiple consensus calling or filtering parameters.  The other is intended for high throughput production environments where performance and throughput take precedence over flexibility.
+
+## Tools used
+
+The steps described below use a mixture of:
+
+- `fgbio` (version 2.0 or higher)
+- `samtools` (version 
+- `bwa mem`
+
```
nh13, line 21: commenting here, but I think we should do two things:

1. use `--compression 1` when writing to disk, and `--compression 0` when piping
2. highlight somewhere `--async-io=true`, since we omit that in a few places

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+Under no circumstances should `-M` be used to `mark shorter split hits as secondary` as this will cause problems with downstream tools.
+
+
+### Step 1.3: Mapped BAM -> Grouped BAM
+
+This step identifies reads or read pairs that originate from the same source molecule based on genomic positions and UMI.  For _most_ data this is the recommended way to run grouping:
+
+```  
+# Group the reads by position and UMI
+fgbio -Xmx8g --compression 1 GroupReadsByUmi \
```
nh13, line 99: async io

docs/best-practice-consensus-pipeline.md
nh13, line 127: async-io please

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+# fgbio Best Practise FASTQ->Consensus Pipeline
+
+This document describes the best practise(s) for building pipelines to go from FASTQ files that contain a mixture of template (genomic) bases and UMI bases, through to aligned and filtered consensus reads.  
+
+The pipeline is broken into two phases:
+
+1. Getting to aligned and grouped raw reads
+2. Going from raw reads to aligned and filtered consensus reads
+
+We provide two alternatives for the second phase.  One intended for R&D environment where it is desirable to be able to branch off from the pipeline and test e.g. multiple consensus calling or filtering parameters.  The other is intended for high throughput production environments where performance and throughput take precedence over flexibility.
```
nh13, line 12: currently R&D/prod only differ in the second step, mention that here?

docs/best-practice-consensus-pipeline.md
```diff
@@ -0,0 +1,217 @@
...
+```bash
+min_reads=3  # Mirrors the setting in 2(a).3 
+
+fgbio -Xmx4g --compression 0 CallMolecularConsensusReads \
+  --input s1.grouped.bam \
+  --output /dev/stdout \
+  --min-reads $min_reads \
+  --min-input-base-quality 20 \
+  --threads 4 \
+  |  fgbio -Xmx8g --compression 1 FilterConsensusReads \
```
nh13, line 192: async io here and above

nh13 (approved):
This is a really nice first pass.  

I made some minor suggestions. I think the next round of this doc needs to have some higher level explanation and motivation of the step, as there may be questions about _why_ certain steps are needed.  But that's a minor quibble.

docs/best-practice-consensus-pipeline.md
tfenne: My 2c is that since `GroupReadsByUmi` discards reads, I like to save the aligned/zippered BAM before pushing it through GroupReadsByUmi.  That way if I want to change my mind about --min-map-q or other settings that affect which reads to use, I don't have to re-align the data.
Also, frankly, a lot of people will branch off at this point to run metrics, alt analysis on non-consensus data, etc.

docs/best-practice-consensus-pipeline.md
tfenne: Added a section on asyncio.  I think there is near-0 benefit when using compression 0 so leaving it out here.  Especially with `--threads`, most of the work is in a threadpool, leaving the main thread idle to write the BAM.

docs/best-practice-consensus-pipeline.md
tfenne: That's not true though.  The R&D version just calls consensus reads and writes to a file as the first step.  The prod version pipes into filter before writing.

PR #815, opened 2022-03-12
Fixed a couple of bugs in the best practice pipeline doc and added a…

tfenne:
…n example snakemake implementation.


docs/FastqToConsensus-RnD.smk
```diff
@@ -0,0 +1,156 @@
+################################################################################
+# Example snakemake file that implements the R&D Best Practices for fgbio's
+# Fastq -> Filtered Consensus Pipeline
+#
+# Required software:
```
nh13, line 5: Consider showing a Snakemake invocation that has mem_gb as resources

PR #817, opened 2022-03-21
docs: update documentation for LogProbability functions

wmchad:
This PR updates the documentation for a few of the functions that deal with `LogProbability`. The documentation should be more correct, but I think a more long term solution would involve changing the names of a few functions. In particular:
- `or` could be more appropriately named `add` or something similar. It is just used to add probabilities, which is appropriate for `Pr(A or B)` only if `A` and `B` are mutually independent. For independent events, `Pr(A or B) = Pr(A) + Pr(B) - Pr(A)*Pr(B)`.
- `aOrNotB` could be more appropriately names `subtract` or similar. It is just used to subtract one probability from another.

https://github.com/fulcrumgenomics/fgbio/blob/560f0898dc803f5d6d16fc9817d4219afe24c71e/src/main/scala/com/fulcrumgenomics/util/NumericTypes.scala#L185
Is there a reason this `and` function takes `Array[Double]` and not `Array[LogProbability]`?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/NumericTypes.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/NumericTypes.scala
@@ -150,7 +150,7 @@ object NumericTypes {
       * See Equation (7) in https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf
       */
     private def log1mexp(value: Double): Double = {
-      if (value < 0) throw new IllegalArgumentException("value was less than zero: " + value)
+      if (value <= 0) throw new IllegalArgumentException("value was less than or equal to zero: " + value)
       else if (value <= LnTwo) log(-expm1(-value))
       else log1p(-exp(-value)) // value > LnTwo
     }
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/util/NumericTypes.scala
```diff
@@ -146,7 +146,7 @@ object NumericTypes {
     }
 
     /**
-      * Precise computation of log(1-exp(x)) for use in the `aOrNotB` method.
+      * Precise computation of log(1-exp(-x)) for use in the `aOrNotB` method.
       * See Equation (7) in https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf
```
nh13, line 150: This is correct, though the R docs says it's for `x > 0`. So I think line `153` should be `if (value <= 0)` with an updated comment

nh13 (approved):
Thank-you for the doc updates, one minor suggestion!

nh13:
Will merge once tests pass!

nh13:
Thank-you @wmchad!

PR #820, opened 2022-03-29
Make SamWriter stop checking sort order when emitting pre-sorted records.

tfenne:
Build will probably fail for a while until the HTSJDK snapshot is published.


nh13 (commented):
how does the output sort order get checked when NOT emitting pre-sorted records?

tfenne:
Well, then it doesn't get checked at all, since it is being sorted.  What am I missing?

nh13 (approved):

Issue #821, opened 2022-04-10
Where to insert Picard MarkIlluminaAdapters in an fgbio workflow?


blackbeerd:
How do you suggest handling the potential presence of adaptor sequence in NGS data processed with your current workflow?

I am putting together an adapted version of the 2.0 fgbio release and workflow you have published and I'm wondering where in the pipeline would be best to carry out Picard's MarkIlluminaAdaptors? A small percentage of my reads often contain adaptor read-through due to short insert length. Best practices for small insert PE reads released by Broad (https://gatk.broadinstitute.org/hc/en-us/articles/360039568932--How-to-Map-and-clean-up-short-read-sequence-data-efficiently) suggests running MarkIlluminaAdaptors right after generation of an unmapped bam and prior to merging. Therefore, my initial thought was to put the MarkIlluminaAdaptors step in just after fgbio's DemuxFastqs (in which I generate uBams) and before Step 1.2 (uBam -> Mapped BAM) of your best practices pipeline (which first converts the uBam to fastq prior to BWA mem). However, Broad's documentation (linked above) points out that some tools like SamToFastq use the XT tag generated by MarkIlluminaAdaptors to handle adaptor read-through in various ways, including by adjusting base quality scores of potential adaptor sequence to 2 (# symbol) when converting uBam to fastqs. There is also the option to convert adaptor bases to "N"s, but I'm not sure which is best for fgbio. (I assume samtools fastq tool has similar options?) I am concerned that converting base quality to 2 might cause issues in the consensus calling and filtering steps later in the pipeline. Since the filtering step relies on probabilities derived from base-quality scores, might this cause the remaining "good" sequence data in reads containing bits of adaptor to be omitted as well? OR, will the min-input-base-quality filter in the CallMolecularConsensusReads step cause those bases with quality scores of "2" to simply be ignored when making consensus sequences? What about the alternative of changing the adaptor bases as "N"s (this method does not change the original base quality score) and would converting to Ns affect the FilterConsensusReads step via the max-no-call-fraction argument?

Do you have any best practices for handling adaptor sequences?

Regards,
Chris

nh13:
@tfenne can weigh in here too, but I am not convinced that adapter _marking_ is all that impactful when you have longer  reads (2 x 75+ bp) and map with an aligner that soft-clips (i.e. `bwa mem`).  In my experience, `bwa mem` will just soft-clip the adapter reads, and that doesn't cause false variant calls later or issues with consensus calling.

[...]

PR #824, opened 2022-04-15
[bugfix] overlapping consensus caller should examine only

nh13:
bases that map in both the read and ref

Fixes issue #821.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
@@ -72,57 +72,34 @@ class OverlappingBasesConsensusCaller(agreementStrategy: AgreementStrategy = Agr
     require(r1.mapped && r2.mapped && r1.paired && r2.paired && r1.name == r2.name && r1.matesOverlap.contains(true))
     require(r1.firstOfPair && r2.secondOfPair)
 
-    // Initialize the iters
-    val maxStartRefPos   = Math.max(r1.start, r2.start)
-    val minEndRefPos     = Math.min(r1.end, r2.end)
-    val r1Iter           = new ReadAndRefPosIterator(r1, minRefPos=maxStartRefPos, maxRefPos=minEndRefPos).buffered
-    val r2Iter           = new ReadAndRefPosIterator(r2, minRefPos=maxStartRefPos, maxRefPos=minEndRefPos).buffered
-    var r1LastReadPos    = r1Iter.head.readPos - 1
-    var r2LastReadPos    = r2Iter.head.readPos - 1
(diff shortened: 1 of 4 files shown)


src/main/scala/com/fulcrumgenomics/bam/OverlappingBasesConsensusCaller.scala
```diff
@@ -72,57 +72,34 @@ class OverlappingBasesConsensusCaller(agreementStrategy: AgreementStrategy = Agr
...
     require(r1.firstOfPair && r2.secondOfPair)
 
-    // Initialize the iters
-    val maxStartRefPos   = Math.max(r1.start, r2.start)
-    val minEndRefPos     = Math.min(r1.end, r2.end)
-    val r1Iter           = new ReadAndRefPosIterator(r1, minRefPos=maxStartRefPos, maxRefPos=minEndRefPos).buffered
-    val r2Iter           = new ReadAndRefPosIterator(r2, minRefPos=maxStartRefPos, maxRefPos=minEndRefPos).buffered
-    var r1LastReadPos    = r1Iter.head.readPos - 1
-    var r2LastReadPos    = r2Iter.head.readPos - 1
+    val iter             = new ReadMateAndRefPosIterator(rec=r1, mate=r2).buffered
```
nh13, line 75: using this new iterator simplifies this method greatly

src/main/scala/com/fulcrumgenomics/bam/ReadAndRefPosIterator.scala
```diff
@@ -179,3 +180,61 @@ object ReadAndRefPosIterator {
...
+      else {
+        nextItem = Some(ReadMateAndRefPos(readPos=nextRec.readPos, matePos=nextMate.readPos, refPos=nextRec.refPos))
+        recIter.next()
+        mateIter.next()
+      }
+    }
+    nextItem.isDefined
+  }
+
+  def next(): ReadMateAndRefPos = this.nextItem match {
```
tfenne, line 224: I feel strongly that this should call/check `hasNext()`.  The way this is done, if I construct an iterator where I _know_ there should be 3 values in it, and then call `iter.next()` three times it will fail because it relies on `.hasNext()` to do the advancing, but nothing says you have to call `.hasNext()`.

I would re-write this as:

```scala
if (!hasNext()) throw new NoSuchElementException
val retval = this.nextItem.get
this.nextItem = None
retval
```

tfenne (approved):
Approved subject to fixing next/hasNext per comment.

src/main/scala/com/fulcrumgenomics/bam/ReadAndRefPosIterator.scala
nh13: done

PR #826, opened 2022-04-20
Add metric builder and sorter

nh13:
Added a new class `MetricSorter` that's useful when we want to sort metrics but cannot hold them all in memory.  Leverages `Sorter`.


src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -205,6 +216,60 @@ object Metric extends LazyLogging {
...
+object MetricSorter {
+  /** The default maximum # of records to keep and sort in memory. */
+  val MaxInMemory: Int = 1e6.toInt
+
+  class MetricSorterCodec[T <: Metric]()(implicit tt: ru.TypeTag[T])
+    extends Sorter.Codec[T] with LazyLogging {
+    private val clazz: Class[T]   = ReflectionUtil.typeTagToClass[T]
+    private val reflectiveBuilder = new ReflectiveBuilder(clazz)
+
+    private def fail(message: String,
```
nh13, line 239: TODO: I want to find a nice way to not duplicate this `fail` method here and in the `Metric` object.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -205,6 +216,60 @@ object Metric extends LazyLogging {
...
+      Metric.build(
+        reflectiveBuilder = reflectiveBuilder,
+        toArg = i => fields(i),
+        fail  = fail
+      )
+
+      // build it.  NB: if arguments are missing values, then an exception will be thrown here
+      // Also, we don't use the default "build()" method since if a collection or option is empty, it will be treated as
+      // missing.
+      reflectiveBuilder.build(reflectiveBuilder.argumentLookup.ordered.map(arg => arg.value getOrElse unreachable(s"Arguments not set: ${arg.name}")))
```
nh13, line 268:
```suggestion
```

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -205,6 +223,41 @@ object Metric extends LazyLogging {
...
+  codec           = new MetricSorter.MetricSorterCodec[T](),
+  keyfunc         = keyfunc,
+  tmpDir          = tmpDir
+)
+
+object MetricSorter {
+  /** The default maximum # of records to keep and sort in memory. */
+  val MaxInMemory: Int = 1e6.toInt
+
+  class MetricSorterCodec[T <: Metric]()(implicit tt: ru.TypeTag[T])
```
nh13, line 241: could make this into a `def codec[T <: Metric](implicit tt: ru.TypeTag[T]): Sorter.Codec[T] with LazyLogging`

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -99,56 +99,74 @@ object Metric extends LazyLogging {
     }
   }
 
+  private[util] def build[T <: Metric](reflectiveBuilder: ReflectiveBuilder[T],
```
nh13, line 102: next two methods refactored out of `def iterator` so `MetricSorterCodec` can use them

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -99,56 +99,74 @@ object Metric extends LazyLogging {
...
+                                       toArg: Int => String,
```
nh13, line 103: was tempted to pass a `Row` here, but then `MetricSorter` has to build one from an array of strings.

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -0,0 +1,137 @@
...
+  private val names             = Metric.names[T]
+
+  /** Builds a metric from a delimited line
+    *
+    * @param line       the line with delimited values
+    * @param delim      the delimiter of the values
+    * @param lineNumber optionally, the line number when building a metric from a line in a file
+    * @return
+    */
+  def fromLine(line: String, delim: String = Metric.DelimiterAsString, lineNumber: Option[Int] = None): T = {
```
nh13, line 55: Note: need to have different names for each of the `from*` methods because I want `lineNumber` to default to `None`, and I can't have multiple functions with the same name AND each with default values.

nh13:
Relies on https://github.com/fulcrumgenomics/commons/pull/82

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  /** Builds a metric of type [[T]]
+    *
+    * @param argMap     map of field names to values.  All required fields must be given.  Can be in any order.
+    * @param lineNumber optionally, the line number when building a metric from a line in a file
+    * @return a new instance of type [[T]]
+    */
+  def fromArgMap(argMap: Map[String, String], lineNumber: Option[Int] = None): T = {
+    reflectiveBuilder.reset() // reset the arguments to their initial values
```
nh13, line 82: Relies on https://github.com/fulcrumgenomics/commons/pull/82

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  /** Builds a metric from a delimited line
+    *
+    * @param line       the line with delimited values
+    * @param delim      the delimiter of the values
+    * @param lineNumber optionally, the line number when building a metric from a line in a file
+    * @return
+    */
+  def fromLine(line: String, delim: String = Metric.DelimiterAsString, lineNumber: Option[Int] = None): T = {
+    fromValues(values = line.split(delim), lineNumber = lineNumber)
```
clintval, line 58: Do you want to use [`StringUtil.split()`](https://github.com/fulcrumgenomics/commons/blob/e9534e08e61ac4da3d5ea8dd6f21a39b105005f2/src/main/scala/com/fulcrumgenomics/commons/util/StringUtil.scala#L166) which could re-use an array and maybe be faster?

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -0,0 +1,142 @@
...
+import scala.util.{Failure, Success}
+
+/** Class for building metrics of type [[T]].
+  *
+  * This is not thread-safe.
+  *
+  * @param source optionally, the source of reading (e.g. file)
+  * @tparam T the metric type
+  */
+class MetricBuilder[T <: Metric](source: Option[String] = None)(implicit tt: ru.TypeTag[T]) extends LazyLogging {
```
clintval, line 44: I see `source` was the name of the parameter beforehand, but it made me think a [`scala.io.Source`](https://www.scala-lang.org/api/3.x/scala/io/Source$.html) was going to be the type at first. Can't think of a better name, though.

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
clintval: Should you also have a public `fromRow`?

The only annoying part for users would be that there is no public way to build a [`Row`](https://github.com/fulcrumgenomics/commons/blob/e9534e08e61ac4da3d5ea8dd6f21a39b105005f2/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala#L38) directly.

So it is up to you. I think a `fromRow` would be nice and you could use it internally in `Metric` since you have a `DelimitedDataParser` there already.

src/test/scala/com/fulcrumgenomics/util/MetricBuilderTest.scala
```diff
@@ -0,0 +1,61 @@
...
+
+    val nameBuilder = new MetricBuilder[MetricBuilderNameMetric]()
+    nameBuilder.fromLine(line = "Foo Bar 42", delim = " ") shouldBe MetricBuilderNameMetric(first="Foo", second="Bar", age=42)
+  }
+
+  it should "throw an FailureException when the # of values are incorrect or wrong type" in {
+    val builder = new MetricBuilder[MetricBuilderNameMetric]()
+    an[FailureException] should be thrownBy builder.fromLine(line="one\ttwo")
+    an[FailureException] should be thrownBy builder.fromLine(line="one\ttwo\t3\tfour", lineNumber=Some(1))
+    an[FailureException] should be thrownBy builder.fromLine(line="one\ttwo\tthree")
```
clintval, line 59:
```suggestion
    a[FailureException] should be thrownBy builder.fromLine(line="one\ttwo")
    a[FailureException] should be thrownBy builder.fromLine(line="one\ttwo\t3\tfour", lineNumber=Some(1))
    a[FailureException] should be thrownBy builder.fromLine(line="one\ttwo\tthree")
```

clintval (approved):
I like it! I caught a few small issues you could fix, and I also couldn't help but leave a few musings. Approving conditional on you zipping through my comments and making fixups where you see fit. I don't need to see this one again.

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
clintval: On second thought, I don't think you need a `fromLine` on this class. It is not used by Metric directly or indirectly. I vote deleting this method and just having a `fromValues` or `fromArgMap`.

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -102,53 +99,14 @@ object Metric extends LazyLogging {
...
-            fail(lineNumber=rowIndex+2, message=s"Did not have a field with name '${names(i)}'.")
-        }
-      }
-
-      // build it.  NB: if arguments are missing values, then an exception will be thrown here
-      // Also, we don't use the default "build()" method since if a collection or option is empty, it will be treated as
-      // missing.
-      reflectiveBuilder.build(reflectiveBuilder.argumentLookup.ordered.map(arg => arg.value getOrElse unreachable(s"Arguments not set: ${arg.name}")))
+      val argMap = names.zipWithIndex.map { case (name, i) => name -> row[String](i) }.toMap
+      builder.fromArgMap(argMap=argMap, lineNumber=Some(rowIndex+2))
```
clintval, line 109: I'm now a tad confused why `fromArgMap` is a necessary method on the Metric builder. The MetricBuilder already knows what the field names are, so the keys to this map are not necessary and you can use a method as simple as `.fromValues()`.

nh13: Since not all values need to specified, for example when the metric has default values.

src/main/scala/com/fulcrumgenomics/util/MetricSorter.scala
```diff
@@ -0,0 +1,70 @@
...
+  *
+  * @param maxObjectsInRam the maximum number of metrics to keep in memory before spilling to disk
+  * @param keyfunc method to convert a metric to an ordered key
+  * @param tmpDir the temporary directory in which to spill to disk
+  * @param tt the type tag for [[T]]
+  * @tparam Key the key to use for sorting metrics
+  * @tparam T the metric type
+  */
+class MetricSorter[Key <: Ordered[Key], T <: Metric](maxObjectsInRam: Int = MetricSorter.MaxInMemory,
+                                                     keyfunc: T => Key,
```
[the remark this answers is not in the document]
nh13: Most metrics are not orderable :/

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -0,0 +1,165 @@
...
+  private val names             = Metric.names[T]
+  private val namesSet          = names.toSet
+
+  /** Builds a metric from a [[ com.fulcrumgenomics.commons.util.Row]] produced from a
+    * [[ com.fulcrumgenomics.commons.util.DelimitedDataParser]].
+    *
+    * @param row        the row to parse.  All required fields must be given.  Can be in any order.
+    * @param lineNumber optionally, the line number when building a metric from a line in a file
+    * @param ignoreExtra ignore extra fields (that are not in the metric)
+    * @return
```
clintval, line 58:
```suggestion
    * @return a new instance of type [[T]]
```

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -0,0 +1,165 @@
...
+  */
+class MetricBuilder[T <: Metric](source: Option[String] = None)(implicit tt: ru.TypeTag[T]) extends LazyLogging {
+  // The main reason why a builder is necessary is to cache some expensive reflective calls.
+  private val clazz: Class[T]   = ReflectionUtil.typeTagToClass[T]
+  private val reflectiveBuilder = new ReflectiveBuilder(clazz)
+  private val names             = Metric.names[T]
+  private val namesSet          = names.toSet
+
+  /** Builds a metric from a [[ com.fulcrumgenomics.commons.util.Row]] produced from a
+    * [[ com.fulcrumgenomics.commons.util.DelimitedDataParser]].
```
clintval, line 53:
```suggestion
  /** Builds a metric from a [[com.fulcrumgenomics.commons.util.Row]] produced from a
    * [[com.fulcrumgenomics.commons.util.DelimitedDataParser]].
```

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -0,0 +1,165 @@
...
+
+import com.fulcrumgenomics.commons.CommonsDef.{forloop, unreachable}
+import com.fulcrumgenomics.commons.reflect.{ReflectionUtil, ReflectiveBuilder}
+import com.fulcrumgenomics.commons.util.{LazyLogging, Row}
+
+import java.io.{PrintWriter, StringWriter}
+import scala.reflect.runtime.{universe => ru}
+import scala.util.{Failure, Success}
+
+case class MetricBuilderException private[util](message: Option[String] = None) extends Exception
```
clintval, line 36: I haven't done it in a while, but I think you need to pass along the message?

```suggestion
case class MetricBuilderException private[util](message: Option[String] = None) extends Exception(message)
```

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -0,0 +1,165 @@
...
+  private val reflectiveBuilder = new ReflectiveBuilder(clazz)
+  private val names             = Metric.names[T]
+  private val namesSet          = names.toSet
+
+  /** Builds a metric from a [[ com.fulcrumgenomics.commons.util.Row]] produced from a
+    * [[ com.fulcrumgenomics.commons.util.DelimitedDataParser]].
+    *
+    * @param row        the row to parse.  All required fields must be given.  Can be in any order.
+    * @param lineNumber optionally, the line number when building a metric from a line in a file
+    * @param ignoreExtra ignore extra fields (that are not in the metric)
```
clintval, line 57: Missing `@param` headers FYI.

But isn't both the builder and `Row` aware of their own headers/names?

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -0,0 +1,165 @@
...
+    * @return
+    */
+  def fromRow(row: Row, headers: Seq[String], lineNumber: Option[Int] = None, ignoreExtra: Boolean = true): T = {
+    val argMap = headers  // NB: filter after map so the row lookup is faster
+      .zipWithIndex
+      .map { case (header, i) => header -> row[String](i) }
```
clintval, line 63: If a user accidentally gives headers that don't exist in the row, then a surprise `NoElementException` will be raised. Since the `Row` should be self-aware of its own header can we use that directly? Alternatively maybe we do:

```suggestion
      .map { case (header, i) => header -> row.get[String](i, allowMissingColumn=ignoreExtra) }
```

clintval (approved):
Couple questions and tiny fixups, but I like how this looks now!

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
nh13: 1. There may be more headers in the input file (row) than the metric
2. Row does know, but it could: https://github.com/fulcrumgenomics/commons/issues/87

nh13: Fixed in https://github.com/fulcrumgenomics/commons/pull/88

PR #842, opened 2022-05-17
[bugfix] SamRecordClipper.clipOverlappingReads now accounts for soft-clipped bases in the positive read

jrm5100:
When the positive strand read has a shorter alignment than the negative strand read the `numBasesExtendingPastMate` function always returned 0. This bugfix reverts to previous (prior to v2) behavior- excess soft-clipped bases are trimmed.


mjhipp (changes requested):
@jrm5100 can you fix compile errors and squash into one commit?

https://github.com/fulcrumgenomics/fgbio/runs/6477469376?check_suite_focus=true

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
```diff
@@ -758,4 +758,18 @@ class SamRecordClipperTest extends UnitSpec with OptionValues {
     mate.start shouldBe 1
     mate.cigar.toString shouldBe "40M20I40M"
   }
+
+  it should "clip FR reads when the F read doesn't align past the R read but contains soft-clipped bases that extend past" in {
+    val builder = new SamBuilder(readLength=50, sort=Some(Queryname))
+    val clipper = new ClipBam(input=dummyBam, output=dummyBam, ref=ref, clipBasesPastMate=true)
+    val Seq(r1, r2) = builder.addPair(start1=20, start2=20, strand1=Minus, strand2=Plus, bases1="A"*10+"CGT"+"C"*37, bases2="CGT"+"C"*35+"A"*12, cigar1="10S40M", cigar2="38M12S")
```
mjhipp, line 765: can you use `pair` helper method instead of builder/addpair?

mjhipp: also this test could test the output of `SamRecordClipper.clipExtendingPastMateEnds` (thats what `it` is in this context)

jrm5100: I got git-lazy.  I made fixes in the original clone and manually brought them over to a fork but didn't pay close enough attention and pasted the test at the end of the wrong file.  Fix incoming.

mjhipp: I do think its a good idea to add tests to `SamRecordClipperTest` that cover these cases since it is closer to the code change you are submitting

jrm5100:
I've found some additional scenarios and will have an update to this PR later today.

mjhipp (approved):

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -340,13 +340,21 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+      else if (rec.negativeStrand && (rec.start <= rec.mateStart)) {
+        // count up to and including one base before where the mate starts
         Math.max(0, rec.readPosAtRefPos(pos=rec.mateStart, returnLastBaseIfDeleted=false) - 1)
+      }
+      else if (rec.negativeStrand && (rec.start > rec.mateStart)) {
+        // record alignment starts after the mate alignment start: remove any excess soft-clipped reads
+        Math.max(0, rec.cigar.leadingSoftClippedBases - (rec.start - rec.mateStart))
       } else {
         // no bases extend past
         0
```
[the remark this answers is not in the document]
nh13: Agreed, the `else` can become `else if` now.  I don't have a preference on if you want to have the `positiveStrand` in one `else if` or keep the two scenarios for `positiveStrand` their own `else if`s.

nh13:
Thank-you @mjhipp and @jrm5100!

PR #845, opened 2022-05-24
Add a missing param to constructor of StreamingPileupBuilder via apply

clintval:
I noticed I forgot to pass through a param as well as document a param.


src/main/scala/com/fulcrumgenomics/bam/pileup/StreamingPileupBuilder.scala
```diff
@@ -98,6 +99,7 @@ object StreamingPileupBuilder {
   * @param includeSupplementalAlignments if true, allow records flagged as supplementary alignments to contribute to a pileup.
   * @param includeMapPositionsOutsideFrInsert if true, include any record of an FR pair where the site requested is outside the insert.
   * @param initialCacheSize the initial size for the internal SAM record cache, set this to your expected pileup depth.
+  * @param source an optional source to additionally close upon closing this pileup builder.
```
nh13, line 102: thank-you for this while you were here!

nh13:
Ready to merge!  Thank-you!

PR #850, opened 2022-06-02
Fix overly aggressive overlap-clipping regressions in ClipBam

### LINKED ISSUE #849: Regression in `ClipBam` in Fgbio v2.
Any soft-clipping at the end being clipped is double-counted.

Input into `ClipBam`

```
1:18773 99  chr17       7674753 60 142M =       7674798 184   ...
1:18773 147 chr17       7674798 60 3S139M       =      7674753  -184  ...
```

Using fgbio v1.4

```
1:18773 99  chr17       7674753 60 94M48H       =      7674847  184  ...
1:18773 147 chr17       7674847 60 52H90M       =      7674753  -184   ...
```

Using latest fgbio snapshot we create a gap of 3 bp (equivalent to the soft-clip length)

```
1:18773 99  chr17       7674753 60 92M50H       =      7674848  184  ...
1:18773 147 chr17       7674848 60 53H89M       =      7674753  -184  ...
```

The alignment looks like this:

```
AGAGACCCCAGTTGCAAACCAGAC Ref
AGAGACCCCAGTHHHHHHHHHHHH Fgbio1 Read 1
HHHHHHHHHHHHTGCAAACCAGAC Fgbio1 Read 2
AGAGACCCCAHHHHHHHHHHHHHH Fgbio2 Read 1
HHHHHHHHHHHHHGCAAACCAGAC Fgbio2 Read 2
```

I've tracked it down and will have a PR immenently.

jrm5100:
Closes #849

1. Because `readEnd` and `mateStart` use `readPosAtRefPos`, soft-clipping is already accounted for- it was being double-counted, adding additional clipping which left gaps between reads.
2. The midpoint of the pair (based on 5' reads) was calculated, but there is no guarantee that the overlapping region covers this position. I've updated it to calculate the midpoint based on 3' ends (the midpoint of the overlap). This makes the results correspond to the documentation ("and is implemented by clipping approximately half the overlapping bases from each read"). This isn't quite true when one read has a deletion, since it is actually based on the midpoint of the alignment not the midpoint of the sequences- it will favor clipping more from the read without the deletion.

An alternative to 2 is to clip more bases from the longer read:

```scala
val midPoint  = { // Need to ensure the midpoint actually occurs in the overlapped region
  val retval = (rec.start + mate.end) / 2
  if (retval > rec.end) rec.end
  else if (retval < mate.start) mate.start
  else retval
}
```

It also could be smarter and move the midpoint to minimize the number of N calls that are kept, or do something related to quality.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
@@ -305,17 +305,22 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean) 
     else if (!rec.isFrPair) (0, 0)
     else if (rec.negativeStrand) clipOverlappingReads(rec=mate, mate=rec).swap // don't for get to swap the results since we swapped inputs
     else {
-      // Pick the mid point in the reference window over which the record and its mate overlap.  The read ends at the
+      // Pick the mid point in the reference window between the 5' ends of the record and its mate.  The read ends at the
       // mid point, or if the mid point is in a deletion, the base prior to the deletion.  The mate ends at the mid
       // point, or if the mid point is in a deletion, the base after the deletion.
-      val midPoint  = (rec.start + mate.end) / 2
+      val midPoint  = { // Need to ensure the midpoint between 5' ends actually occurs in the overlapped region
+        val retval = (rec.start + mate.end) / 2
+        if (retval > rec.end) rec.end // trim only the minus strand read
+        else if (retval < mate.start) mate.start - 1 // trim only positive-strand read
+        else retval
+      }
       val readEnd   = rec.readPosAtRefPos(pos=midPoint, returnLastBaseIfDeleted=true)
       val mateStart = { // NB: need to be careful if the midpoint falls in a deletion
         val retval = mate.readPosAtRefPos(pos=midPoint + 1, returnLastBaseIfDeleted=false)
         if (retval != 0) retval else mate.readPosAtRefPos(pos=midPoint + 1, returnLastBaseIfDeleted=true) + 1
       }
-      val numOverlappingBasesRead = this.clip3PrimeEndOfRead(rec, rec.cigar.trailingClippedBases + rec.length - readEnd)
-      val numOverlappingBasesMate = this.clip3PrimeEndOfRead(mate, mate.cigar.leadingClippedBases + mateStart - 1)
+      val numOverlappingBasesRead = this.clip3PrimeEndOfRead(rec, rec.cigar.trailingHardClippedBases + rec.length - readEnd)
+      val numOverlappingBasesMate = this.clip3PrimeEndOfRead(mate, mate.cigar.leadingHardClippedBases + mateStart - 1)
       (numOverlappingBasesRead, numOverlappingBasesMate)
     }
   }
--- a/src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
@@ -166,21 +166,6 @@ class ClipBamTest extends UnitSpec with ErrorLogLevel with OptionValues {
     r1.end >= r2.start shouldBe false
   }
 
-  // This is a weird test that ensures that things terminate correctly when due to highly clipped
-  // input reads, one of the reads becomes unmapped during clipping
-  it should "handle reads that get unmapped because they are fully overlapped" in {
-    val builder = new SamBuilder(readLength=50)
-    val clipper = new ClipBam(input=dummyBam, output=dummyBam, ref=ref, clipOverlappingReads=true)
-    val Seq(r1a, r2a) = builder.addPair(name="q1", start1=100, start2=100, cigar1="3M47S", cigar2="48S2M")
-    val Seq(r1b, r2b) = builder.addPair(name="q2", start1=100, start2=100, cigar1="2M48S", cigar2="47S3M")
-
-    for ((r1, r2) <- Seq((r1a, r2a), (r1b, r2b))) {
-      clipper.clipPair(r1, r2)
-      val ok = r1.unmapped || r2.unmapped || r1.end < r2.start
-      ok shouldBe true
-    }
-  }
-
   it should "clip a fixed amount on the ends of the reads with reads that do not overlap" in {
     val builder = new SamBuilder(readLength=50)
     val clipper = new ClipBam(input=dummyBam, output=dummyBam, ref=ref,
--- a/src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
@@ -588,6 +588,91 @@ class SamRecordClipperTest extends UnitSpec with OptionValues {
     mate.cigar.toString shouldBe "1S99M"
   }
 
+  it should "clip reads that overlap with a deletion" in {
+    val (rec, mate) = pair(2, "80M10D10M", Plus, 70, "10M10D80M", Minus)
+    clipper(Soft).clipOverlappingReads(rec=rec, mate=mate) shouldBe (10, 10)
+    rec.start shouldBe 2
+    rec.end shouldBe 81
+    rec.cigar.toString shouldBe "80M10S"
+    mate.start shouldBe 90
+    mate.end shouldBe 169
+    mate.cigar.toString shouldBe "10S80M"
+  }
+
+  it should "clip reads that overlap with soft-clipping on the forward read after the midpoint" in {
+    val (rec, mate) = pair(1, "95M5S", Plus, 50, "100M", Minus)
+    clipper(Hard).clipOverlappingReads(rec=rec, mate=mate) shouldBe (20, 26)
+    rec.start shouldBe 1
+    rec.end shouldBe 75
+    rec.cigar.toString shouldBe "75M25H"
+    mate.start shouldBe 76
+    mate.cigar.toString shouldBe "26H74M"
+  }
+
+  it should "clip reads that overlap with soft-clipping on the reverse read before the midpoint" in {
+    val (rec, mate) = pair(1, "100M", Plus, 55, "5S95M", Minus)
+    clipper(Hard).clipOverlappingReads(rec=rec, mate=mate) shouldBe (25, 21)
+    rec.start shouldBe 1
+    rec.end shouldBe 75
+    rec.cigar.toString shouldBe "75M25H"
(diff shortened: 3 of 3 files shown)


jrm5100:
Holding off for now- found another error case

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -308,14 +308,14 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
       // Pick the mid point in the reference window over which the record and its mate overlap.  The read ends at the
       // mid point, or if the mid point is in a deletion, the base prior to the deletion.  The mate ends at the mid
       // point, or if the mid point is in a deletion, the base after the deletion.
-      val midPoint  = (rec.start + mate.end) / 2
+      val midPoint = (mate.start + rec.end) / 2
```
nh13, line 311: the unspoken rule in this repo is to align the equals, sorry for the nit!
```suggestion
      val midPoint  = (mate.start + rec.end) / 2
```

src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
```diff
@@ -166,21 +166,6 @@ class ClipBamTest extends UnitSpec with ErrorLogLevel with OptionValues {
...
-    val clipper = new ClipBam(input=dummyBam, output=dummyBam, ref=ref, clipOverlappingReads=true)
-    val Seq(r1a, r2a) = builder.addPair(name="q1", start1=100, start2=100, cigar1="3M47S", cigar2="48S2M")
-    val Seq(r1b, r2b) = builder.addPair(name="q2", start1=100, start2=100, cigar1="2M48S", cigar2="47S3M")
-
-    for ((r1, r2) <- Seq((r1a, r2a), (r1b, r2b))) {
-      clipper.clipPair(r1, r2)
-      val ok = r1.unmapped || r2.unmapped || r1.end < r2.start
-      ok shouldBe true
-    }
-  }
```
nh13, line 182: where did this test case get moved to?

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
```diff
@@ -588,24 +588,74 @@ class SamRecordClipperTest extends UnitSpec with OptionValues {
...
+    val (rec, mate) = pair(1, "99M1S", Plus, 99, "1S99M", Minus)
+    clipper(Hard).clipOverlappingReads(rec=rec, mate=mate) shouldBe (0, 1)
+    rec.start shouldBe 1
+    rec.end shouldBe 99
+    rec.cigar.toString shouldBe "99M1H"
+    mate.start shouldBe 100
+    mate.cigar.toString shouldBe "2H98M"
+  }
+
+  it should "clip reads that overlap in the first half of the pair, not over the middle" in {
```
nh13, line 621: What does "overlap in the first half of the pair," mean?

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
```diff
@@ -588,24 +588,74 @@ class SamRecordClipperTest extends UnitSpec with OptionValues {
...
+    val (rec, mate) = pair(1, "95M5S", Plus, 90, "20S80M", Minus)
+    clipper(Hard).clipOverlappingReads(rec=rec, mate=mate) shouldBe (3, 3)
+    rec.start shouldBe 1
+    rec.end shouldBe 92
+    rec.cigar.toString shouldBe "92M8H"
+    mate.start shouldBe 93
+    mate.cigar.toString shouldBe "23H77M"
+  }
+
+  it should "clip reads that overlap in the second half of the pair, not over the middle" in {
```
nh13, line 631: ditto

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
nh13: Also, I like your alternative as _I think_ it does a better job of balancing the # of bases trimmed off of each end of the read, which is where the low quality bases are more likely anyhow.

```scala
val midPoint  = { // Need to ensure the midpoint actually occurs in the overlapped region
  val retval = (rec.start + mate.end) / 2
  if (retval > rec.end) rec.end
  else if (retval < mate.start) mate.start
  else retval
}
```

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
```diff
@@ -588,24 +588,74 @@ class SamRecordClipperTest extends UnitSpec with OptionValues {
...
+    rec.end shouldBe 75
+    rec.cigar.toString shouldBe "75M25H"
+    mate.start shouldBe 76
+    mate.cigar.toString shouldBe "11H89M"
+  }
+
   it should "clip reads that overlap with one end having a deletion" in {
     val (rec, mate) = pair(1, "60M10D40M", Plus, 50, "10M10D80M10D10M", Minus)
-    clipper(Soft).clipOverlappingReads(rec=rec, mate=mate) shouldBe (25, 26)
+    clipper(Soft).clipOverlappingReads(rec=rec, mate=mate) shouldBe (30, 21)
```
nh13, line 643: I think you're alternative midpoint calculation will revert the # of bases clipped off of each end, so the two values are as close as possible.

nh13 (changes requested):

src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
jrm5100: This specific test may no longer be necessary since trimming shouldn't fully trim a read anymore.  Specifically, the result now evaluates to `false` because there shouldn't be a way for a read to be entirely trimmed.  That's actually how I tracked down the second regression (found an entirely trimmed read).

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
mjhipp: Those ascii images may be very helpful as comments in the tests

nh13 (approved):
Going to add @tfenne as a reviewer to weigh in on the calculation of the mid point.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -305,17 +305,22 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
+      }
       val readEnd   = rec.readPosAtRefPos(pos=midPoint, returnLastBaseIfDeleted=true)
       val mateStart = { // NB: need to be careful if the midpoint falls in a deletion
         val retval = mate.readPosAtRefPos(pos=midPoint + 1, returnLastBaseIfDeleted=false)
         if (retval != 0) retval else mate.readPosAtRefPos(pos=midPoint + 1, returnLastBaseIfDeleted=true) + 1
       }
-      val numOverlappingBasesRead = this.clip3PrimeEndOfRead(rec, rec.cigar.trailingClippedBases + rec.length - readEnd)
-      val numOverlappingBasesMate = this.clip3PrimeEndOfRead(mate, mate.cigar.leadingClippedBases + mateStart - 1)
+      val numOverlappingBasesRead = this.clip3PrimeEndOfRead(rec, rec.cigar.trailingHardClippedBases + rec.length - readEnd)
+      val numOverlappingBasesMate = this.clip3PrimeEndOfRead(mate, mate.cigar.leadingHardClippedBases + mateStart - 1)
```
tfenne, line 323: Wouldn't it be clearer here to use `clip3PrimeEndOfAlignment` (i.e. ....Alignment() instead of ....Read()) since you just tell that method how many _more_ bases you want clipped, which is what I think we already have?

src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
tfenne: This makes me a little nervous.  I get this test no longer causes a read to be unampped, but is it now theoretically impossible to fully clip a read?  What if the read is e.g. post indel realignment and has a CIGAR that is mostly/all insertion with 1bp overlapping a mate?

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
tfenne: I think for these tests it would be really helpful to add a comment inside the test something like this:

```scala
// Template spans chr1:-169, with mid-point=85
// Second read doesn't overlap the mid-point so all clipping should be done to the first read
```

Though having written that, I'm now curious why the mate/R2 gets clipped at all in this test?

tfenne (commented):
Looks good, but I have a few questions.

src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
jrm5100: A different test could be written (not sure if this is the best place for it or not), but clipping based on `clipOverlappingReads=true` shouldn't ever fully clip a read, unless maybe you have two 1bp reads that are overlapped?

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
jrm5100: That complicates deletion handling, since they are considered `clipped bases`.

For example,

```
val (rec, mate) = pair(2, "80M10D10M", Plus, 70, "10M10D80M", Minus)
```

Finds a midpoint of 85.

Currently we get (10, 10) with CIGARs of `80M10S`, `10S80M`

If calculated like this

```
val numOverlappingBasesRead = this.clip3PrimeEndOfAlignment(rec, rec.end - midPoint)
val numOverlappingBasesMate = this.clip3PrimeEndOfAlignment(mate, midPoint - mate.start + 1)      
```

We get (16,16) with CIGARs of `74M16S`, `16S74M`

src/test/scala/com/fulcrumgenomics/bam/SamRecordClipperTest.scala
jrm5100: Off-by-1 error.  It wasn't technically incorrect since it would do the same amount of trimming, but I've corrected it now so that it only trims the longer read if possible.

src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
tfenne: Fair enough.

tfenne (approved):

PR #864, opened 2022-07-27
[bugfix] fix issue #858 in CallDuplexConsensusReads

### LINKED ISSUE #858: Error in FilterConsensusReads
I get the same error as #847 while using v2.0.2

image

Code:
```
        fgbio -Xmx16g --async-io FilterConsensusReads  \
            -Djava.io.tmpdir=tmp  \
            --input=input  \
            --output=output  \
            --ref=ref \
            --min-reads=0 0 0 \
            --max-read-error-rate=0.05  \
            --max-base-error-rate=0.1  \
            --min-base-quality=50  \
            --max-no-call-fraction=0.05  \
            --require-single-strand-agreement=true
```

How the input to the above was generated:
```
        fgbio -Xmx16g --async-io CallDuplexConsensusReads  \
            -Djava.io.tmpdir=tmp  \
            --input=input  \
            --output=output  \
            --error-rate-pre-umi=45  \
            --error-rate-post-umi=30  \
            --min-input-base-quality=30
```
Also occurs with --min-reads=1 0 0

Here is that read:

image

nh13:
Fixes #858 where the AB strand single strand consensus read could have
zero depth and all Ns, such that the consensus read produced has an aD
value of zero but bD value > 0, causing issues in FilterConsensus Reads.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
@@ -340,15 +340,16 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
   private[umi] def duplexConsensus(ab: Option[VanillaConsensusRead],
                                    ba: Option[VanillaConsensusRead],
                                    sourceReads: Seq[SourceRead]): Option[DuplexConsensusRead] = {
-    (ab, ba) match {
+    // Calculate the length that we'll create a duplex read as "the shorter of the available reads",
+    // and then filter each read to make sure it has some coverage in that section of the read
+    val len = min(ab.map(_.length).getOrElse(Int.MaxValue), ba.map(_.length).getOrElse(Int.MaxValue))
+    val abX = ab.filter(_.depths.iterator.take(len).exists(_ > 0))
+    val baX = ba.filter(_.depths.iterator.take(len).exists(_ > 0))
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -345,6 +345,10 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
       case (None, Some(b))    => Some(DuplexConsensusRead(id=b.id, b.bases, b.quals, b.errors, b, None))
       case (Some(a), Some(b)) =>
         val len = min(a.length, b.length)
+        if (a.depths.take(len).forall(_ == 0)) {
+          require(b.depths.take(len).exists(_ > 0), "Bug: both consensus reads had zero depths across all bases")
+          return duplexConsensus(ba, ab, sourceReads)
+        }
```
nh13, line 351: I don't like returns, but I don't see a way to structure the match better without some ugly refactoring.  Thoughts?

nh13: Alternatively, we could change all single strand consensus reads that are all Ns to `None` we call `duplexConsensus`?

tfenne (approved):
Should we add a test too?

PR #867, opened 2022-07-30
Expose chunkSize parameter in CallDuplexConsensusReads

mjhipp:
I am suggesting that the `chunkSize` part of the `ConsensusCallingIterator` be exposed to the CLI for `CallDuplexConsensusReads`. 

I have hit a use case where the default value of `1024` UMI groups per chunk has prevented me from successfully running the tool in multi-threaded mode for an input file with very large and variable family sizes. By reducing the `chunkSize` to something like `2 * threads` I was able to get it to work.


src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -67,10 +67,10 @@ class ConsensusCallingIterator[ConsensusRead <: SimpleRead](sourceIterator: Iter
...
     }
     else {
-      ParIterator(groupingIterator, threads=threads).flatMap { rs =>
+      ParIterator(groupingIterator, threads=threads, chunkSize=chunkSize).map { rs =>
         val caller = callers.get()
         caller.synchronized { caller.consensusReadsFromSamRecords(rs) }
-      }.toAsync(chunkSize * 8)
+      }.toAsync(chunkSize).flatten
     }
   }
```
mjhipp, line 75: 1. Before, `chunkSize` parameter was not actually passed down to the `ParIterator`, only the `toAsync` part. The default value of `1024` was hard-coded here, which is unexpected as it is an input member of the `ConsensusCallingIterator` class, and would be expected to be passed on.
2. My suggestion is to move the flattening of this mapping operation until after the `.toAsync` call. My reasoning here is that that, according to the documentation, this value should be a multiple of `chunkSize`. If we flatten before calling `toAsync`, we are buffering that number of records, rather than that number of tag families, which the input `ParIterator` is working on. 
  - For example, if your `chunkSize` is 10, and your family size is 100, and we assume two consensus reads per family, each chunk would contain 1000 reads. 
  - If we assume each family produces 2 consensus reads, then each chunk would produce 20 reads.
  - If we flatten before `.toAsync`, then we are using 8*20 reads, which is not a multiple of one chunk of input. In this case, the buffer would not even hold one chunk's worth of input reads in memory.

clintval (commented):
I approve of the code change but am curious about accidentally causing a speed regression.

I will leave final approval on this one up to @tfenne since he recently modified this code.

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -67,10 +67,10 @@ class ConsensusCallingIterator[ConsensusRead <: SimpleRead](sourceIterator: Iter
       groupingIterator.flatMap(caller.consensusReadsFromSamRecords)
     }
     else {
-      ParIterator(groupingIterator, threads=threads).flatMap { rs =>
+      ParIterator(groupingIterator, threads=threads, chunkSize=chunkSize).map { rs =>
         val caller = callers.get()
         caller.synchronized { caller.consensusReadsFromSamRecords(rs) }
-      }.toAsync(chunkSize * 8)
+      }.toAsync(chunkSize).flatten
```
clintval, line 73: I would add some line comment on why we must flatten _**after**_ and not _**before**_ so no one accidentally causes a regression in the future.

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -108,6 +109,10 @@ class CallDuplexConsensusReads
  val maxReadsPerStrand: Option[Int] = None,
  @arg(doc="The number of threads to use while consensus calling.") val threads: Int = 1,
  @arg(doc="Consensus call overlapping bases in mapped paired end reads") val consensusCallOverlappingBases: Boolean = true,
+ @arg(doc="""
+          |The number of consensus groups to include in the buffer when using multiple threads.
+          |For input with very large family sizes, a smaller chunk size, such as 2*<threads> may be suitable.
+          """) val chunkSize: Int = DefaultChunkSize
```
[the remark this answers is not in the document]
mjhipp: I am not sure that making this value a multiple of `threads` will necessarily be most efficient, unless all families are extremely uniform in size. Its more that a larger value will be faster than a smaller value, but use more memory. I tried to address this in my updated comment.

mjhipp:
> curious about accidentally causing a speed regression.

If a user uses a smaller cache, it will pretty much always be slower, but the default is not changed here. 

The only newly hard-coded piece that I can see causing a speed regression is:

```diff
< .toAsync(chunkSize * 8)
> .toAsync(chunkSize)
```
However, my understanding is that this should probably be about equal in ideal cases, since we are flattening after async. So this would be exactly the same number of reads if all families had 8 raw reads. In most cases, the average family size will probably be higher in my experience, so it would potentially cause a small speedup and use slightly more memory.

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -108,6 +109,12 @@ class CallDuplexConsensusReads
  val maxReadsPerStrand: Option[Int] = None,
  @arg(doc="The number of threads to use while consensus calling.") val threads: Int = 1,
  @arg(doc="Consensus call overlapping bases in mapped paired end reads") val consensusCallOverlappingBases: Boolean = true,
+ @arg(doc="""
+            |Pull reads from this many source molecules into memory for multi-threaded processing.
+            |Using a smaller value will require less memory but will negatively impact processing speed.
+            |For very large family sizes, a smaller value may be necessary to reduce memory usage.
+            |This value is only used when `--threads > 1`.
+          """) val maxSourceMoleculesInMemory: Int = DefaultChunkSize
```
clintval, line 117: Do we recommend this be a multiple of threads for when threads > 1? I'm imagining that would be the most efficient guidance if family sizes and the computation time to create a consensus was stable.

clintval (approved):

clintval:
@tfenne are you OK if we merge this?

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
mjhipp: I don't necessarily think it is going to make a difference https://github.com/fulcrumgenomics/fgbio/pull/867#discussion_r935847745

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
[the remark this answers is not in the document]
clintval: Thanks!

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -67,10 +67,11 @@ class ConsensusCallingIterator[ConsensusRead <: SimpleRead](sourceIterator: Iter
...
+      // Flatten AFTER pulling through ParIterator to keep input chunks in phase with output
```
nh13, line 74: If I understand this correctly, we previously got chunks out of order in some cases?
```suggestion
      // Flatten AFTER pulling through ParIterator to keep the iterator in a stable order
```

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -67,10 +67,11 @@ class ConsensusCallingIterator[ConsensusRead <: SimpleRead](sourceIterator: Iter
       groupingIterator.flatMap(caller.consensusReadsFromSamRecords)
     }
     else {
-      ParIterator(groupingIterator, threads=threads).flatMap { rs =>
+      ParIterator(groupingIterator, threads=threads, chunkSize=chunkSize).map { rs =>
         val caller = callers.get()
         caller.synchronized { caller.consensusReadsFromSamRecords(rs) }
-      }.toAsync(chunkSize * 8)
+      }.toAsync(chunkSize).flatten
```
nh13, line 73: I am confused with the chunk size passed to `toAsync` is reduced.  Don't we want this value to multiples of the input `chunkSize`.  Furthermore, since the `ParIterator` is presumably processing `threads` chunks at any one time, I think this should be `chunkSize * threads`?
 
```suggestion
      }.toAsync(chunkSize * threads).flatten
```

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
```diff
@@ -40,7 +40,7 @@ import com.fulcrumgenomics.util.ProgressLogger
   * @param caller the consensus caller to use to call consensus reads
   * @param progress an optional progress logger to which to log progress in input reads
   * @param threads the number of threads to use.
-  * @param chunkSize parallel process in chunkSize units; will cause 8 * chunkSize records to be held in memory
+  * @param chunkSize across the input [[SamRecord]]s from this many source molecules at a time
```
nh13, line 43: The new param doc doesn't track for me.  Based on my comment below:
```suggestion
  * @param chunkSize the number of source molecules to process in a single chunk, where chunks
                     are processed in parallel.  This will cause `2 * threads * chunkSize` source 
                     molecules to be held in memory, where the number of [[SamRecord]]s
                      held in memory depends on the number of [[SamRecord]]s assigned to 
                      each source molecule at any given time.
```

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -142,7 +149,7 @@ class CallDuplexConsensusReads
       maxReadsPerStrand   = maxReadsPerStrand.getOrElse(VanillaUmiConsensusCallerOptions.DefaultMaxReads)
     )
     val progress = ProgressLogger(logger, unit=1000000)
-    val iterator = new ConsensusCallingIterator(inIter, caller, Some(progress), threads)
+    val iterator = new ConsensusCallingIterator(inIter, caller, Some(progress), threads, maxSourceMoleculesInMemory)
```
nh13, line 152: Can you please add the same to `CallMolecularConsensusReads`?  Likely it is not immune from the same issue, so it'd be nice to control there too (and have symmetry of options).

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -108,6 +109,12 @@ class CallDuplexConsensusReads
  val maxReadsPerStrand: Option[Int] = None,
  @arg(doc="The number of threads to use while consensus calling.") val threads: Int = 1,
  @arg(doc="Consensus call overlapping bases in mapped paired end reads") val consensusCallOverlappingBases: Boolean = true,
+ @arg(doc="""
+            |Pull reads from this many source molecules into memory for multi-threaded processing.
```
nh13, line 113: not strictly true, since the `ParIterator` will have `threads * maxSourceMoleculesInMemory`, and so will the `AsyncIterator` that wraps it, so it should be `2 * threads * maxSourceMoleculesInMemory` I think.

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala
tfenne: If this is true, then presumably in `CallDuplexConsensusReads` the number given should be divided before supplying here?  I think it will be very confusing to say "only hold 10 source molecules worth of reads in memory" and then find out that with 8 threads I've got 160 molecules worth.

mjhipp: @tfenne @nh13 Many of these comments, here and elsewhere in the review, go against my interpretation of how `ParIterator` works. The way I understand it, these iterable functions operate on a single chunk at a time (not counting `.toAsync`), using the multi-threaded task support, rather then grabbing one chunk per-thread:

https://github.com/fulcrumgenomics/commons/blob/9372c5585084a1d7486c1b6036d0e06824b2c8d4/src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala#L140-L141

Only one chunk is pulled into memory at a time at the "front" of the `ParIterator` (it does not mention `threads * chunkSize`:

https://github.com/fulcrumgenomics/commons/blob/9372c5585084a1d7486c1b6036d0e06824b2c8d4/src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala#L119-L120

The `parWith` function splits a single chunk across threads:

https://github.com/fulcrumgenomics/commons/blob/9372c5585084a1d7486c1b6036d0e06824b2c8d4/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala#L360-L364

It was based on this interpretation that I do not divide by `threads` before parallelizing. Can you please let me know if I have missed something or where my interpretation is wrong?

PR #868, opened 2022-07-31
Add an option to sort by sequence role in CollectAlternateContigNames

nh13:
This is so that I can have assembled molecules first, whereas in some assembly reports chrM (an assembled molecule) is listed at the end of the report.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
@@ -96,8 +96,9 @@ object SequenceRole extends FgBioEnum[SequenceRole] {
     SequenceRole(roleValue)
   }
 
-  case object AltScaffold         extends SequenceRole { val key: String = "alt-scaffold" }
+  // NB: the order in which these are defined is important for when `--sort-by-sequence-role` is used
   case object AssembledMolecule   extends SequenceRole { val key: String = "assembled-molecule" }
+  case object AltScaffold         extends SequenceRole { val key: String = "alt-scaffold" }
   case object FixPatch            extends SequenceRole { val key: String = "fix-patch" }
   case object NovelPatch          extends SequenceRole { val key: String = "novel-patch" }
   case object UnlocalizedScaffold extends SequenceRole { val key: String = "unlocalized-scaffold" }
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -213,7 +216,18 @@ class CollectAlternateContigNames
...
+          // Use the user-provided roles, otherwise use all role
+          val roles = (if (sequenceRoles.nonEmpty) sequenceRoles.toIndexedSeq else SequenceRole.values)
+            .zipWithIndex
+            .map { case (role, index) =>  role.key -> index }
+            .toMap
+          val metas = metadatas.toSeq.sortBy { metadata =>
+            metadata.get(Column.SequenceRole.tag).map { name => roles(name) }.getOrElse(roles.size)
+          }.zipWithIndex.map { case (metadata, index) => metadata.copy(index=index) }
+          SequenceDictionary(metas:_*)
+        }
```
clintval, line 230: Unit tests recommended for the new code branches you created including whether `sequenceRoles` is defined or not. You have `roles(name)` instead of `roles.get(name)` so I'm assuming code earlier guarantees the metadata doesn't have any invalid role names?

src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -213,7 +216,18 @@ class CollectAlternateContigNames
...
-      case None => SequenceDictionary(metadatas.toSeq:_*)
+      case None       =>
+        if (!sortBySequencingRole) SequenceDictionary(metadatas.toSeq:_*) else {
+          // Use the user-provided roles, otherwise use all role
+          val roles = (if (sequenceRoles.nonEmpty) sequenceRoles.toIndexedSeq else SequenceRole.values)
+            .zipWithIndex
+            .map { case (role, index) =>  role.key -> index }
+            .toMap
+          val metas = metadatas.toSeq.sortBy { metadata =>
+            metadata.get(Column.SequenceRole.tag).map { name => roles(name) }.getOrElse(roles.size)
```
clintval, line 227: Should this be?

```suggestion
            metadata.get(Column.SequenceRole.tag).flatMap { name => roles.get(name) }.getOrElse(roles.size)
```

If I'm understanding the `.getOrElse(roles.size)` correctly, what that is doing is for when a metadata has a role that is not in the provided (or known) list of roles then the metadata entries will be forced to sort to the "end" of the metadata list?

clintval (approved):
Approve contingent on adding some unit tests for this new feature.

PR #871, opened 2022-08-16
Add an option to skip the mate cigar in EstimateRnaSeqInsertSize

nh13:


src/main/scala/com/fulcrumgenomics/rnaseq/EstimateRnaSeqInsertSize.scala
```diff
@@ -90,12 +89,20 @@ class EstimateRnaSeqInsertSize
...
+
+      rec.firstOfPair && // so templates are considered only once
+        rec.pf && // don't want low quality reads
+        rec.paired && // only want paired reads
+        !rec.secondary && !rec.supplementary && // no secondary or supplementary
+        rec.mapq >= minimumMappingQuality && // decent mapping quality
+        (includeDuplicates || !rec.duplicate) && // no duplicates, unless specified
+        rec.mapped && rec.mateMapped && // both ends of a pair are mapped
+        rec.refIndex == rec.mateRefIndex && // both ends of a pair map to the same contig
+        (!skipMissingMateCigar || rec.mateCigar.isDefined) // skip records with missing mate cigars if specified
```
nh13, line 105: moved filters out of the old "htsjdk" filter style, so this could use a double-check

src/main/scala/com/fulcrumgenomics/rnaseq/EstimateRnaSeqInsertSize.scala
```diff
@@ -54,6 +51,7 @@ import scala.jdk.CollectionConverters._
     |chromosomes. Finally, skips transcripts where too few mapped read bases overlap exonic sequence.
     |
     |This tool requires each mapped pair to have the mate cigar (`MC`) tag.  Use `SetMateInformation` to add the mate cigar.
+    |Use the `--skip-missing-mate-cigar` to skip these.
```
NatPRoach, line 54:
```suggestion
    |Use the `--skip-missing-mate-cigar` to skip reads missing the mate cigar tag.
```

NatPRoach (approved):
LGTM, thanks @nh13

clintval:
@nh13 could you clarify why this feature is useful or when someone would want to turn it to `true`? The pre-existing help text for this tool requires that for mapped pairs, both reads have the `MC` tag which seems good and sensible to me. The `MC` tags should exist on mapped pairs if `SetMateInformation` was run.

The new unit test shows a mapped proper pair with the `MC` tags and then another similar pair without the `MC` tags:

https://github.com/fulcrumgenomics/fgbio/blob/24eb2a96c13984b7261459b830b004ea42e96785/src/test/scala/com/fulcrumgenomics/rnaseq/EstimateRnaSeqInsertSizeTest.scala#L354-L357

What real-world situation is this modeling?

I'm concerned that someone might run this with `--skip-missing-mate-cigar=true` after failing to run `SetMateInformation` (or similar) and then no good data will be produced by the tool since all reads will be ignored.

nh13:
@clintval in some corner cases, and in a small subset of reads, some _"aligners"_ may not produce mate cigars.  I'd rather have an option to just skip these problematic reads, since this metric tools is producing gross metrics.   Happy to discuss offline too.

nh13:
I am going to close this PR and keep it here for posterity, in case there ever is a future need.  In that case, we can use that future need as motivation and reopen this PR.  Thank-you for the review and discussion!

PR #881, opened 2022-10-20
Tweak the size of caches for parallel consensus calling down to reduc…

tfenne:
…e memory usage.

@nh13 In my hands this is no slower on "normal" looking data (i.e. 1-10s of reads per tag family), but significantly reduces memory usage when pathological data is present (e.g. tag families with 10,000s of reads).  Do you think it's ok to just tweak these down, or do you think we should add command line parameters to control cache sizing?


clintval:
FYI: https://github.com/fulcrumgenomics/fgbio/pull/867

tfenne:
> FYI: #867
Woah, I'd totally forgotten about that PR!

nh13 (approved):

nh13:
I like #867 and we'd be happy for it to remain open, but this seems like a good quick win.

PR #891, opened 2022-12-11
A CLP to downsample a bam to a specific coverage level.

yfarjoun:
A CLP to downsample a bam to a specific coverage level. 
- keeps "templates" together, and processes the templates in a deterministically random order
- counts no more than coverage 1 per template
- ignores duplicates, pf-fail, secondary and unmapped reads
- allows for a minimal mapping quality threshold for counting toward coverage
- allows for a region of interest to be provided

Also included is a new class "LocusTrack" which allows for a Seq[A] to be colocalized with a locus, including taking sub-loci etc. 

Currently in "personal", but if this is taken in, I'm happy to move it wherever makes most sense.


nh13:
We can re-open this if a client needs this and we can polish it up then?

PR #893, opened 2022-12-29
Tool to downsample a BAM while retaining reads in low coverage areas.

tfenne:
Still needs tests.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
@@ -1,0 +1,206 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2022 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.bam
+
+import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.bam.DownsampleAndNormalizeBam.CoverageTracker
+import com.fulcrumgenomics.bam.api._
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.commons.util.LazyLogging
+import com.fulcrumgenomics.sopt.{arg, clp}
+import com.fulcrumgenomics.util.{Io, ProgressLogger, Sorter}
+import htsjdk.samtools.util.IntervalList.IntervalMergerIterator
+import htsjdk.samtools.util.{CoordMath, Interval, IntervalList, Murmur3, OverlapDetector}
+
+/**
+  * Companion object with helper classes.
+  */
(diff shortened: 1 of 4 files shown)


src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
```diff
@@ -0,0 +1,206 @@
...
+  /**
+    * Coverage counter for a region - all coordinates in the public API are 1-based closed ended.
+    */
+  class RegionCoverage(val chrom: String, val start: Int, val end: Int) {
+    private val offset = start
+    private val counts = new Array[Int](CoordMath.getLength(start, end))
+
+    /** Returns the coverage at the given genomic position, or -1 if the position is not between start:end. */
+    def apply(i: Int): Int = {
+      if (i < start || i > end) -1
```
nh13, line 106: Consider either returning `0` (zero coverage) or `None` (no coverage) instead of `-1`, which feels very Java-like.

src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
```diff
@@ -0,0 +1,206 @@
...
+      * Checks to see if a template has reads that contribute coverage to any bases that are currently
+      * below the coverage threshold.  If so, coverage from all acceptable reads is accumulated and the
+      * method returns true.  Otherwise no coverage is accumulated and the method returns false.
+      *
+      * @param t a Template to accumulate coverage from if it covers any bases that are under the coverage threshold
+      * @return true if the template added coverage, false otherwise
+      */
+    def add(t: Template): Boolean = {
+      val recs = t.allReads
+        .filterNot(_.secondary)
```
nh13, line 65: This made me think about supplementary reads.  Should those be upgraded to "primary" alignments or be filtered out as well?  I'm thinking of a region where we have a large number of both primary and supplementary alignments, and how to pick there.

src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
```diff
@@ -0,0 +1,206 @@
...
+        }
+      }
+
+      if (addsCoverage) {
+        val intervals = {
+          val iter = recs
+            .sortBy(r => (r.refName, r.start, r.end))
+            .iterator
+            .map(r => new Interval(r.refName, r.start, r.end))
+          new IntervalMergerIterator(iter, true, false, false)
```
nh13, line 85: Sad we can't call arguments by name, since what are the three booleans here?

src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
```diff
@@ -0,0 +1,206 @@
...
+        .filterNot(_.secondary)
+        .filterNot(_.unmapped)
+        .filterNot(_.duplicate)
+        .filter(_.mapq >= minMapQ)
+        .toSeq
+
+      val addsCoverage = recs.exists { rec =>
+        detector.getOverlaps(rec.asSam).iterator().exists { cov =>
+          Range.inclusive(math.max(rec.start, cov.start), math.min(rec.end, cov.end)).exists { pos =>
+            cov(pos) <= targetCoverage
```
nh13, line 74: I think we only want to add the record if the "if a template has reads that contribute coverage to any bases that are currently below the coverage threshold":
```suggestion
            cov(pos) < targetCoverage
```

src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
```diff
@@ -0,0 +1,206 @@
...
+    def add(t: Template): Boolean = {
+      val recs = t.allReads
+        .filterNot(_.secondary)
+        .filterNot(_.unmapped)
+        .filterNot(_.duplicate)
+        .filter(_.mapq >= minMapQ)
+        .toSeq
+
+      val addsCoverage = recs.exists { rec =>
+        detector.getOverlaps(rec.asSam).iterator().exists { cov =>
```
nh13, line 72: Perhaps it's time to have `getOverlaps` return an iterator itself?

nh13 (commented):
Just a few comments, one potential bug.

src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
tfenne: I definitely want to include supplementary reads, they're just part of a split-read alignment and therefore contribute to coverage.  E.g. imagine an alignment to a circular genome where anything that maps over the break point will have a primary+supplementary, we want to count both of those.

src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
tfenne: I think perhaps it's time to move on from HTSJDK's OverlapDetector and implement one in scala that suits our needs better :/  But not today.

src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
tfenne: Agreed.  For reference they are:

- combineAbutting
- enforceSameStrand
- concatenateNames

src/main/scala/com/fulcrumgenomics/bam/DownsampleAndNormalizeBam.scala
tfenne: I think actually I can just go back to having it fail if the index is out of bounds.

Issue #896, opened 2023-01-17
Is CallDuplexConsensusReads supposed to outputs reads as unmapped bam?


karlbeutner:
Hello, The output I am getting `fromCallDuplexConsensusReads` has no @SQ tags in the header, and no mapping info (rname, flags, cigar, etc) in the reads. I don't see anywhere in the documentation where the output is described as either aligned, or unaligned so it is hard for me to know if this is the expected result. I assumed the mapping info could/would be inferred from the component reads - but it does not look like this is the case. 

Is there a way to run this command so that the output includes alignment information for the consensus reads?

nh13:
@karlbeutner that's a good point about the docs not saying the consensus reads produced _do not_ have mapping information. You can check out our best-practice recommendations here: https://github.com/fulcrumgenomics/fgbio/blob/main/docs/best-practice-consensus-pipeline.md, where you'll see that one needs to re-align the data after calling consensus reads.

nh13:
Not implemented, and hard.

karlbeutner:
Can you elaborate on what makes it hard, @nh13?

I can think of two approaches for doing this. I would like to hear what makes this hard so that I can think through how/if the approaches I have in mind address what you have already identified as the major challenge(s).

nh13:
It’s a lot of book-keeping so hard in practice not in theory. And then all the SAM tags, and what about secondary and supplementary alignments. 

And since the number of consensus reads are typically much smaller than the raw reads, asking folks to re-align the consensus reads isn’t onerous. So I don’t think this is low hanging fruit.  

Closing this thread as I believe the original question has been answered.

PR #897, opened 2023-01-17
Improve docs for consensus reads being unaligned

nh13:
See #896.  Will apply to `CallMolecularConsensusReads` if this working is accepted.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
@@ -52,6 +52,10 @@ import com.fulcrumgenomics.util.ProgressLogger
     |input they are _ignored_.  Similarly, read pairs for which consensus reads cannot be generated for one or
     |other read (R1 or R2) are omitted from the output.
     |
+    |The consensus reads produced are unaligned, due to the difficulty and error-prone nature of inferring the conesensus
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -52,6 +52,10 @@ import com.fulcrumgenomics.util.ProgressLogger
     |input they are _ignored_.  Similarly, read pairs for which consensus reads cannot be generated for one or
     |other read (R1 or R2) are omitted from the output.
     |
+    |The consensus reads produced are unaligned, due to the difficulty and error-prone nature of inferring the conesensus
+    |alignment.  Consensus reads should therefore be aligned after, which should not be too expensive as likely there
+    |are far fewer conesnsus reads than input raw raws.
```
tfenne, line 57: Sounds good.  Maybe reference the best practices doc for how to do it?

tfenne (approved):

src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
nh13, line 57:
```suggestion
    |are far fewer consensus reads than input raw raws.  Please see how best to use this tool within the best-practice
    |pipeline: https://github.com/fulcrumgenomics/fgbio/blob/main/docs/best-practice-consensus-pipeline.md
```

Issue #906, opened 2023-03-25
Use GroupReadsByUmi for standard duplicate marking?


eboyden:
Few tools exist that will properly perform umi-based duplicate marking.  Picard seems to be the only tool that can mark duplicates with error tolerance _and_ properly sets the bitflag, but its algorithm isn't sophisticated; it just uses a hamming distance.  Samtools recently added umi awareness to markdup, but currently it is not error-tolerant.  Umi-tools can _remove_ duplicates, but its markdup function doesn't work properly on paired end data, and doesn't set the bitflag.  Whereas fgbio GroupReadsByUmi works properly on paired data using a sophisticated algorithm, but doesn't set the bitflag (I realize this is because it is intended to be run upstream of CallMolecularConsensusReads, not for standard duplicate marking).

One workaround is to run GroupReadsByUmi and then pipe the output to Samtools markdup, pointing it to the `MI` tag instead of the `RX` tag; the error tolerance has already been handled so Samtools doesn't need to do the heavy lifting there.  This seems to work well and produces the expected output.  But even when `-m 0` is used, GroupReadsByUmi still filters out a lot of reads (unmapped or N-containing umis), whereas it would be ideal if the final marked bam were lossless.

Could functionality be added to GroupReadsByUmi to simply ignore reads that would otherwise be filtered, and then perform that filtering in the consensus calling tools themselves (or ignore them there as well)?  If it would break something to not apply an `MI` tag, all reads destined for filtering could receive a unique or "N" `MI` tag.  Selecting a "best" read in a family and setting the duplicate bitflag for the rest (thereby eliminating the need for a secondary tool) would be nice but I realize it may be out of scope.

Related: GroupReadsByUmi filters out reads with Ns in the umi.  Could this be modified so that "N" always counts as a mismatch (including to other Ns), and reads within the specified edit distance are still collapsed?  E.g. with `-e 1` ACGN would still be collapsed into ACGT.  (This is how Picard handles Ns in umis.)

nh13:
Sounds like a useful feature. Thanks for the suggestion!

JoeVieira:
@nh13 , @eboyden & I are colleagues at Molecular Loop, I'd be happy to contribute a PR for this functionality if you're willing to accept it. This would help us out & we want to contribute back to the project.

nh13:
[...]

PR #907, opened 2023-04-07
Add option to maintain N in umi

JoeVieira:
Add flag `N` to disable filtering of reads which contain non-ATCG containing UMIs 
Created test case as well.

This was discussed in issue #906

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -482,6 +482,7 @@ class GroupReadsByUmi
   @arg(flag='T', doc="The output tag for UMI grouping.") val assignTag: String = "MI",
   @arg(flag='m', doc="Minimum mapping quality for mapped reads.")         val minMapQ: Int      = 1,
   @arg(flag='n', doc="Include non-PF reads.")            val includeNonPfReads: Boolean = false,
+  @arg(flag='N', doc="Use non-ATCG bases (N) in clustering.")             val includeNonATCG: Boolean = false,
   @arg(flag='s', doc="The UMI assignment strategy.")     val strategy: Strategy,
   @arg(flag='e', doc="The allowable number of edits between UMIs.") val edits: Int = 1,
   @arg(flag='l', doc= """The minimum UMI length. If not specified then all UMIs must have the same length,
@@ -551,7 +552,8 @@ class GroupReadsByUmi
       .filter(r => (r.mapped || (r.paired && r.mateMapped))                         || { filteredPoorAlignment += 1; false })
       .filter(r => (allowInterContig || r.unpaired || r.refIndex == r.mateRefIndex) || { filteredPoorAlignment += 1; false })
       .filter(r => mapqOk(r, this.minMapQ)                                          || { filteredPoorAlignment += 1; false })
-      .filter(r => !r.get[String](rawTag).exists(_.contains('N'))                   || { filteredNsInUmi += 1; false })
+      .filter(r =>
+        (this.includeNonATCG || !r.get[String](rawTag).exists(_.contains('N')))     || { filteredNsInUmi += 1; false })
       .filter { r =>
         this.minUmiLength.forall { l =>
           r.get[String](this.rawTag).forall { umi =>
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -482,6 +482,7 @@ class GroupReadsByUmi
   @arg(flag='T', doc="The output tag for UMI grouping.") val assignTag: String = "MI",
   @arg(flag='m', doc="Minimum mapping quality for mapped reads.")         val minMapQ: Int      = 1,
   @arg(flag='n', doc="Include non-PF reads.")            val includeNonPfReads: Boolean = false,
+  @arg(flag='N', doc="Include UMIs which contain non-ATCG bases (N) in output.")             val includeNonATCG: Boolean = false,
```
nh13, line 485: I don't think we are supporting non-ACGT, for example IUPAC codes.  We're treating Ns as "fifth" base (i.e. an N mismatches an A), so we need to add that to the docstring.  Also see how we calculate the length of the UMI based on how many `[ACGTN]`s there are (not non-ATCGs).   Finally the usage needs updating (see `number of [ACGT]...` and `non-ACGT characters)`).
```suggestion
  @arg(flag='N', doc="Allow UMIs which contain Ns.  Ns will be treated as mismatches to all other non-Ns") val includeNs: Boolean = false,
```

nh13 (changes requested):
Looking good so far.  Thank-you for contributing!

JoeVieira:
Thanks for the feedback! Sorry I missed the usage doc!
Let docs look clear enough, I think they are clear now.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -466,29 +466,32 @@ object Strategy extends FgBioEnum[Strategy] {
...
+ @arg(flag='N', doc="Allow UMIs which contain Ns.  Ns will be treated as mismatches to all other non-Ns")
+  val includeNs: Boolean = false,
+ @arg(flag='s', doc="The UMI assignment strategy.")     val strategy: Strategy,
+ @arg(flag='e', doc="The allowable number of edits between UMIs.") val edits: Int = 1,
+ @arg(flag='l', doc= """The minimum UMI length. If not specified then all UMIs must have the same length,
                        |otherwise discard reads with UMIs shorter than this length and allow for differing UMI lengths.
                        |""")
   val minUmiLength: Option[Int] = None,
-  @arg(flag='x', doc= """
+ @arg(flag='x', doc= """
```
nh13, line 494: Can you revert the indent?

nh13 (approved):
One minor tweak.  @tfenne can you also take a look?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
JoeVieira: Sorry, didn't even notice the white space change.

JoeVieira:
@tfenne since i've your attention - could you wrap this one up too?

tfenne:
So ... I think there's an issue with this.  It's a bit of an edge case, but when running with `--strategy paired` I think you could in theory create a graph of UMIs that are e.g. `ACGT` -> `ACGN` -> `ACNN` -> `ANNN` -> `NNNN`, and that doesn't seem good.

Similarly if you have a bunch of failed UMI reads that are all Ns it will treat the Ns as matches to each other and create groups, when you have no evidence of what the UMI actually was.

What do you think about:

1. When allowing Ns still filter out UMIs that contain more than `edits` Ns?
2. Changing it from `includeNs: boolean` to `maxNsInUmis: Int = 0`, so that users can decide how many Ns are ok?
3. Change the matching logic to say that N mismatches everything (including other Ns) so that the `edit` threshold is breached when there are too many Ns

I think I like (2) the best as it's explicit _and_ (3) might cause a performance problem when we have lots of reads at a given position.

JoeVieira:
Yes, that case 100% exists. Glad you're bringing this up.
A couple additional thoughts.

1.) I think umi's which are fully N need filtering regardless ( they are just non-sense ) 

2.) I personally think that your option 3 is the most internally consistent option, while 2 is explicit it's a bit of an odd usage & puts a fair amount of error possibility into the equation ( it precludes my point number 1 & also explicit would allow the issue you enumerated ) 

Alternatively ( which is what i did in this PR ) assume that if a case like this occurs one has very bad data & it's a "use at your own risk" In my testing with good ( high input germline data ) and hard ( lower input somatic type data ) this case hasn't occured ( of course it could ) 

i'd be happy to do any of these ( other than your 1, i think that's an odd conflation of parameters ), but do prefer what i laid out above assuming it's not performance encumbering.

JoeVieira:
@tfenne updated per ideas above - lmk what you think.

nh13:
We haven't been able to prioritize this volunteer work, so feel free to reach out to us if you'd like to sponsor it.

PR #909, opened 2023-04-12
Ensure FilterSomaticVcf handles PASS variants correctly

clintval:
If the tool was run on data that had `PASS` in the FILTER field then the output could be:

```console
ATailingArtifact;EndRepairFillInArtifact;PASS
```

When the expected output is:

```console
ATailingArtifact;EndRepairFillInArtifact
```

This PR handles variants with `PASS` correctly by unsetting `PASS` if new filters are added.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
+++ b/src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
@@ -26,17 +26,18 @@ package com.fulcrumgenomics.vcf.filtration
 
 import com.fulcrumgenomics.bam.api.SamSource
 import com.fulcrumgenomics.bam.pileup.PileupBuilder
-import com.fulcrumgenomics.bam.pileup.PileupBuilder.{BamAccessPattern, PileupDefaults}
 import com.fulcrumgenomics.bam.pileup.PileupBuilder.BamAccessPattern.RandomAccess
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/vcf/filtration/FilterSomaticVcf.scala
```diff
@@ -182,16 +183,21 @@ import scala.collection.immutable.ListMap
...
             val annotations = filter.annotations(pileup, variant.genotypes(name))
-            allAnnotations ++= annotations
-            allFilters     ++= filter.filters(annotations)
+            newAnnotations ++= annotations
+            newFilters     ++= filter.filters(annotations)
+          }
+
+          val allFilters = {
+            if (variant.filters == PassingFilters && newFilters.nonEmpty) newFilters.toSet
+            else variant.filters ++ newFilters
```
clintval, line 197: Appending filters to a `PASS` in a variant seems like an easy mistake (which I've made!)

What do you think of a helper that would perform a copy and handle `PASS` correctly like:

- `Variant.withFilters()`
- `Variant.withFilter()`

The plural would add all filters in a single copy. The singular would just add one filter.

tfenne (approved):
LGTM.  @clintval would you like me to merge or would you like the honors?

clintval:
Thanks @tfenne! I decided not to implement the helper method since it would do 1 copy, then I would need to do another copy to override the new `INFO` attributes. I suppose I could have made a static helper method without a copy, but it didn't seem worth the energy this time around.

PR #912, opened 2023-05-15
Fix complement of W and S iupac codes.

tfenne:


nh13 (changes requested):
lets chat about this, I think seeing your change `complement` is overloaded and we'd need to change the other mappings too, but I could be wrong so want to chat about it.

nh13 (approved):
Thank-you, the `complement` is not set complement, so we get this.

PR #917, opened 2023-06-21
Add optional validation of kept read ratio to CorrectUmis

mjhipp:
This is a proposed addition of an optional minimum kept ratio input to `CorrectUmis`.
If used, the value must be between 0 and 1, and will cause an error if the ratio of kept/total reads is below the provided minimum value. The error occurs after writing all output files and metrics.

The reasoning for this addition is to catch an instance of a user providing the wrong input UMI list, or a user performing `CorrectUmis` on a library preparation that did not use defined UMIs. In both of those cases, `CorrectUmis` would filter the majority of reads, leaving a very small BAM output file, and complete successfully. In the case of a long pipeline, this UMI incompatibility may not be caught until all other analyses are completed, or at all. With this change, setting a minimum kept ratio as low as 0.01 could catch the error in most cases.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
@@ -134,7 +134,8 @@ class CorrectUmis
   @arg(flag='U', doc="File of UMI sequences, one per line.", minElements=0) val umiFiles: Seq[FilePath] = Seq.empty,
   @arg(flag='t', doc="Tag in which UMIs are stored.") val umiTag: String = ConsensusTags.UmiBases,
   @arg(flag='x', doc="Don't store original UMIs upon correction.") val dontStoreOriginalUmis: Boolean = false,
-  @arg(doc="The number of uncorrected UMIs to cache; zero will disable the cache.") val cacheSize: Int = 100000
+  @arg(doc="The number of uncorrected UMIs to cache; zero will disable the cache.") val cacheSize: Int = 100000,
+  @arg(doc="The minimum ratio of kept UMIs to accept. A ratio below this will cause a failure (but all files will still be written).") val minCorrected: Option[Double] = None
 ) extends FgBioTool with LazyLogging {
 
   validate(umis.nonEmpty || umiFiles.nonEmpty, "At least one UMI or UMI file must be provided.")
@@ -142,6 +143,7 @@ class CorrectUmis
   Io.assertReadable(umiFiles)
   Io.assertCanWriteFile(output)
   rejects.foreach(Io.assertCanWriteFile(_))
+  minCorrected.foreach(m => validate(m >= 0 && m <= 1, "--min-corrected must be between 0 and 1."))
 
   // Construct the cache
   private lazy val cache = new LeastRecentlyUsedCache[String,UmiMatch](maxEntries = cacheSize)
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -134,14 +134,16 @@ class CorrectUmis
...
+  @arg(doc="The number of uncorrected UMIs to cache; zero will disable the cache.") val cacheSize: Int = 100000,
+  @arg(doc="The minimum ratio of kept UMIs to accept. A ratio below this will cause a failure (but all files will still be written).") val minCorrected: Option[Double] = None
 ) extends FgBioTool with LazyLogging {
 
   validate(umis.nonEmpty || umiFiles.nonEmpty, "At least one UMI or UMI file must be provided.")
   Io.assertReadable(input)
   Io.assertReadable(umiFiles)
   Io.assertCanWriteFile(output)
   rejects.foreach(Io.assertCanWriteFile(_))
+  minCorrected.foreach(m => validate(m > 0 && m <= 1, "--min-corrected must be between 0 and 1."))
```
nh13, line 146: Just in case folks want to always specify this option, and set it to zero
```suggestion
  minCorrected.foreach(m => validate(m >= 0 && m <= 1, "--min-corrected must be between 0 and 1."))
```

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -260,6 +262,15 @@ class CorrectUmis
       if (wrongLengthRecords > 0) logger.error(s"# ${wrongLengthRecords} had unexpected UMIs of differing lengths in the BAM file!")
       logger.error("###################################################################")
     }
+
+    minCorrected.foreach{ min =>
```
nh13, line 266:
```suggestion
    minCorrected.foreach { min =>
```

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -260,6 +262,15 @@ class CorrectUmis
...
+      val ratioKept = 1.0 * kept / totalRecords
+      if (ratioKept < min) {
+        logger.error("###################################################################")
+        logger.error(f"# Final ratio of reads kept / total was ${ratioKept}%2.2f (user specified minimum was ${min}%2.2f)")
+        logger.error("###################################################################")
```
nh13, line 271: did you want to throw an exception?

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -260,6 +262,14 @@ class CorrectUmis
       if (wrongLengthRecords > 0) logger.error(s"# ${wrongLengthRecords} had unexpected UMIs of differing lengths in the BAM file!")
       logger.error("###################################################################")
     }
+
+    minCorrected.foreach { min =>
+      val ratioKept = 1.0 * kept / totalRecords
+      validate(ratioKept >= min,
```
nh13, line 268: Validation exceptions are special for the constructor, validation command line parameters.  Sorry if this is not explicit anywhere.  How about assert?

mjhipp:
Failed checks are from 2 test suites being aborted based on env. One is log4j, other is Intel deflater related

nh13:
@mjhipp can you rebase onto main now that #918 is merged?

nh13 (approved):

nh13:
Thanks @mjhipp !

PR #929, opened 2023-08-17
bugfix: ZipperBams should consume any remaining mapped reads

nh13:
So that pipes do not cause failure when unread data exists in bash strict mode.

@jacarey

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
@@ -242,9 +242,11 @@ class ZipperBams
 
     // There really should be no more reads, but lets continue on if so
     if (mappedIter.hasNext) {
-      logger.warning("Processed all unmapped reads but mapped reads remaining, continuing")
-      // consume the remaining so piping in strict mode does not fail
-      mappedIter.foreach { _ => }
+      throw new IllegalStateException(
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -240,6 +240,15 @@ class ZipperBams
       }
     }
 
+    // There really should be no more reads, but lets continue on if so
```
nh13, line 243:
```suggestion
    // There really should be no more mapped reads!
```

tfenne (approved):

src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -240,6 +240,15 @@ class ZipperBams
...
+    // There really should be no more mapped reads!
+    if (mappedIter.hasNext) {
+      throw new IllegalStateException(
+        """Error: processed all unmapped reads but there are mapped reads remaining to be read.
+        |Please ensure the unmapped and mapped reads have the same set of read names in the same
+        |order, and reads with the same name are consecutive (grouped) in each input""".stripMargin
+      )
+    }
+
     out.close()
```
tfenne, line 252: It would probably be kind to move `out.close()` above so that the output BAM is at least somewhat valid/intact?

Issue #931, opened 2023-08-28
Add tag to store barcode quality in sam for FastqToBam


galaxy001:
In `FastqToBam`, there is "--umi-qual-tag" for UMI quality. However, there is no option to save barcode quality to tag.
Please add as STAR require quality input at <https://github.com/alexdobin/STAR/blob/master/docs/STARsolo.md#input-reads-from-bam-files>.

nh13:
Seems like a great first issue if you want to help implement the feature and add a unit test or two!

galaxy001:
Checked #932 with `sbt test`, and all tests passed.

nh13:
Done in #933!

PR #932, opened 2023-08-29
New: added Barcode qualities keeping in FastqToBam

galaxy001:
For issue #931.

This is my first Java code.
The unit test is a bit difficult for me. I tried on real data with and without `-b`, the output bam seems fine.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
@@ -92,6 +92,7 @@ class FastqToBam
   @arg(flag='s', doc="If true, queryname sort the BAM file, otherwise preserve input order.")  val sort: Boolean = false,
   @arg(flag='u', doc="Tag in which to store molecular barcodes/UMIs.")                         val umiTag: String = ConsensusTags.UmiBases,
   @arg(flag='q', doc="Tag in which to store molecular barcode/UMI qualities.")                 val umiQualTag: Option[String] = None,
+  @arg(flag='b', doc="Tag in which to store sample barcode qualities.")                        val smpQualTag: Option[String] = None,
   @arg(flag='n', doc="Extract UMI(s) from read names and prepend to UMIs from reads.")         val extractUmisFromReadNames: Boolean = false,
   @arg(          doc="Read group ID to use in the file header.")                               val readGroupId: String = "A",
   @arg(          doc="The name of the sequenced sample.")                                      val sample: String,
@@ -157,6 +158,7 @@ class FastqToBam
     try {
       val subs = fqs.iterator.zip(structures.iterator).flatMap { case(fq, rs) => rs.extract(fq.bases, fq.quals) }.toIndexedSeq
       val sampleBarcode = subs.iterator.filter(_.kind == SampleBarcode).map(_.bases).mkString("-")
+      val smpQual       = subs.iterator.filter(_.kind == SampleBarcode).map(_.quals).mkString(" ")
       val umi           = subs.iterator.filter(_.kind == MolecularBarcode).map(_.bases).mkString("-")
       val umiQual       = subs.iterator.filter(_.kind == MolecularBarcode).map(_.quals).mkString(" ")
       val templates     = subs.iterator.filter(_.kind == Template).toList
@@ -181,6 +183,7 @@ class FastqToBam
         }
 
         if (sampleBarcode.nonEmpty) rec("BC") = sampleBarcode
+        if (smpQual.nonEmpty) smpQualTag.foreach(rec(_) = smpQual)
 
         // Set the UMI on the read depending on whether we got UMIs from the read names, reads or both
         (umi, umiFromReadName) match {


src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -92,6 +92,7 @@ class FastqToBam
   @arg(flag='s', doc="If true, queryname sort the BAM file, otherwise preserve input order.")  val sort: Boolean = false,
   @arg(flag='u', doc="Tag in which to store molecular barcodes/UMIs.")                         val umiTag: String = ConsensusTags.UmiBases,
   @arg(flag='q', doc="Tag in which to store molecular barcode/UMI qualities.")                 val umiQualTag: Option[String] = None,
+  @arg(flag='b', doc="Tag in which to store sample barcode qualities.")                        val smpQualTag: Option[String] = None,
```
nh13, line 95: since we have a short flag, it's not the worst to have a fully qualified long name 

```suggestion
  @arg(flag='b', doc="Tag in which to store sample barcode qualities.")                        val sampleBarcodeQualTag: Option[String] = None,
```

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -157,6 +158,7 @@ class FastqToBam
     try {
       val subs = fqs.iterator.zip(structures.iterator).flatMap { case(fq, rs) => rs.extract(fq.bases, fq.quals) }.toIndexedSeq
       val sampleBarcode = subs.iterator.filter(_.kind == SampleBarcode).map(_.bases).mkString("-")
+      val smpQual       = subs.iterator.filter(_.kind == SampleBarcode).map(_.quals).mkString(" ")
```
nh13, line 161: 1. updated the name as per above
2. Assuming you'll use `QT` as per [the SAM spec](https://samtools.github.io/hts-specs/SAMtags.pdf), in which case we want it hyphen delimited:
> The recommended implementation concatenates all the barcodes and places a hyphen (‘-’) between the barcodes from the same template.
```suggestion
      val sampleBarcodeQualTag = subs.iterator.filter(_.kind == SampleBarcode).map(_.quals).mkString("-")
```

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
nh13: What do you thin about just making this a `boolean` value, and then hard code `QT` below (like we do for `BC`) since thats the recommended tag from [the SAM spec](https://samtools.github.io/hts-specs/SAMtags.pdf)?

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -181,6 +183,7 @@ class FastqToBam
         }
 
         if (sampleBarcode.nonEmpty) rec("BC") = sampleBarcode
+        if (smpQual.nonEmpty) smpQualTag.foreach(rec(_) = smpQual)
```
nh13, line 186: I'd hard code to `QT` as per above

nh13 (changes requested):
@galaxy001: this is fantastic, and you've done a really good job.  

I'd change the command line option to be a boolean (default false) to opt-in to adding the sample barcode base qualities, and then hardcode the tag as `QT` as per the SAM spec (just like `BC`).  Thoughts?

And then once we have this approved, I can add a unit test so don't worry about that.

galaxy001:
Hard code to `QT` is fine since STAR allow setting the tag name, and the spec is here.
Using hyphen (‘-’) between is recommended for `BC`, and for quality values, the spec always recommends to use space as in `QT`, `QX`, `CY` and `BZ`.

galaxy001:
And for `-q`, the spec says `QX`.
I find `-u` is default to `RX` but remains configurable, is this better to follow this manner ?

nh13:
Lets follow the spec as closely as possible, and I misread it!  I also want to keep backwards compatibility at this point.

How about:

- BC: sample barcode bases, hyphen-delimited (keep-hard-coded)
- QT: sample barcode qualities, space delimited (make hard-coded in this PR, boolean command line option to opt-in)
- RX: molecular barcode bases, hyphen-delimited (keep-hard-coded default as a command line option)
- QX: molecular barcode qualities, space-delimited (keep opt-in with as a command line option)

galaxy001:
> Lets follow the spec as closely as possible, and I misread it! I also want to keep backwards compatibility at this point.
> 
> How about:
> 
>     * BC: sample barcode bases, hyphen-delimited (keep-hard-coded)
> 
>     * QT: sample barcode qualities, space delimited (make hard-coded in this PR, boolean command line option to opt-in)
> 
>     * RX: molecular barcode bases, hyphen-delimited (keep-hard-coded default as a command line option)
> 
>     * QX: molecular barcode qualities, space-delimited (keep opt-in with as a command line option)

If we choose to follow the spec recommendation, the `BC` - `QT` pair can be hard-coded, but the `RX` - `QX` pair is not, since there is also `OX` -`BZ` pair for Raw (uncorrected) UMI. The spec in fact stats two sets of UMI tags, thus it is better to make them at least switchable.

I think the `OX` -`BZ` pair is less common just because there is no famous correcting tools available. So considering the future, we should provide them.

So, there are two options:
1. offer the default value and allow user to configure, either to `OX` -`BZ` or to anything else.
2. hard-coded four value, add another option for raw UMI to switch `RX` - `QX` to `OX` -`BZ`.

nh13:
It's probably easier for me to explain in code, how about this: https://github.com/fulcrumgenomics/fgbio/pull/933?

galaxy001:
It is fine to consider barcode only this time. And barcode is preferred to use hard-coded tag name.

I made a comment on L95.

PR #933, opened 2023-08-31
feat: add an option to store sample base qualities in the QT for FastqToBam

nh13:
Alternative to #932 to implement #931

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
@@ -92,6 +92,7 @@ class FastqToBam
   @arg(flag='s', doc="If true, queryname sort the BAM file, otherwise preserve input order.")  val sort: Boolean = false,
   @arg(flag='u', doc="Tag in which to store molecular barcodes/UMIs.")                         val umiTag: String = ConsensusTags.UmiBases,
   @arg(flag='q', doc="Tag in which to store molecular barcode/UMI qualities.")                 val umiQualTag: Option[String] = None,
+  @arg(doc="Store the sample barcode qualities in the QT Tag.")                                val storeSampleBarcodeQualities: Boolean = false,
   @arg(flag='n', doc="Extract UMI(s) from read names and prepend to UMIs from reads.")         val extractUmisFromReadNames: Boolean = false,
   @arg(          doc="Read group ID to use in the file header.")                               val readGroupId: String = "A",
   @arg(          doc="The name of the sequenced sample.")                                      val sample: String,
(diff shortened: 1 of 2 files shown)


src/test/scala/com/fulcrumgenomics/fastq/FastqToBamTest.scala
```diff
@@ -322,6 +322,24 @@ class FastqToBamTest extends UnitSpec {
     recs(3).basesString shouldBe "CCCCCCCCCC"
   }
 
+  it should "extract sample barcode qualities when requested" in {
```
nh13, line 325: @galaxy001, take a look at the inputs and outputs and see if this suffices.

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -92,6 +92,7 @@ class FastqToBam
   @arg(flag='s', doc="If true, queryname sort the BAM file, otherwise preserve input order.")  val sort: Boolean = false,
   @arg(flag='u', doc="Tag in which to store molecular barcodes/UMIs.")                         val umiTag: String = ConsensusTags.UmiBases,
   @arg(flag='q', doc="Tag in which to store molecular barcode/UMI qualities.")                 val umiQualTag: Option[String] = None,
+  @arg(doc="Store the sample barcode qualities in the QT Tag.")                                val storeSampleBarcodeQualities: Boolean = false,
```
galaxy001, line 95: Why there is no `flag` ? What will the option char for enabling `QT` ?

nh13: Use the full command line option `--store-sample-barcode-qualities`.  I don't think there's a flag that makes sense, and likely since folks are implementing this in a workflow software, I don't think it's too onerous.  If you really want one, I'd pick `x` since it's sufficiently different than the others.

tfenne: I'd probably just add `-Q` if it were up to me

src/test/scala/com/fulcrumgenomics/fastq/FastqToBamTest.scala
```diff
@@ -322,6 +322,24 @@ class FastqToBamTest extends UnitSpec {
...
+    val recs = readBamRecs(bam)
+    recs should have size 3
+    recs(0).apply[String]("BC") shouldBe "AAA-AAA"
+    recs(0).apply[String]("QT") shouldBe "ABC GHI"
+    recs(1).apply[String]("BC") shouldBe "TAA-AAA"
+    recs(1).apply[String]("QT") shouldBe "BCD HIJ"
+    recs(2).apply[String]("BC") shouldBe "GAA-TCG"
+    recs(2).apply[String]("QT") shouldBe "CDE IJK"
+  }
+
```
tfenne, line 342: Could you either add a second pass here, or assert in one of the other tests, that if you don't supply a value for `storeSampleBarcodeQualities` that the `QT` tag doesn't get set with anything?

tfenne (approved):
Requested one minor test addition.

Issue #920, opened 2023-07-07
Retaining secondary and supplementary alignments


calcasey:
Hi,

I am working with RNA-seq data to identify gene fusions. I'm currently using fgbio for UMI grouping and consensus sequence calling.

As well as using the tools in the latest fgbio release to process the UMIs, I would like to test an adapted version of GroupReadsByUmi which does not filter out secondary and supplementary alignments, to see how this affects the fusion calling results. 

However, I am new to scala and sbt, and I am unsure how to build an executable JAR file to test the adapted tool.

I have removed this line in `fgbio-2.1.0/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala`:
```
.filter(r => !r.secondary && !r.supplementary)
```
which I am hoping will do the trick. I would be very grateful if you could advise me on how to build an executable JAR file so I can test this out. I noticed `README.md` mentions `sbt assembly`, but do I need to edit `build.sbt` before running this?

Additionally, I would appreciate your opinion on what unintended negative consequences I might run into downstream in my analysis if I retain secondary and/or supplementary alignments, or whether it might be better to retain only one or the other. I am also fairly new to fusion calling, so I am not sure whether the benefits of retaining secondary and/or supplementary alignments for identifying fusion breakpoints outweigh the disadvantages of retaining potential alignment artifacts.

Thank you!

nh13:
See this link for how to build: https://github.com/fulcrumgenomics/fgbio#building.

I'll leave this issue open in the case we have time in the future to answer your additional questions.  My apologies as we are prioritized paid client work at this time.

calcasey:
Hi @bonitalam, sure, here is a summary of what I did next:

- I commented out the line that filters out secondary and supplementary reads in GroupReadsByUmi, built the executable JAR file, and ran the edited GroupReadsByUmi 
- However, I ran into this error (I substituted the name of the read for XXXX):
 
```
[2023/07/11 14:02:07 | FgBioMain | Fatal] #################################################################################################################################
[2023/07/11 14:02:07 | FgBioMain | Fatal] Execution failed!
[2023/07/11 14:02:07 | FgBioMain | Fatal] Mate mapping quality (MQ) tag not present on read XXXX.
[2023/07/11 14:02:07 | FgBioMain | Fatal] #################################################################################################################################
[2023/07/11 14:02:07 | FgBioMain | Info] GroupReadsByUmi failed. Elapsed time: 0.05 minutes.
```
[...]

PR #934, opened 2023-09-01
Mark Duplicate Bitflag & include secondary & Supplimentary Reads

### LINKED ISSUE #906: Use GroupReadsByUmi for standard duplicate marking?

JoeVieira:
Closes #906 & #920

This exposes a few new options

`-d` Set duplicate bit flag on all non-representative reads
`-D` representative read strategy, accepts strategies from [htslib](https://github.com/samtools/htsjdk/blob/6d3fc7bc1f613ecfce1c22d368f3ae17cb86823d/src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java#L36-L39). 
Representative read is chosen by the sum scores ( from the scoring strategy ) across all primary reads, ties are broken with sum of mapping qualities from primary reads.

`-S` include secondary reads in output ( also mark duplicates on as needed ) 
`-U` include supplementary reads in output ( mark as needed )

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -482,6 +482,8 @@ class GroupReadsByUmi
  @arg(flag='d', doc="Mark Duplicate Flag will be set on primary read.")     val markDup: Boolean = false,
  @arg(flag='D', doc="If -d is set, primary read assignment strategy, default = Sum Of Base Qualities.")
   val dupStrategy: DuplicateScoringStrategy.ScoringStrategy = DuplicateScoringStrategy.ScoringStrategy.SUM_OF_BASE_QUALITIES,
+ @arg(flag='S', doc="Include secondary reads.")         val includeSecondary: Boolean = false,
+ @arg(flag='U', doc="Include supplementary reads.")     val includeSupplementary: Boolean = false,
  @arg(flag='m', doc="Minimum mapping quality for mapped reads.")         val minMapQ: Int      = 1,
  @arg(flag='n', doc="Include non-PF reads.")            val includeNonPfReads: Boolean = false,
  @arg(flag='s', doc="The UMI assignment strategy.")     val strategy: Strategy,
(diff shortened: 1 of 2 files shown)


JoeVieira:
@tfenne @nh13 feedback welcome.

tfenne:
My apologies @JoeVieira - this is definitely on my list to look at, I've just been swamped recently.  In all honestly it'll probably be another week or so until I have the time to do a thorough review.

JoeVieira:
@tfenne thanks for the response - no worries, appreciate the transparency!

PR #936, opened 2023-09-09
adds a barcode option to FastqToBam

bwlang:


src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -96,8 +96,9 @@ class FastqToBam
   @arg(          doc="Read group ID to use in the file header.")                               val readGroupId: String = "A",
   @arg(          doc="The name of the sequenced sample.")                                      val sample: String,
   @arg(          doc="The name/ID of the sequenced library.")                                  val library: String,
+  @arg(          doc="Library or Sample barcode sequence.")                                    val barcode: Option[String] = None,
   @arg(          doc="Sequencing Platform.")                                                   val platform: String = "illumina",
-  @arg(doc="Platform unit (e.g. '<flowcell-barcode>.<lane>.<sample-barcode>')")                val platformUnit: Option[String] = None,
+  @arg(doc="Platform unit (e.g. '<flowcell-barcode>.<lane>')")                                 val platformUnit: Option[String] = None,
```
nh13, line 101: I don’t like this change. It won’t be a unique identifier as required by the spec of we merge samples pooled across lanes or flowcells for example. There’s no text that forbids sample barcode as part of this unique identifier either. 

see:
IMG_5792

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -135,6 +136,7 @@ class FastqToBam
     val rg = new SAMReadGroupRecord(this.readGroupId)
     rg.setSample(sample)
     rg.setLibrary(library)
+    this.barcode.foreach(bc => rg.setBarcodes(util.Arrays.asList(bc)))
```
nh13, line 139: I like this change!  Thank you!

nh13 (changes requested):
Let’s leave out the change for platform unit as I don’t think it’s wrong to have the sample barcode as part of the platform unit, and in some cases is useful (we did this at my time at Broad, often)

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
bwlang: I found [a post](https://gatk.broadinstitute.org/hc/en-us/articles/360035890671-Read-groups) re sample-barcode  in GATK. Seems strange to me - since "platform unit" sound to me like a physical thing. reverting the doc change.

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -96,6 +96,7 @@ class FastqToBam
   @arg(          doc="Read group ID to use in the file header.")                               val readGroupId: String = "A",
   @arg(          doc="The name of the sequenced sample.")                                      val sample: String,
   @arg(          doc="The name/ID of the sequenced library.")                                  val library: String,
+  @arg(          doc="Library or Sample barcode sequence.")                                    val barcode: Option[String] = None,
```
nh13, line 99: Thinking about this further, how would you feel about making this a `Seq[String]` with default `Seq.empty`, then joining them with a hyphen?  That way folks don't have to do so themselves?
```suggestion
  @arg(          doc="Library or Sample barcode sequence.", minElements=0)                                    val barcode: Seq[String] = Seq.empty,
```

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
nh13, line 139: not in an idea, but this could work:
```
    this.barcode.foreach(bc => rg.setBarcodes(util.Arrays.asList(bc)))
    if (this.barcode.nonEmpty) rg.setBarcodes(...)
```

nh13 (changes requested):
One more question, thanks for your patience!

bwlang:
I tried this
`if (this.barcode.nonEmpty) rg.setBarcodes(util.Arrays.asList(this.barcode.mkString("+")))`
testing like this:
`barcode=Seq("TATA", "GAGA"),`
which works.

but i was surprised to see that ... 
`if (this.barcode.nonEmpty) rg.setBarcodes(util.Arrays.asList(this.barcode.mkString("-")))`
gets split by htsjdk's getBarcodes on "-"( i guess???) , producing  a list of 2 barcodes ["TATA", "GAGA"]

I've done like you asked but it's a bit too much magic for my taste...

nh13:
@bwlang want to revert to the pre-`Seq` changes then I'll run the CI and merge if it passes?

bwlang:
I should say. it doesn't really matter which way we pass the barcodes in. 
I didn't appreciate that until testing with 'TATA-GAGA'... so whichever way you think is clearer.

nh13 (approved):

nh13:
Thank-you for your patience @bwlang !  Making sure unit tests pass here, then I'll merge.

Issue #878, opened 2022-10-09
feature request: switch order of operations for ClipBam


eboyden:
ClipBam `--clip-overlapping-reads` can occasionally orphan one read in a pair if it is much longer than its mate and extends past the mate's start, because equal trimming causes the mate to be trimmed to length 0.  The new function `--clip-bases-past-mate` solves this problem when ClipBam is run twice sequentially, which is great but requires 2X sorting in-between (because currently MD/NM/UQ tag repair is required).  Both functions can be called simultaneously, but testing shows that the order of operations is `--clip-overlapping-reads` before `--clip-bases-past-mate` (or possibly `--clip-overlapping-reads` supercedes `--clip-bases-past-mate` since the former renders the latter superfluous).  It would be better if `--clip-bases-past-mate` occurred first, which would ensure that paired reads (virtually) never become orphaned regardless of their respective lengths.

nh13:
I am not opposed to allowing the user to specify the order of functions and having the current be the default.

> because currently MD/NM/UQ tag repair is required).
What if you only repaired on the second pass?

eboyden:
That gets to my question in https://github.com/fulcrumgenomics/fgbio/issues/877 - I wasn't sure if tag repair was required for subsequent steps to behave properly.  If not, then making tag repair optional in all tools would certainly be beneficial, since then you can just do it once at the end without a lot of extraneous upstream sorting.  In this case, I don't know if simply running ClipBam twice in succession would be any slower than calling both functions (in the preferred order) in a single ClipBam instance.  Being able to call both functions simultaneously would be a bit simpler though.

If at this point you think it's too much effort for too little benefit to make the function order customizable (or just switching the order, since I can't think of any case where upstream `--clip-overlapping-reads` doesn't make downstream `--clip-bases-past-mate` irrelevant, so I can't imagine anyone wanting to run both functions in that order), then I would suggest at least making the function order clear in the documentation.

nh13:
I think all of those are good suggestions, I’ll again be transparent (and it’s totally fine to ask questions/suggestions) that supporting fgbio is on a volunteer basis and right now personally I’m prioritizing client work.

JoeVieira:
@nh13 - I'll take this feature.

clintval:
This should be closed now with:

- https://github.com/fulcrumgenomics/fgbio/pull/1026

We are now using the record **_and the record's mate_** directly instead of relying on the `MC` tag.

[...]

PR #937, opened 2023-09-20
Bug Fix - SamRecordClipper.numBasesExtendingPastMate

JoeVieira:
@nh13 @tfenne 

When operating on SamRecords with multiple serialized operations, use of mate cigar record information is relied upon for clipping past mate end, but it is not synchronized between operations with it's corresponding SamRecord object's data.

This results in an inconsistent data model for clipping.

With previous code this test case would fail.

```
it should "clip FR reads that extend past mate in asymmetrical read" in {
    val builder = new SamBuilder(readLength = 200, sort = Some(Queryname))
    val clipper = new ClipBam(input = dummyBam, output = dummyBam, ref = ref, clipBasesPastMate = true,
      readTwoFivePrime = 50, readOneFivePrime = 10)
    val Seq(r1, r2) = builder.addPair(start1 = 100, start2 = 90)

    r1.end shouldBe 299
    r2.end shouldBe 289
    clipper.clipPair(r1, r2)
    r2.end shouldBe r2.end
    r1.start shouldBe r2.start
  }
```

It would erroneously calculate r2.start == 100 & r1.start != r2.start, because the mate.start data from the MC tag was used, which was not updated with the 5' clipping from the first operation. 

https://github.com/fulcrumgenomics/fgbio/blob/2af51acea8cd55fbc393ce435b5b13d7a32fc9ae/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala#L358

This reliance on a possibly dirty MC tag is the cause of this #878 inconsistent behavior. I believe this fix, makes this issue irrelevant & it certainly might fix some other oddities that people seem to have pointed out.

I've removed convenience functions which allow this to happen, and enforce either explicit passing of mates or start / end. 

I do this rather than updating the MC record, because that extra step between each operation doesn't seem worthwhile, when we have the object loaded in memory already, the tags should be updated after all operations are complete a single time.

The exception to this is a single method which is used in consensus calling, where mate isn't easily available - this pattern still requires getting from mate cigar, which might still result in this bug occurring, as each mate could have clipping applied but again the data isn't synchronized to the MC tag.

I'll also work on that bug also, which I believe should be handled by operating on mates together, so all relevant data is loaded in memory explicitly, rather than using the tag.

As a sidenote: I don't understand the white space formatting rules for the project, do you happen to have an intellj profile i could use to keep these consistent for your formatting?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
@@ -86,7 +86,6 @@ class ClipBam
   @arg(          doc="Require at least this number of bases to be clipped on the 3' end of R2") val readTwoThreePrime: Int = 0,
   @arg(          doc="Clip overlapping reads.") val clipOverlappingReads: Boolean = false,
   @arg(          doc="Clip reads in FR pairs that sequence past the far end of their mate.") val clipBasesPastMate: Boolean = false,
-  @arg(          doc="Clip bases past mate before clipping overlapping reads") val clipPastFirst: Boolean = false,
   @arg(flag='S', doc="The sort order of the output. If not given, output will be in the same order as input if the input.")
   val sortOrder: Option[SamOrder] = None
 ) extends FgBioTool with LazyLogging {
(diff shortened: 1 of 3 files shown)


eboyden:
Bumping this.  As it pertains to a stealthy bug that can produce incorrect output, rather than a new feature, I hope it can be prioritized.

clintval:
Most of this was solved in this recently merged PR (and release [v2.5.0](https://github.com/fulcrumgenomics/fgbio/releases/tag/2.5.0) now exists):

- https://github.com/fulcrumgenomics/fgbio/pull/1026

I still have a lingering issue with how hard-clipping is handled:

I have another PR up to be more careful with hard clipping when it should be ignored:

- https://github.com/fulcrumgenomics/fgbio/pull/1065

Once that is merged, the spirit of the tests you outlined in this PR will pass (most already do now on `main`).

PR #938, opened 2023-09-25
Fix link in DemuxFastqs.scala

PeteHaitch:
Fix markdown syntax


nh13 (approved):
Thank-you @PeteHaitch, you may consider looking at https://github.com/fulcrumgenomics/fqtk as a faster alternative that we've built!

PeteHaitch:
I did take a look at `fqtk` but conda was failing to resolve the dependencies on the machine I'm using. 
I'll take another look when I'm not under such time constraints to get these particular samples demultiplexed.

nh13:
@PeteHaitch happy to help when you have time. DemuxFastqs was built a long time ago for a lot less data!

PR #940, opened 2023-09-28
Alternative PR to add duplicate marking to GroupReadsByUmi

tfenne:
@JoeVieira and @nh13 this is my attempt to provide a cleaned up version of #934.  I figured this was easier than trying to shepherd the original PR through with lots of comments - though I greatly appreciate you making that PR Joe as it showed me how small the changes actually are and motivated me to do this.

Some of the changes stylistic / scala idiom stuff.  Functional changes include:

- Auto-defaulting of minMapQ and including of secondary/supplementary based on whether `--mark-duplicates` is provided
- I chose not to expose the duplicate ScoringStrategy as I'm not sure I want to tie ourselves to that, and just picked one to use for now
- It now also reset the duplicate flag to false on the chosen representative template's reads

Edited to add: Oh, and also changing the code downstream to avoid the consensus callers seeing secondary/supplementary reads should a BAM generated with `--mark-duplicates` get sent downstream.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -482,6 +482,8 @@ class GroupReadsByUmi
  @arg(flag='d', doc="Mark Duplicate Flag will be set on primary read.")     val markDup: Boolean = false,
  @arg(flag='D', doc="If -d is set, primary read assignment strategy, default = Sum Of Base Qualities.")
   val dupStrategy: DuplicateScoringStrategy.ScoringStrategy = DuplicateScoringStrategy.ScoringStrategy.SUM_OF_BASE_QUALITIES,
+ @arg(flag='S', doc="Include secondary reads.")         val includeSecondary: Boolean = false,
+ @arg(flag='U', doc="Include supplementary reads.")     val includeSupplementary: Boolean = false,
  @arg(flag='m', doc="Minimum mapping quality for mapped reads.")         val minMapQ: Int      = 1,
  @arg(flag='n', doc="Include non-PF reads.")            val includeNonPfReads: Boolean = false,
  @arg(flag='s', doc="The UMI assignment strategy.")     val strategy: Strategy,
@@ -548,7 +550,8 @@ class GroupReadsByUmi
     // Filter and sort the input BAM file
     logger.info("Filtering the input.")
     val filteredIterator = in.iterator
-      .filter(r => !r.secondary && !r.supplementary)
+      .filter(r => includeSecondary || !r.secondary  )
+      .filter(r => includeSupplementary || !r.supplementary )
       .filter(r => (includeNonPfReads || r.pf)                                      || { filteredNonPf += 1; false })
       .filter(r => (r.mapped || (r.paired && r.mateMapped))                         || { filteredPoorAlignment += 1; false })
       .filter(r => (allowInterContig || r.unpaired || r.refIndex == r.mateRefIndex) || { filteredPoorAlignment += 1; false })
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -752,14 +752,32 @@ class GroupReadsByUmi
             }
         }
     }
-    group.sortBy { t =>
-      calculateScore(t.primaryReads)
+    val deDupGroup = group.sortWith{ (t1, t2) =>
+      compareScore(t1.primaryReads, t2.primaryReads)
     }
-    group.slice(1,group.length).foreach { t => t.allReads.foreach( r => r.duplicate = true) }
+    deDupGroup.slice(1,deDupGroup.length).foreach { t => t.allReads.foreach( r => r.duplicate = true) }
   }
 
-  private def calculateScore(read: Iterator[SamRecord]): Int = {
-    read.foldLeft(0)((acc, obj) => acc + DuplicateScoringStrategy.computeDuplicateScore(obj.asSam, this.dupStrategy, true))
+  private def compareScore(priRead1: Iterator[SamRecord], priRead2: Iterator[SamRecord]): Boolean = {
+    var score1 = 0
+    var score2 = 0
+    var r1MapQ = 0
+    var r2MapQ = 0
+    priRead1.foreach{ r =>
+      score1 = score1 + DuplicateScoringStrategy.computeDuplicateScore(r.asSam, this.dupStrategy, true)
+      r1MapQ = r1MapQ + r.mapq
+    }
+    priRead2.foreach { r =>
+      score2 = score2 + DuplicateScoringStrategy.computeDuplicateScore(r.asSam, this.dupStrategy, true)
+      r2MapQ = r2MapQ + r.mapq
+    }
+    if (score1 == score2) {
+      // tie break on mapQ
+      r1MapQ > r2MapQ
+    }
+    else {
+      score1 > score2
+    }
   }
 
 }
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -439,8 +439,9 @@ object Strategy extends FgBioEnum[Strategy] {
     |   4. Read Name
     |
     |During grouping, reads are filtered out if a) all reads with the same queryname are unmapped, b) any primary
-    |read has mapping quality < `min-map-q` (default=1), or c) the primary mappings for R1 and R2 are on different
-    |chromosomes and `--allow-inter-contig` has been set to false.
+    |read has mapping quality < `min-map-q` (default=1), c) the primary mappings for R1 and R2 are on different
+    |chromosomes and `--allow-inter-contig` has been set to false., or d.) all non-primary reads are filtered,
+    |if `--includeSecondary` and\or `--includeSupplementary` are set to false (default=false)
     |
     |Grouping of UMIs is performed by one of four strategies:
     |
@@ -469,8 +470,21 @@ object Strategy extends FgBioEnum[Strategy] {
     |(i.e. does not count dashes and other non-ACGT characters). This option is not implemented for reads with UMI pairs
     |(i.e. using the paired assigner).
     |
+    |If mark duplicates flag is set, will set duplicate bit flag on all non-representative reads inside of a UMI group
+    |which are output.
+    |
+    |Strategies for choosing representative read is independent of UMI grouping strategy.
+    |
+    |Choosing representative read is preformed by one of three strategies:
+    |
+    |In each strategy, ties are broken by mapping quality
+    |
+    |1. **SUM_OF_BASE_QUALITIES**:  Sum base quality across entire read, the read with the highest combined base quality is chosen.
+    |2. **TOTAL_MAPPED_REFERENCE_LENGTH**: Sum mapped reference across each paired read ( if paired , the longest read is chosen.
+    |3. **RANDOM**: Choose a representative read at random.
+    |
     |If the input is not template-coordinate sorted (i.e. `SO:unsorted GO:query SS:unsorted:template-coordinate`), then
-    |this tool will re-sort the input. The ouitput will be written in template-coordinate order.
+    |this tool will re-sort the input. The output will be written in template-coordinate order.
   """
 )
 class GroupReadsByUmi
@@ -479,8 +493,8 @@ class GroupReadsByUmi
  @arg(flag='f', doc="Optional output of tag family size counts.") val familySizeHistogram: Option[FilePath] = None,
  @arg(flag='t', doc="The tag containing the raw UMI.")  val rawTag: String    = "RX",
  @arg(flag='T', doc="The output tag for UMI grouping.") val assignTag: String = "MI",
- @arg(flag='d', doc="Mark Duplicate Flag will be set on primary read.")     val markDup: Boolean = false,
- @arg(flag='D', doc="If -d is set, primary read assignment strategy, default = Sum Of Base Qualities.")
+ @arg(flag='d', doc="Mark duplicates, duplicate bitflag will be set on non-representative reads.")     val markDup: Boolean = false,
+ @arg(flag='D', doc="If -d is set, representative read assignment strategy, default = Sum Of Base Qualities.")
   val dupStrategy: DuplicateScoringStrategy.ScoringStrategy = DuplicateScoringStrategy.ScoringStrategy.SUM_OF_BASE_QUALITIES,
  @arg(flag='S', doc="Include secondary reads.")         val includeSecondary: Boolean = false,
  @arg(flag='U', doc="Include supplementary reads.")     val includeSupplementary: Boolean = false,
@@ -498,6 +512,7 @@ class GroupReadsByUmi
   @deprecated val allowInterContig: Boolean = true
 )extends FgBioTool with LazyLogging {
   import GroupReadsByUmi._
+
 
   require(this.minUmiLength.forall(_ => this.strategy != Strategy.Paired), "Paired strategy cannot be used with --min-umi-length")
 
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -733,4 +733,28 @@ class GroupReadsByUmi
         r1[String](this.rawTag)
     }
   }
+
+  private def assignRepReadToGroup(group: Seq[Template]): Unit = {
+    //check all primary reads have SAME UMI defined
+    group.headOption match {
+      case Some(firstObject) =>
+        val groupUmi = firstObject.r1.get[String](this.assignTag)
+        group.foreach {
+          t =>
+            t.primaryReads.foreach { r =>
+              if (r.get[String](this.assignTag).getOrElse("") != groupUmi)
+                fail(s"Record '$r' has mismatched UMI from group '${groupUmi}'")
+            }
+        }
+    }
+    group.sortBy { t =>
+      calculateScore(t.primaryReads)
+    }
+    group.slice(1,group.length).foreach { t => t.allReads.foreach( r => r.duplicate = true) }
+  }
+
+  private def calculateScore(read: Iterator[SamRecord]): Int = {
+    read.foldLeft(0)((acc, obj) => acc + DuplicateScoringStrategy.computeDuplicateScore(obj.asSam, this.dupStrategy, true))
+  }
+
 }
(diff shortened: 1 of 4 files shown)


tfenne:
Tests could still use some reformatting to follow convention of other code.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -439,9 +439,13 @@ object Strategy extends FgBioEnum[Strategy] {
...
     |
+    |If the input is not template-coordinate sorted (i.e. `SO:unsorted GO:query SS:unsorted:template-coordinate`), then
+    |this tool will re-sort the input. The output will be written in template-coordinate order.
+    |
     |During grouping, reads are filtered out if a) all reads with the same queryname are unmapped, b) any primary
-    |read has mapping quality < `min-map-q` (default=1), or c) the primary mappings for R1 and R2 are on different
-    |chromosomes and `--allow-inter-contig` has been set to false.
+    |read has mapping quality < `min-map-q` (default=1), c) the primary mappings for R1 and R2 are on different
+    |chromosomes and `--allow-inter-contig` has been set to false., or d.) all non-primary reads are filtered,
+    |if `--includeSecondary` and\or `--includeSupplementary` are set to false (default=false)
```
nh13, line 448:
```suggestion
    |if `--include-secondary` and\or `--include-supplementary` are set to false (default=false)
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -439,9 +439,13 @@ object Strategy extends FgBioEnum[Strategy] {
...
     |   4. Read Name
     |
+    |If the input is not template-coordinate sorted (i.e. `SO:unsorted GO:query SS:unsorted:template-coordinate`), then
+    |this tool will re-sort the input. The output will be written in template-coordinate order.
+    |
     |During grouping, reads are filtered out if a) all reads with the same queryname are unmapped, b) any primary
-    |read has mapping quality < `min-map-q` (default=1), or c) the primary mappings for R1 and R2 are on different
-    |chromosomes and `--allow-inter-contig` has been set to false.
+    |read has mapping quality < `min-map-q` (default=1), c) the primary mappings for R1 and R2 are on different
+    |chromosomes and `--allow-inter-contig` has been set to false., or d.) all non-primary reads are filtered,
```
nh13, line 447:
```suggestion
    |chromosomes and `--allow-inter-contig` has been set to false, or d.) all non-primary reads are filtered,
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: I don't grok 446-447 after multiple attempts at reading the usage (before looking at the code).  Perhaps you don't mean "non-primary"?  And if not, then I'm really confused!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -470,27 +474,38 @@ object Strategy extends FgBioEnum[Strategy] {
...
-                       |""")
-  val minUmiLength: Option[Int] = None,
-  @arg(flag='x', doc= """
-                       |DEPRECATED: this option will be removed in future versions and inter-contig reads will be
-                       |automatically processed.""")
+(@arg(flag='i', doc="The input BAM file.")              val input: PathToBam  = Io.StdIn,
+ @arg(flag='o', doc="The output BAM file.")             val output: PathToBam = Io.StdOut,
+ @arg(flag='f', doc="Optional output of tag family size counts.") val familySizeHistogram: Option[FilePath] = None,
+ @arg(flag='t', doc="The tag containing the raw UMI.")  val rawTag: String    = "RX",
+ @arg(flag='T', doc="The output tag for UMI grouping.") val assignTag: String = "MI",
```
nh13, line 494: Shall we finally update these to use https://github.com/fulcrumgenomics/fgbio/blob/d748c76f06c6b1af478e87a032b81f7732f234ba/src/main/scala/com/fulcrumgenomics/umi/ConsensusTags.scala#L46 and https://github.com/fulcrumgenomics/fgbio/blob/d748c76f06c6b1af478e87a032b81f7732f234ba/src/main/scala/com/fulcrumgenomics/umi/ConsensusTags.scala#L54?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -731,4 +756,18 @@ class GroupReadsByUmi
         r1[String](this.rawTag)
     }
   }
+
+  /** Sets the duplicate flags on all reads within all templates.  */
+  private def setDuplicateFlags(group: Seq[Template]): Unit = {
+    val nonDuplicateTemplate = group.minBy { template =>
```
nh13, line 762: Do we want `maxBy` here?  To my eye, htsjdk's duplicate score returns higher values as better:

https://github.com/samtools/htsjdk/blob/38dbd0e6b87a2bd5791bb81399f2b5ec1008773e/src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java#L73-L116

For example, if the duplicate score is the sum of base qualities, then don't we want the template with the highest sum of base qualities?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -731,4 +756,18 @@ class GroupReadsByUmi
...
+      template.primaryReads.sumBy { r =>
+        DuplicateScoringStrategy.computeDuplicateScore(r.asSam, ScoringStrategy.SUM_OF_BASE_QUALITIES)
+      }
+    }
+
+    group.foreach { template =>
+      val flag = !(template eq nonDuplicateTemplate)
```
nh13, line 769:
```suggestion
      val flag = template ne nonDuplicateTemplate
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -731,4 +756,18 @@ class GroupReadsByUmi
         r1[String](this.rawTag)
     }
   }
+
+  /** Sets the duplicate flags on all reads within all templates.  */
+  private def setDuplicateFlags(group: Seq[Template]): Unit = {
+    val nonDuplicateTemplate = group.minBy { template =>
+      template.primaryReads.sumBy { r =>
+        DuplicateScoringStrategy.computeDuplicateScore(r.asSam, ScoringStrategy.SUM_OF_BASE_QUALITIES)
```
nh13, line 764: As a compromise, consider setting the ScoringStrategy as a constructor parameter that's not exposed on the command line?

nh13 (changes requested):
A few things I'm confused about, but looking really good!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Good catch!

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I really don't want to expose this anywhere - I want the freedom to change this if/when we decide using HTSJDK's duplicate scoring stuff was a bad idea in the future, without breaking any APIs.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -499,6 +519,10 @@ class GroupReadsByUmi
 
   private val assigner = strategy.newStrategy(this.edits)
 
+  // Give values to unset parameters that are different in duplicate marking mode
```
JoeVieira, line 522: Coupling these seems odd to me & deviates from the intended change set.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -470,27 +479,38 @@ object Strategy extends FgBioEnum[Strategy] {
...
-    |If the input is not template-coordinate sorted (i.e. `SO:unsorted GO:query SS:unsorted:template-coordinate`), then
-    |this tool will re-sort the input. The ouitput will be written in template-coordinate order.
+    |If the `--mark-duplicates` option is given, reads will also have their duplicate flag set in the BAM file.
+    |Each tag-family is treated separately, and a single template within the tag family is chosen to be the "unique"
+    |template and marked as non-duplicate, while all other templates in the tag family are then marked as duplicate.
+    |
+    |Several parameters have different defaults depending on whether duplicates are being marked or not (all are
+    |directly settable on the command line):
+    |
+    |  1. `--min-map-q` defaults to 0 in duplicate marking mode and 1 otherwise
```
JoeVieira, line 489: Why?

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
JoeVieira: A few things in here - 
1.) hard-coding this seems like a really odd choice, we could wrap it if you're concerned about leaking API.
2.) you don't have a tie break, which is why mapQ was included in my implementation. 

this is more tidy logic, nice.

JoeVieira:
@tfenne appreciate you tackling it. 

I certainly would have preferred feedback rather than this, but i'm happy it's getting attention - this introduces some oddities that were not present in my change set, while a couple of suggested edits on my PR would have gotten your point of using groupby instead of sorting across just fine.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
JoeVieira: Just as a point re duplicate scoring strategies - the core strategies for choosing the representative read, much like UMI clustering algorithms are fairly established in literature & while having others seems possible, not supporting established ones, seems like a limitation which would be self imposed, I'm happy to wrap these enums if that's a need from the projects perspective so that the HTSJDK api isn't leaking.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: Can you provide a bit more context of why it seems odd?  My intuition is that if you're marking duplicates with this tool, you want to keep all reads (i.e. don't pre-filter based on mapq).  Similarly, you want to set the duplicate flag on secondary and supplementary reads (and don't prefilter them) too.  What am I missing?

Also, what's the "intended change set"?

JoeVieira: Given that the tool allows for multiple versions of filtering, allowing flexibility to choose what one is filtering & not seems like it's giving appropriate control over the functionality, rather than forcing a workflow. 

It's true that, if one is keeping all reads, one must update all reads, but forcing all reads to stay, when marking doesn't seem like a requirement for that.

nh13: I don't think any option is "forced", it just has a default when the option is unset.  You can always set the option and this will override the default.  For example, you can explicitly set `--include-suppementary=true` or `--include-suppementary=false`.  I think that's why the types on the command line parameters are `Option[Boolean]` with default `None`:

1. `None` -> use a default
2. `Some(true)` -> include supplementary
3. `Some(false)` -> filter out supplementary

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -470,27 +479,38 @@ object Strategy extends FgBioEnum[Strategy] {
...
+    |  2. `--include-secondary` defaults to true in duplicate marking mode and false otherwise
+    |  3. `--include-suppementary` defaults to true in duplicate marking mode and false otherwise
```
nh13, line 491:
```suggestion
    |  3. `--include-supplementary` defaults to true in duplicate marking mode and false otherwise
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
JoeVieira: Fair. I'm being overly dramatic with "forced" & you're right, it can be overidden, you're the maintainers so the paradigm the tool uses is ultimately up to you. 

I just strikes me as odd to change defaults when semi related parameters are passed - they aren't strictly coupled, parameters, if they were strictly coupled i'd agree - i think this makes for a usage that implies a specific workflow & causes users to have to more closely read the docs if their workflow isn't that one. 

As a user, when defaults change because of other parameters it can result in unexpected results - which are often hard to debug.

nh13: Fair, the alternative is to provide a second command line tool that performs the duplicate marking that is just a facade over group reads by umi, but that seems a bit heavy handed.  I'll let @tfenne take this discussion over, as I don't have strong feelings either way, but would lean to the current implementation as it is well-documented in the usage.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: Lol, this did make me laugh a little.  I would agree if deep thought had gone into the original duplicate scoring strategies ... but as the author of that original code (in Picard that moved into HTSJDK) I can tell you that it was me making it up on the fly and coming up with something vaguely sensible!

Over time I tend to think that the duplicate marking stuff in Picard has gotten _far_ too complicated.

I'd also argue that there's little to no value tie-breaking on mapq.  If the reads are considered duplicates of each other and have either a) the same sum of base qualities > Q15, or b) the same mapped reference length, there's a very high probability they also have the same sum of mapqs.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I made this change because otherwise, if one wants to use `GroupReadsByUmi` as an equivalent to `MarkDuplicates` in picard, one would have had to add:

`--mark-duplicates --min-mapq 0 --include-secondary --include-supplementary`

which to me seems error prone.  I don't see the harm in defaulting arguments one way when the intended usage is for consensus calling and another when the intended usage is for duplicate marking.  The user is always free to set them explicitly.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
JoeVieira: I don't necessarily mean the algorithm was fully flushed out, i mean it's usage is well established in literature. 
Regardless - I'd be just as happy to implement a version of UMItools which is just as or more vetted. https://umi-tools.readthedocs.io/en/latest/reference/dedup.html#selecting-the-representative-read

I'd tend to argue that baseQ isn't a very important metric at this stage anyway mapping quality is far more important, since it has already taken the base q into consideration when being aligned. 

Agree most of picards stuff is too complex - which is why i did the simplest thing i could think of.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -731,4 +761,18 @@ class GroupReadsByUmi
         r1[String](this.rawTag)
     }
   }
+
+  /** Sets the duplicate flags on all reads within all templates.  */
+  private def setDuplicateFlags(group: Seq[Template]): Unit = {
+    val nonDuplicateTemplate = group.maxBy { template =>
+      template.primaryReads.sumBy { r =>
+        DuplicateScoringStrategy.computeDuplicateScore(r.asSam, ScoringStrategy.SUM_OF_BASE_QUALITIES)
```
JoeVieira, line 769:
```suggestion
        r.mapq
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
tfenne: I'm happy to use something different when picking the representative, though would also like in this first version to just have one way of doing it, and not a user option.

My intuition though is that if you used mapq (summed across primary reads?), then in the vast majority of cases it just becomes a random selection.  The most prevelant short read aligners (bwa, star, minimap2) don't use base quality in computing mapping quality, so reads that are duplicates of one another are likely to have identical mapping qualities _except_ in regions that are hard to map, where mapping a few more or fewer bases is likely to change the confidence.

How about mapq + total number of mapped query bases?  Or something along those lines?

tfenne: Alternatively @JoeVieira I'd be happy to be proved wrong - if you want to take a look at some real data and report how frequently there are differing maqps within a tag family.

JoeVieira: One options seems fine. I think it's perfectly fine that that is sum of base q, nothing about picards implementation of these scoring strategies is complex, the usages deep in the code get to be rather complex, but this simple score calculator just seems to be good code reuse to me. 

I'm happy to look at how map q changes of note my testing is with bowtie2 mostly - so that certainly is impacting my perspective here.

I'll stop pushing here & just test a few more things, give some perspective with data.

tfenne: Thanks @JoeVieira ... please let me know when you've had a chance to look at your data.

JoeVieira: @tfenne it'll be a little bit before I'm able to dig back into that data. But will in the next week or so.

msto: I think it would be sensible to choose the read with highest mapq and breaking ties randomly.

@JoeVieira did you have any suggestions after reviewing some data?

JoeVieira: Sorry all, I have no been able to review data at scale yet. It's on my list for next week i promise.

nh13 (approved):
LGTM.  Until we have a second duplicate scoring strategy that is not random, I agree that setting to sum of base qualities from htsjdk is fine.  It is also a proxy for aligned length, as well as read length, which can vary even when the outer coordinates are the same.

Aside: I would agree that "by mapping quality" is effectively random, since read pairs that are duplicates of each other are almost surely aligned the same by virtue of being duplicates of each other and therefore having the same mapq by current aligners.

nh13 (approved):
LGTM.  I did add a commit (I hope that's ok) to make it work with codecov (which now require s a token, even for public repos)

PR #954, opened 2023-12-30
Allow the PairedAssigner to use UMI pairs where one is absent

clintval:


nh13:
Instead of empty UMIs, why not just use Ns?

clintval:
@nh13 because an IUPAC N stands for either A, C, G, T and we may have none of those in the UMI for one end of the molecule. It will also affect metrics about the UMIs (see [`DuplexUmiMetric`](https://github.com/fulcrumgenomics/fgbio/blob/9d5a42224ff98bef05b873cf37ce64249a97e113/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala#L182-L213)). This PR seems relatively harmless to consider and the output data will be more accurate to the true biochemical situation.

clintval:
But the real show-stopper is the logic scattered about that removes or ignores reads/templates with Ns in the UMI:

https://github.com/fulcrumgenomics/fgbio/blob/9d5a42224ff98bef05b873cf37ce64249a97e113/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala#L516

https://github.com/fulcrumgenomics/fgbio/blob/9d5a42224ff98bef05b873cf37ce64249a97e113/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala#L654

Instead of regressing on how we choose to filter/handle UMIs, this PR seemed like a compelling alternative. (The other idea is to use a non-N IUPAC code placeholder UMI, but that would be disingenuous and potentially confusing).

nh13:
What do you meant by “none of those”? What other base would the missing UMI be if you could have observed it?  Could you update the description to have a better explanation why the UMI is omitted, as clearly I’m missing something?

nh13:
Why not an opt-in option to allow Ns?

clintval:
> Why not an opt-in option to allow Ns?

Because I favored the approach where the data now most closely represents the actual biochemical situation (see first comment here https://github.com/fulcrumgenomics/fgbio/pull/954#issuecomment-1872616872).

If you want an opt-in option to allow Ns, let me know and I can send up a new PR. I personally favor this approach because now all metrics I could possible generate on the UMIs are going to be more sensible for this type of biochemical situation.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -530,7 +531,7 @@ class CollectDuplexSeqMetrics
     val uniqueTotal      = metrics.map(_.unique_observations).sum.toDouble
 
     metrics.foreach { m =>
-      val Array(umi1, umi2)                   = m.umi.split('-')
+      val Array(umi1, umi2)                   = UmiSeparatorPattern.split(m.umi, -1)
```
nh13, line 534: Do we have to worry about performance being reduced, since we're moving from an optimized single character split to something that's not?  See: https://stackoverflow.com/questions/11001330/java-split-string-performances

Should we implement our own `split` that adds a trailing empty string if the string ends with the single-char delimiter

nh13 (changes requested):
I am a bit concerned that we are using a regular expression split versus an optimized single-character split.  I don't think we need the former. Thoughts?

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
clintval: I wondered about this and figured it might impart a slowdown. I'm frustrated Java provides a `.split()` with the option to preserve empty strings but only for regex patterns and not for single character delimiters. Frustrating! Looks like there are few fast alternatives I'll have to compare and contrast. We do have access to Google's `Splitter` via an htsjdk dependency:

https://guava.dev/releases/snapshot/api/docs/com/google/common/base/Splitter.html#on-java.lang.String-

clintval: Turns out most JDKs have a fastpath! We have to use a 1-length character string (`"-"`) and `index = -1`:

<details>
<summary> <code>String.java</code> </summary>

```java
public String[] split(String regex, int limit) {
    /* fastpath if the regex is a
     * (1) one-char String and this character is not one of the
     *     RegEx's meta characters ".$|()[{^?*+\\", or
     * (2) two-char String and the first char is the backslash and
     *     the second is not the ascii digit or ascii letter.
     */
    char ch = 0;
    if (((regex.length() == 1 &&
         ".$|()[{^?*+\\".indexOf(ch = regex.charAt(0)) == -1) ||
         (regex.length() == 2 &&
          regex.charAt(0) == '\\' &&
          (((ch = regex.charAt(1))-'0')|('9'-ch)) < 0 &&
          ((ch-'a')|('z'-ch)) < 0 &&
          ((ch-'A')|('Z'-ch)) < 0)) &&
        (ch < Character.MIN_HIGH_SURROGATE ||
         ch > Character.MAX_LOW_SURROGATE))
    {
        int off = 0;
        int next = 0;
        boolean limited = limit > 0;
        ArrayList<String> list = new ArrayList<>();
        while ((next = indexOf(ch, off)) != -1) {
            if (!limited || list.size() < limit - 1) {
                list.add(substring(off, next));
                off = next + 1;
            } else {    // last one
                //assert (list.size() == limit - 1);
                int last = length();
                list.add(substring(off, last));
                off = last;
                break;
            }
        }
        // If no match was found, return this
        if (off == 0)
            return new String[]{this};

        // Add remaining segment
        if (!limited || list.size() < limit)
            list.add(substring(off, length()));

        // Construct result
        int resultSize = list.size();
        if (limit == 0) {
            while (resultSize > 0 && list.get(resultSize - 1).isEmpty()) {
                resultSize--;
            }
        }
        String[] result = new String[resultSize];
        return list.subList(0, resultSize).toArray(result);
    }
    return Pattern.compile(regex).split(this, limit);
}
```

</details>

clintval:
@nh13 back to you!

nh13 (approved):
Thank-you!

PR #956, opened 2024-01-05
PileupBuilder should not report insertions when checking the final mapped base before soft-clipping

jrm5100:
I was using the PileupBuilder and ran into an error when attempting to collect an insertion sequence:

```
case entry: InsertionEntry if entry.rec.start < pos  =>
  // Get insertion bases
  val insStart     = entry.rec.readPosAtRefPos(pos, returnLastBaseIfDeleted = true)
  val insEnd       = entry.rec.readPosAtRefPos(pos + 1, returnLastBaseIfDeleted = true)
  val alleleString = entry.rec.basesString.substring(insStart - 1, insEnd - 1)
```

The call to `substring` failed because `insEnd` was `0`.

I tracked this down to the `PileupBuilder` making an incorrect check. `rec.length` includes soft-clipped bases, so checking `offset < rec.length - 1` will always be true at a pileup for the last mapped base as long as there is at least 1 soft-clipped base. The second half of the check (`rec.refPosAtReadPos(offset + 1)`) will also be true, but for the wrong reason (it's soft-clipped, not an insertion).

I shoehorned in a test where it seems appropriate.

### DIFF
--- a/src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
@@ -171,13 +171,19 @@ class PileupBuilderTest extends UnitSpec {
       p4.iterator.collect{ case x: InsertionEntry => x }.map(_.rec.name).next() shouldBe "q5"
       p4.baseIterator.toSeq should contain theSameElementsAs p4.withoutIndels.iterator.toSeq
 
-      // Locus with the end of a read that is soft-clipped
+      // Locus with the remainder of a read that is soft-clipped
       val p5 = piler.pileup(Chr1, 247)
       p5.depth shouldBe 1
-      p5.iterator.size shouldBe 1 // should not report an insertion due to the remaining bases
+      p5.iterator.size shouldBe 1 // should not report an insertion due to the remaining soft-clipped bases
       p5.baseIterator.size shouldBe 1
       p5.baseIterator.toSeq should contain theSameElementsAs p5.withoutIndels.iterator.toSeq
 
+      // Locus at the site of a single read that is soft-clipped
+      val p6 = piler.pileup(Chr1, 248)
+      p6.depth shouldBe 0
+      p6.iterator.size shouldBe 0
+      p6.baseIterator.size shouldBe 0
+      p6.baseIterator.toSeq should contain theSameElementsAs p5.withoutIndels.iterator.toSeq
       source.safelyClose()
       piler.safelyClose()
     }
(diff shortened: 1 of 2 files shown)


src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
```diff
@@ -170,6 +171,13 @@ class PileupBuilderTest extends UnitSpec {
       p4.iterator.collect{ case x: InsertionEntry => x }.map(_.rec.name).next() shouldBe "q5"
       p4.baseIterator.toSeq should contain theSameElementsAs p4.withoutIndels.iterator.toSeq
 
+      // Locus with the end of a read that is soft-clipped
+      val p5 = piler.pileup(Chr1, 247)
+      p5.depth shouldBe 1
+      p5.iterator.size shouldBe 1 // should not report an insertion due to the remaining bases
+      p5.baseIterator.size shouldBe 1
+      p5.baseIterator.toSeq should contain theSameElementsAs p5.withoutIndels.iterator.toSeq
+
```
clintval, line 180: I also noticed we don't have a test yet for piling up at a locus that is soft-clipped in the read read so would you consider adding a small additional test?

```suggestion
      // Locus with the remainder of a read that is soft-clipped
      val p5 = piler.pileup(Chr1, 247)
      p5.depth shouldBe 1
      p5.iterator.size shouldBe 1 // should not report an insertion due to the remaining soft-clipped bases
      p5.baseIterator.size shouldBe 1
      p5.baseIterator.toSeq should contain theSameElementsAs p5.withoutIndels.iterator.toSeq

      // Locus at the site of a single read that is soft-clipped
      val p6 = piler.pileup(Chr1, 248)
      p6.depth shouldBe 0
      p6.iterator.size shouldBe 0
      p6.baseIterator.size shouldBe 0
      p6.baseIterator.toSeq should contain theSameElementsAs p5.withoutIndels.iterator.toSeq
```

clintval (approved):
Took me a bit to walk through the logic using the underlying functions. This does indeed appear to be a bug fix! Thanks for this @jrm5100!

src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
```diff
@@ -170,6 +171,19 @@ class PileupBuilderTest extends UnitSpec {
...
+      p5.iterator.size shouldBe 1 // should not report an insertion due to the remaining soft-clipped bases
+      p5.baseIterator.size shouldBe 1
+      p5.baseIterator.toSeq should contain theSameElementsAs p5.withoutIndels.iterator.toSeq
+
+      // Locus at the site of a single read that is soft-clipped
+      val p6 = piler.pileup(Chr1, 248)
+      p6.depth shouldBe 0
+      p6.iterator.size shouldBe 0
+      p6.baseIterator.size shouldBe 0
+      p6.baseIterator.toSeq should contain theSameElementsAs p5.withoutIndels.iterator.toSeq
```
clintval, line 186: Sorry about that.

```suggestion
      p6.baseIterator.toSeq should contain theSameElementsAs p6.withoutIndels.iterator.toSeq
```

jrm5100: Ha, just caught this running the test locally.

Issue #957, opened 2024-01-19
CopyUmiFromReadName should support reverse-complemented UMIs prefixed by "r"


msto:
## Problem

Some UMIs produced by BCL convert are prefixed with "r", indicating "reverse complement."[^1]

> When sequencing a run with Unique Molecular Identifier (UMI) situated in the index2 (i5) on a NextSeq1000/2000 instrument, BCL Convert will put a leading "r" in front of the reverse-complemented UMI in the FASTQ header. 

`CopyUmiFromReadName` enforces that the UMI sequence contains only valid bases (A/C/G/T/N[^2]) or a delimiter between multiple UMIs (`+` or `-`). UMIs prefixed with "r" fail this validation.

## Proposed solution

I think it would be sensible to add the following features to `CopyUmiFromReadName`:

- `--umi-delimiter` (`Char`, default=`+`)
    - The default should be `+`, as this is the default delimiter in Illumina FASTQs.[^2]
    - If this character appears in the UMI sequence, split the sequence into multiple UMIs and validate each separately.
    - Join multiple UMIs with a hyphen (`-`) before storing them in the `RX` tag, per SAM spec.[^3]

- Support reverse complemented UMIs.
    - For each UMI, if it begins with "r", remove the "r" and (optionally?) reverse-complement the remaining sequence
    - (NB: This could be turned off by default, e.g. with `--allow-reverse-umis` as a flag. @clintval raised the concern that degenerate UMIs could include `r` as a masked A or G[^4], although this does not appear to be permitted under the current Illumina FASTQ spec.[^2])

[^1]: https://knowledge.illumina.com/software/general/software-general-reference_material-list/000007945

[^2]: https://support.illumina.com/help/BaseSpace_Sequence_Hub_OLH_009008_2/Source/Informatics/BS/FileFormat_FASTQ-files_swBS.htm
    > Restricted characters: A/T/G/C/N
    > UMI sequences for Read 1 and Read 2, separated by a plus [+].
    
[^3]: https://samtools.github.io/hts-specs/SAMtags.pdf
    > In the case of multiple unique molecular identifiers (e.g., one on each end of the template) the
recommended implementation concatenates all the barcodes with a hyphen (‘-’) between the different
barcodes.

[^4]: https://www.bioinformatics.org/sms/iupac.html

PR #958, opened 2024-01-19
feat: add --umi-prefix to CopyUmiFromReadName

msto:
https://github.com/fulcrumgenomics/fgbio/issues/957


src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
```diff
@@ -42,11 +42,16 @@ import com.fulcrumgenomics.util.{Io, ProgressLogger}
...
+    |To obtain behavior similar to `umi_tools`' `--umi-separator=":r"`, specify the delimiter and
+    |prefix separately, i.e. `--umi-delimiter=":"` and `--umi-prefix="r"`.
   """)
 class CopyUmiFromReadName
 ( @arg(flag='i', doc="The input BAM file") input: PathToBam,
   @arg(flag='o', doc="The output BAM file") output: PathToBam,
-  @arg(doc="Remove the UMI from the read name") removeUmi: Boolean = false
+  @arg(doc="Remove the UMI from the read name") removeUmi: Boolean = false,
+  @arg(doc="Delimiter between the read name and UMI.") umiDelimiter: Char = ':',
+  @arg(doc="Any characters preceding the UMI sequence in the read name.") umiPrefix: Option[String] = None,
```
nh13, line 54: Sorry, I don't understand the behavior. What does this option do?  While I could go read the documentation of `umi_tools` (and I like that you mentioned the equivalency in the usage), I would like the documentation to explicit about the behavior.

Perhaps rename this to `removeUmiPrefix` or `umiPrefixToRemove`?

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -41,9 +41,9 @@ object Umis {
     * @param delimiter the delimiter of fields within the read name
     * @return the modified record
     */
-  def copyUmiFromReadName(rec: SamRecord, removeUmi: Boolean = false, delimiter: Char = ':'): SamRecord = {
+  def copyUmiFromReadName(rec: SamRecord, removeUmi: Boolean = false, delimiter: Char = ':', prefix: Option[String] = None): SamRecord = {
```
nh13, line 44: add to the docs

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -80,8 +80,9 @@ object Umis {
       Some(name.substring(idx + 1, name.length))
     }
 
-    val umi   = rawUmi.map(raw => (if (raw.indexOf('+') > 0) raw.replace('+', '-') else raw).toUpperCase)
-    val valid = umi.forall(u => u.forall(isValidUmiCharacter))
+    val umiSeq = rawUmi.map(seq => (if (prefix.isEmpty) seq else seq.stripPrefix(prefix.get)))
+    val umi = umiSeq.map(raw => (if (raw.indexOf('+') > 0) raw.replace('+', '-') else raw).toUpperCase)
+    val valid  = umi.forall(u => u.forall(isValidUmiCharacter))
```
nh13, line 85: 1. Align the equals on the second assignment to match the code base.  We call this tim-format
2. Avoid the use of `get` on options (this is considered bad form), instead use `match`
```suggestion
    val umiSeq = prefix match {
      case None.     => rawUmi
      case Some(pre) => rawUmi.map(_.stripPrefix(pre))
    }
    val umi    =  umiSeq.map(raw => (if (raw.indexOf('+') > 0) raw.replace('+', '-') else raw).toUpperCase)
    val valid  = umi.forall(u => u.forall(isValidUmiCharacter))
```

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -80,8 +80,9 @@ object Umis {
       Some(name.substring(idx + 1, name.length))
     }
 
-    val umi   = rawUmi.map(raw => (if (raw.indexOf('+') > 0) raw.replace('+', '-') else raw).toUpperCase)
-    val valid = umi.forall(u => u.forall(isValidUmiCharacter))
+    val umiSeq = rawUmi.map(seq => (if (prefix.isEmpty) seq else seq.stripPrefix(prefix.get)))
```
clintval, line 83: Although `.get()` exists, try to avoid it in Scala if possible since it relies on a guarantee that somewhere earlier (near or far away) there is a value contained in the option. That type of logical decoupling can introduce bugs/regressions in the future if a refactoring removes that assumption by accident.

```suggestion
    val umiSeq = rawUmi.flatMap(raw => prefix.map(raw.stripPrefix))
```

nh13 (changes requested):
Thank-you, just a few requests, and I want to just review the usage before we accept

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -80,8 +80,9 @@ object Umis {
...
+    val umi = umiSeq.map(raw => (if (raw.indexOf('+') > 0) raw.replace('+', '-') else raw).toUpperCase)
+    val valid  = umi.forall(u => u.forall(isValidUmiCharacter))
```
clintval, line 86: We also tend to align equals in blocks in the codebase (I have an IntelliJ plugin to do this):

```suggestion
    val umiSeq = rawUmi.flatMap(raw => prefix.map(raw.stripPrefix))
    val umi    = umiSeq.map(raw => (if (raw.indexOf('+') > 0) raw.replace('+', '-') else raw).toUpperCase)
    val valid  = umi.forall(u => u.forall(isValidUmiCharacter))

```

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
tfmorris: > Align the equals on the second assignment to match the code base. We call this tim-format

@nh13 out of curiosity, what tools support this format for linting/formatting/etc?

nh13: Not sure, but something folks have tried previously. @clintval ?

nh13: Found it! https://scalameta.org/scalafmt/

clintval: I use this plugin while selecting the lines I want aligned.

https://plugins.jetbrains.com/plugin/13903-smart-align

It works most of the time pretty well.

nh13 (changes requested):
A number of small changes to format some code similar to the current codebase (I know I know), changing `rc` to `reverseComplement`, and remove `normalizeRcUmis` throughout.

tfmorris (commented):
Thanks for the opportunity to review, but I only commented on the PR to ask a questiion about tooling, so I'm +0

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
```diff
@@ -36,17 +36,30 @@ import com.fulcrumgenomics.util.{Io, ProgressLogger}
...
-( @arg(flag='i', doc="The input BAM file") input: PathToBam,
-  @arg(flag='o', doc="The output BAM file") output: PathToBam,
-  @arg(doc="Remove the UMI from the read name") removeUmi: Boolean = false
+( @arg(flag='i', doc="The input BAM file.") input: PathToBam,
+  @arg(flag='o', doc="The output BAM file.") output: PathToBam,
+  @arg(doc="Remove the UMI from the read name.") removeUmi: Boolean = false,
+  @arg(doc="Delimiter between the read name and UMI.") fieldDelimiter: Char = ':',
+  @arg(doc="Delimiter between UMI sequences.") umiDelimiter: Char = '+',
+  @arg(doc="The prefix to a UMI sequence that indicates it is reverse-complemented.") rcPrefix: Option[String] = None,
+  @arg(doc="Whether to reverse-complement UMI sequences with the '--rc-prefix'.") normalizeRcUmis: Boolean = false,
```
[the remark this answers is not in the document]
msto: To replicate `umi_tools` behavior.

`umi_tools` does not support reverse complementing the UMI. We have used the `--umi-separator` flag to separate the UMI from the read name (i.e. `--umi-separator ":r"`). 

In the [original issue](https://github.com/fulcrumgenomics/fgbio/issues/957) I suggested that `fgbio` could additionally support reverse-complementing the UMI sequence when the UMI is prefixed with `r`. I suggested that this could be optional to maintain compatibility with `umi_tools`.

> - Support reverse complemented UMIs.
>      - For each UMI, if it begins with "r", remove the "r" and (optionally?) reverse-complement the remaining sequence

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
```diff
@@ -36,17 +36,30 @@ import com.fulcrumgenomics.util.{Io, ProgressLogger}
...
 class CopyUmiFromReadName
-( @arg(flag='i', doc="The input BAM file") input: PathToBam,
-  @arg(flag='o', doc="The output BAM file") output: PathToBam,
-  @arg(doc="Remove the UMI from the read name") removeUmi: Boolean = false
+( @arg(flag='i', doc="The input BAM file.") input: PathToBam,
+  @arg(flag='o', doc="The output BAM file.") output: PathToBam,
+  @arg(doc="Remove the UMI from the read name.") removeUmi: Boolean = false,
+  @arg(doc="Delimiter between the read name and UMI.") fieldDelimiter: Char = ':',
+  @arg(doc="Delimiter between UMI sequences.") umiDelimiter: Char = '+',
+  @arg(doc="The prefix to a UMI sequence that indicates it is reverse-complemented.") rcPrefix: Option[String] = None,
```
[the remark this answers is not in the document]
jdidion: As currently designed, if it is defaulted to 'r' then it will always check for the prefix. Since this is only going to happen with NextSeq data (a minority case), it seemed better to have this off by default. Another option would be to default this to 'r' but have another option to enable/disable the prefix handling.

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -38,17 +39,30 @@ object Umis {
...
     * @return the modified record
     */
-  def copyUmiFromReadName(rec: SamRecord, removeUmi: Boolean = false, delimiter: Char = ':'): SamRecord = {
+  def copyUmiFromReadName(rec: SamRecord,
+                          removeUmi: Boolean = false,
+                          fieldDelimiter: Char = ':',
+                          umiDelimiter: Char = '+',
+                          rcPrefix: Option[String] = None,
+                          normalizeRcUmis: Boolean = false): SamRecord = {
     // Extract and set the UMI
```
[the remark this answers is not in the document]
jdidion: Done

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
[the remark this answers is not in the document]
jdidion: @nh13 does this change your opinion on whether to have this option?

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
```diff
@@ -36,17 +36,30 @@ import com.fulcrumgenomics.util.{Io, ProgressLogger}
...
+  @arg(doc="Whether to reverse-complement UMI sequences with the '--rc-prefix'.") normalizeRcUmis: Boolean = false,
 ) extends FgBioTool with LazyLogging {
```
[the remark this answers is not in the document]
jdidion: Done

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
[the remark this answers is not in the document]
nh13: I don't think we ever had the goal of "being compatible" with umi-tools.  Given that, I would remove it.

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
```diff
@@ -36,29 +36,51 @@ import com.fulcrumgenomics.util.{Io, ProgressLogger}
...
     |valid bases (ACGTN), the program will report the error and fail.
     |
-    |If a read name contains multiple UMIs they may be delimited by either hyphens (`-`) or pluses (`+`). The
-    |resulting UMI in the `RX` tag will always be hyphen delimited.
+    |If a read name contains multiple UMIs they may be delimited (typically by a hyphen (`-`) or plus (`+`)).
+    |The `--umi-delimiter` option specifies the delimiter on which to split.  The resulting UMI in the `RX` tag
+    |will always be hyphen delimited.
+    |
+    |Some tools (e.g. BCL Convert) may reverse-complement UMIs on R2 and add a prefix to indicate that the sequence
+    |has been reverse-complemented.  The `--rc-prefix` option specifies the prefix character(s) and causes them to
```
nh13, line 48: usage needs to be updated

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
```diff
@@ -36,29 +36,51 @@ import com.fulcrumgenomics.util.{Io, ProgressLogger}
...
-( @arg(flag='i', doc="The input BAM file") input: PathToBam,
-  @arg(flag='o', doc="The output BAM file") output: PathToBam,
-  @arg(doc="Remove the UMI from the read name") removeUmi: Boolean = false
+( @arg(flag='i', doc="The input BAM file.") input: PathToBam,
+  @arg(flag='o', doc="The output BAM file.") output: PathToBam,
+  @arg(doc="Remove the UMI from the read name.") removeUmi: Boolean = false,
+  @arg(doc="Delimiter between the read name and UMI.") fieldDelimiter: Char = ':',
+  @arg(doc="Delimiter between UMI sequences.") umiDelimiter: Char = '+',
+  @arg(flag='p', doc="The prefix to a UMI sequence that indicates it is reverse-complemented.") reverseComplementPrefix: Option[String] = None,
+  @arg(flag='r', doc="Whether to reverse-complement UMI sequences with the '--reverse-complement-prefix'.") normalizeReverseComplementUmis: Boolean = false,
```
nh13, line 62: Please remove, and condition on if `reverseComplementPrefix` is defined.

nh13 (changes requested):

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
jdidion: Done

src/main/scala/com/fulcrumgenomics/umi/CopyUmiFromReadName.scala
jdidion: Done

nh13 (approved):
LGTM, thank-you for your patience.

PR #964, opened 2024-01-30
GroupReadsByUmi may fail when marking duplicates including secondary/supplementary reads

nh13:
There's an open issue in hts-specs about how we want to handle getting the primary alignment information when looking at a secondary or supplementary read: https://github.com/samtools/hts-specs/issues/755

This PR adds the _read primary_ "rp" tag to store the primary alignment for end of the current secondary/supplementary alignment, in the same format as the "SA" tag.  The _mate's primary_ alignment is stored in the "mp" tag.  Both are currently lowercase as they are not reserved tags.

I have tested that `ZipperBams` will now add these, that `SortBam` will correctly sort in template-coordinate, and finally that `GroupReadsByUmi` passes.  I added tests for `GroupReadsByUmi` and `SamOrder`.

Also, in my hands, secondary and supplementary records will never be output by `GroupReadsByUmi` as [currently only primary alignments are output](https://github.com/fulcrumgenomics/fgbio/blob/afa634e8ef3067fe1c5dec1276049d7cf99bbc36/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala#L722).

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
+++ b/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
@@ -208,7 +208,7 @@ case class Cigar(elems: IndexedSeq[CigarElem]) extends Iterable[CigarElem] {
   def trailingSoftClippedBases: Int = stats.trailingSoftClippedBases
 
   /** Returns the number of bases that are hard-clipped at the start of the sequence. */
-  def leadingHardClippedBases = this.headOption.map { elem =>
+  def leadingHardClippedBases: Int = this.headOption.map { elem =>
     if (elem.operator == CigarOperator.H) elem.length else 0
   }.getOrElse(0)
 
@@ -216,12 +216,22 @@ case class Cigar(elems: IndexedSeq[CigarElem]) extends Iterable[CigarElem] {
   def leadingClippedBases: Int = leadingHardClippedBases + leadingSoftClippedBases
 
   /** Returns the number of bases that are hard-clipped at the end of the sequence. */
-  def trailingHardClippedBases = this.lastOption.map { elem =>
+  def trailingHardClippedBases: Int = this.lastOption.map { elem =>
     if (elem.operator == CigarOperator.H) elem.length else 0
   }.getOrElse(0)
 
   /** Returns the number of bases that are clipped at the end of the sequence. */
   def trailingClippedBases: Int = trailingHardClippedBases + trailingSoftClippedBases
+
+  /** Returns the alignment start position adjusted for leading soft and hard clipped bases. */
+  def unclippedStart(start: Int): Int = {
+    start - this.iterator.takeWhile(_.operator.isClipping).map(_.length).sum
+  }
+
+  /** Returns the alignment end position adjusted for trailing soft and hard clipped bases. */
+  def unclippedEnd(end: Int): Int = {
+    end + this.reverseIterator.takeWhile(_.operator.isClipping).map(_.length).sum
+  }
 
   /** Truncates the cigar based on either query or target length cutoff. */
   private def truncate(len: Int, shouldCount: CigarElem => Boolean): Cigar = {
--- a/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/Bams.scala
@@ -25,6 +25,7 @@
 package com.fulcrumgenomics.bam
 
 import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.alignment.Cigar
 import com.fulcrumgenomics.bam.api.SamOrder.Queryname
 import com.fulcrumgenomics.bam.api._
 import com.fulcrumgenomics.commons.collection.{BetterBufferedIterator, SelfClosingIterator}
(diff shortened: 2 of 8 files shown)


src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
```diff
@@ -175,24 +178,50 @@ object SamOrder {
...
+        if (readChrom < mateChrom || (readChrom == mateChrom && readPos < matePos) ||
+             (readChrom == mateChrom && readPos == matePos && !readNeg)) {
+          TemplateCoordinateKey(readChrom, mateChrom, readPos, matePos, readNeg, mateNeg, lib, mid, rec.name, false)
+        }
+        else {
+          TemplateCoordinateKey(mateChrom, readChrom, matePos, readPos, mateNeg, readNeg, lib, mid, rec.name, true)
+        }
+      } else {
+        val primary   = Supplementary(rec[String]("rp"))
+        val mate      = Supplementary(rec[String]("mp"))
```
nh13, line 205: Todo, better error message or fallback

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -41,6 +42,36 @@ import htsjdk.samtools.util.{CloserUtil, CoordMath, Murmur3, SequenceUtil}
...
+  /** Returns a formatted alignment as per the SA tag: `(rname ,pos ,strand ,CIGAR ,mapQ ,NM ;)+` */
+  def toString(rec: SamRecord): String = {
+    val strand = if (rec.positiveStrand) '+' else '-'
+    f"${rec.refName},${rec.start},${strand},${rec.cigar},${rec.mapq},${rec.getOrElse(SAMTag.NM.name(),0)}"
+  }
+
+
+  def apply(sa: String): Supplementary = {
+    val parts = sa.split(",")
+    Supplementary(parts(0), parts(1).toInt, parts(2) == "+", Cigar(parts(3)), parts(4).toInt, parts(5).toInt)
```
msto, line 71:
```suggestion
    Supplementary(parts(0), parts(1).toInt - 1, parts(2) == "+", Cigar(parts(3)), parts(4).toInt, parts(5).toInt)
```

**suggestion** We need to subtract 1 if `start` is zero-based.

Without scaladoc I'm not sure, but I'm assuming it's zero based; and the `SA` tag is 1-based.

> Pos is a 1-based coordinate.

https://samtools.github.io/hts-specs/SAMtags.pdf

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -107,11 +138,21 @@ case class Template(r1: Option[SamRecord],
...
+      nonPrimary("mp") = Supplementary.toString(primary)
+      r2.foreach(r => nonPrimary("rp") = Supplementary.toString(r))
     }
-    for (primary <- r2; supp <- r1Supplementals) {
-      SamPairUtil.setMateInformationOnSupplementalAlignment(supp.asSam, primary.asSam, true)
+    for (primary <- r2; nonPrimary <- r1NonPrimary) {
+      SamPairUtil.setMateInformationOnSupplementalAlignment(nonPrimary.asSam, primary.asSam, true)
+      nonPrimary(SAMTag.MQ.name()) = primary.mapq
+      nonPrimary("mp") = Supplementary.toString(primary)
+      r1.foreach(r => nonPrimary("rp") = Supplementary.toString(r))
```
msto, line 155: **question** Would you find it more legible to extract these for loops into a helper so we don't repeat it twice?

src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
```diff
@@ -175,24 +178,50 @@ object SamOrder {
...
-        if (index >= 0) m.substring(0, index) else m
-      }.getOrElse("")
-
-      if (readChrom < mateChrom || (readChrom == mateChrom && readPos < matePos) ||
-           (readChrom == mateChrom && readPos == matePos && !readNeg)) {
-        TemplateCoordinateKey(readChrom, mateChrom, readPos, matePos, readNeg, mateNeg, lib, mid, rec.name, false)
-      }
-      else {
-        TemplateCoordinateKey(mateChrom, readChrom, matePos, readPos, mateNeg, readNeg, lib, mid, rec.name, true)
+      // For non-secondary/non-supplementary alignments, use the info in the record.  For secondary and supplementary
```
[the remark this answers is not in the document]
msto: if you additionally set `mp` on the primary alignments, not just the supplementaries, (and also take my suggestion to define an `apply` for `SamRecord` 🙂 ) you could do the following:

```scala
val primary = if (!rec.secondary && !rec.supplementary) Supplementary(rec) else Supplementary(rec[String]("rp"))
val mate = Supplementary(rec[String]("mp"))

// Just the second branch, using the info from `Supplementary` instead of `SamRecord`
...
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -719,7 +719,7 @@ class GroupReadsByUmi
 
       // Then output the records in the right order (assigned tag, read name, r1, r2)
       templatesByMi.keys.toSeq.sortBy(id => (id.length, id)).foreach(tag => {
-        templatesByMi(tag).sortBy(t => t.name).flatMap(t => t.primaryReads).foreach(rec => {
+        templatesByMi(tag).sortBy(t => t.name).flatMap(t => t.allReads).foreach(rec => {
```
msto, line 722: **question** Where are `--include-supplementary` and `--include-secondary` taken into consideration?

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
```diff
@@ -41,6 +42,36 @@ import htsjdk.samtools.util.{CloserUtil, CoordMath, Murmur3, SequenceUtil}
 import java.io.Closeable
 import scala.math.{max, min}
 
+
+
+case class Supplementary(refName: String, start: Int, positiveStrand: Boolean, cigar: Cigar, mapq: Int, nm: Int) {
+  def negativeStrand: Boolean = !positiveStrand
+  def refIndex(header: SAMFileHeader): Int = header.getSequence(refName).getSequenceIndex
+
+  def end: Int = start + cigar.lengthOnTarget - 1
```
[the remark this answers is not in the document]
nh13: `fgbio` is generally 1-based inclusive, but I don't think we say that anywhere.

src/main/scala/com/fulcrumgenomics/bam/Bams.scala
nh13: ditto: `fgbio` is generally 1-based inclusive

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
nh13: In the initial filter of the BAM file (see the `_includeSecondaryReads` variabel)

src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala
```diff
@@ -208,21 +208,31 @@ case class Cigar(elems: IndexedSeq[CigarElem]) extends Iterable[CigarElem] {
   def trailingSoftClippedBases: Int = stats.trailingSoftClippedBases
 
   /** Returns the number of bases that are hard-clipped at the start of the sequence. */
-  def leadingHardClippedBases = this.headOption.map { elem =>
+  def leadingHardClippedBases: Int = this.headOption.map { elem =>
```
nh13, line 211: public defs should have return types here and below

PR #976, opened 2024-04-03
Add support for multi-allelic variants

jdidion:


src/main/scala/com/fulcrumgenomics/vcf/DownsampleVcf.scala
```diff
@@ -97,69 +98,106 @@ object DownsampleVcf extends LazyLogging {
...
       )
+      // compute genotype log-likelihoods      
+      (0 until numAlleles).flatMap(b =>
+        (0 to b).map(a =>
+          (0 until numAlleles).map(allele =>
+            logProbs(Array(a, b).count(_ == allele)) * alleleDepths(allele)
+          ).sum
+        )
+      )
+    }
```
TedBrookings, line 159: I see that this passes the equivalency test with the biallelic case, but I found it very difficult to understand. It might help to have a bit more commentary here. I'd suggest at least something like:
```
// compute genotype log-likelihoods in a flat array pairing each allele with all alleles earlier or equal to it
// e.g. lp_AA, lp_BA, lp_BB, lp_CA, lp_CB, lp_CC, ...
```

src/main/scala/com/fulcrumgenomics/vcf/DownsampleVcf.scala
```diff
@@ -97,69 +98,106 @@ object DownsampleVcf extends LazyLogging {
...
     /**
       * Returns the likelihoods as a list of phred-scaled integers (i.e, the value of the PL tag).
       * @return a list of phred-scaled likelihooodS for AA, AB, BB.
       */
-    def pls = IndexedSeq(aa.round.toInt, ab.round.toInt, bb.round.toInt)
+    def pls: IndexedSeq[Int] = {
+      Likelihoods.logToPhredLikelihoods(genotypeLikelihoods)
+    }
+
+    def mostLikelyGenotype: Option[(Int, Int)] = {
```
TedBrookings, line 181:
```suggestion
    private def mostLikelyGenotype: Option[(Int, Int)] = {
```

src/main/scala/com/fulcrumgenomics/vcf/DownsampleVcf.scala
```diff
@@ -13,6 +13,7 @@ import com.fulcrumgenomics.vcf.DownsampleVcf.{downsampleAndRegenotype, winnowVar
 
 import scala.math.log10
 import scala.util.Random
+import scala.tools.nsc.doc.html.HtmlTags
```
TedBrookings, line 16: Do we need this?

TedBrookings (approved):
It looks good to me. A couple very minor issues, but I think the functional organization is actually improved and easier to understand even with the more complex likelihood model.

PR #981, opened 2024-04-24
Make `PileupBuilder.includeMapPositionsOutsideFrInsert` intuitively correct

### LINKED ISSUE #980: bug: the logic for the PileupBuilder `includeMapPositionsOutsideFrInsert` filter is wrong
`positionIsOutsideFrInsert` is actually `positionIsInsideFrInsert`

https://github.com/fulcrumgenomics/fgbio/blob/3a74fd28ff630b417410953541ff107a1b7b0fb7/src/main/scala/com/fulcrumgenomics/bam/pileup/PileupBuilder.scala#L123-L128

The filter is used in a way that is reversed so that the current tests still work as designed.

https://github.com/fulcrumgenomics/fgbio/blob/3a74fd28ff630b417410953541ff107a1b7b0fb7/src/main/scala/com/fulcrumgenomics/bam/pileup/PileupBuilder.scala#L203-L205

Additionally there is a redundant `rec.isFrPair` call (this is done inside the method).

I came across this because I was tracking down some aberrant alleles in a pileup. Should reads mapped to separate chromosomes be rejected by this filter, or should that be a separate filter (likely in `quickAcceptRecord`)?  Currently these are accepted even if `includeMapPositionsOutsideFrInsert` is `false`, which i think may be misleading.

clintval:
Closes #980

### DIFF
--- a/src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
@@ -289,7 +289,7 @@ class PileupBuilderTest extends UnitSpec {
       piler.safelyClose()
     }
 
-    it should "not filter out single-end records when we are removing records outside the insert of FR pairs" in {
+    it should "not filter out single-end records when we are not filter for mapped pairs only and we are not removing records where the position is outside the insert of FR pairs" in {
       val builder = new SamBuilder(readLength = ReadLength, sd = Some(TestSequenceDictionary), sort = Some(Coordinate))
 
       builder.addFrag(name = "q1", contig = Chr1Index, start = 100)
--- a/src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
@@ -289,6 +289,26 @@ class PileupBuilderTest extends UnitSpec {
       piler.safelyClose()
     }
 
+    it should "not filter out single-end records when we are removing records outside the insert of FR pairs" in {
+      val builder = new SamBuilder(readLength = ReadLength, sd = Some(TestSequenceDictionary), sort = Some(Coordinate))
+
+      builder.addFrag(name = "q1", contig = Chr1Index, start = 100)
+
+      val source = builder.toSource
+      val piler  = PileupBuilder(
+        source,
+        accessPattern                      = accessPattern,
+        mappedPairsOnly                    = false,
+        includeMapPositionsOutsideFrInsert = true
+      )
+
+      piler.pileup(Chr1, 100).depth shouldBe 1
+      piler.pileup(Chr1, 100 + ReadLength - 1).depth shouldBe 1
+
+      source.safelyClose()
+      piler.safelyClose()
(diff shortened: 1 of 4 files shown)


nh13 (approved):

clintval:
@jrm5100 could I get your 👀 on this one?

jrm5100 (commented):
See discussion here: https://github.com/fulcrumgenomics/fgbio/issues/980#issuecomment-2075049301

src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
```diff
@@ -323,6 +343,24 @@ class PileupBuilderTest extends UnitSpec {
       piler.safelyClose()
     }
 
+    it should "exclude records that appear to be in an FR pair but are on different chromosomes" in {
```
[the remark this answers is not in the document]
clintval: Yeah, this test is overly protective since it is for a read pair that "appears to be FR" based only on start/end positions and strand (but not chromosome) to always ensure reference sequence is taken into consideration. I added it to double-check some behavior while investigating and decided to keep it now that it is written.

clintval:
We've settled on how the filter is supposed to work and I've updated the documentation.

The filter should now be more obvious in how it is supposed to be used.

No functional change will occur when this PR is merged. The only updates:

- A variable name was updated to be logically correct
- An additional unit test was added for sanity sake
- Better documentation exists for the filter's intended use case

src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
```diff
@@ -289,6 +289,26 @@ class PileupBuilderTest extends UnitSpec {
...
+      val builder = new SamBuilder(readLength = ReadLength, sd = Some(TestSequenceDictionary), sort = Some(Coordinate))
+
+      builder.addFrag(name = "q1", contig = Chr1Index, start = 100)
+
+      val source = builder.toSource
+      val piler  = PileupBuilder(
+        source,
+        accessPattern                      = accessPattern,
+        mappedPairsOnly                    = false,
+        includeMapPositionsOutsideFrInsert = true
```
jrm5100, line 302:
```suggestion
```

src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
```diff
@@ -289,6 +289,26 @@ class PileupBuilderTest extends UnitSpec {
       piler.safelyClose()
     }
 
+    it should "not filter out single-end records when we are removing records outside the insert of FR pairs" in {
```
jrm5100, line 292:
```suggestion
    it should "not filter out single-end records when mappedPairsOnly is false" in {
```

jrm5100 (commented):
This looks solid. The one test just needs updating (though I may have the intention of it wrong).

src/test/scala/com/fulcrumgenomics/bam/pileup/PileupBuilderTest.scala
clintval: I updated the test name to be more clear about that!

PR #987, opened 2024-05-20
Add conda install instructions to homepage

clintval:


msto (approved):

nh13 (approved):

clintval:
@nh13 yes I think we should, but because of the way GH Pages works, I'll have to send up a separate PR and cannot add it to this one. Thanks!

PR #989, opened 2024-05-20
feat: FastqToBam can extract UMI(s) from the comment in the read name

nh13:


src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -117,6 +123,7 @@ class FastqToBam
   validate(input.length == actualReadStructures.length, "input and read-structure must be supplied the same number of times.")
   validate(1 to 2 contains actualReadStructures.flatMap(_.templateSegments).size, "read structures must contain 1-2 template reads total.")
   validate(!extractUmisFromReadNames || umiQualTag.isEmpty, "Cannot extract UMI qualities when also extracting UMI from read names.")
+  validate(!extractUmisFromReadComment || umiQualTag.isEmpty, "Cannot extract UMI qualities when also extracting UMI from read description.")
```
nh13, line 126:
```suggestion
  validate(!extractUmisFromReadComment || umiQualTag.isEmpty, "Cannot extract UMI qualities when also extracting UMI from the comment in the read name.")
```

src/test/scala/com/fulcrumgenomics/umi/UmisTest.scala
```diff
@@ -91,7 +126,7 @@ class UmisTest extends UnitSpec with OptionValues {
     copyUmiFromReadName(rec=rec("UMI:C:ACC+GGT"), removeUmi=true).nameAndUmi shouldBe ("UMI:C", "ACC-GGT")
   }
   
-  it should "split on a different name delimiter if specified" in {
+  it should "split on a different delimiter if specified" in {
```
nh13, line 129:
```suggestion
  it should "split on a different delimiter name if specified" in {
```

src/test/scala/com/fulcrumgenomics/umi/UmisTest.scala
```diff
@@ -101,7 +136,7 @@ class UmisTest extends UnitSpec with OptionValues {
     copyUmiFromReadName(rec=rec("UMI:C:ACC+GGT")).nameAndUmi shouldBe ("UMI:C:ACC+GGT", "ACC-GGT")
   }
 
-  it should "fail if the read name has only one field" in {
+  it should "fail if the read has only one field" in {
```
nh13, line 139:
```suggestion
  it should "fail if the read name has only one field" in {
```

src/test/scala/com/fulcrumgenomics/umi/UmisTest.scala
```diff
@@ -43,7 +43,7 @@ class UmisTest extends UnitSpec with OptionValues {
     Umis.extractUmisFromReadName("1:2:3:4:5:6:7:ACGTACGT", strict=true).value shouldBe "ACGTACGT"
   }
 
-  it should "return None if the read name has only 7 parts in strict mode" in {
+  it should "return None if the read has only 7 parts in strict mode" in {
```
nh13, line 46:
```suggestion
  it should "return None if the read name has only 7 parts in strict mode" in {
```

src/test/scala/com/fulcrumgenomics/umi/UmisTest.scala
```diff
@@ -91,7 +126,7 @@ class UmisTest extends UnitSpec with OptionValues {
     copyUmiFromReadName(rec=rec("UMI:C:ACC+GGT"), removeUmi=true).nameAndUmi shouldBe ("UMI:C", "ACC-GGT")
   }
   
-  it should "split on a different name delimiter if specified" in {
+  it should "split on a different delimiter name if specified" in {
```
nh13, line 129:
```suggestion
  it should "split on a different name delimiter if specified" in {
```

src/test/scala/com/fulcrumgenomics/fastq/FastqToBamTest.scala
```diff
@@ -383,6 +383,37 @@ class FastqToBamTest extends UnitSpec {
...
+      FastqRecord("q1:2:3:4:5:6:7:8", "GGNCCGAAAAAAA", "=============", comment=Some("0:1:2:ACGT+CGTA")),
+      FastqRecord("q2:2:3:4:5:6:7:8", "TANAACAAAAAAA", "=============", comment=Some("0:1:2:TTGA+TAAT")),
+    )
+    val rs  = ReadStructure("2M1S2M+T")
+    val bam = makeTempFile("fastqToBamTest.", ".bam")
+    new FastqToBam(input=Seq(r1), readStructures=Seq(rs), output=bam, sample="s", library="l", extractUmisFromReadComment=true).execute()
+    val recs = readBamRecs(bam)
+    recs should have size 2
+    recs(0).apply[String]("RX") shouldBe "ACGT-CGTA-GG-CC"
+    recs(1).apply[String]("RX") shouldBe "TTGA-TAAT-TA-AA"
```
msto, line 413: Why are these suffixed with `-GG-CC` and `-TA-AA`?

nh13: As per the method docs, the UMIs may be extracted from the read names, the read sequences, or both.  In this case, the read structure shows UMI bases in the read sequences themselves, as well as the comment in the read name header, so we get four (!) UMI segments, two from the read sequences, and two from the comment in the read header.

src/main/scala/com/fulcrumgenomics/fastq/FastqToBam.scala
```diff
@@ -166,7 +173,11 @@ class FastqToBam
       val templates     = subs.iterator.filter(_.kind == Template).toList
 
       // If requested, pull out the UMI(s) from the read name
-      val umiFromReadName = if (extractUmisFromReadNames) Umis.extractUmisFromReadName(fqs.head.name, strict=true) else None
+      val umiFromReadName = {
+        if (extractUmisFromReadNames) Umis.extractUmisFromReadName(fqs.head.name, strict=true)
+        else if (extractUmisFromReadComment) fqs.head.comment.flatMap(comment => Umis.extractUmisFromReadComment(comment, strict=true))
```
msto, line 178: Since `strict` is set to `true` and is not configurable at the CLI, should the CLI usage describe the constraints imposed by strict extraction?

PR #993, opened 2024-05-30
Add TemplateCoordinate sort order to the usage of SortBam

nh13:


src/main/scala/com/fulcrumgenomics/bam/SortBam.scala
```diff
@@ -41,6 +41,10 @@ import com.fulcrumgenomics.util.Io
     |and several
     |4. **RandomQuery**: sorts the reads into a random order but keeps reads with the same
     |   queryname together. The ordering is deterministic for any given input.
+    |5. **TemplateCoordinate**: sorts the reads The sort order used by `GroupReadByUmi`. Sorts reads by
```
nh13, line 44: copied verbatim from: https://github.com/fulcrumgenomics/fgbio/blob/d5b38ca053d7b6dcd3661c7d2b6192cfc40f57b1/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala#L168-L170

msto, line 44:
```suggestion
    |5. **TemplateCoordinate**: The sort order used by `GroupReadByUmi`. Sorts reads by
```

I think this was a copy/paste artifact from the preceding line

msto (approved):

PR #994, opened 2024-05-30
Validate IO in SortBam to provide nicer exceptions

nh13:
Using the `Io.assert*` within the constructor will yield better error messages for the user.

Also added an explicit close to the input `SamSource`

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/SortBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/SortBam.scala
@@ -59,13 +59,15 @@ class SortBam
   @arg(flag='s', doc="Order into which to sort the records.") val sortOrder: SamOrder = SamOrder.Coordinate,
   @arg(flag='m', doc="Max records in RAM.") val maxRecordsInRam: Int = SamWriter.DefaultMaxRecordsInRam
 ) extends FgBioTool with LazyLogging {
+  
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
+  
   override def execute(): Unit = {
-    Io.assertReadable(input)
-    Io.assertCanWriteFile(output)
-
     val in  = SamSource(input)
     val out = SamWriter(output, in.header.clone(), sort=Some(sortOrder), maxRecordsInRam=maxRecordsInRam)
     out ++= in
     out.close()
+    in.safelyClose()
   }
 }


src/main/scala/com/fulcrumgenomics/bam/SortBam.scala
```diff
@@ -59,13 +59,15 @@ class SortBam
   @arg(flag='s', doc="Order into which to sort the records.") val sortOrder: SamOrder = SamOrder.Coordinate,
   @arg(flag='m', doc="Max records in RAM.") val maxRecordsInRam: Int = SamWriter.DefaultMaxRecordsInRam
 ) extends FgBioTool with LazyLogging {
+  
+  Io.assertReadable(input)
+  Io.assertCanWriteFile(output)
```
msto, line 64: Naive question, what changes when these are moved to the constructor/class body? 

Since the constructor doesn't do anything else, and nothing happens in `execute` before the assert, I would have thought that asserting at either location would be functionally the same

src/main/scala/com/fulcrumgenomics/bam/SortBam.scala
```diff
@@ -59,13 +59,15 @@ class SortBam
...
+  
   override def execute(): Unit = {
-    Io.assertReadable(input)
-    Io.assertCanWriteFile(output)
-
     val in  = SamSource(input)
     val out = SamWriter(output, in.header.clone(), sort=Some(sortOrder), maxRecordsInRam=maxRecordsInRam)
     out ++= in
     out.close()
+    in.safelyClose()
```
msto, line 71: Also a naive question, why doesn't `out` also take `safelyClose()`?

src/main/scala/com/fulcrumgenomics/bam/SortBam.scala
nh13: Exceptions in the constructor are [`ValidationException`s](https://github.com/fulcrumgenomics/sopt/blob/fa5a8db01f6db59c9f9f2e40414677296d9ff8fc/src/main/scala/com/fulcrumgenomics/sopt/cmdline/ValidationException.scala#L30), and any exceptions (typically validation errors of the arguments) are nicely formatted on the command line, versus a huge stacktrace.  And it will print out the usage too.  But none of those things will happen when _running_ the tool

src/main/scala/com/fulcrumgenomics/bam/SortBam.scala
nh13: The idea is that we care that the _output_ files are closed, but do not want the tool to fail/crash/except when we're cleaning up resources after the main effort of the tool has completed, so we do not propagate errors during closing inputs in this case.

tfenne (approved):

PR #996, opened 2024-06-28
Update Bams.scala 

yfarjoun:
-u is no longer supported by `samtools sort`


msto:
I don't see this deprecation mentioned in the docs or any of the recent release notes

http://www.htslib.org/doc/samtools-sort.html#OPTIONS
https://github.com/samtools/samtools/releases

yfarjoun:
odd. my version doesn't have it and when I looked for it I didn't see it! sorry for the noise.

nh13:
Still alive https://github.com/samtools/samtools/blob/972c1889942a4f07d8f62e93330f723da919c271/bam_sort.c#L1611

PR #1003, opened 2024-07-22
feat: raise exception if CollectDuplexSeqMetrics run on consensus BAM

### LINKED ISSUE #992: Warn or raise an exception when CollectDuplexSeqMetrics is run on a consensus BAM
It is possible to run the tool `CollectDuplexSeqmetrics` on a consensus BAM file but the results will not be intuitively correct since metrics collection is expecting the input be **_raw reads_** (and not consensus sequences) annotated with the `MI` tag:

https://github.com/fulcrumgenomics/fgbio/blob/d5b38ca053d7b6dcd3661c7d2b6192cfc40f57b1/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala#L223-L225

We should either support gathering metrics over the consensus BAM or raise an exception for unsupported behavior if any of the consensus SAM tags are encountered.

znorgaard:
Closes #992.

This PR adds a new `isConsensusRead` method to `Umis.scala` which is then used in `CollectDuplexSeqMetrics` to raise an exception if the first record is a consensus record.

In addition to the unit tests, I ran this on a consensus BAM locally and generated the expected error message. Running on the preceding grouped BAM generated the expected outputs as well.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -270,10 +270,6 @@ trait SamRecord {
       SamPairUtil.getPairOrientation(this) == PairOrientation.FR
   }
 
-  def isConsensus: Boolean = {
-    AllPerReadTags.exists(this.contains)
-  }
-
   /** Clone method that does a "reasonably deep" clone. The bases and quals are cloned as is the attributes map,
     * though not the values in the attributes map. */
   override def clone(): SamRecord = {
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -27,6 +27,7 @@ package com.fulcrumgenomics.bam.api
 
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.alignment.Cigar
+import com.fulcrumgenomics.umi.ConsensusTags.PerRead.AllPerReadTags
 import htsjdk.samtools
 import htsjdk.samtools.SamPairUtil.PairOrientation
 import htsjdk.samtools._
@@ -267,6 +268,10 @@ trait SamRecord {
       mateMapped &&
       refIndex == mateRefIndex &&
       SamPairUtil.getPairOrientation(this) == PairOrientation.FR
+  }
+
+  def isConsensus: Boolean = {
+    AllPerReadTags.exists(this.contains)
   }
 
   /** Clone method that does a "reasonably deep" clone. The bases and quals are cloned as is the attributes map,
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
@@ -27,7 +27,6 @@ package com.fulcrumgenomics.bam.api
 
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.alignment.Cigar
-import com.fulcrumgenomics.umi.ConsensusTags.PerRead.AllPerReadTags
 import htsjdk.samtools
 import htsjdk.samtools.SamPairUtil.PairOrientation
 import htsjdk.samtools._
--- a/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
@@ -306,18 +306,19 @@ class CollectDuplexSeqMetrics
   override def execute(): Unit = {
     // Build the iterator we'll use based on whether or not we're restricting to a set of intervals
     val in = SamSource(input)
-    val _filteredIterator = in.iterator.filter(r => r.paired && r.mapped && r.mateMapped && r.firstOfPair && !r.secondary && !r.supplementary)
+    val _filteredIterator = in
+      .iterator
+      .filter(r => r.paired && r.mapped && r.mateMapped && r.firstOfPair && !r.secondary && !r.supplementary)
       .bufferBetter
     // Ensure the records are not consensus records
     _filteredIterator.headOption.foreach {
-      r =>
-        def exceptionString = s"Input BAM file to CollectDuplexSeqMetrics ($input) appears to contain consensus reads." +
+      rec =>
+        def exceptionString = s"Input BAM file to CollectDuplexSeqMetrics ($input) appears to contain consensus sequences. " +
           "CollectDuplexSeqMetrics cannot run on consensus BAMs, and instead requires the UMI-grouped BAM generated " +
           "prior to consensus calling. The UMI-grouped BAM is the output of running GroupReadsByUmi." +
-          s"\nFirst record in $input has consensus tags present:\n${_filteredIterator.head}"
+          s"\nFirst record in $input has consensus SAM tags present:\n$rec"
 
-        if (Umis.isConsensusRead(r)) throw new IllegalArgumentException(exceptionString)
-        else r
+        if (Umis.isConsensusRead(rec)) throw new IllegalArgumentException(exceptionString)
     }
     val iterator = intervals match {
       case None       => _filteredIterator
--- a/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
@@ -313,7 +313,7 @@ class CollectDuplexSeqMetrics
     // Ensure the records are not consensus records
     _filteredIterator.headOption.foreach {
       rec =>
-        def exceptionString = s"Input BAM file to CollectDuplexSeqMetrics ($input) appears to contain consensus sequences. " +
+        val exceptionString = s"Input BAM file to CollectDuplexSeqMetrics ($input) appears to contain consensus sequences. " +
           "CollectDuplexSeqMetrics cannot run on consensus BAMs, and instead requires the UMI-grouped BAM generated " +
           "prior to consensus calling. The UMI-grouped BAM is the output of running GroupReadsByUmi." +
           s"\nFirst record in $input has consensus SAM tags present:\n$rec"
--- a/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
@@ -307,11 +307,18 @@ class CollectDuplexSeqMetrics
     // Build the iterator we'll use based on whether or not we're restricting to a set of intervals
     val in = SamSource(input)
     val _filteredIterator = in.iterator.filter(r => r.paired && r.mapped && r.mateMapped && r.firstOfPair && !r.secondary && !r.supplementary)
-      .tapEach {
-        r =>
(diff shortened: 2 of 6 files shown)


src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -269,6 +270,10 @@ trait SamRecord {
       SamPairUtil.getPairOrientation(this) == PairOrientation.FR
   }
 
+  def isConsensus: Boolean = {
+    AllPerReadTags.exists(this.contains)
+  }
+
```
clintval, line 276: I would prefer this not on the `SamRecord` API since it is specific to the consensus tooling and not SAM records in general. The way this is implemented, it will return true if the SAM record is an fgbio-style duplex consensus, but would return false if it was an fgbio-style single strand consensus. I'd suggest moving the helper function to another consensus-specific namespace like `Umis.scala` and the logic could be simpler with pseudo-code: `hasTag(cd) | (hasTag(ad) and hasTag(bd))` so it would detect either a single-strand consensus or a duplex consensus.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -340,6 +340,7 @@ class CollectDuplexSeqMetrics
 
       // Assign reads a random number between 0 and 1 inclusive based on their read name
       group.foreach { rec =>
+        if (rec.isConsensus) throw new IllegalArgumentException("Found consensus record. Expected UMI-grouped BAM")
```
clintval, line 343: This validation could be raised much earlier and external to the `collect()` scope. Perhaps `.tapEach()` and make the exception raise on the `filteredIterator` in the tool itself? You could also wrap the filtered iterator in a peekable iterator, perform the validation on the first record if it exists, and then send the peekable iterator into `.collect()`.

src/test/scala/com/fulcrumgenomics/bam/api/SamRecordTest.scala
```diff
@@ -289,4 +290,17 @@ class SamRecordTest extends UnitSpec with OptionValues {
...
+    builder.addFrag(start=100).exists(_.isConsensus) shouldBe false
+    builder.addPair(start1=100, start2=100, unmapped2=true).exists(_.isConsensus) shouldBe false
+  }
+
+  it should "return true for reads with consensus tags" in {
+    val builder = new SamBuilder(sort=Some(SamOrder.Coordinate), readLength=10, baseQuality=20)
+    AllPerReadTags.forall(
+      tag => builder.addFrag(start=10, attrs=Map(tag -> 123)).exists(_.isConsensus)
+    ) shouldBe true
+  }
```
clintval, line 305: For a function with a name as generic as `isConsensus` I'd also want a test on a fgbio-style single-strand consensus too.

clintval (changes requested):

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
znorgaard: I think change the name as well to something like `isConsensusRead` because we'll lose some context.

Should we be looking for `cd`, `ad`, and `bd` (per base tags) or `cD`, `aD`, and `bD` (per read tags)? Or does it not matter?

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -27,6 +27,7 @@ package com.fulcrumgenomics.bam.api
 
 import com.fulcrumgenomics.FgBioDef._
 import com.fulcrumgenomics.alignment.Cigar
+import com.fulcrumgenomics.umi.ConsensusTags.PerRead.AllPerReadTags
```
tfenne, line 30: Can you please re-remove this import?

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -26,6 +26,8 @@
 package com.fulcrumgenomics.umi
 
 import com.fulcrumgenomics.bam.api.SamRecord
+import com.fulcrumgenomics.umi.ConsensusTags.PerBase.AbRawReadCount
+import com.fulcrumgenomics.umi.ConsensusTags.PerRead.{BaRawReadCount, RawReadCount}
```
tfenne, line 30: I think I would suggest just importing `ConsensusTags` and then using qualified names below like `ConsensusTags.PerRead.RawReadCount`.

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -127,4 +129,13 @@ object Umis {
...
     ch == 'A' || ch == 'C' || ch == 'G' || ch == 'T' || ch == 'N' || ch == '-'
   }
+
+  /** Tests if a record is a consensus or not
+   *
+   * @param rec the record to test
+   * @return boolean indicating if the record is a consensus or not
+   */
+  def isConsensusRead(rec: SamRecord): Boolean = {
+    rec.contains(RawReadCount) | (rec.contains(AbRawReadCount) & rec.contains(BaRawReadCount))
```
tfenne, line 139:
```suggestion
    rec.contains(RawReadCount) || (rec.contains(AbRawReadCount) && rec.contains(BaRawReadCount))
```

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -307,6 +307,11 @@ class CollectDuplexSeqMetrics
     // Build the iterator we'll use based on whether or not we're restricting to a set of intervals
     val in = SamSource(input)
     val _filteredIterator = in.iterator.filter(r => r.paired && r.mapped && r.mateMapped && r.firstOfPair && !r.secondary && !r.supplementary)
+      .tapEach {
+        r =>
+          if (Umis.isConsensusRead(r)) throw new IllegalArgumentException("Found consensus record. Expected UMI-grouped BAM")
+          else r
+      }
```
tfenne, line 314: A couple of thoughts here:

1. Is it necessary to check each and every read?  Or could you add `.bufferBetter` to the construction of `_filteredIterator` and then just check the first element?
2. I think this is a case where a much more comprehensive error message would be useful, something like:

```
Input BAM file to CollectDuplexSeqMetrics ($input) appears to contain consensus reads.
CollectDuplexSeqMetrics cannot run on consensus BAMs, and instead requires the UMI-grouped BAM prior to consensus calling.  The UMI-grouped BAM is the output of running
GroupReadsByUmi.

First record in $input has consensus tags present:
${_filteredIterator.peek()}
```

tfenne (changes requested):

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
znorgaard: For the first point, I think it depends on what we're trying to guard against. If we're only guarding against a well formed consensus BAM being input then I think checking the first record should work.

If we're guarding against all the chaos a user may concoct, then this will not be sufficient. But, I think guarding against all those possibilities is probably unnecessary and unrealistic.

clintval (commented):
Looking good! I do like the buffered and peek approach since a SAM/BAM is likely to have all the records with the tags, or none of them (in the non-corrupt example), so checking the first should be sufficient to catch 99%+ of improper inputs.

src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala
```diff
@@ -307,6 +307,18 @@ class CollectDuplexSeqMetrics
...
     val in = SamSource(input)
     val _filteredIterator = in.iterator.filter(r => r.paired && r.mapped && r.mateMapped && r.firstOfPair && !r.secondary && !r.supplementary)
+      .bufferBetter
+    // Ensure the records are not consensus records
+    _filteredIterator.headOption.foreach {
+      r =>
+        def exceptionString = s"Input BAM file to CollectDuplexSeqMetrics ($input) appears to contain consensus reads." +
+          "CollectDuplexSeqMetrics cannot run on consensus BAMs, and instead requires the UMI-grouped BAM generated " +
+          "prior to consensus calling. The UMI-grouped BAM is the output of running GroupReadsByUmi." +
+          s"\nFirst record in $input has consensus tags present:\n${_filteredIterator.head}"
```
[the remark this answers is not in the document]
znorgaard: I think it's solid, starts with the query name and provides a little extra info.

```
Exception in thread "main" java.lang.IllegalArgumentException: Input BAM file to CollectDuplexSeqMetrics ([/full/path/to/]TEST-SAMPLE.consensus.clipped.bam) appears to contain consensus reads. CollectDuplexSeqMetrics cannot run on consensus BAMs, and instead requires the UMI-grouped BAM generated prior to consensus calling. The UMI-grouped BAM is the output of running GroupReadsByUmi.
First record in [/full/path/to/]/TEST-SAMPLE.consensus.clipped.bam has consensus tags present:
TEST-SAMPLE:1084 1/2 45b aligned to chr1:810063-810102.

**Review by clintval — APPROVED** [review]
This works well for me! Thanks Zach! @tfenne could you give a final review?

File: src/main/scala/com/fulcrumgenomics/umi/CollectDuplexSeqMetrics.scala [hunk 9]
```diff
@@ -306,7 +306,20 @@ class CollectDuplexSeqMetrics
...
+    val _filteredIterator = in
+      .iterator
+      .filter(r => r.paired && r.mapped && r.mateMapped && r.firstOfPair && !r.secondary && !r.supplementary)
+      .bufferBetter
+    // Ensure the records are not consensus records
+    _filteredIterator.headOption.foreach {
+      rec =>
+        val exceptionString = s"Input BAM file to CollectDuplexSeqMetrics ($input) appears to contain consensus sequences. " +
+          "CollectDuplexSeqMetrics cannot run on consensus BAMs, and instead requires the UMI-grouped BAM generated " +
+          "prior to consensus calling. The UMI-grouped BAM is the output of running GroupReadsByUmi." +
```
nh13, line 318:
```suggestion
          "CollectDuplexSeqMetrics cannot run on consensus BAMs, and instead requires the UMI-grouped BAM generated " +
          "by GroupReadsByUmi which is run prior to consensus calling." +
```

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -127,4 +128,14 @@ object Umis {
   @inline private def isValidUmiCharacter(ch: Char): Boolean = {
     ch == 'A' || ch == 'C' || ch == 'G' || ch == 'T' || ch == 'N' || ch == '-'
   }
+
+  /** Tests if a record is a consensus or not
```
nh13, line 132:
```suggestion
  /** Returns True if the record appears to be a consensus read, typically produced by fgbio's CallMolecularConsensusReads or CallDuplexConsensusReads.
```

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
```diff
@@ -127,4 +128,14 @@ object Umis {
...
+   *
+   * @param rec the record to test
```
nh13, line 134:
```suggestion
   * @param rec the record to check
```

nh13 (approved):
A few suggestions, but lgtm

PR #1010, opened 2024-08-29
Ensure `--min-reads` is a hard filter in CallDuplexConsensusReads

### LINKED ISSUE #1009: The option `--min-reads` in CallDuplexConsensusReads is a pre-filter and not a post-filter
It is possible for consensus with read counts fewer than the value set by `--min-reads` to be output by the tool `CallDuplexConsensusReads`. The CLI parameter acts as a "pre-filter" of molecular groups prior to consensus calling. However, during the consensus calling process, additional reads may be filtered out if they do not conform to the most common alignment ([docs](https://github.com/fulcrumgenomics/fgbio/blob/6ca773317c64c989189e19d9fa554e306b67dd88/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala#L323-L337)). These situations should be rare but are counterintuitive to users.

I suggest we add a "post-filter" that removes any final duplex consensus that does not match the minimum read count values using the duplex consensus SAM tags.

---

To track the logic start with the consensus calling iterator which runs the function `consensusReadsFromSamRecords`:

https://github.com/fulcrumgenomics/fgbio/blob/6ca773317c64c989189e19d9fa554e306b67dd88/src/main/scala/com/fulcrumgenomics/umi/ConsensusCallingIterator.scala#L70-L79

Which is defined here and calls `consensusSamRecordsFromSamRecords`:

https://github.com/fulcrumgenomics/fgbio/blob/6ca773317c64c989189e19d9fa554e306b67dd88/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala#L286-L291

Which for duplex calling is defined here:

https://github.com/fulcrumgenomics/fgbio/blob/6ca773317c64c989189e19d9fa554e306b67dd88/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala#L180-L205

Importantly, the min-reads check occurs at the bottom of this function but prior to the duplex consensus calling process:

https://github.com/fulcrumgenomics/fgbio/blob/6ca773317c64c989189e19d9fa554e306b67dd88/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala#L196-L202

The minimum read parameters are define using `--min-reads` here:

https://github.com/fulcrumgenomics/fgbio/blob/6ca773317c64c989189e19d9fa554e306b67dd88/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala#L109

Then, for reads in a molecular group that pass the minimum read check they are sent to duplex consensus calling:

https://github.com/fulcrumgenomics/fgbio/blob/6ca773317c64c989189e19d9fa554e306b67dd88/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala#L244-L309

However, reads may be further removed before a consensus is actually formed here:

https://github.com/fulcrumgenomics/fgbio/blob/6ca773317c64c989189e19d9fa554e306b67dd88/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala#L273-L274

clintval:
Closes #1009

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
@@ -198,6 +198,9 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
         if (consensus.forall(duplexHasMinimumNumberOfReads)) {
           consensus
         } else {
+          // Even though we had enough reads going into consensus calling, we can
+          // lose some in the process (e.g. to CIGAR filtering), and end up with 
+          // consensus reads that will than fail the min-reads check
           rejectRecords(groups.flatten, FilterMinReads)
           Nil
         }
@@ -221,9 +224,8 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
     this.minTotalReads <= numXy + numYx && this.minXyReads <= numXy && this.minYxReads <= numYx
   }
 
-  /** Returns true if there enough are enough reads composing the consensus according to the minReads option. */
+  /** Returns true if there are enough reads composing the consensus according to the minReads option. */
   private def duplexHasMinimumNumberOfReads(consensus: SamRecord): Boolean = {
-    if (!Umis.isFgbioDuplexConsensus(consensus)) throw new IllegalArgumentException(s"Record is not a duplex consensus: $consensus")
     val (numXy: Int, numYx: Int) = {
       val numAb = consensus[Int](ConsensusTags.PerRead.AbRawReadCount)
       val numBa = consensus[Int](ConsensusTags.PerRead.BaRawReadCount)
--- a/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
@@ -194,7 +194,13 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
       val y = groups.lift(1).getOrElse(Seq.empty)
 
       if (hasMinimumNumberOfReads(x, y)) {
-        callDuplexConsensusRead(x, y)
+        val consensus = callDuplexConsensusRead(x, y)
+        if (consensus.forall(duplexHasMinimumNumberOfReads)) {
+          consensus
+        } else {
+          rejectRecords(groups.flatten, FilterMinReads)
+          Nil
+        }
       }
       else {
         rejectRecords(groups.flatten, FilterMinReads)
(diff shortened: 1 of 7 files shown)


src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -369,7 +369,7 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
   private def simplifyCigar(cigar: Cigar) = {
     import CigarOperator._
     if (cigar.forall(e => e.operator == M || e.operator == I || e.operator == D)) {
-      cigar
+      cigar.coalesce
```
clintval, line 372: Previously, the consensus caller would treat `5M5M` and `10M` as structurally different!

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -194,7 +194,13 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
       val y = groups.lift(1).getOrElse(Seq.empty)
 
       if (hasMinimumNumberOfReads(x, y)) {
-        callDuplexConsensusRead(x, y)
+        val consensus = callDuplexConsensusRead(x, y)
+        if (consensus.forall(duplexHasMinimumNumberOfReads)) {
+          consensus
+        } else {
+          rejectRecords(groups.flatten, FilterMinReads)
```
clintval, line 201: We reject the original reads in the group, and not the consensus.

src/test/scala/com/fulcrumgenomics/umi/DuplexConsensusCallerTest.scala
```diff
@@ -171,18 +171,43 @@ class DuplexConsensusCallerTest extends UnitSpec with OptionValues {
...
+
+    // Test for a non-coalesced but simplified cigar
+    val similar2 = builder.addPair(name="q6", start1=200, start2=100, strand1=Minus, strand2=Plus, bases1="CCCCCCCCCC", bases2="AAAAAAAAAA", attrs=Map(MI -> "foo/B"), cigar1 = "5M5M")
+    caller(minReads=Seq(3)).consensusReadsFromSamRecords(records ++ similar2) should have size 2 // the new read pair is considered in the final consensus
+    caller(minReads=Seq(2)).consensusReadsFromSamRecords(records ++ similar2) should have size 2
+
+    val dissimilar = builder.addPair(name="q6", start1=200, start2=100, strand1=Minus, strand2=Plus, bases1="CCCCCCCCCC", bases2="AAAAAAAAAA", attrs=Map(MI -> "foo/B"), cigar1 = "5M1D5M")
+    caller(minReads=Seq(3)).consensusReadsFromSamRecords(records ++ dissimilar) shouldBe empty // the new read pair is *not* considered in the final consensus
+    caller(minReads=Seq(2)).consensusReadsFromSamRecords(records ++ dissimilar) should have size 2
+
```
clintval, line 210:
```suggestion
```

src/test/scala/com/fulcrumgenomics/umi/UmisTest.scala
```diff
@@ -112,19 +113,34 @@ class UmisTest extends UnitSpec with OptionValues {
...
-        attrs=Map(ConsensusTags.PerRead.AbRawReadCount -> 10, ConsensusTags.PerRead.BaRawReadCount -> 10)
-    ).exists(Umis.isFgbioStyleConsensus) shouldBe true
+    builder.addFrag(start=10, attrs=Map(RawReadCount -> 10)).exists(Umis.isFgbioStyleConsensus) shouldBe true
+    builder.addFrag(start=10, attrs=Map(AbRawReadCount -> 10, BaRawReadCount -> 10)).exists(Umis.isFgbioStyleConsensus) shouldBe true
+  }
+
+  "Umis.isFgbioSimplexConsensus" should "return true for reads with simplex only consensus tags" in {
+    val builder = new SamBuilder(sort=Some(SamOrder.Coordinate), readLength=10, baseQuality=20)
+    builder.addFrag(start=10, attrs=Map(RawReadCount -> 10)).exists(Umis.isFgbioSimplexConsensus) shouldBe true
+
```
clintval, line 131:
```suggestion
```

tfenne (approved):
I took a brief look in `VanillaUmiConsensusCaller` and it doesn't seem to share this bug, right?  I.e. it already re-checks to see if it has enough reads after filtering to the most common alignment.

Feel free to merge after fixing up minor suggestions.

src/main/scala/com/fulcrumgenomics/umi/DuplexConsensusCaller.scala
```diff
@@ -215,6 +221,17 @@ class DuplexConsensusCaller(override val readNamePrefix: String,
     this.minTotalReads <= numXy + numYx && this.minXyReads <= numXy && this.minYxReads <= numYx
   }
 
+  /** Returns true if there enough are enough reads composing the consensus according to the minReads option. */
+  private def duplexHasMinimumNumberOfReads(consensus: SamRecord): Boolean = {
+    if (!Umis.isFgbioDuplexConsensus(consensus)) throw new IllegalArgumentException(s"Record is not a duplex consensus: $consensus")
```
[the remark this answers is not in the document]
clintval: Good point

clintval:
> I took a brief look in VanillaUmiConsensusCaller and it doesn't seem to share this bug, right? I.e. it already re-checks to see if it has enough reads after filtering to the most common alignment.

Yes, you are correct!

Documented the reasoning in the original issue too: https://github.com/fulcrumgenomics/fgbio/issues/1009#issuecomment-2316265005

We should be all set after this PR is merged.

PR #1021, opened 2024-12-16
feat: output umi grouping metrics

### LINKED ISSUE #1020: GroupReadsByUmi should emit an additional metrics file
There are some great metrics generated which may be lost in the log. It would be helpful to have them output to a metrics file.

Reads Accepted for grouping
non-PF reads
Filtered out due to mapping issues
Filtered out due to one or more Ns in UMIs
Filtered out due to UMIs that were too short

znorgaard:
Closes #1020 

This PR optionally emits a new metric file with read filtering statistics generated by GroupReadsByUmi.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -468,17 +468,17 @@ case class TagFamilySizeMetric(family_size: Int,
 
 /**
  * Metrics produced by `GroupReadsByUmi` to describe reads passed through UMI grouping
- * @param sam_records The number of SAM records accepted for grouping.
- * @param filtered_non_pf The number of non-PF SAM records.
- * @param filtered_poor_alignment The number of SAM records discarded for poor alignment.
- * @param filtered_ns_in_umi The number of SAM records discarded due to one or more Ns in the UMI.
- * @param filtered_umis_to_short The number of SAM records discarded due to a shorter than expected UMI.
+ * @param accepted_sam_records The number of SAM records accepted for grouping.
+ * @param discarded_non_pf The number of discarded non-PF SAM records.
+ * @param discarded_poor_alignment The number of SAM records discarded for poor alignment.
+ * @param discarded_ns_in_umi The number of SAM records discarded due to one or more Ns in the UMI.
+ * @param discarded_umis_to_short The number of SAM records discarded due to a shorter than expected UMI.
  */
-case class UmiGroupingMetric(sam_records: Long,
-                             filtered_non_pf: Long,
-                             filtered_poor_alignment: Long,
-                             filtered_ns_in_umi: Long,
-                             filtered_umis_to_short: Long) extends Metric
+case class UmiGroupingMetric(accepted_sam_records: Long,
+                             discarded_non_pf: Long,
+                             discarded_poor_alignment: Long,
+                             discarded_ns_in_umi: Long,
+                             discarded_umis_to_short: Long) extends Metric
 
 /** The strategies implemented by [[GroupReadsByUmi]] to identify reads from the same source molecule.*/
 sealed trait Strategy extends EnumEntry {
@@ -760,13 +760,13 @@ class GroupReadsByUmi
 
     // Write out UMI grouping metrics
     this.groupingMetrics.foreach { path =>
-        val groupingMetrics = UmiGroupingMetric(
-          sam_records             = kept,
-          filtered_non_pf         = filteredNonPf,
-          filtered_poor_alignment = filteredPoorAlignment,
-          filtered_ns_in_umi      = filteredNsInUmi,
-          filtered_umis_to_short  = filterUmisTooShort,
-        )
+      val groupingMetrics = UmiGroupingMetric(
+        accepted_sam_records     = kept,
+        discarded_non_pf         = filteredNonPf,
+        discarded_poor_alignment = filteredPoorAlignment,
+        discarded_ns_in_umi      = filteredNsInUmi,
+        discarded_umis_to_short  = filterUmisTooShort,
+      )
       Metric.write(path, groupingMetrics)
     }
   }
--- a/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
@@ -466,6 +466,20 @@ case class TagFamilySizeMetric(family_size: Int,
                                var fraction: Proportion = 0,
                                var fraction_gt_or_eq_family_size: Proportion = 0) extends Metric
 
+/**
+ * Metrics produced by `GroupReadsByUmi` to describe reads passed through UMI grouping
+ * @param reads The number of reads accepted for grouping
+ * @param filteredNonPf The number of non-PF reads
+ * @param filteredPoorAlignment The number of templates discarded for poor
+ * @param filteredNsInUmi The number of templates discarded due to one or more Ns in the UMI
+ * @param filterUmisTooShort The number of templates discarded due to a shorter than expected UMI
+ */
+case class UmiGroupingMetric(reads: Long,
+                             filteredNonPf: Long,
+                             filteredPoorAlignment: Long,
+                             filteredNsInUmi: Long,
+                             filterUmisTooShort: Long) extends Metric
+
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -742,6 +757,20 @@ class GroupReadsByUmi
...
+    case Some(p) =>
+      val groupingMetrics = UmiGroupingMetric(
+        reads                 = kept,
+        filteredNonPf         = filteredNonPf,
+        filteredPoorAlignment = filteredPoorAlignment,
+        filteredNsInUmi       = filteredNsInUmi,
+        filterUmisTooShort    = filterUmisTooShort,
+      )
+      Metric.write(p, groupingMetrics)
+    }
```
clintval, line 773:
```suggestion
    // Write out UMI grouping metrics
    this.groupingMetrics.foreach { path =>
      val groupingMetrics = UmiGroupingMetric(
        reads                 = kept,
        filteredNonPf         = filteredNonPf,
        filteredPoorAlignment = filteredPoorAlignment,
        filteredNsInUmi       = filteredNsInUmi,
        filterUmisTooShort    = filterUmisTooShort,
      )
      Metric.write(path, groupingMetrics)
    }
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -466,6 +466,20 @@ case class TagFamilySizeMetric(family_size: Int,
...
+ * @param filteredNonPf The number of non-PF reads
+ * @param filteredPoorAlignment The number of templates discarded for poor
+ * @param filteredNsInUmi The number of templates discarded due to one or more Ns in the UMI
+ * @param filterUmisTooShort The number of templates discarded due to a shorter than expected UMI
+ */
+case class UmiGroupingMetric(reads: Long,
+                             filteredNonPf: Long,
+                             filteredPoorAlignment: Long,
+                             filteredNsInUmi: Long,
+                             filterUmisTooShort: Long) extends Metric
```
clintval, line 481: Metrics are the style exception and should be `snake_case`:

```suggestion
case class UmiGroupingMetric(reads: Long,
                             filtered_non_pf: Long,
                             filtered_poor_alignment: Long,
                             filtered_ns_in_umi: Long,
                             filter_umis_too_short: Long) extends Metric
```

src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
```diff
@@ -267,6 +268,8 @@ class GroupReadsByUmiTest extends UnitSpec with OptionValues with PrivateMethodT
 
       hist.toFile.exists() shouldBe true
 
+      metrics.toFile.exists() shouldBe true
```
clintval, line 271: You should assert the contents of this file are equal to an expected Metric.

You can do this with `val expected: UmiGroupingMetric = ...` and a call to `Metric.read()`.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -466,6 +466,20 @@ case class TagFamilySizeMetric(family_size: Int,
                                var fraction: Proportion = 0,
                                var fraction_gt_or_eq_family_size: Proportion = 0) extends Metric
 
+/**
+ * Metrics produced by `GroupReadsByUmi` to describe reads passed through UMI grouping
+ * @param reads The number of reads accepted for grouping
+ * @param filteredNonPf The number of non-PF reads
+ * @param filteredPoorAlignment The number of templates discarded for poor
+ * @param filteredNsInUmi The number of templates discarded due to one or more Ns in the UMI
+ * @param filterUmisTooShort The number of templates discarded due to a shorter than expected UMI
```
clintval, line 475: Some of the descriptions might need updating.

`filteredPoorAlignment` has a sentence fragment.

I also notice all the other metric descriptions end in periods, but these do not.

And are reads _**reads**_, or read pairs, or are they SAM records?

znorgaard: They're SAM records. I was aligning with `logStats()`. If we're going to be explicit in the metrics file I think the logging should be updated as well.

clintval: The way the tool works I think the context for this would mean `SAM records == reads`.

So it's probably OK to keep it as **_reads_** since that should be more intuitive for users.

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -466,6 +466,20 @@ case class TagFamilySizeMetric(family_size: Int,
...
+ * @param filtered_non_pf The number of non-PF SAM records.
+ * @param filtered_poor_alignment The number of SAM records discarded for poor alignment.
+ * @param filtered_ns_in_umi The number of SAM records discarded due to one or more Ns in the UMI.
+ * @param filtered_umis_to_short The number of SAM records discarded due to a shorter than expected UMI.
+ */
+case class UmiGroupingMetric(sam_records: Long,
+                             filtered_non_pf: Long,
+                             filtered_poor_alignment: Long,
+                             filtered_ns_in_umi: Long,
+                             filtered_umis_to_short: Long) extends Metric
```
clintval, line 481: I'm wondering if we should have slightly easier to understand metric names:

```suggestion
case class UmiGroupingMetric(accepted_sam_records: Long,
                             discarded_non_pf: Long,
                             discarded_poor_alignment: Long,
                             discarded_ns_in_umi: Long,
                             discarded_umis_to_short: Long) extends Metric
```

src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala
```diff
@@ -742,14 +757,26 @@ class GroupReadsByUmi
...
+
+    // Write out UMI grouping metrics
+    this.groupingMetrics.foreach { path =>
+        val groupingMetrics = UmiGroupingMetric(
+          sam_records             = kept,
+          filtered_non_pf         = filteredNonPf,
+          filtered_poor_alignment = filteredPoorAlignment,
+          filtered_ns_in_umi      = filteredNsInUmi,
+          filtered_umis_to_short  = filterUmisTooShort,
+        )
```
clintval, line 769:
```suggestion
      val groupingMetrics = UmiGroupingMetric(
        sam_records             = kept,
        filtered_non_pf         = filteredNonPf,
        filtered_poor_alignment = filteredPoorAlignment,
        filtered_ns_in_umi      = filteredNsInUmi,
        filtered_umis_to_short  = filterUmisTooShort,
      )
```

src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
```diff
@@ -267,6 +269,10 @@ class GroupReadsByUmiTest extends UnitSpec with OptionValues with PrivateMethodT
 
       hist.toFile.exists() shouldBe true
 
+      // TODO: Create a more comprehensive test case
```
clintval, line 272:
```suggestion
      // TODO: Consider creating more unit tests that vary the following metric fields
```

clintval (approved):
A few final suggestions from me. Looks good!

@tfenne or @nh13 do you want to give a pass?

src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala
nh13: @znorgaard how about adding to existing tests?  E.g. https://github.com/fulcrumgenomics/fgbio/blob/e1dcbabb6ad8dadc825537ef22914b1872e85337/src/test/scala/com/fulcrumgenomics/umi/GroupReadsByUmiTest.scala#L426

znorgaard: @nh13, just add the metric generation and check to those other tests? 

That would work. I generally like to see a "complex" test case where non-zero values are generated for each metric entry at once. But, in this specific case we're pulling from existing variables, so individual tests might be good enough.

nh13 (approved):

Issue #1016, opened 2024-12-02
Soft clip disappeared after `CallMolecularConsensusReads`


jonghwanyoon:
hello

When calling Molecular consensus reads using fgbio's `CallMolecularConsensusReads`, the soft clip disappears. 
I went through `CorrectUmis`, `GroupReadsByUmi`, `CallMolecularConsensusReads`, `alignment` on a bam file mapped with raw fastq, and noticed that the soft clips disappeared on important **singleton** reads. 

Given the presence of soft clips in the bam file resulting from `GroupReadsByUmi`, it seems that the issue is in `CallMolecularConsensusReads`. 

I only gave `CallMolecularConsensusReads` the java option and the `--min-reads=1` parameter.
There doesn't seem to be any parameter associated with `CallMolecularConsensusReads` (I could be wrong)
This is the same in `2.1.1`, when collapsing between chromosomes became possible, and in `2.4.0`, the current latest version.

 Although it is a singleton read, it is an important read that represents a clinically meaningful fusion, so I don't want the soft clips to disappear. 
May I request your opinion on this issue?

Thank you!

![image](https://github.com/user-attachments/assets/50dd4279-a8cd-4b0d-bf31-faff604d4e78)

nh13:
@jonghwanyoon I've added a test in #1026 to show where this occurs in the consensus calling process.    In particular, it happens here:
https://github.com/fulcrumgenomics/fgbio/blob/f93fdfbb427da5f5d60304de5761b19ea8209b33/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala#L255

A read is clipped/trimmed based on if it extends past its mate.  In the case that the same end is clipped on both reads, perhaps we shouldn't clip, we'll have to think about this further.

In the meantime, if you'd like to sponsor this work, feel free to reach out to us.

nh13:
@jonghwanyoon I think I found an easy fix, will need a review from the team

PR #1026, opened 2025-02-06
Do not trim reads when both ends are clipped in consensus calling

nh13:
Closes https://github.com/fulcrumgenomics/fgbio/issues/1016

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
@@ -188,9 +188,7 @@ class ClipBam
 
     val (numExtendingPastMateStartReadOne, numExtendingPastMateStartReadTwo) = {
       if (clipBasesPastMate && r1.isFrPair) {
-        val clip1 = this.clipper.clipExtendingPastMateEnd(rec=r1, mateEnd=r2.end)
-        val clip2 = this.clipper.clipExtendingPastMateEnd(rec=r2, mateEnd=r1.end)
-        (clip1, clip2)
+        this.clipper.clipExtendingPastMateEnds(rec=r1, mate=r2)
       }
       else (0, 0)
     }
--- a/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
@@ -327,20 +327,6 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean) 
 
   /** Returns the number of bases extending past the mate end for FR pairs including any soft-clipped bases, zero otherwise.
     *
-    * If the mate-cigar (MC SAM tag) cigar is present, then the _unclipped_ mate start and end will be used (i.e.
-    * `mateUnclippedStart` and `mateUnclippedEnd`).  Otherwise,the mate start will be `mateStart` and the mate end will
-    * be computed from this records start plus the insert size (minus one).
-    *
-    * @param rec the record to examine
-    */
-  def numBasesExtendingPastMate(rec: SamRecord): Int = {
-    val mateUnclippedStart = rec.mateUnclippedStart.getOrElse(rec.mateStart)
-    val mateUnclippedEnd   = rec.mateUnclippedEnd.getOrElse(rec.start + abs(rec.insertSize) - 1)
-    numBasesExtendingPastMate(rec=rec, mateUnclippedStart=mateUnclippedStart, mateUnclippedEnd=mateUnclippedEnd)
-  }
-
-  /** Returns the number of bases extending past the mate end for FR pairs including any soft-clipped bases, zero otherwise.
-    *
     * @param rec the record to examine
     * @param mateUnclippedStart the smallest mapped genomic coordinate considering clipping
     * @param mateUnclippedEnd the largest mapped genomic coordinate considering clipping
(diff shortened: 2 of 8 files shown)


src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -344,18 +344,22 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
     */
   def numBasesExtendingPastMate(rec: SamRecord, mateEnd: Int): Int = {
     if (!rec.isFrPair) 0 else rec.positiveStrand match {
-      case true if rec.end >= mateEnd => 
+      case true if rec.end >= mateEnd =>
+        // account for any soft-clipping on the same side of the template
+        val trailingClippedBases = if (rec.end == mateEnd) Math.min(rec.cigar.trailingClippedBases, rec.mateCigar.get.trailingClippedBases) else 0
```
nh13, line 349: I think this is wrong, mate unclipped end should be used...

src/main/scala/com/fulcrumgenomics/bam/ClipBam.scala
```diff
@@ -188,9 +188,7 @@ class ClipBam
 
     val (numExtendingPastMateStartReadOne, numExtendingPastMateStartReadTwo) = {
       if (clipBasesPastMate && r1.isFrPair) {
-        val clip1 = this.clipper.clipExtendingPastMateEnd(rec=r1, mateEnd=r2.end)
-        val clip2 = this.clipper.clipExtendingPastMateEnd(rec=r2, mateEnd=r1.end)
-        (clip1, clip2)
+        this.clipper.clipExtendingPastMateEnds(rec=r1, mate=r2)
```
clintval, line 191: I think this new change is OK and the docs **_do not_** need updating:


```console
--clip-bases-past-mate
Clip reads in FR pairs that sequence past the far end of their mate.
```

Since "far end" is unspecified with regard to the mate's soft-clipping or not.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -379,20 +381,22 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
     * @return the additional number of bases clipped (3' end in sequencing order)
     */
   def clipExtendingPastMateEnd(rec: SamRecord): Int = {
-    val mateEnd = rec.mateEnd.getOrElse(rec.start + abs(rec.insertSize) - 1)
-    clipExtendingPastMateEnd(rec=rec, mateEnd=mateEnd)
+    val mateUnclippedStart = rec.mateUnclippedStart.getOrElse(rec.mateStart)
+    val mateUnclippedEnd   = rec.mateUnclippedEnd.getOrElse(rec.start + abs(rec.insertSize) - 1)
```
clintval, line 385: This function's documentation describes how the mate end is calculated but needs to also have a similar description for how we are calculating the mate start. 

Also, given how important the `MC` tag is to this subtle behavior change, I'm thinking we should fixup the `ClipBam` tool help message to say that the `MC` tag really should be present (and how to add it if it does not exist!).

Also, there are no unit tests covering this functionality according to codecov. Could you add some?

clintval: EDIT: on second thought, since there are no unit tests covering this function I suspect this function is not really used. Can we delete it? I am worrisome it will be used accidentally without the `MC` tag present and the data will not be optimal. I have a similar concern about the `numBasesExtendingPastMate` that infers the mate's start/end (biasing towards the unclipped positions).

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -333,29 +333,31 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
     * @param rec the record to examine
     */
   def numBasesExtendingPastMate(rec: SamRecord): Int = {
-    val mateEnd = rec.mateEnd.getOrElse(rec.start + abs(rec.insertSize) - 1)
-    numBasesExtendingPastMate(rec=rec, mateEnd=mateEnd)
+    val mateUnclippedStart = rec.mateUnclippedStart.getOrElse(rec.mateStart)
+    val mateUnclippedEnd   = rec.mateUnclippedEnd.getOrElse(rec.start + abs(rec.insertSize) - 1)
```
clintval, line 337: In both places in this PR where you access the unclipped mate start/end positions I think you want to not consider any hard clip sequences and _**only**_ consider soft clip sequences. E.g. we don't want a read's soft-clip sequence to extend past, and into, the mates hard clip sequence.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -252,7 +252,13 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
       var index = if (!rec.isFrPair) trimToLength - 1 else {
         // Get the number of mapped bases to clip that maps beyond the mate's end, including any soft-clipped bases. Use
         // that to compute where in the read to keep.
-        val clipPosition = rec.length - this.clipper.numBasesExtendingPastMate(rec=rec)
+        val mateUnclippedStart = rec.mateUnclippedStart.getOrElse(throw new IllegalStateException(f"Mate cigar (MC SAM tag) needed for read: ${rec.name}"))
+        val mateUnclippedEnd   = rec.mateUnclippedEnd.getOrElse(throw new IllegalStateException(f"Mate cigar (MC SAM tag) needed for read: ${rec.name}"))
```
clintval, line 256: Given the `UmiConsensusCaller` doesn't really have access to the mate here, I think it makes sense to raise an exception as you have done. Mind adding this requirement to the scaladoc with a `@raises IllegalArgumentException`?

The exceptions should be for an illegal argument (correct type, bad value):

```suggestion
        val mateUnclippedStart = rec.mateUnclippedStart.getOrElse(throw new IllegalArgumentException(f"Mate cigar (MC SAM tag) needed for read: ${rec.name}"))
        val mateUnclippedEnd   = rec.mateUnclippedEnd.getOrElse(throw new IllegalArgumentException(f"Mate cigar (MC SAM tag) needed for read: ${rec.name}"))
```

clintval (approved):
A few suggestions (docs / variable naming / test coverage) but nothing to block you from merging. This looks better! Thanks also for opening the hard-clip issue:

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -276,6 +282,26 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
     }
   }
 
+  /**
+    * Adds mate-cigars to records without mate cigars.
+    */
+  private def updateMateCigars(recs: Seq[SamRecord]): Unit = {
```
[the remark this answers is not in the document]
nh13: Added

PR #1028, opened 2025-02-14
feat: SamSource.query should handle out of range queries

### LINKED ISSUE #943: SamReader.query() should handle when coordinates are given < 1 or > end of contig
The underlying SAMFileReader throws exceptions back if you ask to query off the ends of contigs, rather than just gracefully capping the query at the ends of the contig.  I think it would be nice to have SamReader min/max the query coordinates with 1/contig_length on the way through to prevent these kinds of exceptions.

This often comes up when running queries with `position - flank` to `position + flank` and forgetting to limit the positions.

nh13:
Fixes #943

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamSource.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamSource.scala
@@ -92,7 +92,8 @@ class SamSource private(private val reader: SamReader) extends View[SamRecord] w
   override def iterator: SamIterator = new SamIterator(reader.getFileHeader, reader.iterator())
 
   private def newQueryInterval(region: Locatable): QueryInterval = {
-    val contig = dict(region.getContig)
+    val contig = dict.get(region.getContig).getOrElse(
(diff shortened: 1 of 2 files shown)


nh13:
@clintval would you be able to take a quick peek?

src/main/scala/com/fulcrumgenomics/bam/api/SamSource.scala
```diff
@@ -91,9 +91,16 @@ class SamSource private(private val reader: SamReader) extends View[SamRecord] w
   /** Returns an iterator over all the records in the source. */
   override def iterator: SamIterator = new SamIterator(reader.getFileHeader, reader.iterator())
 
+  private def newQueryInterval(region: Locatable): QueryInterval = {
+    val contig = dict(region.getContig)
```
tfenne, line 95:
```suggestion
    val contig = dict.get(region.getContig).getOrElse(
      throw new NoSuchElementException("Contig '${region.getContig}' not in SAM/BAM header.")
```

Or pick your favorite exception to throw .. just makes the error nicer if you query a contig that's not in the BAM header.

tfenne (approved):

PR #1032, opened 2025-02-20
feat: a default base quality may be provided in consensus callers whe…

### LINKED ISSUE #1031: How should we handle missing base qualities in the consensus caller?
Consider a read where the base qualities are not stored, we could:

1.  Except, with a better message than we currently have (I believe we'll get an index out of bounds exception).
2. Add an option to the consensus callers to use a default base quality when base qualities are missing (e.g. Q30).
3. Others?

The issue with (1) is that would force folks to filter out such reads if they want to get the consensus caller to work.  The issue with (2 )is that someone produces a BAM file accidentally with missing qualities and fgbio handles it gracefully, and they would have preferred to have known about it (no one reads the logs).

nh13:
…n base qualities are missing.

Fixes: #1031

This has the following improvements:

1.  A better error message is displayed, with the name of the read.  Previously, a generic `ArrayIndexOutOfBoundsException` with on mention of which read.
2. If the `--missing-base-quality` is not given, then the exception occurs (no change in current behavior)
3. If the `--missing-base-quality` is given, then it is used (gracefully)

My only question is if we should have a non-None value by default.  See #1031 for a discussion on why that might be a bad idea.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
@@ -113,6 +113,7 @@ class CallDuplexConsensusReads
  val maxReadsPerStrand: Option[Int] = None,
  @arg(doc="The number of threads to use while consensus calling.") val threads: Int = 1,
  @arg(doc="Consensus call overlapping bases in mapped paired end reads") val consensusCallOverlappingBases: Boolean = true,
+ @arg(doc = "The phred-scaled base quality to use when base qualities are missing in the input") val missingBaseQuality: Option[PhredScore] = None,
 ) extends FgBioTool with LazyLogging {
 
   Io.assertReadable(input)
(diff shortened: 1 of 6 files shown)


src/main/scala/com/fulcrumgenomics/umi/CallDuplexConsensusReads.scala
```diff
@@ -113,6 +113,7 @@ class CallDuplexConsensusReads
  val maxReadsPerStrand: Option[Int] = None,
  @arg(doc="The number of threads to use while consensus calling.") val threads: Int = 1,
  @arg(doc="Consensus call overlapping bases in mapped paired end reads") val consensusCallOverlappingBases: Boolean = true,
+ @arg(doc = "The phred-scaled base quality to use when base qualities are missing in the input") val missingBaseQuality: Option[PhredScore] = None,
```
nh13, line 116:
```suggestion
 @arg(doc="The phred-scaled base quality to use when base qualities are missing in the input") val missingBaseQuality: Option[PhredScore] = None,
```

PR #1033, opened 2025-02-21
feat: FilterBam optionally writes to a rejects file

### LINKED ISSUE #971: feature request: add option to FilterBam to write filtered reads to separate file
Self-explanatory.  Useful for example for preserving unmapped reads, operating on aligned reads, and then adding the unmapped reads back to create a "lossless" bam.  Command could be run serially if finer control is desired, e.g. permanently discard secondary alignments and preserve unmapped reads only.

nh13:
Fixes #971

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
@@ -85,7 +85,6 @@ class FilterBam
   }
 
   override def execute(): Unit = {
-    val progress = ProgressLogger(logger, verb="written", unit=5e6.toInt)
     val in       = SamSource(input)
     val iterator = buildInputIterator(in, intervals)
     val out      = SamWriter(output, in.header)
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -87,6 +89,7 @@ class FilterBam
     val in       = SamSource(input)
```
clintval, line 89: I noticed on line 88 we have an unused `ProgressLogger`! Mind deleting, or wiring it in?

src/main/scala/com/fulcrumgenomics/bam/FilterBam.scala
```diff
@@ -98,6 +101,7 @@ class FilterBam
           minMappedBases.exists(count => countMappedBases(rec) < count)
 
         if (throwOut) {
+          rejects.foreach { writer => writer += rec }
```
clintval, line 104: This reads a little easier:

```suggestion
          rejects.foreach(writer.write)
```

clintval (approved):

PR #1040, opened 2025-03-05
feat: improve error message in consensus calling when base qualities are missing

nh13:
Motivated by:

- https://github.com/fulcrumgenomics/fgbio/issues/1031#issuecomment-2701736760

### DIFF
--- a/src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
@@ -26,6 +26,7 @@
 package com.fulcrumgenomics.umi
 
 import com.fulcrumgenomics.alignment.Cigar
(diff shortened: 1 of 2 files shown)


src/test/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCallerTest.scala
```diff
@@ -617,6 +618,12 @@ class VanillaUmiConsensusCallerTest extends UnitSpec with OptionValues {
     s2.cigar.toString shouldBe "40M20I40M"
   }
 
+  it should "except when the reads do not have base qualities and a minimum base quality is not given" in {
```
clintval, line 621: The exception will occur regardless of minimum base quality:

```suggestion
  it should "except when the reads do not have base qualities" in {
```

clintval (approved):

PR #1042, opened 2025-03-12
Add support for collections in Metric.

nh13:
Collections will be comma-separated lists, so values in the list should not contain commas.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/Metric.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/Metric.scala
@@ -25,6 +25,7 @@
 
 package com.fulcrumgenomics.util
 
+import com.fulcrumgenomics.commons.CommonsDef._
 import com.fulcrumgenomics.commons.io.{Writer => CommonsWriter}
 import com.fulcrumgenomics.commons.reflect.{ReflectionUtil, ReflectiveBuilder}
 import com.fulcrumgenomics.commons.util.DelimitedDataParser
@@ -42,6 +43,7 @@ import scala.reflect.runtime.{universe => ru}
 object Metric {
   val Delimiter: Char = '\t'
   val DelimiterAsString: String = s"$Delimiter"
+  val DefaultCollectionDelimiter: Char = ','
 
   /** A typedef for [[scala.Long]] to be used when representing counts. */
   type Count = Long
@@ -163,9 +165,12 @@ object Metric {
 /**
   * Base trait for metrics.
   *
-  * All classes extending this class should be a case class.  By convention, all fields should be lower case with
+  * All classes extending this class should be a case class.  By convention, all fields should be lowercase with
   * words separated by underscores.
-  */
+ *
+ * To use a custom delimiter for collection values, define a value named `CollectionDelimiter` in the
+ * companion object of the Metric that gives a `Char` delimiter.
+ */
 trait Metric extends Product with Iterable[(String,String)] {
   /** Get the names of the arguments in the order they were defined. */
   def names: Seq[String] = Metric.names(getClass)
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -234,6 +258,9 @@ trait Metric extends Product with Iterable[(String,String)] {
   /** Gets an iterator over the fields of this metric in the order they were defined.  Returns tuples of names and values */
   override def iterator: Iterator[(String,String)] = this.names.zip(this.values).iterator
 
+  /** The delimiter for collection types. */
+  protected def collectionDelimiter: Char = Metric.DefaultCollectionDelimiter
```
clintval, line 262: Is this so that the delimiter can be overridden by child classes? If so, wouldn't you also want to make this the default for the functions above that perform iteration and reading instead of the static `DefaultCollectionDelimiter`? Is that possible?

clintval (approved):
One question about a class-level default being used for other function signatures, but otherwise this looks good to me. Thank you!

src/main/scala/com/fulcrumgenomics/util/Metric.scala
```diff
@@ -163,9 +165,12 @@ object Metric {
 /**
   * Base trait for metrics.
   *
-  * All classes extending this class should be a case class.  By convention, all fields should be lower case with
+  * All classes extending this class should be a case class.  By convention, all fields should be lowercase with
   * words separated by underscores.
-  */
+ *
+ * To use a custom delimiter for collection values, define a field named `CollectionDelimiter` in the
+ * companion object of the metric that gives the `Char` delimiter.
```
clintval, line 172:
```suggestion
 * To use a custom delimiter for collection values, define a value named `CollectionDelimiter` in the
 * companion object of the Metric that gives a `Char` delimiter.
```

src/main/scala/com/fulcrumgenomics/illumina/RunInfo.scala
```diff
@@ -50,7 +50,11 @@ case class RunInfo
   run_date: Iso8601Date,
   read_structure: ReadStructure,
   num_lanes: Int
-) extends Metric
+) extends Metric {
+  override protected def formatValue(value: Any): String = {
+    if (value.isInstanceOf[ReadStructure]) value.toString else super.formatValue(value)
+  }
+}
```
clintval, line 57: How did this change become necessary?

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -28,18 +28,47 @@ package com.fulcrumgenomics.util
...
+      val typeSignature   = instanceMirror.symbol.typeSignature
+      val fieldSymbol     = typeSignature.member(TermName("CollectionDelimiter"))
+      if (fieldSymbol == NoSymbol) {
+        DefaultCollectionDelimiter
+      } else {
+        try {
+          instanceMirror.reflectField(fieldSymbol.asMethod).get.asInstanceOf[Char]
+        } catch {
+          case ex: ClassCastException =>
+            throw new ClassCastException(f"`CollectDelimiter` in the `${companionObject.name}` companion object must be of type `Char`: $ex")
```
nh13, line 64:
```suggestion
            throw new ClassCastException(f"`CollectionDelimiter ` in the `${companionObject.name}` companion object must be of type `Char`: $ex")
```

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -28,18 +28,47 @@ package com.fulcrumgenomics.util
...
 
 /** Exception for errors building metrics. */
 case class MetricBuilderException
 (private val message: Option[String] = None, private val cause: Option[Throwable] = None)
 extends Exception(message.getOrElse(""), cause.orNull)
 
 
+object MetricBuilder {
+  /** Gets the collection delimiter for the metric class.  Looks for a field in the
+   * companion object with name `CollectionDelimiter`. */
```
clintval, line 46:
```suggestion
  /** Gets the collection delimiter for the Metric class.  Looks for a value in the
   * companion object with name `CollectionDelimiter` and return type `Char`. */
```

src/main/scala/com/fulcrumgenomics/util/MetricBuilder.scala
```diff
@@ -53,6 +82,7 @@ class MetricBuilder[T <: Metric](source: Option[String] = None)(implicit tt: ru.
   private val reflectiveBuilder = new ReflectiveBuilder(clazz)
   private val names             = Metric.names[T]
   private val namesSet          = names.toSet
+  private val delim             = MetricBuilder.findCollectionDelimiter(clazz)
```
clintval, line 85:
```suggestion
  private lazy val delim        = MetricBuilder.findCollectionDelimiter(clazz)
```

clintval (approved):

src/main/scala/com/fulcrumgenomics/illumina/RunInfo.scala
nh13: `ReadStructure` is a collection type, so the by default, the read segments will be comma-delimited (e.g. `8M,100T`) whereas we just want it represented as a string (e.g. `8M100T`).

Issue #1043, opened 2025-03-18
Automate building metric docs


nh13:
Based on a [previous discussion,](https://github.com/fulcrumgenomics/fgbio/pull/811#pullrequestreview-2665806363) I would like three PRs:

1. [x] On any PR, we test that docs and metrics compile without error, and are available (as artifacts) so folks can review them if needed.  #1044 
2. [x] On any commit to main, the docs and metrics are built and published to the gh-pages branch as the latest "develop" version. To avoid race conditions, we should set concurrency. #1046
3. [ ] On any tagged release, we publish the documentation and metrics as per the wiki. #1046

nh13:
It's done!

PR #1044, opened 2025-03-18
ci: add preview for gh-pages to each PR

nh13:
1/3 for #1043

** for the reviewers **

1. most important of all, is that _every commit to a PR will make a corresponding commit to the `gh-pages` branch_.  Once the PR is closed (merged or otherwise), the preview sub-folder on `gh-pages` will be removed, but the `git` history will include all the commits adding and removing the previews.
1. we set concurrency on the unit tests to be at the branch level, so that we do not end up having race conditions when uploading commits in short succession
1. we use `sdkman` to install java and scala as other actions do not provide support for Java 8 and scaladoc.  In the tests workflow we use the `olafurpg/setup-scala` [which doesn't support intalling scaladoc](https://github.com/olafurpg/setup-scala/issues/47).  The recommended solution of `coursier` doesn't have the java version (8) I am looking for.
1. we use `bash -l {0}` as shell so the profile is reloaded each time (needed for `sdkman`)

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -2,13 +2,18 @@ name: unit tests
 
 on: [push, pull_request]
 
+concurrency:
+  group: docs-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   test:
+    name: Unit Tests
     runs-on: ubuntu-24.04
     environment: github-actions
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/unittests.yaml
```diff
@@ -59,3 +64,87 @@ jobs:
       - name: Upload to Sonatype
         run: |
           sbt +publish
+  docs:
```
nh13, line 67: I may split this out to it's own workflow later, but keeping it here for now.

nh13: I split it out in https://github.com/fulcrumgenomics/fgbio/pull/1046

PR #1047, opened 2025-04-17
Use strict compiler flags and turn on inlining

clintval:
I have been chipping away at this for almost a year, and have finally finished!

## Inlining

Inlining in [Scala 2.x is off by default](https://www.scala-lang.org/api/2.13.4/scala/inline.html)!

> Note that by default, the Scala optimizer is disabled and no callsites are inlined. See -opt:help for information on how to enable the optimizer and inliner.
>
> When inlining is enabled, the inliner will always try to inline methods or callsites annotated @inline (under the condition that inlining from the defining class is allowed, see -opt-inline-from:help). If inlining is not possible, for example because the method is not final, an optimizer warning will be issued. See -opt-warnings:help for details.

In order to take advantage of inlining we must turn on the optimizer and ask it to inline code using Scala compiler flags. I have also added some of my favorite inlining-related compiler flags so we get log messages when inlining happens and warnings when an `@inline` annotation is used but the compiler could not perform the optimization.

## Linting and Bugs

I've long wanted stricter compiler flags in `fgbio` so that all warnings are addressed and no future warnings are possible to add without manual override (e.g. with a conscious `@nowarn`). Turning on my preferred collection of compiler flags revealed multiple minor bugs and faulty unit tests.

## Implicit Type Coercion and Narrowing

By default the Scala compiler will narrow or widen numeric types (e.g. a `Byte` to `Int`). I want to disallow this behavior because it can create hidden bugs such as when an `Int` is used where a `Byte` is required and under/over flow occurs. Unintentional loss of precision can also be a consequence.

Luckily for us, I did not spot any bugs created by automatic narrow or widening.

## Impact

1. Inlining should result in a speed enhancement for many aspects of `fgbio`
2. All open PRs will need to rebase and address newly found compiler warnings
3. No tool or public API regressions should have been made with this PR
4. In most cases I was able to solve bugs, however, in 2 cases I couldn't (yet!)


src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -642,6 +639,7 @@ private[vcf] object PhaseCigar {
       case _ => true
     }.flatMap { case (t, c) => // collect metrics but only keep sites where either variant context (i.e. truth or call) is phased.
       val ctxs                  = Seq(t, c).flatten // NB: can be 1 or 2 contexts here
+      @unused // TODO: should this be unused or should it be used to set something upon the metric?
```
clintval, line 642: There is a variable here that _**looks**_ like it should be used to update the metric output of `AssessPhasing` but I'm uncertain about the expected implementation due to my unfamiliarity with this tool.

src/main/scala/com/fulcrumgenomics/vcf/DownsampleVcf.scala
```diff
@@ -99,7 +98,7 @@ object DownsampleVcf extends LazyLogging {
   def downsampleAndRegenotype(gt: Genotype, proportion: Double, random: Random, epsilon: Double): Genotype = {
     val oldAds = gt.getOrElse[IndexedSeq[Int]]("AD", throw new Exception(s"AD tag not found for sample ${gt.sample}"))
     val newAds = downsampleADs(oldAds, proportion, random)
-    val likelihoods = Likelihoods(newAds)
+    val likelihoods = Likelihoods(newAds, epsilon = epsilon)
```
clintval, line 101: This was a "bug" whereby `epsilon` was not passed down the call stack so only the default was being used.

src/test/scala/com/fulcrumgenomics/alignment/MatrixTest.scala
```diff
@@ -72,12 +72,12 @@ class MatrixTest extends UnitSpec {
...
       """
-        |0  1
-        |0  2
-        |0  3
-      """.stripMargin.trim.replaceAll(" +", "\t")
+        |0 0
+        |0 1
+        |0 2
+      """.stripMargin.trim.replaceAll(" ", "\t")
+    m.toString shouldBe expected
```
clintval, line 81: This test was never testing anything but now that it is, I had to change the "expected" test case in order to get it to pass.

src/test/scala/com/fulcrumgenomics/bam/FindSwitchbackReadsTest.scala
```diff
@@ -101,7 +101,7 @@ class FindSwitchbackReadsTest extends UnitSpec with OptionValues {
     noClipHit shouldBe None
 
     val fourClip = Template(builder.addFrag(contig=0, start=71, cigar="4S46M", bases="ccagCTGGACTCCTTCCACTCGCGGGTCGTCTCCAGAGCTCAGAAAATGA").iterator)
-    val fourClipHit = FindSwitchbackReads.findReadBasedSwitchback(refMap, noClip, minLength=5, maxOffset=20, maxErrorRate=0.1)
+    val fourClipHit = FindSwitchbackReads.findReadBasedSwitchback(refMap, fourClip, minLength=5, maxOffset=20, maxErrorRate=0.1)
```
clintval, line 104: This was a bugged test, but now is testing the right thing.

src/test/scala/com/fulcrumgenomics/basecalling/ExtractBasecallingParamsForPicardTest.scala
```diff
@@ -98,24 +98,6 @@ class ExtractBasecallingParamsForPicardTest extends UnitSpec with ErrorLogLevel
...
-      |20000101-EXPID-4,Sample_Name_4,,,GATTACAG,GATTACAGA,I7_4,GATTACAACGT,Sample_Project_4,Description_4
-      |20000101-EXPID-5,Sample_Name_5,,,GATTACAG,GATTACAGA,I7_5,GATTACAACGT,Sample_Project_5,Description_5
-      |20000101-EXPID-6,Sample_Name_6,,,GATTACAG,GATTACAGA,I7_6,GATTACAACGT,Sample_Project_6,Description_6
-      |20000101-EXPID-7,Sample_Name_7,,,GATTACAG,GATTACAGA,I7_7,GATTACAACGT,Sample_Project_7,Description_7
-      |20000101-EXPID-8,Sample_Name_8,,,GATTACAG,GATTACAGA,I7_8,GATTACAACGT,Sample_Project_8,Description_8
-      |20000101-EXPID-9,Sample_Name_9,,,GATTACAG,GATTACAGA,I7_9,GATTACAACGT,Sample_Project_9,Description_9
-      |20000101-EXPID-10,Sample_Name_10,,,GATTACAG,GATTACAGA,I7_10,GATTACAACGT,Sample_Project_10,Description_10
-      |20000101-EXPID-11,Sample_Name_11,,,GATTACAG,GATTACAGA,I7_11,GATTACAACGT,Sample_Project_11,Description_11
-      |20000101-EXPID-12,Sample_Name_12,,,GATTACAG,GATTACAGA,I7_12,GATTACAACGT,Sample_Project_12,Description_12"""
-      .stripMargin.split("\n").toIndexedSeq
```
clintval, line 116: This was never used, but maybe should have been?

src/test/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmisTest.scala
```diff
@@ -101,21 +101,24 @@ class AnnotateBamWithUmisTest extends UnitSpec {
...
+//    val out      = makeTempFile("with_umis.", ".bam")
+//    val longFq   = makeTempFile(s"extra_umis.", ".fq.gz")
+//    val lines  = Io.readLines(fq) ++ Seq("@not_a_flowcell:1:1101:10060:3200/2 2:N:0:19","GATCTTGG","+","-,86,,;:")
+//    Io.writeLines(longFq, lines)
+//    val annotator = new AnnotateBamWithUmis(input=sam, fastq=Seq(longFq), output=out, attribute=umiTag, sorted=true)
+//    annotator.execute()
+//    SamSource(out).foreach(rec => {
+//      rec[String](umiTag) shouldBe rec.basesString.substring(0,8)
+//    })
+//  }
```
clintval, line 121: The test that was here was a copy-paste of the previous test, but without actually executing the tool! What this meant was that there were zero reads so the `.foreach()` asserted nothing. Now that the tool runs, the test fails because it is not setup correctly. I'm inclined to open an issue and someone with a future PR could fix this test.

src/test/scala/com/fulcrumgenomics/util/UpdateDelimitedFileContigNamesTest.scala
```diff
@@ -138,18 +138,20 @@ class UpdateDelimitedFileContigNamesTest extends UnitSpec {
...
+      val expected = actual.take(outputFirstNumLines) ++ allModified.drop(outputFirstNumLines)
       runTest(
         delimiter = ',',
         columns   = Seq(0),
         actual    = actual,
         expected  = expected,
-        comment   = "$"
+        comment   = "$",
+        outputFirstNumLines = outputFirstNumLines,
       )
```
clintval, line 155: This was a bugged test but is now working correctly.

src/test/scala/com/fulcrumgenomics/vcf/AssessPhasingTest.scala
```diff
@@ -205,6 +207,10 @@ class AssessPhasingTest extends ErrorLogLevel {
     val header = new SAMFileHeader
     header.setSequenceDictionary(Header.getSequenceDictionary)
     val intervalList = new IntervalList(header)
+    // TODO: BUG!!! These variables are never passed down into the interval list.
+    locally { val _ = start }
+    locally { val _ = end }
+    // TODO: BUG!!! This is meant to be start and end, but that is not happening.
```
clintval, line 213: > [!WARNING]
> This concerns me. It is a bug in the test. When you replace `start` and `end` in the interval that is added to the interval instead of the default `1` and `10`, many tests start to fail. I don't know if the implementation of `AssessPhasing` is wrong or if the tests that were written were copy-pasted from the outcome of the bugged test info.

I will open an issue so someone can solve in a future PR.

src/test/scala/com/fulcrumgenomics/vcf/DownsampleVcfTest.scala
```diff
@@ -222,15 +221,17 @@ class DownsampleVcfTest extends UnitSpec {
...
       (IndexedSeq(0, 1), IndexedSeq(e, 0.5, 1 - e)),
       (IndexedSeq(1, 1), IndexedSeq((1 - e) * e, 0.25, (1 - e) * e)),
       (IndexedSeq(2, 0), IndexedSeq(math.pow((1 - e), 2), 0.25, math.pow(e, 2))),
     )
-    cases.foreach { case (input, output) =>
+    cases.foreach { case (input, _) =>
       val biallelic = Likelihoods(2, DownsampleVcf.Likelihoods.biallelic(input(0), input(1), e))
       val generalized = Likelihoods(2, DownsampleVcf.Likelihoods.generalized(input, e))
+      biallelic.pls should contain theSameElementsInOrderAs biallelic.pls
       biallelic.pls should contain theSameElementsInOrderAs generalized.pls
```
clintval, line 235: "bilallelic" was never tested so I added that. But what I don't understand is what the second item in the seq was meant for. I think this test is not working correctly but I lack the familiarity with this code to know how to fix it.

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
clintval, line 1: The warnings from this file concerned me because a character can numeric widen to an integer which is an allowable type for `lastIndexOf` and `indexOf`. However, what I think is meant with this code is that the character is to be referred to as a string. Either way, the implementation may be the same but we should pick a coercion so it is clear. Perhaps `.toInt()` would have been more performant than the calls to `.toString()`?

.travis.install-git-lfs.sh
clintval, line 1: This file is no longer needed.

CODE_OF_CONDUCT.md
clintval, line 1: We no longer need this because we have an org-level file here:

- https://github.com/fulcrumgenomics/.github/blob/main/CODE_OF_CONDUCT.md

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
clintval, line 1: The warnings from this file concerned me because a character can numeric widen to an integer which is an allowable type for lastIndexOf and indexOf. However, what I think is meant with this code is that the character is to be referred to as a string. Either way, the implementation may be the same but we should pick a coercion so it is clear. Perhaps .toInt() would have been more performant than the calls to .toString()?

src/main/scala/com/fulcrumgenomics/personal/tfenne/SummarizeGff.scala
clintval, line 1: Technically this tool has an API break but it is in the "personal" space.

src/test/scala/com/fulcrumgenomics/vcf/DownsampleVcfTest.scala
nh13, line 235: I think there's two methods for calculating per-allele likelihoods.  The first using a biallelic model, whereas the second uses a multiallelic model (for triallelic and so on).  They should be the same when two alleles are given.  Furthermore, the second item in the list is the _expected_ computed likelihoods, so we'll want to use that.

```suggestion
    cases.foreach { case (alleleDepths, likelihoods) =>
      val biallelic = Likelihoods(2, DownsampleVcf.Likelihoods.biallelic(alleleDepths(0), alleleDepths(1), e))
      val generalized = Likelihoods(2, DownsampleVcf.Likelihoods.generalized(alleleDepths, e))
      // Check that the biallelic likelihoods match
      biallelic.pls should contain theSameElementsInOrderAs likelihoods
      // Check that both methods of computing the likelihoods match
      biallelic.pls should contain theSameElementsInOrderAs generalized.pls
```

src/test/scala/com/fulcrumgenomics/vcf/AssessPhasingTest.scala
nh13: Yes, please open an issue, since there's a bug here in that case.

nh13 (approved):
Very minor suggestions, and nothing blocking.  What a Herculean effort!

src/main/scala/com/fulcrumgenomics/bam/ErrorRateByReadPosition.scala
```diff
@@ -192,7 +191,7 @@ class ErrorRateByReadPosition
   }
 
   /** Generates a variant mask object. */
-  private def buildVariantMask(variants: Option[PathToVcf], intervals: Option[IntervalList], dict: SequenceDictionary): VariantMask = {
+  private def buildVariantMask(intervals: Option[IntervalList], dict: SequenceDictionary): VariantMask = {
     this.variants match {
```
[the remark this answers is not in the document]
clintval: I thought about that but simplified because:

1. The function is private and is only called by other local code (and unit tests)
2. The function could/should have been "pure" and put in the companion object, but it is only ever used for the `variants` that are parameterized at the class level and I didn't want to do a bigger refactor in this PR.
3. Adding back `variants` and using the local definition would be shadowing the class level variable of the same name and be confusing to future reads.

src/main/scala/com/fulcrumgenomics/bam/TrimPrimers.scala
```diff
@@ -255,8 +255,8 @@ class TrimPrimers
       val plusTrim  = plus.end   - minus.end
       val minusTrim = plus.start - minus.start
 
-      if (plusTrim  > 0) this.clipper.clip3PrimeEndOfAlignment(plus, plusTrim)
-      if (minusTrim > 0) this.clipper.clip3PrimeEndOfAlignment(minus, minusTrim)
+      if (plusTrim  > 0) { val _ = this.clipper.clip3PrimeEndOfAlignment(plus, plusTrim) }
+      if (minusTrim > 0) { val _ = this.clipper.clip3PrimeEndOfAlignment(minus, minusTrim) }
```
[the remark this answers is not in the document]
clintval: That would move the same problem to the call site and `val _ = ` would be needed there instead. Given this is a private function I kept the "fix" as close to the code that triggers the issue. At least now it is clear that _something_ is being returned by the clippers but we are intentionally choosing to ignore that return value(s).

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -49,18 +49,19 @@ private[api] trait SamRecordIntermediate extends SAMRecord {
...
+  protected[api] def cigarChanged(): Unit = {}
 }
 
 /** Class that is used to provide a nice API to transient attributes in the SamRecord. */
 class TransientAttrs(private val rec: SamRecord) {
   def apply[A](key: Any): A = rec.asSam.getTransientAttribute(key).asInstanceOf[A]
   def update(key: Any, value: Any): Unit = {
-    if (value == null) rec.asSam.removeTransientAttribute(key) else rec.asSam.setTransientAttribute(key, value)
+    if (value == null) { val _ = rec.asSam.removeTransientAttribute(key) }
+    else { val _ = rec.asSam.setTransientAttribute(key, value) }
```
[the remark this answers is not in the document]
clintval: There are 2 ways to solve this one.

I added a commit that solves it the more readable way which has the slight effect of changing the type signature of `update()` from `update(): Unit` to `update(): Any`. This should be a backwards compatible type signature change too.

https://github.com/fulcrumgenomics/fgbio/pull/1047/commits/a5b6b7ba983b17e3c6bf013a2cd75b8a23228037

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -343,7 +343,7 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
   }
 
   /** Used it [[filterToMostCommonAlignment()]] to store a cigar string and a set of flags for which reads match. */
-  private final case class AlignmentGroup(cigar: Cigar, flags: mutable.BitSet, var size: Int = 0) {
+  private final class AlignmentGroup(val cigar: Cigar, val flags: mutable.BitSet, var size: Int = 0) {
```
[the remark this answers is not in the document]
clintval: Two reasons:

1. There is a [compiler bug](https://github.com/scala/bug/issues/4440) that makes it so this case class cannot be type checked correctly at runtime because it is a "final inner" class that is within an outer class. I believe the root of the issue has to do with the codegen for case classes in combo with the `final` modifier.
2. Case classes have more features than a typical class and it turns out this private class didn't need any of those extra features anyway.

If we did want to keep it `private` and `final` and `case` then we'd have to move it to the companion object or to its own class file but I intentionally avoided refactoring code in this PR and (1) and (2) from above biased me to simply remove the `case` modifier (as it triggered a compiler warning and was also unnecessary).

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
```diff
@@ -950,7 +949,7 @@ private[vcf] class PhaseCigar private(val cigar: Seq[PhaseCigarOp]) {
     /** Computes the # of long switch errors in the HMM output phased list. */
     def numLongSwitchErrors(states: List[Int]): Int = {
       if (states.length == 1) 0
-      else states.sliding(2).count { case Seq(first, last) => first != last }
+      else states.sliding(2).filter(_.length == 2).count(seq => seq.head != seq(1))
```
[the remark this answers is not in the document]
clintval: This one is tricky! The compiler is unable to _**prove**_ that `states` is non-empty. So, the `case` match approach would raise an exception if an empty `states` were to ever be encountered. The pattern match would not be exhaustive. Filtering for sub-slices that are of length 2 and then unpacking the tuple makes it such that no exception can be raised. I notice the previous code didn't guard for an empty list so I will add that in now too.
```diff
--- a/src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
+++ b/src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
-      if (states.length == 1) 0
-      else states.sliding(2).filter(_.length == 2).count(seq => seq.head != seq(1))
+      if (states.isEmpty | states.length == 1) 0
+      else states.sliding(2).count(seq => seq.head != seq(1))
```

src/test/scala/com/fulcrumgenomics/vcf/MakeTwoSampleMixtureVcfTest.scala
```diff
@@ -24,19 +24,22 @@
...
-import com.fulcrumgenomics.FgBioDef._
 import htsjdk.variant.vcf.VCFFileReader
 
+import scala.annotation.nowarn
+
+@nowarn("msg=class VariantContextSetBuilder in package testing is deprecated")
 class MakeTwoSampleMixtureVcfTest extends UnitSpec {
   private val builder = new VariantContextSetBuilder(sampleNames = List("s1", "s2"))
-  private val (_A, _C, _G, _T, _N) = ("A", "C", "G", "T", Allele.NO_CALL_STRING)
+  private val (_A, _C,  _T, _N) = ("A", "C", "T", Allele.NO_CALL_STRING)
```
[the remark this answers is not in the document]
clintval: LOL, Ok I'm adding it back in as `@unused` because I feel bad for it.

src/main/scala/com/fulcrumgenomics/vcf/AssessPhasing.scala
clintval: https://github.com/fulcrumgenomics/fgbio/issues/1048

src/test/scala/com/fulcrumgenomics/umi/AnnotateBamWithUmisTest.scala
clintval: https://github.com/fulcrumgenomics/fgbio/issues/1049

src/test/scala/com/fulcrumgenomics/vcf/AssessPhasingTest.scala
clintval: https://github.com/fulcrumgenomics/fgbio/issues/1050

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
clintval: @nh13 because the code routes directly to a [Java call](https://www.scala-lang.org/api/2.12.9/scala/collection/immutable/StringOps.html#lastIndexOf(x$1:Int):Int) of the same name with these type signatures:

Screenshot 2025-04-17 at 2 47 49 PM

I've done [more research](https://stackoverflow.com/questions/469657/string-lastindexof-is-there-a-difference-between-a-char-and-a-string-with-a-si) now which is making me think widening to an `Int` might be faster than building a single character string. I now think I need to manually widen to an `Int` since that is how Java would have behaved:

- https://stackoverflow.com/a/30718111/3727678

This will unfortunately make the code less readable to a human, but it is explicit as to what the compiler is going to do.

src/test/scala/com/fulcrumgenomics/vcf/DownsampleVcfTest.scala
clintval: Unfortunately that's not quite right because the `.pls` are a list of integers but the second element in the test data is a list of doubles. I also tried comparing the doubles to the `Liklihood.genotypeLikelihoods` list which matches the type, but none of the numbers were even close to being correct.

I will have to open an issue to track this since it's out of scope for this PR.

clintval: https://github.com/fulcrumgenomics/fgbio/issues/1051

src/main/scala/com/fulcrumgenomics/umi/Umis.scala
nh13: I think it's fine to make it less readable in this instance since it'll be called once per read in this instance, so we want it to be performant.  Up to you!

PR #1054, opened 2025-04-18
fix: DownsampleVcf test was not using an expected variable

### LINKED ISSUE #1051: Unused variable in a `DownsampleVcf` test may indicate a bug, or lack of true test coverage
The variable `output` (second element in the tuples) is never used in this test:

https://github.com/fulcrumgenomics/fgbio/blob/0e32d740ab4a8dc7b13622367f215740a442b84a/src/test/scala/com/fulcrumgenomics/vcf/DownsampleVcfTest.scala#L223-L236

I tried comparing it to the `.pls` field of the Likelihoods class but the types are incorrect (`Int` vs `Double`). I also compared it to the `genotypeLikelihoods` field of the class which is the correct type, but none of the value were even close to matching.

Observed in:

- https://github.com/fulcrumgenomics/fgbio/pull/1047

nh13:
Fixes #1051


clintval (approved):
Thank you!

PR #1057, opened 2025-04-18
Test building with all working versions of the JDK

clintval:
We cannot successfully build with 23 or greater.

We still target version 8:

https://github.com/fulcrumgenomics/fgbio/blob/48751481b5a838a7da8c4c4cb6ac1f5b7fd00e2a/build.sbt#L92

Picard supports Java 17 at a minimum now, so should we consider trying to migrate up?

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -42,9 +42,7 @@ jobs:
         run: |
           set -e
           sbt +clean coverage +test
-          sbt coverageReport coverageAggregate 
-          (find "$HOME/.sbt" -name "*.lock" -print0 | xargs -0 rm) || echo "No sbt lock files found in $HOME/.sbt"
-          (find "$HOME/.ivy2" -name "ivydata-*.properties" -print0 | xargs -0 rm) || echo "No ivy properties found in $HOME/.ivy2"
+          sbt coverageReport coverageAggregate
       - name: Code Coverage
         uses: codecov/codecov-action@v5.1.1
         with:
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -19,7 +19,10 @@ concurrency:
 jobs:
   test:
     name: Unit Tests
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java-version: [ 8, 11, 17, 21, 22 ]
     environment: github-actions
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/unittests.yaml
```diff
@@ -28,21 +31,13 @@ jobs:
...
+        run: git lfs checkout
+      - uses: actions/setup-java@v4
         with:
-          path: ~/.sbt
-          key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.properties') }}
+          distribution: 'temurin'
+          java-version: ${{ matrix.java-version }}
+          cache: sbt
+      - uses: sbt/setup-sbt@v1
       - name: Unit Tests
```
clintval, line 41: @nh13 do you think we can delete lines 46 and 47?

Do you recall what they are for?

.github/workflows/unittests.yaml
```diff
@@ -28,21 +31,13 @@ jobs:
...
-      - name: Cache .sbt
-        uses: actions/cache@v4
+        run: git lfs checkout
+      - uses: actions/setup-java@v4
         with:
-          path: ~/.sbt
-          key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.properties') }}
+          distribution: 'temurin'
+          java-version: ${{ matrix.java-version }}
+          cache: sbt
```
clintval, line 39: I removed all the manual caching since the `setup-java` action has functionality for sbt-related caching builtin.

.github/workflows/unittests.yaml
nh13: Those files may cause unnecessary cache misses, causing us to not use the cache: https://www.scala-sbt.org/1.x/docs/Travis-CI-with-sbt.html#Caching

clintval: Then I do think I can delete them since the GitHub Action to cache is smart and only looks for certain keys, instead of requiring we delete stuff that will dirty the cache.

https://github.com/actions/setup-java/blob/f4f1212c880fdec8162ea9a6493f4495191887b4/README.md?plain=1#L128

Thanks!

nh13: I thinks was also a hold over from using travis-ci

.github/workflows/unittests.yaml
```diff
@@ -28,28 +31,18 @@ jobs:
...
-      - name: Cache .ivy2
-        uses: actions/cache@v4
-        with:
-          path: ~/.ivy2
-          key: ${{ runner.os }}-ivy2-${{ hashFiles('**/build.sbt') }}
-      - name: Cache .sbt
-        uses: actions/cache@v4
+        run: git lfs checkout
+      - uses: actions/setup-java@v4
         with:
```
nh13, line 36: Nice find!

.github/workflows/unittests.yaml
```diff
@@ -28,28 +31,19 @@ jobs:
...
-        uses: actions/cache@v4
-        with:
-          path: ~/.ivy2
-          key: ${{ runner.os }}-ivy2-${{ hashFiles('**/build.sbt') }}
-      - name: Cache .sbt
-        uses: actions/cache@v4
-        with:
-          path: ~/.sbt
-          key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.properties') }}
+          apps: scala scalac scaladoc sbt
```
nh13, line 40:
```suggestion
          apps: scala:2-13.14 scalac:2.13.14 scaladoc:2-13.14 sbt:1.10.1
```

.github/workflows/unittests.yaml
```diff
@@ -69,17 +63,20 @@ jobs:
...
           fetch-depth: 0
-      - name: Setup Scala
-        uses: olafurpg/setup-scala@v12
+      - name: Setup cache
+        uses: coursier/cache-action@v6
+      - name: setup Scala
+        uses: coursier/setup-action@v1
         with:
-          java-version: adopt-openj9@1.8.0-292
+          apps: scala scalac scaladoc sbt
```
nh13, line 71:
```suggestion
          apps: scala:2-13.14 scalac:2.13.14 scaladoc:2-13.14 sbt:1.10.1
```

clintval:
Closing in favor of another attempt due to not wanting to rebase this PR given there are now many changes on main. Will be easier to submit a new, clean, PR.

PR #1060, opened 2025-04-25
feat: add a cached version of `Aligner`

nh13:
In profiling the `Aligner`, we (@TimD1 and myself) found that 40-50% of the time was spent re-allocating memory for the alignment matrices.  This PR adds a new `CachedAligner` class that stores the alignment matrices and re-allocates (enlarges) only when necessary.

I'd be ok with changing `Aligner` itself versus having a new class, but I know we use `Aligner` _a lot_ so I am being conservative in this PR.


src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
```diff
@@ -86,19 +85,86 @@ object Aligner {
...
+      CachedAlignmentMatrix(
+        direction    = dir,
+        scoring      = Matrix[Int](initQueryLength+1, initTargetLength+1),
+        trace        = Matrix[Direction](initQueryLength+1, initTargetLength+1),
+        queryLength  = initQueryLength,
+        targetLength = initTargetLength
+      )
+    }.toArray
+
+    override protected def allocMatrices(query: Array[Byte], target: Array[Byte]): Array[AlignmentMatrix] = {
```
clintval, line 146: Scaladoc on these two members would be especially helpful for future readers.

src/test/scala/com/fulcrumgenomics/alignment/AlignerTest.scala
```diff
@@ -550,4 +551,37 @@ class AlignerTest extends UnitSpec {
 
     System.out.println(s"Run median=${counter.median()}, mean=${counter.mean()}")
   }
+
+  Seq(true, false).foreach { useDefault =>
+    "CachedAligner.align(Global)" should f"align two identical sequences with all matches (default constructor: $useDefault)" in {
```
clintval, line 556: Is there any reason why we can't run all the unit tests over both aligner kinds?

src/main/scala/com/fulcrumgenomics/alignment/Aligner.scala
```diff
@@ -86,19 +85,86 @@ object Aligner {
...
+            scoring      = Matrix[Int](xLength, yLength),
+            trace        = Matrix[Direction](xLength, yLength),
+            queryLength  = query.length,
+            targetLength = target.length
+          )
+        }.toArray
+      } else if (this.matrices(0).queryLength != query.length || this.matrices(0).targetLength != target.length) {
+        this.matrices = this.matrices.map { matrix =>
+          matrix.copy(queryLength=query.length, targetLength=target.length)
+        }
```
clintval, line 162: This will leave the matrices `scoring` and `trace` with different sizes than the stated `queryLength` and `targetLength`. Is that OK?

PR #1061, opened 2025-04-29
Test building with known working versions of the JDK

clintval:
This PR runs all _**unit tests**_ using a matrix of JDKs that are known to work for building this project. I confirmed that sbt is correctly picking up the set Java distribution in each of the tests. E.g. here is the log line for the [Java 22 test](https://github.com/fulcrumgenomics/fgbio/actions/runs/14739164674/job/41373013950):

```console
[info] welcome to sbt 1.10.4 (Eclipse Adoptium Java 22.0.2)
```

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -25,16 +25,16 @@ jobs:
         jvm: ["temurin:1.8.0-442", "temurin:1.11.0.27", "temurin:1.17.0.15", "temurin:1.21.0.7", "temurin:1.22.0.2"]
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
           lfs: true
       - name: Checkout LFS objects
         run: git lfs checkout        
       - name: Setup cache
-        uses: coursier/cache-action@v6
+        uses: coursier/cache-action@4e2615869d13561d626ed48655e1a39e5b192b3c # v6.4.7
       - name: setup Scala
-        uses: coursier/setup-action@v1
+        uses: coursier/setup-action@039f736548afa5411c1382f40a5bd9c2d30e0383 # v1.3.9
         with:
           apps: scala:2.13.14 scalac:2.13.14 scaladoc:2.13.14 sbt:1.10.1
           jvm: ${{ matrix.jvm }}
@@ -44,7 +44,7 @@ jobs:
           sbt clean coverage test
           sbt coverageReport coverageAggregate
       - name: Code Coverage
-        uses: codecov/codecov-action@v5.1.1
+        uses: codecov/codecov-action@ad3126e916f78f00edff4ed0317cf185271ccc2d # v5.4.2
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           flags: unittests # optional
@@ -65,11 +65,12 @@ jobs:
       SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
       SONATYPE_PASS: ${{ secrets.SONATYPE_PASS }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
+      # TODO: The setup-scala action is deprecated and should be migrated to actions/setup-java or coursier/setup-action!
       - name: Setup Scala
-        uses: olafurpg/setup-scala@v12
+        uses: olafurpg/setup-scala@32ffa16635ff8f19cc21ea253a987f0fdf29844c # v14
         with:
           java-version: adopt-openj9@1.8.0-292
       - name: Upload to Sonatype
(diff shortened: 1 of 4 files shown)


.github/workflows/unittests.yaml
```diff
@@ -19,8 +19,10 @@ concurrency:
 jobs:
   test:
     name: Unit Tests
-    runs-on: ubuntu-24.04
-    environment: github-actions
```
clintval, line 23: Removing this environment since it is not needed.

This has the benefit of each PR spamming us less.

README.md
```diff
@@ -1,10 +1,12 @@
...
+[![Language](https://img.shields.io/badge/language-scala-c22d40.svg)](https://www.scala-lang.org/)
+[![Java Version](https://img.shields.io/badge/java-8,11,17,21,22-c22d40.svg)](https://github.com/AdoptOpenJDK/homebrew-openjdk)
+[![License](http://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/fulcrumgenomics/fgbio/blob/main/LICENSE)
+
 [![Bioconda](https://img.shields.io/conda/dn/bioconda/fgbio.svg?label=Bioconda)](http://bioconda.github.io/recipes/fgbio/README.html)
+[![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.fulcrumgenomics/fgbio_2.13/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.fulcrumgenomics/fgbio_2.13)
 [![Javadocs](http://javadoc.io/badge/com.fulcrumgenomics/fgbio_2.13.svg)](http://javadoc.io/doc/com.fulcrumgenomics/fgbio_2.13)
-[![License](http://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/fulcrumgenomics/fgbio/blob/main/LICENSE)
-[![Language](http://img.shields.io/badge/language-scala-brightgreen.svg)](http://www.scala-lang.org/)
 [![DOI](https://zenodo.org/badge/53011104.svg)](https://zenodo.org/doi/10.5281/zenodo.10456900)
```
clintval, line 10: I added a Java supported versions badge but it caused the badges to wrap and it looked ugly so I tried to group them and put them on two different lines. If you don't feel this looks good, lmk how we should think about re-organizing them!

.github/workflows/unittests.yaml
nh13: question: Don't we need it so we can use the `CODECOV_TOKEN`?   Or is it pickup it up an organizational token?

What do you mean "PR spamming us less"?

.github/workflows/unittests.yaml
```diff
@@ -29,27 +31,18 @@ jobs:
...
-        uses: actions/cache@v4
-        with:
-          path: ~/.ivy2
-          key: ${{ runner.os }}-ivy2-${{ hashFiles('**/build.sbt') }}
-      - name: Cache .sbt
-        uses: actions/cache@v4
+      - name: Setup cache
+        uses: coursier/cache-action@v6
+      - name: setup Scala
+        uses: coursier/setup-action@v1
```
nh13, line 37: suggestion: while we're here, shall we start the practice of using git hashes for versions instead?

.github/workflows/unittests.yaml
```diff
@@ -29,27 +31,18 @@ jobs:
...
         with:
-          path: ~/.sbt
-          key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.properties') }}
+          apps: scala:2.13.14 scalac:2.13.14 scaladoc:2.13.14 sbt:1.10.1
+          jvm: ${{ matrix.jvm }}
       - name: Unit Tests
         run: |
           set -e
           sbt +clean coverage +test
```
nh13, line 44: chore: while you're here... we don't build 2.12 anymore, so any "+<command>" can be updated
```suggestion
          sbt clean coverage test
```

README.md
nh13: suggestion: consider a different color for java (anything but red)

README.md
```diff
@@ -13,7 +15,7 @@ fgbio
 A set of tools to analyze genomic data with a focus on Next Generation Sequencing.
 
 <p>
-<a href float="left"="https://fulcrumgenomics.com"><img src=".github/logos/fulcrumgenomics.svg" alt="Fulcrum Genomics" height="100"/></a>
+<a href="https://fulcrumgenomics.com"><img src=".github/logos/fulcrumgenomics.svg" alt="Fulcrum Genomics" height="100"/></a>
```
nh13, line 18: huh, I thought I fixed these

nh13 (approved):

.github/workflows/unittests.yaml
clintval: Codecov still works, you can see the active GitHub Action check and also the Codecov comment on this PR:

- https://github.com/fulcrumgenomics/fgbio/pull/1061#issuecomment-2840052006

This is because we had `CODECOV_TOKEN` as an org-level secret for a while now:

- https://github.com/organizations/fulcrumgenomics/settings/secrets/actions

As far as spam, I would have thought you'd notice by now! We had the same issue in prymer and a few other OSS repos. Basically, anytime unit tests are run you'll get a PR message that a "temporary deployment" was made which is unnecessary PR clutter. It is especially advised to remove this behavior for this PR because we are now running unit tests `n` times over all the JDKs so it will `n`-times the amount of temporary deployment spam.

Screenshot 2025-04-30 at 6 46 32 AM

README.md
clintval: Sure thing. I had picked red so it blended in nicely next to Scala, but the next best choice is to consider making it the default Java color that GitHub uses:


Screenshot 2025-04-30 at 7 07 07 AM

It's a bit ugly, but maybe that was done on purpose 🤔

README.md
clintval: I did too 🤷

nh13 (approved):

PR #1062, opened 2025-04-29
Use a Git cliff TOML for releases

clintval:
I noticed a typo in the release action; we don't have a `pyproject.toml` to read from.

### DIFF
--- a/.cliff.toml
+++ b/.cliff.toml
@@ -1,0 +1,29 @@
+[changelog]
+header = ""
+trim = true
+body = """
+{% for group, commits in commits | group_by(attribute="group") %}
+    ## {{ group | upper_first }}
+    {% for commit in commits %}
+        - {{ commit.message | upper_first }} ({{ commit.id | truncate(length=8, end="") }})\
+    {% endfor %}
+{% endfor %}\n
+"""
+
+[git]
+conventional_commits = true
+commit_parsers = [
+    { message = "^.+!:*", group = "Breaking"},
+    { message = "^security*", group = "Security"},
+    { message = "^feat*", group = "Features"},
+    { message = "^fix*", group = "Bug Fixes"},
+    { message = "^docs*", group = "Documentation"},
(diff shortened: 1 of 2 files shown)


.cliff.toml
```diff
@@ -0,0 +1,29 @@
...
+    { message = "^docs*", group = "Documentation"},
+    { message = "^perf*", group = "Performance"},
+    { message = "^refactor*", group = "Refactor"},
+    { message = "^style*", group = "Styling"},
+    { message = "^test*", group = "Testing"},
+    { message = "^chore\\(release\\):*", skip = true},
+    { message = "^chore*", group = "Miscellaneous Tasks"},
+    { body = ".*security", group = "Security"}
+]
+filter_commits = false
```
clintval, line 29: Even if conventional commits are not used, the changelog will still look good! This configuration is "conventional commit aware" but will not fail if we don't use that standard. This is the same configuration we use in other template repos for that reason.

nh13 (approved):
I'd like @tfenne to weigh in too please

.github/workflows/release.yaml
```diff
@@ -148,7 +148,7 @@ jobs:
         uses: orhun/git-cliff-action@v3
         id: git-cliff
         with:
-          config: pyproject.toml
+          config: .cliff.toml
```
clintval, line 151: The alternative solution (instead of having a Git Cliff TOML) is to delete the TOML and use whatever default settings the changelog generator runs with. This would work and avoid the need to have a new configuration file in the root of the repo.

Although we have a mix of standard commit messages and conventional commits in our mainline history, so I do see some benefit now in being "conventional commit"-aware when generating release notes so that the releases are more readable and the conventional commit prefixes are stripped.

https://github.com/fulcrumgenomics/fgbio/commits/main/

tfenne:
> I'd like @tfenne to weigh in too please

I can't say I fully understand this, but I'd say let's merge it and see how it works for the next release, and then adjust?

Issue #1030, opened 2025-02-19
Why do we use of hard-clipping when calculating the mate's start and end?


nh13:
Should we be using hard-clapping the mate's start and end in our consensus calling tools?  I see two places where we do (or will do):

1. [GroupReadsByUmi](https://github.com/fulcrumgenomics/fgbio/blob/f93fdfbb427da5f5d60304de5761b19ea8209b33/src/main/scala/com/fulcrumgenomics/umi/GroupReadsByUmi.scala#L101)
2. In #1026 (see [this discussion with @clintval](https://github.com/fulcrumgenomics/fgbio/pull/1026#discussion_r1962091807))

I make up that the reason for adjusting the start and end based on soft-clipping is because those bases _could_ be aligned, and may actually be aligned in the mate, which can happen if we have short inserts.  But why do we adjust it _also based on hard-clipping_?  Those bases are removed.  Perhaps if there's hard clipping on only one of the reads in a pair on the one end of the molecule?  Something else?

Note: there are other places we adjust based on hard-clipping as well:

1. https://github.com/fulcrumgenomics/fgbio/blob/f93fdfbb427da5f5d60304de5761b19ea8209b33/src/main/scala/com/fulcrumgenomics/util/AmpliconDetector.scala#L177C15-L181
2. https://github.com/fulcrumgenomics/fgbio/blob/f93fdfbb427da5f5d60304de5761b19ea8209b33/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala#L183

And of course, there are a number of other places that use the [unclipped start](https://github.com/search?q=repo%3Afulcrumgenomics%2Ffgbio%20unclippedStart&type=code) and [unclipped end](https://github.com/search?q=repo%3Afulcrumgenomics%2Ffgbio+unclippedEnd&type=code) for the current record.  I think examining those is worthwhile, but we should focus on the consensus calling tools first.

PR #1065, opened 2025-05-08
Often ignore hard clipping when fetching unclipped ends

clintval:
Motivated by:

Closes https://github.com/fulcrumgenomics/fgbio/issues/878
Closes https://github.com/fulcrumgenomics/fgbio/pull/937
Closes https://github.com/fulcrumgenomics/fgbio/issues/1030

_**Complete PR justification comment coming soon, opening as draft for now...**_

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
@@ -327,23 +327,23 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean) 
   /** Returns the number of bases extending past the mate end for FR pairs including any soft-clipped bases, zero otherwise.
     *
     * @param rec the record to examine
-    * @param mateUnclippedStart the smallest mapped genomic coordinate considering clipping
-    * @param mateUnclippedEnd the largest mapped genomic coordinate considering clipping
+    * @param mateUnSoftClippedStart the smallest mapped genomic coordinate considering *ONLY* soft-clipping
+    * @param mateUnSoftClippedEnd the largest mapped genomic coordinate considering *ONLY* soft-clipping
     */
-  def numBasesExtendingPastMate(rec: SamRecord, mateUnclippedStart: Int, mateUnclippedEnd: Int): Int = {
+  def numBasesExtendingPastMate(rec: SamRecord, mateUnSoftClippedStart: Int, mateUnSoftClippedEnd: Int): Int = {
     if (!rec.isFrPair) 0 else rec.positiveStrand match {
-      case true if rec.end >= mateUnclippedEnd =>
-        // positive strand record is aligned to/past the mate alignment end: count any bases aligned/softclipped after
-        Math.max(0, rec.length - rec.readPosAtRefPos(pos=mateUnclippedEnd, returnLastBaseIfDeleted=false))
+      case true if rec.end >= mateUnSoftClippedEnd =>
+        // positive strand record is aligned to/past the mate alignment end: count any bases aligned/soft-clipped after
+        Math.max(0, rec.length - rec.readPosAtRefPos(pos=mateUnSoftClippedEnd, returnLastBaseIfDeleted=false))
       case true =>
         // positive strand record alignment ends before the mate alignment end: remove any excess soft-clipped reads
-        Math.max(0, rec.cigar.trailingSoftClippedBases - (mateUnclippedEnd - rec.end))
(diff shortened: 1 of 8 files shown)


src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
```diff
@@ -186,13 +189,30 @@ trait SamRecord {
...
+      mateStart - cigar.leadingSoftClippedBases
+    }
+  }
+  @inline final def mateUnSoftClippedEnd: Option[Int] = {
+    require(paired && mateMapped, "Cannot get mate un-soft-clipped end position on read without a mapped mate.")
+    get[String]("MC").map { cig =>
+      val cigar = Cigar(cig)
+      mateStart + cigar.lengthOnTarget - 1 + cigar.trailingSoftClippedBases
     }
   }
```
clintval, line 217: I'm starting to think we should cache the mate cigar like we do with the record's cigar. It would require similar trickery to detect when a new mate cigar is updated upon the record since we would have to intercept any call to `setAttribute`, check if the attribute is being set with the `MC` tag, and then reset the in-memory mate cigar to the new value. If I implement this suggestion, it will be in a separate PR.

src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -327,23 +327,23 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
   /** Returns the number of bases extending past the mate end for FR pairs including any soft-clipped bases, zero otherwise.
     *
     * @param rec the record to examine
-    * @param mateUnclippedStart the smallest mapped genomic coordinate considering clipping
-    * @param mateUnclippedEnd the largest mapped genomic coordinate considering clipping
+    * @param mateUnSoftClippedStart the smallest mapped genomic coordinate considering *ONLY* soft-clipping
+    * @param mateUnSoftClippedEnd the largest mapped genomic coordinate considering *ONLY* soft-clipping
```
clintval, line 331: I don't love the new names for these but really couldn't think of anything better.

src/test/scala/com/fulcrumgenomics/bam/ClipBamTest.scala
clintval, line 1: No old tests need to be updated which makes me feel confident this edge case hasn't been considered or explored much yet.

src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -287,12 +287,12 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
       var index = if (!rec.isFrPair) trimToLength - 1 else {
         // Get the number of mapped bases to clip that maps beyond the mate's end, including any soft-clipped bases. Use
         // that to compute where in the read to keep.
-        val mateUnclippedStart = rec.mateUnclippedStart.getOrElse(throw new IllegalArgumentException(f"Mate cigar (MC SAM tag) needed for read: ${rec.name}"))
-        val mateUnclippedEnd   = rec.mateUnclippedEnd.getOrElse(throw new IllegalArgumentException(f"Mate cigar (MC SAM tag) needed for read: ${rec.name}"))
+        val mateUnSoftClippedStart = rec.mateUnSoftClippedStart.getOrElse(throw new IllegalArgumentException(f"Mate cigar (MC SAM tag) needed for read: ${rec.name}"))
+        val mateUnSoftClippedEnd   = rec.mateUnSoftClippedEnd.getOrElse(throw new IllegalArgumentException(f"Mate cigar (MC SAM tag) needed for read: ${rec.name}"))
```
nh13: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
consider a note or TODO that the MC is parsed twice

clintval: I will open an issue!

This will be fixed when the mate cigar can be cached. I'm going to open an issue to track that change so I don't need to add a comment to the `UmiConsensusCaller` at this time. The double-parse also existed before and after this PR too.

src/main/scala/com/fulcrumgenomics/bam/api/SamRecord.scala
clintval: https://github.com/fulcrumgenomics/fgbio/issues/1115

PR #1072, opened 2025-05-29
DRAFT PR to add a consensus caller for CODEC sequencing

tfenne:
This PR is not yet ready for review and is a work-in-progress towards a CODEC consensus caller.


src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
```diff
@@ -354,9 +354,14 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean)
...
-    val basesClipped2 = clipExtendingPastMateEnd(rec=mate, mateUnclippedStart=rec.unclippedStart, mateUnclippedEnd=rec.unclippedEnd)
-    (basesClipped1, basesClipped2)
+    if (rec.isFrPair) {
+      val basesClipped1 = clipExtendingPastMateEnd(rec=rec, mateUnclippedStart=mate.unclippedStart, mateUnclippedEnd=mate.unclippedEnd)
+      val basesClipped2 = clipExtendingPastMateEnd(rec=mate, mateUnclippedStart=rec.unclippedStart, mateUnclippedEnd=rec.unclippedEnd)
+      (basesClipped1, basesClipped2)
+    }
+    else {
+      (0, 0)
+    }
```
clintval, line 364: The two calls to `clipExtendingPastMateEnd` already performed this guard and returned 0 each. Are you doing this so that we don't have to spend time computing the unclipped end positions if we don't need them?

src/main/scala/com/fulcrumgenomics/umi/CallCodecConsensusReads.scala
```diff
@@ -0,0 +1,131 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
```
nh13, line 4:
```suggestion
 * Copyright (c) 2025 Fulcrum Genomics LLC
```

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -0,0 +1,196 @@
...
+    val (pairs, frags) = recs.partition(_.paired)
+    rejectRecords(frags, FilterFragments)
+
+    if (pairs.isEmpty) {
+      Nil
+    }
+    else {
+      // Extract just the primary alignments and then clip where they extend past the ends of their mates
+      // TODO: Remove the isFrPair here and handle chimeric reads or reads with one unmapped etc.
+      val primaries = pairs.filterNot(r => r.secondary || r.supplementary).filter(_.isFrPair)
```
nh13, line 91: question: Most aligners prefer to to have short inserts between R1 and R2, so this may not be an issue, but do we want to mention in the tool docs that we expect R1 and R2 to be almost fully overlapping, so alignment settings should be changed (e.g. insert size penalty) to really prefer overlapping alignments?

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -0,0 +1,196 @@
...
+      else {
+        // Calculate the max overlap between R1 and R2, and if it's too short then filter those reads
+        val longestR1Alignment = r1s.flatMap(_.sam).maxBy(_.cigar.lengthOnTarget)
+        val longestR2Alignment = r2s.flatMap(_.sam).maxBy(_.cigar.lengthOnTarget)
+
+        val (longestPosAln, longestNegAln) =
+          if (longestR1Alignment.positiveStrand) (longestR1Alignment, longestR2Alignment)
+          else (longestR2Alignment, longestR1Alignment)
+
+        val (overlapStart, overlapEnd) = (longestNegAln.start, longestPosAln.end)
```
nh13, line 109: question: the reads sequence towards each other still (innies), is that why we use negative for start, positive for end?  Perhaps a comment or two here?

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -0,0 +1,196 @@
...
+              rejectRecords((r1s.view ++ r2s.view).flatMap(_.sam), FilterIndelError)
+              Nil
+            case n if n < r1Consensus.length || n < r2Consensus.length =>
+              // TODO remove this branch once SamRecordClipper has been updated
+              rejectRecords((r1s.view ++ r2s.view).flatMap(_.sam), "Temporary Issue: Fix Clipping Issue")
+              Nil
+            case consensusLength =>
+              // Pad out the vanilla consensus reads to the full length
+              val paddedR1 = r1Consensus.padded(newLength = consensusLength, left = longestR1Alignment.negativeStrand, qual = 0)
+              val paddedR2 = r2Consensus.padded(newLength = consensusLength, left = longestR2Alignment.negativeStrand, qual = 0)
```
nh13, line 146:
```suggestion
              val paddedR1 = r1Consensus.padded(newLength = consensusLength, left = longestR1Alignment.negativeStrand, qual = PhredScore.MinValue)
              val paddedR2 = r2Consensus.padded(newLength = consensusLength, left = longestR2Alignment.negativeStrand, qual = PhredScore.MinValue)
```

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -0,0 +1,196 @@
...
+    *
+    * Returns -1 if the length couldn't be computed because the overlap of the reads ends on an indel.
+    */
+  private def computeConsensusLength(pos: SamRecord, neg: SamRecord): Int = {
+    val refOverlapEnd = pos.end
+    val posReadPos = pos.readPosAtRefPos(refOverlapEnd, returnLastBaseIfDeleted=false)
+    val negReadPos = neg.readPosAtRefPos(refOverlapEnd, returnLastBaseIfDeleted=false)
+    if (posReadPos == 0 || negReadPos == 0) -1 else {
+      posReadPos + neg.length - negReadPos
+    }
```
nh13, line 174: question: is it true that we assume that the two reads are sequenced the same # of cycles?  If so, do we have to worry about differences in lengths here and above?

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -204,6 +204,18 @@ object Sequences {
...
+  @inline @specialized def reverse[T](xs: Array[T]): Unit = {
+    var (i, j) = (0, xs.length - 1)
+    while (i < j) {
+      val tmp = xs(i)
+      xs(i) = xs(j)
+      xs(j) = tmp
+      i += 1
+      j -= 1
+    }
+  }
```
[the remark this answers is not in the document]
tfenne: I think there are two issues with this.  The first is that microbenchmarking is *hard*.  The biggest difference between the two implementations is that `array.reverse` allocates a new array and reverses the elements into it, while `reverse(arr)` here reverses in place.  Otherwise they are *very* similar - a while loop with some simple arithmetic and assignments.  I find it hard to believe that in heavy use (we'd be reversing ~15 arrays per consensus read) in a JVM that's actually doing a lot of other allocation and work, that the GC overhead of all those allocations wouldn't add up.  But I guess the question is, is it "not faster" or "slower"?

The second is that I'm doing this on case classes with `val array`s, and there's no way to make `Array.reverse` work in place.

src/main/scala/com/fulcrumgenomics/umi/CallCodecConsensusReads.scala
```diff
@@ -0,0 +1,131 @@
...
+class CallCodecConsensusReads
+(@arg(flag='i', doc="The input SAM or BAM file.") val input: PathToBam,
+ @arg(flag='o', doc="Output SAM or BAM file to write consensus reads.") val output: PathToBam,
+ @arg(flag='p', doc="The prefix all consensus read names") val readNamePrefix: Option[String] = None,
+ @arg(flag='R', doc="The new read group ID for all the consensus reads.") val readGroupId: String = "A",
+ @arg(flag='1', doc="The Phred-scaled error rate for an error prior to the UMIs being integrated.") val errorRatePreUmi: PhredScore = DefaultErrorRatePreUmi,
+ @arg(flag='2', doc="The Phred-scaled error rate for an error post the UMIs have been integrated.") val errorRatePostUmi: PhredScore = DefaultErrorRatePostUmi,
+ @arg(flag='m', doc="Ignore bases in raw reads that have Q below this value.") val minInputBaseQuality: PhredScore = DefaultMinInputBaseQuality,
+ @arg(flag='S', doc="The sort order of the output, the same as the input if not given.") val sortOrder: Option[SamOrder] = None,
+ @arg(flag='M', minElements=1, maxElements=3, doc="The minimum number of codec read pairs to form a consensus read.") val minReadsPerStrand: Int = 1,
```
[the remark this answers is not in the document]
tfenne: Good idea on the name change, thanks!

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
tfenne: I think this only happens in repetitive regions right?  I don't think I have enough experience with CODEC yet to be offering advice on alignment ... and I think I'd prefer to just not analyze reads in regions where the alignments are not of high confidence (after consensus calling).

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
tfenne: Yeah - I'll add a comment, but everything here is predicated on the reads being an FR pair that overlaps.

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
tfenne: No ... there's no assumption anywhere that reads are all the same length.  The only assumption is that by the time we get here the reads have been trimmed so that they don't extend past each other - which I guess _is_ an assumption they are the same length _in the case when they are fully overlapped_.

PR #1073, opened 2025-05-30
Shorter version of refPosAtReadPos.

tfenne:
@nh13 I hope you don't mind - I took at shot at writing a version of refPosAtReadPos that doesn't require a locally defined function and is a bit shorter.  I _think_ it should be more efficient?  All the tests pass :)

I think readPosAtRefPos could be reworked the same way if you like this.


nh13 (approved):

nh13 (approved):
@tfenne lgtm, I can add in `readPosAtRefPos`.

PR #1075, opened 2025-05-30
Change FilterConsensusReads to allow for a fraction of Ns OR a count of Ns.

tfenne:
For CODEC which will produce reads of highly varying length, it would be useful to be able to set the threshold of no-calls as a *count* instead of a fraction.  This PR changes the name of the parameter from `--max-no-call-fraction` to `--max-no-calls` and interprets the value as a fraction if it is `0 < value < 1.0` and a count if `value >= 1.0`.


src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
```diff
@@ -128,7 +129,7 @@ class FilterConsensusReads
   Io.assertCanWriteFile(output)
   if (maxReadErrorRate.exists(e => e < 0 || e > 1)) fail("max-read-error-rate must be between 0 and 1.")
   if (maxBaseErrorRate.exists(e => e < 0 || e > 1)) fail("max-base-error-rate must be between 0 and 1.")
-  if (maxNoCallFraction < 0 || maxNoCallFraction > 1) fail("max-no-call-fraction must be between 0 and 1.")
+  if (maxNoCalls < 0) fail("max-no-calls must be greater than or equal to 0.")
```
nh13, line 132:
```suggestion
  if (maxNoCalls < 0) fail("max-no-calls must be greater than or equal to 0.")
  if (maxNoCalls >= 1 && maxNoCalls != maxNoCalls.toInt.toFloat) fail("max-no-calls must be an integer when greater than or equal to 1.")
```

nh13 (approved):
LGTM with one minor suggestion

PR #1080, opened 2025-06-20
Add end and single-strand quality lowering to codec consensus calling.

tfenne:


src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -193,6 +200,7 @@ class CodecConsensusCaller(readNamePrefix: String,
 
               val consensus = duplexConsensus(Some(paddedR1), Some(paddedR2), sourceReads = None)
               if (longestR1Alignment.negativeStrand) consensus.foreach(_.revcomp())
+              consensus.foreach(maskCodecConsensusQuals)
```
nh13, line 203: [suggestion(non-blocking)](https://pullpo.io/cc?l=suggestion&d=non-blocking) 

This may be a bit of a pre-mature optimization, but I included here regardless:

```suggestion
              if (this.singleStrandQual.isDefined || this.outerBasesLength > 0) consensus.foreach(maskCodecConsensusQuals)
```

nh13 (approved):

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
tfenne: I'd rather have all the masking related decisions inside the function - can come back and optimize (or inline!) if I'm wrong.

PR #1081, opened 2025-06-20
Implementation of discarding consensus reads with too much disagreement.

tfenne:


src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -198,15 +206,31 @@ class CodecConsensusCaller(readNamePrefix: String,
               val paddedR1 = r1Consensus.padded(newLength = consensusLength, left = longestR1Alignment.negativeStrand, qual = 0)
               val paddedR2 = r2Consensus.padded(newLength = consensusLength, left = longestR2Alignment.negativeStrand, qual = 0)
 
-              val consensus = duplexConsensus(Some(paddedR1), Some(paddedR2), sourceReads = None)
-              if (longestR1Alignment.negativeStrand) consensus.foreach(_.revcomp())
-              consensus.foreach(maskCodecConsensusQuals)
+              // Calculate number and rate of duplex errors ... but don't really if the thresholds won't need them
+              val (duplexErrors, duplexErrorRate) = {
+                if (this.maxDuplexDisagreementRate < 1 && this.maxDuplexDisagreements < overlapLength) (0, 0.0)
```
nh13: [issue](https://pullpo.io/cc?l=issue) 

It probably could use a test ;P

```suggestion
                if (this.maxDuplexDisagreementRate >= 1 && this.maxDuplexDisagreements > overlapLength) (0, 0.0)
```

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -198,15 +206,31 @@ class CodecConsensusCaller(readNamePrefix: String,
               val paddedR1 = r1Consensus.padded(newLength = consensusLength, left = longestR1Alignment.negativeStrand, qual = 0)
               val paddedR2 = r2Consensus.padded(newLength = consensusLength, left = longestR2Alignment.negativeStrand, qual = 0)
 
-              val consensus = duplexConsensus(Some(paddedR1), Some(paddedR2), sourceReads = None)
-              if (longestR1Alignment.negativeStrand) consensus.foreach(_.revcomp())
-              consensus.foreach(maskCodecConsensusQuals)
+              // Calculate number and rate of duplex errors ... but don't really if the thresholds won't need them
+              val (duplexErrors, duplexErrorRate) = {
+                if (this.maxDuplexDisagreementRate >= 1 && this.maxDuplexDisagreements >= overlapLength) (0, 0.0)
```
nh13, line 211: [issue(non-blocking)](https://pullpo.io/cc?l=issue&d=non-blocking) 
could probably use a test, your call

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -275,4 +299,30 @@ class CodecConsensusCaller(readNamePrefix: String,
...
+    */
+  private[umi] def computeDuplexCounts(a: VanillaConsensusRead, b: VanillaConsensusRead): (Int, Int) = {
+    var duplexBases  = 0
+    var duplexErrors = 0
+
+    forloop (from=0, until=a.length) { i =>
+      val aBase = a.bases(i)
+      val bBase = b.bases(i)
+
+      if (aBase != 'n' && aBase != 'N' && bBase != 'n' && bBase != 'N') {
```
nh13, line 319: [nitpick(non-blocking)](https://pullpo.io/cc?l=nitpick&d=non-blocking) 
Can't we just use [`NoCall`](https://github.com/fulcrumgenomics/fgbio/blob/153d4508fafe7e686e13b61162e34e1accf82817/src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala#L36) or similar?  I.e. when would we be producing `'n'`?

nh13 (approved):

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
tfenne: In the codec caller I'm using `n` when I pad out the reads when they don't fully overlap.  That way, in the ss reads in the tags, you can clearly see the "padding" vs. any bases that are `N` due to disagreement/poor quality/etc.

PR #1086, opened 2025-07-29
Remove primary name if also found in alternate names in CollectAlternateContigNames

tfenne:
I was trying to use CollectAlternateContigNames on [this assembly report](https://ftp.ncbi.nlm.nih.gov/genomes/all/GCF/018/350/175/GCF_018350175.1_F.catus_Fca126_mat1.0/GCF_018350175.1_F.catus_Fca126_mat1.0_assembly_report.txt).  It was failing, because the last line has the sequence name `MT` and also assigned-molecule as `MT` and that was causing redundancy in the generated sequence dictionary.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
+++ b/src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
@@ -184,7 +184,7 @@ class CollectAlternateContigNames
               None
             case alternate => Some(alternate)
           }
-      }.distinct
+      }.distinct.filter(_ != name)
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/fasta/CollectAlternateContigNames.scala
```diff
@@ -184,7 +184,7 @@ class CollectAlternateContigNames
               None
             case alternate => Some(alternate)
           }
-      }.distinct
+      }.distinct.filter(_ != name)
```
clintval, line 187: Should we add a unit test to protect this case from future regression?

clintval (approved):

PR #1097, opened 2025-09-24
Make it so FastqToBam can stream input

clintval:
I'd like to stream with `FastqToBam` so I don't have to generate temporary FASTQ files before format conversion.

Currently, `FastqToBam` opens all input files, samples them to determine the quality encoding, and closes them before re-opening them for actual iteration. If we add a fully peekable iterator over the FASTQ streams, then we can peek forward as many records as we need for quality encoding detection and then re-use the iterators for doing the actual work.

I happened to have a fully peekable iterator in my personal portfolio, but if that dependency is unwanted, I can copy-paste the code file over the fgbio project.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/QualityEncoding.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/QualityEncoding.scala
@@ -25,8 +25,11 @@
 package com.fulcrumgenomics.fastq
 
 import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.commons.collection.BetterBufferedIterator
 import com.fulcrumgenomics.commons.util.{NumericCounter, SimpleCounter}
+import com.fulcrumgenomics.fastq.QualityEncodingDetector.DefaultSampleCount
 import enumeratum.EnumEntry
+import io.cvbio.collection.FullyPeekableIterator
 
 import scala.collection.immutable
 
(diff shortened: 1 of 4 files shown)


src/main/scala/com/fulcrumgenomics/fastq/QualityEncoding.scala
```diff
@@ -110,12 +113,22 @@ class QualityEncodingDetector {
...
-    while (counter.total < goal && ss.hasNext) add(ss.next())
+  def sample(ss: Iterator[String], count: Int = DefaultSampleCount): Unit = {
+    ss match {
+      case iter: BetterBufferedIterator[String] if count == 1 =>
+        iter.nextOption().foreach(add)
+      case iter: FullyPeekableIterator[String] =>
+        iter.liftMany(0, count).flatten.foreach(add)
+      case iter: Iterator[String] =>
+        iter.take(count).foreach(add)
+    }
```
clintval, line 131: This wasn't essential for the task at hand, but I felt would be a safe addition for future users. I can remove this if desired.

nh13:
I think either adding the new collection to [`commons`](https://github.com/fulcrumgenomics/commons/tree/main/src/main/scala/com/fulcrumgenomics/commons/collection) or to just buffer manually is the easiest way to go.

clintval:
@nh13 I'll re-work this!

PR #1100, opened 2025-09-29
Personal tool to split up large intervals in an interval list.

tfenne:
Personal tool used once to break up an interval list for a project, that's here for posterity.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/personal/tfenne/SplitIntervals.scala
+++ b/src/main/scala/com/fulcrumgenomics/personal/tfenne/SplitIntervals.scala
@@ -1,0 +1,52 @@
+package com.fulcrumgenomics.personal.tfenne
+
+import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.sopt.{arg, clp}
+import htsjdk.samtools.util.{Interval, IntervalList}
+
+@clp(group=ClpGroups.Personal, description=
+  """
+    |Splits intervals within an interval list.  Consumes an input interval list and attempts to break it
+    |into _at least_ `--num-intervals` intervals.  The resulting interval list will likely end up with additional
+    |intervals.
+    |
+    |Calculates a max interval size by summing the length of all input intervals and dividing by the desired
+    |number of output intervals.  Then each input interval is is broken up into approximately equals size,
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/personal/tfenne/SplitIntervals.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import com.fulcrumgenomics.FgBioDef._
+import com.fulcrumgenomics.cmdline.{ClpGroups, FgBioTool}
+import com.fulcrumgenomics.sopt.{arg, clp}
+import htsjdk.samtools.util.{Interval, IntervalList}
+
+@clp(group=ClpGroups.Personal, description=
+  """
+    |Splits intervals within an interval list.  Consumes an input interval list and attempts to break it
+    |into _at least_ `--num-intervals` intervals.  The resulting interval list will likely end up with additional
```
nh13, line 11:
```suggestion
    |into _at least_ `--num-intervals` intervals of approximately equal size.  The resulting interval list will likely end up with additional
```

src/main/scala/com/fulcrumgenomics/personal/tfenne/SplitIntervals.scala
```diff
@@ -0,0 +1,52 @@
...
+    |intervals.
+    |
+    |Calculates a max interval size by summing the length of all input intervals and dividing by the desired
+    |number of output intervals.  Then each input interval is is broken up into approximately equals size,
```
nh13, line 15:
```suggestion
    |number of output intervals.  Then each input interval is is broken up into approximately equal size,
```

src/main/scala/com/fulcrumgenomics/personal/tfenne/SplitIntervals.scala
```diff
@@ -0,0 +1,52 @@
...
+    |non-overlapping, sub-intervals of no more than the max size.
+  """)
```
nh13, line 17: I found the usage very confusing, sorry.  Perhaps something like, takes in a interval list, and tries splitting intervals such that the output interval list has at least `--num-intervals` intervals.  If the interval list already has at least this number, nothing is done (is this true?)

nh13 (approved):

yfarjoun (approved):
The code does what I expect, the documentation I leave to the english speakers...

PR #1103, opened 2025-10-03
Support cell barcodes in ReadStructure and FastqToBam

clintval:
## Introduction

I'd like to support extracting cell barcodes with `FastqToBam`.

This requires that the `ReadStructure` specification has support for a cell barcode identifier. I propose we add the `C` operator to the read structure specification. Once we have a `C` operator we can extract cell barcodes from a read and their quality sequences too.

The [SAM spec.](https://samtools.github.io/hts-specs/SAMtags.pdf) recommends we store cell barcode information similarly to molecular indices: hypen-separated for the cell barcode sequences and space-separated for the associated qualities. The recommended tags for cell barcode info is as follows:

- `CB`: cell identifier
- `CR`: cell barcode sequence bases (uncorrected)
- `CY`: Phred qualities of the cell barcode sequence

Along with this PR, I propose we make these changes to the Wiki on [`ReadStructure`](https://github.com/fulcrumgenomics/fgbio/wiki/Read-Structures):

```diff
20c20
< * `FastqToBam` in fgbio to convert from fastq to BAM while preserving sample barcode and UMI information
---
> * `FastqToBam` in fgbio to convert from fastq to BAM while preserving sample barcode, cell barcode, and UMI information
28a29
> * `C` or Cell Barcode: the bases in the segment are a cell bar index sequence used to identify the cell being sequenced
51a53,55
> * A 2x150bp single-cell sequencing run with two cell-specific barcodes separated by a skipped linker and a UMI:
>   * `5C30S5C3S8M99T8B150T`
>   * [`5C30S5C3S8M+T`, `8B`, `+T`]
62c66
< <operator>           ::= "T" | "B" | "M" | "S"
---
> <operator>           ::= "T" | "B" | "M" | "C" | "S"
```

And this change to the read structure validator:

- https://github.com/fulcrumgenomics/fgbio/compare/gh-pages...cv_cell_barcodes_js

## Out of Scope

At this time I did not feel that supporting cell barcodes in `DemuxFastqs` was worth the effort since we'd rather users adopt `fqtk`. So when a cell barcode is encountered, a validation exception is raised. I also did not add cell barcode specific functionality to `ExtractUmisFromBam` since I don't think that's necessary for the function of the tool.

I also intend to skip Picard's [`IlluminaBaseCallsToSam`](https://gatk.broadinstitute.org/hc/en-us/articles/360037225472-IlluminaBasecallsToSam-Picard) and [`IlluminaBaseCallsToFastq`](https://gatk.broadinstitute.org/hc/en-us/articles/360037051752-IlluminaBasecallsToFastq-Picard) because of [numerous issues](https://github.com/broadinstitute/picard/issues?q=is%3Aissue%20state%3Aopen%20(IlluminaBaseCallsToSam%20OR%20IlluminaBaseCallsToFastq)), many of which I've experienced such that I would no longer recommend those tools be used in the era of NovaSeq.

## Next Steps

If this PR is merged and we update the wiki and Javascript validator, then I'd go and make changes to the rest of the ecosystem to ensure better support for cell barcodes in read structures:

- [ ] Python [`ReadStructure`](https://github.com/fulcrumgenomics/fgpyo/blob/39a7f58bcbe6eac7bd9ee667c889a8e558b8b237/fgpyo/read_structure.py#L193) in `fgpyo`
- [ ] Rust [`SegmentType`](https://github.com/fulcrumgenomics/read-structure/blob/295aa041621cd8d95ef88af4d30c03b5d2bb4348/src/segment_type.rs#L18) in `read-structure`
- [ ] Newer Rust demultiplexer [`fqtk`](https://github.com/fulcrumgenomics/fqtk)
- [ ] _**Others?**_

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -377,6 +377,10 @@ class DemuxFastqs
  val omitControlReads: Boolean = false,
  @arg(doc="Mask bases with a quality score below the specified threshold as Ns") val maskBasesBelowQuality: Int = 0,
 ) extends FgBioTool with LazyLogging {
+  validate(
+    !readStructures.exists(_.cellBarcodeSegments.nonEmpty),
+    "DemuxFastqs does not support cell barcodes in read structures!",
+  )
 
   private val fastqStandards: FastqStandards = {
     FastqStandards(
--- a/src/test/scala/com/fulcrumgenomics/fastq/FastqToBamTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/fastq/FastqToBamTest.scala
@@ -383,6 +383,22 @@ class FastqToBamTest extends UnitSpec {
     recs(1).apply[String]("RX") shouldBe "TTGA-TAAT-TA-AA"
   }
 
+  it should "extract cell barcodes in read sequences" in {
+    val r1 = fq(
+      FastqRecord("q1:2:3:4:5:6:7:ACGT+CGTA", "GGNCCGAAAAAAA", "============="),
+      FastqRecord("q2:2:3:4:5:6:7:TTGA+TAAT", "TANAACAAAAAAA", "============="),
+    )
+    val rs  = ReadStructure("2C1S2C+T")
+    val bam = makeTempFile("fastqToBamTest.", ".bam")
+    new FastqToBam(input=Seq(r1), readStructures=Seq(rs), output=bam, sample="s", library="l", cellQualTag = Some("CY")).execute()
+    val recs = readBamRecs(bam)
+    recs should have size 2
+    recs(0).apply[String]("CB") shouldBe "GG-CC"
+    recs(0).apply[String]("CY") shouldBe "== =="
+    recs(1).apply[String]("CB") shouldBe "TA-AA"
+    recs(1).apply[String]("CY") shouldBe "== =="
+  }
+
   it should "fail when read names don't match up" in {
     val r1 = fq(FastqRecord("q1", "AAAAAAAAAA", "=========="))
     val r2 = fq(FastqRecord("x1", "CCCCCCCCCC", "##########"))
--- a/src/test/scala/com/fulcrumgenomics/util/ReadStructureTest.scala
+++ b/src/test/scala/com/fulcrumgenomics/util/ReadStructureTest.scala
@@ -76,7 +76,7 @@ class ReadStructureTest extends UnitSpec with OptionValues {
     an[Exception] shouldBe thrownBy { ReadStructure("+M+T") }
     an[Exception] shouldBe thrownBy { ReadStructure("+M70T") }
   }
-  
+
   it should "be built from segments while resetting their offset" in {
     compareReadStructuresResetOffset(Seq(T(Int.MaxValue, 1)), ReadStructure("1T"))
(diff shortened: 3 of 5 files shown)


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -377,6 +377,10 @@ class DemuxFastqs
  val omitControlReads: Boolean = false,
  @arg(doc="Mask bases with a quality score below the specified threshold as Ns") val maskBasesBelowQuality: Int = 0,
 ) extends FgBioTool with LazyLogging {
+  validate(
+    !readStructures.exists(_.cellBarcodeSegments.nonEmpty),
+    "DemuxFastqs does not support cell barcodes in read structures!",
```
nh13, line 382: [todo(blocking)](https://pullpo.io/cc?l=todo&d=blocking) 
Can you  alsoadd valdiation to:
1. `AnnotateBamWithUmis`
2. `ExtractUmisFromBam`

And any other tool that takes a read structure as a command line argument?

src/test/scala/com/fulcrumgenomics/fastq/FastqToBamTest.scala
```diff
@@ -383,6 +383,22 @@ class FastqToBamTest extends UnitSpec {
...
+    )
+    val rs  = ReadStructure("2C1S2C+T")
+    val bam = makeTempFile("fastqToBamTest.", ".bam")
+    new FastqToBam(input=Seq(r1), readStructures=Seq(rs), output=bam, sample="s", library="l", cellQualTag = Some("CY")).execute()
+    val recs = readBamRecs(bam)
+    recs should have size 2
+    recs(0).apply[String]("CB") shouldBe "GG-CC"
+    recs(0).apply[String]("CY") shouldBe "== =="
+    recs(1).apply[String]("CB") shouldBe "TA-AA"
+    recs(1).apply[String]("CY") shouldBe "== =="
```
nh13, line 399: 👍

src/test/scala/com/fulcrumgenomics/util/ReadStructureTest.scala
```diff
@@ -187,6 +193,7 @@ class ReadStructureTest extends UnitSpec with OptionValues {
     ReadStructure("5B101T").length shouldBe 2
     ReadStructure("123456789T").length shouldBe 1
     ReadStructure("10T10B10B10S10M").length shouldBe 5
+    ReadStructure("10T10B10B10S10M10C").length shouldBe 6
```
nh13, line 196: <chuckles> _length is defined on a read structure_

nh13 (approved):
LGTM, though I want @tfenne to weigh in too!

src/test/scala/com/fulcrumgenomics/util/ReadStructureTest.scala
clintval: I know... I don't like it.

src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
clintval: I think we're OK to leave those tools as-is. Both should ignore non-UMI non-template operators and they do already. I found a usage for `ExtractIlluminaRunInfo` but the tool's behavior is invariant of if you have a cell barcode or not. It's only `DemuxFastqs` I want to protect against since I'm not adding support for cell barcode extraction.

PR #1105, opened 2025-10-15
Log the JVM maximum runtime memory at startup

nh13:
Requested by @yfarjoun

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
+++ b/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
@@ -42,7 +42,6 @@ import java.net.InetAddress
 import java.nio.file.Paths
 import java.text.DecimalFormat
 import scala.annotation.unused
-import java.lang.Runtime
 
 /**
   * Main program for fgbio that loads everything up and runs the appropriate sub-command
@@ -171,8 +170,11 @@ class FgBioMain extends LazyLogging {
     val snappy     = if (new SnappyLoader().isSnappyAvailable) "with snappy" else "without snappy"
     val inflater   = if (IntelCompressionLibrarySupported) "IntelInflater" else "JdkInflater"
     val deflater   = if (IntelCompressionLibrarySupported) "IntelDeflater" else "JdkDeflater"
-    val maxMemory  = Runtime.getRuntime().maxMemory()
-    logger.info(s"Executing $tool from $name version $version as $user@$host on JRE $jreVersion with Max Heap Memory (-Xmx) $maxMemory with $snappy, $inflater, and $deflater")
+    val maxMemory  = {
+      val mib = Runtime.getRuntime.maxMemory() / (1024 * 1024)
+      if (mib >= 1024) f"${mib / 1024.0}%.1fg" else s"${mib}m"
+    }
+    logger.info(s"Executing $tool from $name version $version as $user@$host on JRE $jreVersion with max heap memory $maxMemory, $snappy, $inflater, and $deflater")
   }
 
   /** Prints a line of useful information when a tool stops executing. */


src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
```diff
@@ -170,7 +171,8 @@ class FgBioMain extends LazyLogging {
     val snappy     = if (new SnappyLoader().isSnappyAvailable) "with snappy" else "without snappy"
     val inflater   = if (IntelCompressionLibrarySupported) "IntelInflater" else "JdkInflater"
     val deflater   = if (IntelCompressionLibrarySupported) "IntelDeflater" else "JdkDeflater"
-    logger.info(s"Executing $tool from $name version $version as $user@$host on JRE $jreVersion $snappy, $inflater, and $deflater")
+    val maxMemory  = Runtime.getRuntime().maxMemory()
+    logger.info(s"Executing $tool from $name version $version as $user@$host on JRE $jreVersion with Max Heap Memory (-Xmx) $maxMemory with $snappy, $inflater, and $deflater")
```
clintval: Bytes make for a very large number. It's hard to grok how much memory was allowed as a human reader. You also have 2x `with` words for Snappy as well:

```console
[2025/10/15 05:49:14 | FgBioMain | Info] Executing Tool from fgbio version 3.1.0-cd465d0-SNAPSHOT as clint@macbook on JRE 17.0.14+0 with Max Heap Memory (-Xmx) 17179869184 with with snappy, JdkInflater, and JdkDeflater
```

Could we do?

```suggestion
    val maxMemMiB  = Runtime.getRuntime().maxMemory() / (1024 * 1024)
    logger.info(s"Executing $tool from $name version $version as $user@$host on JRE $jreVersion with max runtime memory of ${maxMemMiB}m $snappy, $inflater, and $deflater")
```

Which prints:

```console
[2025/10/15 05:49:14 | FgBioMain | Info] Executing Tool from fgbio version 3.1.0-cd465d0-SNAPSHOT as clint@macbook on JRE 17.0.14+0 with max runtime memory of 16384m with snappy, JdkInflater, and JdkDeflater
```

I avoided `-Xmx` in the log statement since `Runtime.maxMemory()` doesn't return always (but often does) return the value that was passed in as `-Xmx` but instead is platform-dependent and returns the maximum amount of heap memory the garbage collector thinks remains ([link](https://stackoverflow.com/a/79435094/3727678)).

nh13: I'd rather keep `-Xmx` in the logging statement since that's _why_ I want this logging statement for the users (who are not me) in the first place.  I.e. someone uses `JAVA_OPTS="-Xmx2g" fgbio` (from bioconda) and is surprised when the maximum memory is 1g (since bioconda has a wrapper script).  It's not perfect, but it's close enough.

I do like make it it MiB though, and good catch on 2x.

clintval: Understood 🫡 . I'm OK with that. It should be extremely rare someone is using a different garbage collector implementation, right?

Maybe we also update the wrapper script so it respects `JAVA_OPTS` too? I notice the script doesn't use `JAVA_OPTS` but does have a feature for `_JAVA_OPTIONS` which seems unconventional to me:

- https://github.com/bioconda/bioconda-recipes/blob/master/recipes/fgbio/fgbio.py

Here's a wrapper script I've been using:

- https://gist.github.com/clintval/bc760c2227b373a1957a23d2e551984d#file-make-jar-executable-sh-L19-L21

clintval (approved):
I like this much better, thanks!

PR #1106, opened 2025-10-23
Ensure file toRealPath() does not raise for file descriptors

clintval:
It would be nice to use an official release of commons, but in the meantime I'm opening this PR to track a feature I'd like fgbio to have. Specifically, the ability to use bash process substitution for file handles:

```bash
tool --input <( ... )
```


clintval:
> I hope it was ok to edit and replace commons after I did a release!

Very much OK! The progress of science shouldn't be slowed by being shy in Git trees.

PR #1108, opened 2025-10-30
feat: speedup CorrectUmis and add a new --revcomp option

nh13:


src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -157,7 +172,7 @@ class CorrectUmis
 
       val lengths = set.map(_.length)
       validate(lengths.size == 1, s"UMIs of multiple lengths found. Lengths: ${lengths.mkString(", ")}")
-      (set.toArray, lengths.head)
+      (set.map(_.toUpperCase).toArray, lengths.head)
```
nh13: here is where we make sure all the known/fixed UMIs are uppercase

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -185,7 +200,7 @@ class CorrectUmis
           missingUmisRecords += 1
           rejectOut.foreach(w => w += rec)
         case Some(umi: String) =>
-          val sequences = umi.split("-", -1)
+          val sequences = if (revcomp) umi.split("-", -1).map(Sequences.revcomp) else umi.split("-", -1)
```
nh13: here's where the new `--revcomp` argument is used

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -278,12 +293,25 @@ class CorrectUmis
     cachedResult match {
       case Some(result) => result
       case None         =>
-        val mismatches = umis.map(umi => Sequences.countMismatches(bases, umi))
-        val min        = mismatches.min
-        val matched    = (min <= maxMismatches) && (mismatches.count(m => m < min + this.minDistanceDiff) == 1)
-        val umiMatch   = UmiMatch(matched, umis(mismatches.indexOf(min)), min)
```
nh13: changes below:
1. The while condition: when searching across UMIs, if we find a perfect match, and then we can break the loop if we find another match that is within the minimum distance
2. Pass the next-best distance found so far to the `countMismatches` method so it can abort early if possible
3.  Don't keep all the results, just the min and next-best, since that's all we need to create the `UmiMatch`

ameynert (changes requested):
[issue](https://pullpo.io/cc?l=issue) 
It'd be good to see unit tests on `countMismatches`, though not required if you are satisfied that the existing tests on `findBestMatches` are sufficient.

Definitely needs an expansion of the unit tests on the end-to-end runs to test the `revcomp` flag.

src/main/scala/com/fulcrumgenomics/util/Sequences.scala
```diff
@@ -108,6 +108,26 @@ object Sequences {
...
+   * @note '''Preconditions:'''
+   *       - `s1` and `s2` must have the same length
+   *       - `max` is greater than or equal to zero
+   * */
+  @inline
+  def countMismatchesWithMax(s1: String, s2: String, max: Int): Int = {
+    var count = 0
+    var i = 0
+    while (i < s1.length && count < max) {
+      if (s1.charAt(i) != s2.charAt(i)) count += 1
```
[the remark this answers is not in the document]
nh13: Good idea.  Unfortunately, it didn't speed it up in my hands.  I think we'd want to bit encode the bases to compute hamming distance (see [fqtk](https://github.com/fulcrumgenomics/fqtk/blob/f74eb6c4939c2db123067e50c7e76c07777ebdc2/src/lib/mod.rs#L49) and [here](https://github.com/fulcrumgenomics/fqtk/blob/f74eb6c4939c2db123067e50c7e76c07777ebdc2/src/lib/bitenc.rs#L432)).  I still am not sure that's faster given the short length of these sequences.

I do like the idea of using a prefix trie or something like that when we have many sequences against which we're comparing, but that could be for a later PR.

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -77,8 +83,8 @@ object CorrectUmis {
     */
   def findUmiPairsWithinDistance(umis: Seq[String], distance: Int): Seq[(String,String,Int)] = {
     umis.tails.flatMap {
-      case x +: ys => ys.map(y => (x, y, Sequences.countMismatches(x, y))).filter(_._3 <= distance)
-      case _      => Seq.empty
+      case x +: ys => ys.map(y => (x, y, Sequences.countMismatchesWithMax(x.getBytes, y.getBytes, distance+1))).filter(_._3 <= distance)
+      case _       => Seq.empty
     }.toList
   }
```
tfenne: Looking at this code anew, I wonder why we chose to do `umi.tails.flatMap`.  I think this would likely be considerably more efficient as something like:

```suggestion
val builder = Seq[(String, String,Int)].builder()
forloop (from=0, until=umis.length) { i =>
  forloop (from=i+1, until=umis.length) { j => 
    val d = Sequences.countMismatchesWithMax(umis(i).getBytes, umis(j).getBytes, distance+1)
    if (d <= distance) builder += (umis(i), umis(j), d)
  }
}

builder.result()
```

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
```diff
@@ -272,18 +280,46 @@ class CorrectUmis
...
-        val umiMatch   = UmiMatch(matched, umis(mismatches.indexOf(min)), min)
+        var min = bases.length + 1
+        var minIndex = -1
+        var nextBest = bases.length + 1
+        var i = 0
+        val bytes = bases.getBytes
+        // Note: when the best match has `min == 0`, we still need to scan to see if we find
+        // the next best match (`nextBest`) with `nextBest < min + this.minDistanceDiff`.  If we do
+        // find `min == 0` and `nextBest < this.minDistanceDiff`, we are guaranteed not to match.
+        while (i < umis.length && !(min == 0 && nextBest < this.minDistanceDiff)) {
```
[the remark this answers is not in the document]
tfenne: And now I realize that `min == 0` is required ... which makes (1) irrelevant but emphasizes (2)

tfenne (approved):

src/main/scala/com/fulcrumgenomics/umi/CorrectUmis.scala
nh13: I'd actually prefer your idea of reversing when there are multiple barcodes.  The client I'm implementing does not have (and will not have) dual UMIs, so I prefer the intuitive approach as it does not affect the client.  Good catch!

PR #1114, opened 2025-11-25
fix: Template.fixMateInfo() to set primary mate info before supplemen…

nh13:
…tary info

Set primary mate info before supplementary alignment info so that supplementary alignments inherit correctly computed TLEN values instead of stale values from the aligner.


clintval (approved):
Good find, I agree!

PR #1116, opened 2025-11-29
fix: the consensus RX (UMI) should be called from filtered records only

nh13:


src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -239,7 +239,7 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
   }
 
   /** Creates a consensus read from the given records.  If no consensus read was created, None is returned. */
-  protected[umi] def consensusFromSamRecords(records: Seq[SamRecord]): Option[VanillaConsensusRead] = {
+  protected[umi] def consensusFromSamRecords(records: Seq[SamRecord]): Option[(VanillaConsensusRead, Seq[SourceRead])] = {
```
nh13: We now return the _filtered_ (by common alignment) source reads, from which we call the consensus UMI (to put into the RX tag).

tfenne (approved):

PR #1119, opened 2025-11-30
fix: preserve input order in filterToMostCommonAlignment

nh13:
## Summary

`filterToMostCommonAlignment` sorts reads by descending length internally but previously returned them in that sorted order rather than the original input order. This caused non-deterministic consensus calling in near-tie scenarios.

## Motivation

Even with Kahan summation (PR #1120), floating-point arithmetic is not perfectly associative—accumulation order can still affect which base wins in near-tie scenarios. Preserving input order ensures:

- **Determinism**: Same input always produces same output, regardless of read lengths
- **Consistency**: VanillaUmiConsensusCaller and DuplexConsensusCaller behave identically

## Changes

- Sort indices instead of records to track original positions
- Use a BitSet to mark kept indices during filtering  
- Reconstruct output in original order via linear scan (O(n) vs previous O(k log k) sort)

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
@@ -399,17 +399,19 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
     * NOTE: filtered out reads are sent to the `rejectsMethod` method and do not need further handling
     */
   protected[umi] def filterToMostCommonAlignment(recs: Seq[SourceRead]): Seq[SourceRead] = if (recs.size < 2) recs else {
-    val groups = new ArrayBuffer[AlignmentGroup]
-    val sorted = recs.sortBy(r => -r.length).toIndexedSeq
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/umi/VanillaUmiConsensusCaller.scala
```diff
@@ -246,7 +246,7 @@ class VanillaUmiConsensusCaller(override val readNamePrefix: String,
     }
     else {
       val sourceRecords   = records.flatMap(toSourceRead(_, this.options.minInputBaseQuality, this.options.qualityTrim))
-      val filteredRecords = filterToMostCommonAlignment(sourceRecords)
+      val filteredRecords = filterToMostCommonAlignment(sourceRecords).sortBy { rec => sourceRecords.indexOf(rec) }
```
nh13: I'd also be open to not requiring a sort, by instead returning a list of the _index_ of the read in the original list for those that are kept.

clintval (approved):

PR #1120, opened 2025-12-05
fix: improve the numeric stability of the consensus caller

nh13:
The first commit shows that we get different results when we change the order in which the bases are added to consensus caller.

The second commit makes the consensus caller numerically stable by storing the terms for the summation of each likelihood, and then when the sum is needed, sorting them (by magnitude) from smallest to biggest, and summing them in that order.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
@@ -29,7 +29,6 @@ import com.fulcrumgenomics.util.NumericTypes._
 import htsjdk.samtools.util.SequenceUtil
 
 import java.util
-import scala.collection.mutable.ArrayBuffer
 
 object ConsensusCaller {
   type Base = Byte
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/umi/ConsensusCaller.scala
```diff
@@ -81,12 +82,15 @@ class ConsensusCaller(errorRatePreLabeling:  PhredScore,
     */
   class ConsensusBaseBuilder {
     private val observations = new Array[Int](DnaBaseCount)
-    private val likelihoods = new Array[LogProbability](DnaBaseCount)
+
+    // Note: to ensure numerical stability, we store the terms we want to eventually sum, rather than storing the sum
+    // itself. We can then sum smallest (in magnitude) to largest (in magnitude) for numerical stability.
+    private val likelihoods = Range.inclusive(1, DnaBaseCount).map { _ => new ArrayBuffer[LogProbability](256) }.toArray
```
nh13: I _think_ sorting isn't that bad in terms of increased performance (run time), but if others disagree, how would you feel about merging this PR, then following up with another PR with those changes.

For example, instead a array of all values, since we're converting Phred values, we could have a fixed size array of zeros and increment the offsets based on the Phred value.  Then when we sum, we can instead sort the unique values rather than all the values.  We would convert the phred value to log probability right before the sum.

PR #1121, opened 2025-12-11
feat: cram support for AssignPrimers

FelixMoelder:
As the AssignPrimers subcommand does not allow CRAM input and output files this PR should enable support of these files.
When testing locally I still ran in an error returning: `Exception in thread "main" java.lang.ClassCastException: class htsjdk.samtools.SAMRecord cannot be cast to class com.fulcrumgenomics.bam.api.SamRecord (htsjdk.samtools.SAMRecord and com.fulcrumgenomics.bam.api.SamRecord are in unnamed module of loader 'app')`.

As I am not experienced in Scala I would be happy if someone would help out to get this running.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/util/AssignPrimers.scala
+++ b/src/main/scala/com/fulcrumgenomics/util/AssignPrimers.scala
@@ -1,6 +1,6 @@
 package com.fulcrumgenomics.util
 
-import com.fulcrumgenomics.FgBioDef.{FilePath, PathToBam, SafelyClosable}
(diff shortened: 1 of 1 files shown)


FelixMoelder:
@nh13 Thanks for your suggestion. I think adding the reference as a common argument would make much more sense to enable general support for all subcommands.
I just gave it a try by providing `-Dsamjdk.reference_fasta=<ref_path>` but I still get the error message mentioned above. It seems to be thrown in general when providing a cram file as input.

PR #1125, opened 2025-12-17
we don't want reads to reappear in zipperbam output bams if they were  removed (e.g. by adapter trimming) prior to mapping. 

bwlang:
removed (e.g. by adapter trimming) prior to mapping.


src/main/scala/com/fulcrumgenomics/bam/ZipperBams.scala
```diff
@@ -240,7 +241,9 @@ class ZipperBams
       }
       else {
         logger.debug(s"Found an unmapped read with no corresponding mapped read: ${template.name}")
-        out ++= template.allReads
+        if (!excludeMissingReads) {
+          out ++= template.allReads
```
nh13: **question:** would it be helpful to count the number of reads that were excluded when using this option, and log it at the end?

bwlang: yep - added a patch for that.

nh13 (approved):

nh13:
Thanks Brad, this will be helpful!

PR #1137, opened 2026-02-17
fix: prevent double-counting of rejected reads in consensus stats

### LINKED ISSUE #1135: Logic Error/Negative Values in CallMolecularConsensusReads Stats File
I am seeing inconsistent and negative values in the metrics output of CallMolecularConsensusReads for 2 out of 16 samples processed using the standard fgbio workflow. While consensus reads are successfully emitted, the summary statistics show a "Negative Used" count because the total rejected reads exceed the total considered reads. This is targeted data from germline and ctDNA aimed to be 1000x post-deduplication.

E.g.

key     value   description
raw_reads_considered    53487904        Total raw reads considered from input file
raw_reads_rejected      56735091        Total number of raw reads rejected before consensus calling
raw_reads_used  -3247187        Total count of raw reads used in consensus reads
frac_raw_reads_used     -0.060709       Fraction of raw reads used in consensus reads
raw_reads_rejected_for_insufficient_support     28299863        Insufficient reads to generate a consensus
raw_reads_rejected_for_minority_alignment       81740   Reads has a different, and minority, set of indels
raw_reads_rejected_for_orphan_consensus 28353462        Only one of R1 or R2 consensus generated
raw_reads_rejected_for_zero_bases_post_trimming 26      Read or mate had zero bases post trimming
consensus_reads_emitted 10677500        Total number of consensus reads (R1+R2=2) emitted.

Environment:

fgbio version: 3.1.1. Using conda installation

Command used: 
fgbio CallMolecularConsensusReads  --input=my.bam --output=consensus.bam --stats=consensus_stats.tsv --error-rate-post-umi 40  --error-rate-pre-umi 45  --output-per-base-tags false  --min-reads 2  --sort-order=queryname --max-reads 50  --min-input-base-quality 20  --read-name-prefix='consensus'

nh13:
## Summary
- Fixes #1135 — `CallMolecularConsensusReads` could produce negative `raw_reads_used` in the stats file because reads were double-counted as rejected
- Reads rejected individually (`ZeroPostAfterTrimming`, `MinorityAlignment`, `InsufficientSupport`) inside `consensusFromSamRecords()` were being re-rejected as `OrphanConsensus` in the paired-end orphan handler
- Made `rejectRecords()` idempotent: it now skips records already tagged with `rr`, so duplicate rejections are impossible regardless of call site
- Extracted `RejectReasonTag` constant for the `rr` SAM tag

## Test plan
- [x] Added test for `(None, None)` case: both ends fail — previously 12 rejections for 6 reads, now correctly 6
- [x] Added test for `(None, Some)` case: one end fails, other has partial individual rejections — previously 7 rejections for 6 reads, now correctly 6
- [x] Both tests assert exact `totalFiltered` values (not just `<= totalReads`)
- [x] All 50 `VanillaUmiConsensusCallerTest` tests pass

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
@@ -53,6 +53,9 @@ object UmiConsensusCaller {
     val ReadTypeKey: String = "__read_type__"
   }
 
+  /** SAM tag used to record the rejection reason on rejected records. */
+  val RejectReasonTag: String = "rr"
+
   /** Enum entry type for why we reject SamRecords during consensus calling. */
   sealed trait RejectionReason extends EnumEntry {
     val code: String
@@ -222,11 +225,12 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
 
   /** Records that the supplied records were rejected, and not used to build a consensus read. */
   protected def rejectRecords(recs: Iterable[SamRecord], reason: RejectionReason) : Unit = {
-    this._filteredReads.count(reason, recs.size.toLong)
+    val toReject = recs.filterNot(_.contains(UmiConsensusCaller.RejectReasonTag)).toIndexedSeq
+    toReject.foreach(_(UmiConsensusCaller.RejectReasonTag) = reason.code)
+    this._filteredReads.count(reason, toReject.size.toLong)
     this.rejectsWriter.foreach { writer =>
-      recs.foreach(_("rr") = reason.code)
       writer.synchronized {
-        writer ++= recs
+        writer ++= toReject
       }
     }
   }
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/UmiConsensusCaller.scala
```diff
@@ -222,11 +225,12 @@ trait UmiConsensusCaller[ConsensusRead <: SimpleRead] {
 
   /** Records that the supplied records were rejected, and not used to build a consensus read. */
   protected def rejectRecords(recs: Iterable[SamRecord], reason: RejectionReason) : Unit = {
-    this._filteredReads.count(reason, recs.size.toLong)
+    val toReject = recs.filterNot(_.contains(UmiConsensusCaller.RejectReasonTag)).toIndexedSeq
```
clintval: On looking at this again, I started to think we should partition the collection and append the new reject reasons to the value at `rr` (making it a array of strings) so we have all rejection reasons. But then I noticed we're writing them out to a rejects BAM. So I don't think we can do that since the first time a record is rejected, it's written and it is gone.

PR #1142, opened 2026-03-03
feat: consider a cell barcode when sorting by template coordinate

clintval:
For single-cell technologies, a cell barcode may be present on SAM records and should be used as a grouping key during `--template-coordinate` sort. This ensures templates at the same locus are grouped together successfully when those templates are known to come from different cells.

Companion PR: https://github.com/samtools/samtools/pull/2314

@nh13 would you be willing to ensure this is added to `fgumi` too or would you welcome me to make that issue/PR?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
@@ -27,8 +27,7 @@ package com.fulcrumgenomics.bam.api
 import com.fulcrumgenomics.umi.ConsensusTags
 import htsjdk.samtools.SAMFileHeader.{GroupOrder, SortOrder}
 import htsjdk.samtools.util.Murmur3
-import htsjdk.samtools.{SAMFileHeader, SAMTag, SAMUtils}
-
+import htsjdk.samtools.{SAMFileHeader, SAMTag}
 
 /** Trait for specifying BAM orderings. */
 sealed trait SamOrder extends Product {


src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala
```diff
@@ -164,8 +164,9 @@ object SamOrder {
 
   /**
     * The sort order used by GroupReadByUmi. Sorts reads by the earlier unclipped 5' coordinate of the read
-    * pair, the higher unclipped 5' coordinate of the read pair, library, the molecular identifier (see
-    * [[com.fulcrumgenomics.umi.ConsensusTags.MolecularId]]), read name, and if R1 has the lower coordinates of the pair.
+    * pair, the higher unclipped 5' coordinate of the read pair, library, cellular barcode, the molecular identifier
+    * (see [[com.fulcrumgenomics.umi.ConsensusTags.MolecularId]]), read name, and if R1 has the lower coordinates of
+    * the pair.
```
nh13: **nitpick:** check if the order matches below

clintval:
I don't regard this as a **_breaking change_** since the tool's contract never specified anything about cell barcodes. The output of our template coordinate sort will still be backwards compatible with ecosystem tooling. When the `CB` tag is not present, the tool will behave exactly the same as it did before. When the tags are present, templates that would have otherwise been grouped across cells will now be grouped by cell which could be considered an an enhancement for correctness-sake.

The samtools PR was merged so I think we should be OK to merge this PR as well:

- https://github.com/samtools/samtools/pull/2314

PR #1150, opened 2026-03-27
#960: bump htsjdk to 4.2.0, sbt-cov to 2.4.4, fix sbt issues

### LINKED ISSUE #960: SetMateInformation should add the MQ tag to any supplementary alignments
Currently, `SetMateInformation` does not add the `MQ` tag to supplementary alignments. 

I'm not sure if this is intended behavior - the docs suggest that supplementary alignments would have this tag added. 

> Adds and/or fixes mate information on paired-end reads. Sets the MQ (mate mapping quality), 'MC' (mate cigar string),
ensures all mate-related flag fields are set correctly, and that the mate reference and mate start position are
correct.  
>
>  Supplementary records are handled correctly (updated with their mate's non-supplemental attributes). Secondary
alignments are passed through but are not updated.

`GroupReadsByUmi` requires that all alignments, including supplementary alignments, have the `MQ` tag set, so it would be helpful if `SetMateInformation` produced a compatible BAM 🙂

roland-ewald:
This bumps htsjdk from 3.0.5. to 4.2.0 (to fix #960); there are no compile errors and all tests pass.

It also fixes a broken link in the README.

## Other changes (to fix the build)

- Running 'sbt assembly' runs into an issue because sbt-scoverage 1.6.1 requires another scala-xml version (so this bumps the sbt-coverage version to 2.4.4).

- Running 'sbt assembly' also runs into trouble with duplicate 'module-info.class' files inside the dependencies, so this updates the merge strategy to discard those files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Contributions are welcome and encouraged.
 We will do our best to provide an initial response to any pull request or issue within one-week.
 For urgent matters, please contact us directly.
 
-See [Contributing](Contributing.md) for more details.
+See [Contributing](CONTRIBUTING.md) for more details.
 
 ## Authors
 
--- a/build.sbt
+++ b/build.sbt
@@ -236,7 +236,7 @@ val customMergeStrategy: String => MergeStrategy = {
         MergeStrategy.filterDistinctLines
       case ("spring.schemas" :: Nil) | ("spring.handlers" :: Nil) =>
         MergeStrategy.filterDistinctLines
-      case ("versions" :: "9" :: "module-info.class" :: Nil) =>
+      case ("versions" :: _ :: "module-info.class" :: Nil) =>
         MergeStrategy.discard
       case _ => MergeStrategy.deduplicate
     }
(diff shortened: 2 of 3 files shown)


nh13 (changes requested):
Thank-you for putting together this PR, contributions are very welcome!

1. Can you squash the commits into one commit (the changes are small)?
2. Can you fix the nitpick?
3. Can you update the badge in the readme to drop Java 8/11 and show 17+ only?
4. Can you update the CI to test 17+ (should we see if there are newer CI versions)?

Before we merge this, I will need to investigate:
  - Are there any htsjdk API deprecations or behavioral changes between 3.0.5 and 4.2.0 that could affect fgbio at runtime but wouldn't be caught by existing tests?                                                                                                                                    
  - htsjdk 4.x dropped Java 8 support and moved to Java 17 minimum: does this cause any issues to downstream users?

roland-ewald:
@nh13 Wow, thanks! :+1: Sorry, did not yet have time to work in your change requests :sweat_smile: --- great to see #1151 merged! :rocket:

PR #1154, opened 2026-04-04
Switch from intel inflater/deflater to libdeflate.

tfenne:
Running on a test BAM (rewriting it with RemoveSamTags), the tool runs about 2X as fast on a recent intel AWS instance (figure that's giving the intel inflater/deflater a fighting chance).


src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
```diff
@@ -105,10 +103,10 @@ class FgBioMain extends LazyLogging {
...
     htsjdk.samtools.util.Log.setGlobalLogLevel(htsjdk.samtools.util.Log.LogLevel.WARNING)
 
-    // Use the Intel Inflater/Deflater if available
-    if (IntelCompressionLibrarySupported) {
-      BlockCompressedOutputStream.setDefaultDeflaterFactory(new IntelDeflaterFactory)
-      BlockGunzipper.setDefaultInflaterFactory(new IntelInflaterFactory)
+    // Use jlibdeflate for accelerated compression/decompression if available
+    if (LibdeflateSupport.isSupported) {
+      BlockCompressedOutputStream.setDefaultDeflaterFactory(new LibdeflateDeflaterFactory)
+      BlockGunzipper.setDefaultInflaterFactory(new LibdeflateInflaterFactory)
```
nh13, line 109: **question (blocking):** what do you think about having an opt-out system property that falls back to the intel compression library, in case of any issues for a quarter, since the libdeflate dependency was just created ;)

You could:
1. `-Dfgbio.use.libdeflate=false`
2. and/or add a common arg `--use-intel-deflate`

src/main/scala/com/fulcrumgenomics/util/LibdeflateSupport.scala
```diff
@@ -0,0 +1,180 @@
...
+    inputLen = len
+  }
+
+  /** Decompresses the input into the output buffer in a single call. The caller must provide
+    * the exact expected uncompressed size as `len` (as HTSJDK does from the BGZF block footer).
+    * Returns `len` on success.
+    */
+  override def inflate(output: Array[Byte], off: Int, len: Int): Int = {
+    decompressor.deflateDecompress(inputBuf, inputOff, inputLen, output, off, len)
+    len
```
nh13: **question:** 
does `deflateCompress` return a value in case of an error (e.g. `-1` or a value `<len`)?  Can you verify?

src/main/scala/com/fulcrumgenomics/util/LibdeflateSupport.scala
```diff
@@ -0,0 +1,180 @@
...
+  * @param nowrap if true, use raw DEFLATE (no zlib/gzip wrapper). Must be true for BGZF.
+  */
+class LibdeflateInflater(nowrap: Boolean) extends Inflater(nowrap) {
+  // Release the JDK's native zlib allocation immediately; we manage our own native handle
+  super.end()
+
+  private val decompressor = new LibdeflateDecompressor()
+  private var inputBuf: Array[Byte] = _
+  private var inputOff: Int = 0
+  private var inputLen: Int = 0
```
nh13, line 119: what, no `=` alignment? :P

src/test/scala/com/fulcrumgenomics/util/LibdeflateSupportTest.scala
```diff
@@ -0,0 +1,440 @@
...
+    val bam = makeTempFile("benchmark", ".bam")
+    builder.write(bam)
+    bam
+  }
+
+  // ---------------------------------------------------------------------------
+  // LibdeflateDeflater unit tests
+  // ---------------------------------------------------------------------------
+
+  Seq(0, 1, 5, 6, 9).foreach { level =>
```
nh13, line 106: **thought:** I've found that libdeflate's level 1 compresses almost as well as level 6, but much faster.  Given this PR, consider a followup effort to validate this.  It could reduce speed for a marginal increase in output size.

nh13 (changes requested):
Just a couple things worth discussing, but nothing that I don't think will be a big block.  Thanks for putting this together!

src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala
tfenne: I'd really rather not do this.  I think the IntelDeflater is both outclassed by libdeflate _and_ supported on fewer platorms (no aarch64 support).  I'd trust that the new route works, and issue a patch quickly if problems show up than clutter the command line with options that would break back-compat when we remove them (or have to leave them around forever).

nh13: 👍

nh13 (approved):

PR #1155, opened 2026-04-17
fix(codec): avoid MatchError on dovetail FR pairs with asymmetric isFrPair

nh13:
## Summary

- `CallCodecConsensusReads` previously threw `scala.MatchError` deep in a worker thread after hundreds of millions of records on some CODEC datasets. The crash was deterministic once a template with a specific geometry was encountered.
- The failure site is a per-record `filter(_.isFrPair)` immediately followed by a rigid `Seq(rec, mate)` pattern match at `CodecConsensusCaller.scala:176`. When a read-name bucket contained only one record, that match blew up.
- Root cause is an asymmetry in htsjdk's `SamPairUtil.getPairOrientation`: for dovetail FR pairs whose aligned ends coincide (producing TLEN=±1 after htsjdk's own insert-size math) the call returns FR for R1 but RF for R2 on the same pair. The CODEC caller kept one mate and dropped the other, then crashed on the resulting singleton bucket.
- A secondary fix in this PR also preserves BAM-iteration order of templates when grouping primaries by read name. The previous `groupBy(_.name).toSeq` made `longestR1Alignment` / `longestR2Alignment` selection non-deterministic across JVMs and could pair records from different templates, producing spurious `IndelErrorBetweenStrands` rejections.

## Fix

Group primary alignments by read name **first**, then keep only buckets that form a single primary FR pair, where the FR check is computed at the pair level from the unclipped 5′ positions of both records (so the answer is symmetric in its arguments). Templates that do not form a clean primary FR pair — singletons, tandem, RF, cross-contig, or with an unmapped mate — are now rejected cleanly with a new `NotPrimaryFrPair` reason instead of throwing.

## Secondary fix: deterministic template ordering

`consensusSamRecordsFromSamRecords` selects the longest R1 and R2 alignments via `maxBy(_.cigar.lengthOnTarget)`, which returns the first element in iteration order on ties. The previous `pairs.filterNot(...).groupBy(_.name).toSeq` pattern produced a template order that depended on the JVM's `String#hashCode` and `HashMap` layout rather than BAM-iteration order. On ties this could pick R1 and R2 from different templates whose geometry did not agree, producing spurious `IndelErrorBetweenStrands` rejections.

A new `CodecConsensusCaller.orderedPrimaryPairs` helper groups by read name while preserving BAM-iteration order (order of first occurrence of each read name via `Iterator#distinct`). The caller now delegates to the helper. Verified locally that `Seq("zulu", "alpha", "mike", "bravo", "yankee").groupBy(identity).toSeq.map(_._1)` yields `List(alpha, yankee, bravo, zulu, mike)` on OpenJDK 17 — a concrete example of the hash-order/BAM-order mismatch that the new test locks down.

## Minimal repro

The failing stack trace:

```
Exception in thread "main" scala.MatchError: (<readname>, Vector(<readname> 1/2 129b aligned to <chrom>:X-Y.)) (of class scala.Tuple2)
  at com.fulcrumgenomics.umi.CodecConsensusCaller.$anonfun$consensusSamRecordsFromSamRecords$8(CodecConsensusCaller.scala:176)
```

The offending template has both R1 and R2 present as primary alignments, both mapped to the same contig, same MI, same UMI, and an identical aligned boundary — the reverse-strand read's `alignmentEnd` coincides with the forward-strand read's `alignmentStart`, and heavy soft-clipping at both ends causes htsjdk to compute `TLEN=±1`.

Schematically:

```
R1  FLAG=97   POS=X     CIGAR=68S53M8S   TLEN=+1   (forward)
R2  FLAG=145  POS=X-47  CIGAR=28S48M53S  TLEN=-1   (reverse)
```

With `X = R2.alignmentEnd = R1.alignmentStart`, htsjdk's `getPairOrientation` returns FR on R1 (comparing `R1.alignmentStart` to `R1.alignmentStart + TLEN = X+1`) and RF on R2 (comparing `R1.alignmentStart` to `R2.alignmentEnd`, both `X`, strict `<` is false). Per-record disagreement → one mate filtered → singleton bucket → `MatchError`.

The geometry is exercised directly in a new unit test using `SamBuilder` — no real data or BAMs committed. A separate smoke-test against a 2-record BAM extracted from a real offending template (not committed) confirmed the tool no longer crashes and instead rejects the template as `IndelErrorBetweenStrands` downstream, which is expected — the CODEC caller isn't wired for dovetails whose aligned regions barely overlap; that is a separate, larger change.

## Tests

- `CodecConsensusCaller.isPrimaryFrPair`: unit tests for symmetry, canonical FR, RF, cross-chromosomal, one-mate-unmapped, tandem.
- `CodecConsensusCaller.orderedPrimaryPairs`: unit tests locking down BAM-iteration order using names that `HashMap` would reorder, plus a test documenting the "caller pre-filters secondary/supplementary" contract.
- `consensusSamRecordsFromSamRecords` regression test that reproduces the dovetail geometry (`68S53M8S` / `28S48M53S`, aligned ends coinciding) and asserts the caller does not throw.
- Singleton-bucket regression test (defensive path) that asserts the caller rejects cleanly instead of crashing.

Full `sbt test` passes locally.

## Follow-ups (out of scope)

- The per-record asymmetry in htsjdk's `SamPairUtil.getPairOrientation` is arguably a bug; a separate htsjdk issue/PR is planned.
- Other uses of `SamRecord.isFrPair` elsewhere in fgbio inherit the same per-record asymmetry but do not crash; they may produce latent correctness issues on dovetail pairs and are worth auditing separately.

## Test plan

- [x] Reproduce MatchError with a SamBuilder test built on current `main`.
- [x] Verify fix prevents the MatchError in that test.
- [x] Lock down BAM-iteration order of primary-pair templates via a dedicated unit test that would fail against the broken `groupBy(_.name).toSeq` path.
- [x] Verify fix on an offline smoke-test BAM extracted from a real failing MI family (not committed).
- [x] Full `sbt test` passes.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
@@ -170,10 +170,24 @@ class CodecConsensusCaller(readNamePrefix: String,
       Nil
     }
     else {
-      // Extract just the primary alignments and then clip where they extend past the ends of their mates
-      // TODO: Remove the isFrPair here and handle chimeric reads or reads with one unmapped etc.
-      val primaries = pairs.filterNot(r => r.secondary || r.supplementary).filter(_.isFrPair)
-      primaries.groupBy(_.name).foreach { case (_, Seq(rec, mate)) => clipper.clipExtendingPastMateEnds(rec, mate) }
+      // Group the primary alignments by read name and retain only templates that form a single
+      // primary FR pair. The FR check is performed at the pair level (with both records in hand)
+      // because htsjdk's per-record `SamPairUtil.getPairOrientation` can disagree between mates
+      // on dovetail pairs whose aligned ends coincide (TLEN=+/-1), which previously caused a
+      // `scala.MatchError` on a singleton read-name bucket here.
+      // Template order is preserved from BAM-iteration order (not hash order) so that downstream
+      // `maxBy` tie-breaks on `cigar.lengthOnTarget` are stable across JVMs.
+      // TODO: handle chimeric reads or reads with one unmapped etc.
+      val primaryPairs = CodecConsensusCaller.orderedPrimaryPairs(
+        pairs.filterNot(r => r.secondary || r.supplementary)
+      )
+      val (frPairs, nonFrPairs) = primaryPairs.partition {
+        case (_, Seq(a, b)) => CodecConsensusCaller.isPrimaryFrPair(a, b)
+        case _              => false
+      }
+      rejectRecords(nonFrPairs.flatMap(_._2), RejectionReason.NotPrimaryFrPair)
+      frPairs.foreach { case (_, Seq(rec, mate)) => clipper.clipExtendingPastMateEnds(rec, mate) }
+      val primaries = frPairs.flatMap(_._2)
 
       val r1s = filterToMostCommonAlignment(primaries.filter(_.firstOfPair).map(toSourceReadForCodec))
       val r2s = filterToMostCommonAlignment(primaries.filter(_.secondOfPair).map(toSourceReadForCodec))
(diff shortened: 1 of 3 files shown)


src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -170,10 +170,23 @@ class CodecConsensusCaller(readNamePrefix: String,
...
-      // Extract just the primary alignments and then clip where they extend past the ends of their mates
-      // TODO: Remove the isFrPair here and handle chimeric reads or reads with one unmapped etc.
-      val primaries = pairs.filterNot(r => r.secondary || r.supplementary).filter(_.isFrPair)
-      primaries.groupBy(_.name).foreach { case (_, Seq(rec, mate)) => clipper.clipExtendingPastMateEnds(rec, mate) }
+      // Group the primary alignments by read name and retain only templates that form a single
+      // primary FR pair. The FR check is performed at the pair level (with both records in hand)
+      // because htsjdk's per-record `SamPairUtil.getPairOrientation` can disagree between mates
+      // on dovetail pairs whose aligned ends coincide (TLEN=+/-1), which previously caused a
+      // `scala.MatchError` on a singleton read-name bucket here.
+      // TODO: handle chimeric reads or reads with one unmapped etc.
```
tfenne: Can you link to the HTSJDK PR here, so we know when we can revert to it's FrPair code?

src/main/scala/com/fulcrumgenomics/umi/CodecConsensusCaller.scala
```diff
@@ -375,3 +388,21 @@ class CodecConsensusCaller(readNamePrefix: String,
...
+    */
+  private[umi] def isPrimaryFrPair(a: SamRecord, b: SamRecord): Boolean = {
+    if (!a.mapped || !b.mapped) false
+    else if (a.refIndex != b.refIndex) false
+    else if (a.positiveStrand == b.positiveStrand) false
+    else {
+      val (fwd, rev) = if (a.positiveStrand) (a, b) else (b, a)
+      fwd.unclippedStart <= rev.unclippedEnd
+    }
+  }
```
tfenne, line 433: Could you just find the higher coordinate of the two reads, and then call htsjdk's method here?  Your main problem was lack of consistency, and doing _anything_ on the pair gets you there.

PR #1156, opened 2026-04-19
Rename duplex-filter tier locals to best/worst

nh13:
## Summary

- Rename locals in `filterDuplexConsensusRead` and `DuplexConsensusPerBaseValues.maskBaseAt` from `abMaxDepth`/`baMaxDepth`/`abError`/`baError` (and the per-base equivalents) to `bestDepth`/`worstDepth`/`bestError`/`worstError`.
- Add comments clarifying that the `ab`/`ba` prefix on the threshold structs (`abFilters` / `baFilters`) is a **tier label** (strict / lenient), not a biological strand reference — the locals being sorted are per-metric extremes across the two strands, independently per metric.
- Pure rename. No behavior change.

## Why

The existing names conflate strand identity with filter tier. A reader (or a static analyzer) can easily misread `abMaxDepth < abFilters.minReads` as "AB strand depth vs AB threshold" and then flag the `abError` pairing (which is "min error across both strands") as a bug where the "AB strand's error" is being mismatched to the "BA strand's error rate". In practice neither local is a biological AB/BA strand — they're the best and worst per-metric values. The new names make that visible at each call site without changing semantics.

Parallel Rust rename in fgumi: fulcrumgenomics/fgumi#299.

## Test plan
- [ ] CI (sbt `FilterConsensusReadsTest`) passes

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
+++ b/src/main/scala/com/fulcrumgenomics/umi/FilterConsensusReads.scala
@@ -302,11 +302,17 @@ class FilterConsensusReads
   private def filterDuplexConsensusRead(rec: SamRecord): FilterResult = {
     val failsReadLevelChecks = {
       import ConsensusTags.PerRead._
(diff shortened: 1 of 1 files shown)


clintval (approved):
Makes sense to me!

PR #1157, opened 2026-04-23
feat!: allow non-terminal + segment in read structures

tfenne:
The `+` (indefinite-length) segment can now appear at any position within a read structure, not only the last, mirroring the Rust `read-structure` crate PR #14 on the Scala side. Extract resolves middle- and leading-+ structures by walking backward from the end of the read for post-+ segments.

Breaking API changes:
 - ReadSegment.offset removed; ReadStructure tracks offsets internally via a private signed array. Constructors now take (length, kind) instead of (offset, length, kind).
 - ReadStructure.apply(segments, resetOffsets) simplified to apply(segments); offsets are always computed.
 - Fixed-length structures reject over-long reads instead of silently truncating. Callers that may see variable-length input should route it through withVariableLastSegment.
 - extract gained an includeSkips: Boolean = false parameter; Skip segments are now omitted from extract results by default. All in-tree callers were already filtering by kind, so no behavior change for them.

ExtractUmisFromBam.updateClippingInformation now walks segments accumulating a running non-template-length sum before the clip position. It requires all but the last segment to be fixed-length, since translating a clip past a non-terminal + would need the read length.

The + segment still means zero-or-more bases here, deliberately differing from the Rust crate's one-or-more; DemuxFastqs relies on zero-or-more to handle short/trimmed reads via
withVariableLastSegment.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -244,9 +244,10 @@ object DemuxFastqs {
       |3. `M` identifies a unique molecular index read
       |4. `S` identifies a set of bases that should be skipped or ignored
       |
-      |The last `<number><operator>` pair may be specified using a `+` sign instead of number to denote "all remaining
-      |bases". This is useful if, e.g., fastqs have been trimmed and contain reads of varying length. Both reads must
-      |have template bases.  Any molecular identifiers will be concatenated using
+      |At most one `<number><operator>` pair in a read structure may use a `+` sign in place of the number to denote
+      |"all remaining bases", and it may appear at any position in the read structure.  This is useful if, e.g.,
+      |fastqs have been trimmed and contain reads of varying length. Both reads must have template bases.  Any
+      |molecular identifiers will be concatenated using
       |the `-` delimiter and placed in the given SAM record tag (`RX` by default).  Similarly, the sample barcode bases
       |from the given read will be placed in the `BC` tag.
       |
--- a/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
+++ b/src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
@@ -245,11 +245,10 @@ object DemuxFastqs {
       |4. `S` identifies a set of bases that should be skipped or ignored
       |
       |At most one `<number><operator>` pair in a read structure may use a `+` sign in place of the number to denote
-      |"all remaining bases", and it may appear at any position in the read structure.  This is useful if, e.g.,
-      |fastqs have been trimmed and contain reads of varying length. Both reads must have template bases.  Any
-      |molecular identifiers will be concatenated using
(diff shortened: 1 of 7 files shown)


src/main/scala/com/fulcrumgenomics/fastq/DemuxFastqs.scala
```diff
@@ -244,9 +244,10 @@ object DemuxFastqs {
...
       |4. `S` identifies a set of bases that should be skipped or ignored
       |
-      |The last `<number><operator>` pair may be specified using a `+` sign instead of number to denote "all remaining
-      |bases". This is useful if, e.g., fastqs have been trimmed and contain reads of varying length. Both reads must
-      |have template bases.  Any molecular identifiers will be concatenated using
+      |At most one `<number><operator>` pair in a read structure may use a `+` sign in place of the number to denote
+      |"all remaining bases", and it may appear at any position in the read structure.  This is useful if, e.g.,
+      |fastqs have been trimmed and contain reads of varying length. Both reads must have template bases.  Any
+      |molecular identifiers will be concatenated using
       |the `-` delimiter and placed in the given SAM record tag (`RX` by default).  Similarly, the sample barcode bases
```
clintval: Line lengths are a bit broken here.

Should render fine, but looks odd in the source code.

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
```diff
@@ -427,5 +427,25 @@ class ExtractUmisFromBamTest extends UnitSpec with OptionValues {
...
+
+    // clip position exactly on the next segment's start when that segment is a template
+    record("XT") = 20
+    ExtractUmisFromBam.updateClippingInformation(record, Some("XT"), ReadStructure("10M10T80T"))
+    record[Int]("XT") shouldBe 10
+
+    // multiple template segments with a mol barcode between them; clip lands in the second template
+    record("XT") = 40
+    ExtractUmisFromBam.updateClippingInformation(record, Some("XT"), ReadStructure("20T10M70T"))
+    record[Int]("XT") shouldBe 30
```
clintval, line 449: Decent tests, but did you mean to add any read segments with the `+` operator before/after the UMI segment to flex the new functionality?

clintval (approved):
Why does the Scala implementation allow a zero-length `+` operator but the Rust implementation does not? Would you be willing to update our wiki docs on read structures here too?

https://github.com/fulcrumgenomics/fgbio/wiki/Read-Structures

src/test/scala/com/fulcrumgenomics/umi/ExtractUmisFromBamTest.scala
tfenne: I've added a positive test for a trailing `+T`, but the tool actually explicitly checks for and forbids a `+` segment in the non terminal position (separate and apart from the read-structure parsing), so I've added a test that that check still fires.  

Throughout this PR I've avoided changing behavior other than the RS parsing.  I think we could entertain separate PRs to loosen restrictions in tools where it makes sense.

PR #1172, opened 2026-08-14
fix(bam): clip past the mate's end using query, not reference, distances

### LINKED ISSUE #1090: Update SamRecordClipper's functions to "clip past end of mate" to use distances in query bases, not reference bases
SamRecordClipper has function(s) to trim reads that extend beyond the ends of their mates.  I.e. if you have an FR pair where the reads and longer than the insert size, to trim the reads back to the insert size.

A change was made to change the trim point to [take into account softclipping](https://github.com/fulcrumgenomics/fgbio/pull/1026) and reduce how aggressive we are in clipping.  

However, with the change it is possible to create situations where a read is now either over- or under-clipped.  This only comes up in rather infrequent cases involving:

- A read pair that sequences all the way through the insert and into the adapters
- A mapping location in the reference where at least 1-2 bases of adapter will align to the reference
- An indel of sufficient size / proximity to one end of the read to disrupt the alignment in the presence of other sequencing errors

For example, say we have 150bp reads from an insert that a) matches to 90bp of the reference sequence and b) has a 10bp insertion 30bp from one side.  We might get alignments that look like this:

```
r1 strand=+  start=1234500  cigar=70M10I23M47S  end=1234592 // 3bp of adapter sequence matches reference
r2 strand=-  start=1234500  cigar=50S70M30S     end=1234569
```

When go to see if R1 extends past the start of R2, we calculate the "unclipped" end of R2 as `end+clipping` which comes out to `1234569 + 30` = 1234599.  We see that is beyond the mapped end position of R1, into the soft-clipping, therefore a) don't remove any aligned bases and b) if in hard-clipping mode, remove only part of the soft-clipped bases at the end of R1.

Alternatively, if a deletion is present instead of an insertion, then over-clipping occurs.

All of this occurs because we are using distances on the reference and distances on the query sequence interchangeably in the calculation of where to clip, which breaks down in the presences of indels.  I believe what we should be doing instead is as follows:

1. Quick/efficient check to see if clipping is definitely not needed because a) the read being examined has no soft-clipping at it's end _and_ it's end position is not past the start position of it's mate
2. Calculate the last position at which the read and it's mate share a _reference_ position (in the example above this would be at `1234569` when clipping R1)
3. Calculate how many query bases each read has past the last shared reference position (in the example above this would be 80 for R1 and 30 for R2)
4. Remove bases such that there are the same number of query bases after the last shared reference position (in the example above, remove 50 bases from the end of R1, bringing it's cigar to `70M10I20M50H`)

nh13:
Fixes #1090.

`SamRecordClipper`'s "clip past the far end of the mate" functions mixed reference-space and query-space distances. `numBasesExtendingPastMate` took the mate's *un-soft-clipped* end — a reference coordinate produced by adding the mate's trailing soft-clip *length* (a query distance) to its alignment end — and then converted that coordinate back into a query offset on the record with `readPosAtRefPos`. With no indels the two spaces coincide and the answer is right; with an indel near either end of either alignment they diverge, and the read is over- or under-clipped.

Two concrete failure modes, both reachable in duplex consensus calling of libraries with adapter read-through:

- **Over-clipping.** If the synthesized coordinate lands inside a deletion in the record, `readPosAtRefPos(..., returnLastBaseIfDeleted=false)` returns `0`, so the computed clip becomes the record's entire length and the record is un-mapped outright. Worse, the mate was then clipped against the just-un-mapped record and so was left unclipped, which can take out a whole template.
- **Under-clipping.** If the record has an insertion before that coordinate, its alignment end falls short of where its query bases reach, so the record keeps bases past the end of the insert (the example in the issue).

## What changed

`numBasesExtendingPastMate` now works entirely in query space, using a shared reference position only as a landmark, as described in the issue:

1. Find the reference positions covered by both alignments.
2. Count how many query bases each read has beyond the relevant end of that shared span — the last shared position for a positive-strand record (its 3' end is on the right), the first shared position for a negative-strand record.
3. Both reads sequenced the same insert, so they must have the same number of query bases beyond that landmark. The excess is what is clipped.

### When the two alignments share no reference position

There is no landmark to count from, so this case keeps the estimate `numBasesExtendingPastMate` has applied since #842: project the record's soft-clipped tail into reference space at one query base per reference base and compare it against the mate's un-soft-clipped far end.

Answering `0` here — which an earlier revision of this PR did — is wrong, and the reason is worth stating, because "no overlap means no read-through" is the intuitive reading and it is false. Both alignments lie *within* the insert, so a pair that read through into the adapter can still leave them disjoint: it happens whenever each alignment stops short of the other's, which soft clipping at a variant or a mis-called base is enough to produce. Concretely, two 100 base reads over a 39 base insert, aligned `20M80S` at 1000 and `80S20M` at 1019, share exactly one reference position and give up 61 bases each. Move the mate one base right and they share none — the insert is one base longer, so the answer must be 60, not 0. The extrapolation is the analytic continuation of the query-space count, which reduces to exactly that expression as the shared span shrinks to a single position.

It is also safe against the defect this PR fixes. Neither alignment places a base between the two, so neither can place an indel between them, and the reference/query conflation cannot arise. An indel in the *sample* that falls in the gap is invisible to both reads and shifts the estimate by its own length in either direction, but the result is capped by the record's own soft clipping, so no aligned base is ever removed on the strength of it. #842 added this branch for exactly this reason: without it, "the `numBasesExtendingPastMate` function always returned 0 [...] generally resulting in reads now failing when they used to pass" in `FilterConsensusReads`.

A record whose 3' end points *away* from a mate it does not overlap is a different case and is not clipped at all — that describes an outward-facing template rather than read-through. The mate's alignment is supplied by the caller, so it can contradict what the record's own flags say about the pair, and measuring such a record against the landscape like any other would count every one of its query bases as past the mate and take its aligned ones.

### Counting

Counting is done by walking the cigar directly rather than by calling `readPosAtRefPos`, which makes the deletion case explicit instead of relying on the `returnLastBaseIfDeleted` flag: a position inside a deletion has no query base, and every query base past the gap is counted. Soft-clipped bases still count as insert bases, preserving the less-aggressive behavior introduced in #1026; hard-clipped bases do not, since they are not present in the record.

Two smaller correctness fixes in the same functions, both needed to make the above land correctly:

- `clipExtendingPastMateEnds` now computes both clipping amounts *before* clipping either read. Previously the mate was measured against an alignment the first call had already shortened (or un-mapped), which made the result order-dependent.
- Existing hard-clipping at the 3' end is now added back in when calling `clipEndOfRead` / `clipStartOfRead`, which take the *total* clipping desired including what is already there. Without this, requested clipping is silently absorbed by pre-existing hard-clipping — reachable in `ClipBam`, which applies fixed 3' clipping before this step. This mirrors what `clipOverlappingReads` already does.

## API change

`numBasesExtendingPastMate` now takes the mate's alignment start and cigar instead of its un-soft-clipped start and end. The old parameters cannot express the mate's indel structure, which the query-space count requires. Both in-tree callers already have the mate's cigar available (`UmiConsensusCaller` via the `MC` tag, which it already ensures is set).

## Notes

The `ClipOverlapFailed` branch in `CodecConsensusCaller` carries a `// TODO remove this branch once SamRecordClipper has been updated` comment referring to this bug. It is left in place: `computeConsensusLength` has its own reference/query-space assumptions and can still return a length shorter than one of the consensus reads, so the branch has not been proven dead. Removing it is a separate change.

## Tests

All new tests fail on `main` and pass here, except where noted:

- `clip back to the mate's soft-clipped end when neither read contains an indel` — control; passes before and after, pins that the indel-free path (including soft-clip-aware trimming from #1026) is unchanged.
- `clip using query distances when a deletion falls at the mate's un-soft-clipped end` — the over-clipping case. Was `(127, 0)` with the record un-mapped; now `(2, 0)`.
- `clip the mate even when the record's deletion falls at the mate's un-soft-clipped end` — pins the knock-on effect on the mate. Was `(127, 0)` with the mate untouched; now `(2, 2)`.
- `clip using query distances when the record contains an insertion before the mate's end` — the example from the issue, which produces exactly the `70M10I20M50H` the issue asks for. Was `(0, 0)`.
- `clip past the mate's end when the read has already been hard-clipped at its 3' end` — pins the hard-clipping fix. Was `(20, 20)`; now `(30, 30)`.
- `count query bases, not reference bases, when an indel is present near the end of an alignment` — unit coverage of `numBasesExtendingPastMate` itself for both the deletion and insertion cases.
- `clip the read-through of a pair whose alignments share no reference position` — the continuity pair above (61 with one shared position, 60 with none), on both strands, plus a hard-clip assertion pinning that neither record loses an aligned base. Passes before and after; it fails against a version of this PR that answers 0 when the alignments are disjoint.
- `not clip a record lying entirely past the mate alignment it is given` — the outward-facing guard.

All existing `SamRecordClipper` tests are unchanged and still pass.

## Verification beyond the unit tests

The new counts were checked out-of-tree against an independent oracle built from htsjdk's per-base `getReferencePositionAtReadPosition` — a different code path from the cigar walk — over a randomized sweep of cigars:

- 3,911 random FR pairs (1,995 of them containing an indel): `numBasesExtendingPastMate` matched the oracle on every one.
- 3,925 random indel-free FR pairs: matched the **pre-fix** formula exactly, confirming the change is a no-op when neither alignment has an indel near the relevant end.
- 2,863 overlapping FR pairs, in both soft- and hard-clip mode: neither read is ever un-mapped, at least the computed excess is clipped from each 3' end, and a second pass clips nothing further.
- Across 29,246 indel-bearing evaluations the pre-fix formula disagreed with the fix 1,338 times (4.6%), by a mean of 6.7 query bases and a maximum of 92 (an entire read). 92 of those were whole-read over-clips — cases where the old code returned a clip at least as long as the read and would therefore have un-mapped it.

A second sweep, over a set of 20,000 randomly generated FR pairs (100 base reads, leading and trailing soft clips 0-80, mate offsets −160 to +260, half carrying an indel) that deliberately spans overlapping, touching and **disjoint** geometries rather than only overlapping ones:

- 40,000 evaluations, both strands. **662 differ from the pre-fix formula, and every one of them is on a pair carrying an indel** — all 9,210 indel-free evaluations are unchanged. The largest disagreement is 99 bases, a whole-read over-clip the pre-fix formula produced from a deletion at the synthesized boundary.
- The same 20,000 pairs were run against [fulcrumgenomics/fgumi#759](https://github.com/fulcrumgenomics/fgumi/pull/759), an independent rework of the same arithmetic in a different language, with **zero disagreements**.

That second sweep is what caught the disjoint case. The earlier randomized sweep above generated only overlapping pairs, so it could not see it: an intermediate revision of this PR answered 0 for disjoint pairs, which differed from the pre-fix formula on 28 indel-free evaluations in the new set.

One pre-existing behavior worth noting, unchanged here: `clip()` never splits an insertion, so if the clip boundary falls inside one the whole insertion is clipped. The applied clipping can therefore exceed the computed excess by up to (insertion length − 1).

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
+++ b/src/main/scala/com/fulcrumgenomics/bam/SamRecordClipper.scala
@@ -324,27 +324,150 @@ class SamRecordClipper(val mode: ClippingMode, val autoClipAttributes: Boolean) 
     }
   }
 
-  /** Returns the number of bases extending past the mate end for FR pairs including any soft-clipped bases, zero otherwise.
+  /** Returns the number of query bases at the 3' end of a record that extend past the 3' end of its mate, for FR pairs,
+    * and zero otherwise.
+    *
+    * The two reads of an FR pair sequence the same insert from opposite ends, so beyond the last reference position
+    * that both alignments cover they must have sequenced the same number of insert bases.  Any query bases a record has
+    * beyond that count are read-through past the end of the insert (i.e. into the adapter).  Both counts are taken in
+    * _query_ space so that an indel near the end of either alignment does not skew the comparison; the shared reference
+    * position is used only as the landmark from which to count.
+    *
+    * Soft-clipped bases are counted, since they are query bases that may still belong to the insert.  Hard-clipped
+    * bases are not, since they are not present in the record.
+    *
+    * When the two alignments share no reference position there is no landmark to count from, and the count falls back
+    * to `numBasesExtendingPastDisjointMate` below.
     *
     * @param rec the record to examine
-    * @param mateUnSoftClippedStart the smallest mapped genomic coordinate considering *ONLY* soft-clipping
-    * @param mateUnSoftClippedEnd the largest mapped genomic coordinate considering *ONLY* soft-clipping
+    * @param mateStart the alignment start of the mate
+    * @param mateCigar the alignment cigar of the mate
+    * @return the number of query bases to clip from the 3' end of the record (in sequencing order)
     */
-  def numBasesExtendingPastMate(rec: SamRecord, mateUnSoftClippedStart: Int, mateUnSoftClippedEnd: Int): Int = {
-    if (!rec.isFrPair) 0 else rec.positiveStrand match {
-      case true if rec.end >= mateUnSoftClippedEnd =>
-        // positive strand record is aligned to/past the mate alignment end: count any bases aligned/soft-clipped after
-        Math.max(0, rec.length - rec.readPosAtRefPos(pos=mateUnSoftClippedEnd, returnLastBaseIfDeleted=false))
-      case true =>
-        // positive strand record alignment ends before the mate alignment end: remove any excess soft-clipped reads
-        Math.max(0, rec.cigar.trailingSoftClippedBases - (mateUnSoftClippedEnd - rec.end))
-      case false if rec.start > mateUnSoftClippedStart =>
-        // negative strand record alignment starts after the mate alignment start: remove any excess soft-clipped reads
-        Math.max(0, rec.cigar.leadingSoftClippedBases - (rec.start - mateUnSoftClippedStart))
-      case false =>
-        // negative strand record alignment starts at or before the mate start: count up to and including one base before
-        Math.max(0, rec.readPosAtRefPos(pos=mateUnSoftClippedStart, returnLastBaseIfDeleted=false) - 1)
+  def numBasesExtendingPastMate(rec: SamRecord, mateStart: Int, mateCigar: Cigar): Int = {
+    if (!rec.isFrPair) 0 else {
+      val mateEnd      = mateStart + mateCigar.lengthOnTarget - 1
+      val overlapStart = Math.max(rec.start, mateStart)
+      val overlapEnd   = Math.min(rec.end, mateEnd)
+      if (overlapEnd < overlapStart) numBasesExtendingPastDisjointMate(rec, mateStart, mateCigar, mateEnd)
+      else if (rec.positiveStrand) {
+        // The 3' end of a positive strand record is its right-hand end, so count from the last shared position.
+        val recBases  = numQueryBasesAfter(start=rec.start, cigar=rec.cigar, refPos=overlapEnd)
+        val mateBases = numQueryBasesAfter(start=mateStart, cigar=mateCigar, refPos=overlapEnd)
+        Math.max(0, recBases - mateBases)
+      }
+      else {
+        // The 3' end of a negative strand record is its left-hand end, so count from the first shared position.
+        val recBases  = numQueryBasesBefore(start=rec.start, cigar=rec.cigar, refPos=overlapStart)
+        val mateBases = numQueryBasesBefore(start=mateStart, cigar=mateCigar, refPos=overlapStart)
+        Math.max(0, recBases - mateBases)
+      }
     }
+  }
+
+  /** Returns the number of query bases at the 3' end of a record that extend past the 3' end of its mate, when the two
+    * alignments cover no reference position in common.
+    *
+    * With no shared position there is no landmark to count query bases from, so the record's own soft-clipped tail is
+    * projected into reference space at one query base per reference base and compared against the mate's
+    * un-soft-clipped far end — the estimate this function has applied since #842.  Two reads can absolutely have read
+    * through into each other's adapter here: both alignments lie within the insert, so a short insert read from both
+    * ends leaves them disjoint whenever each alignment stops short of the other's, which soft clipping at a variant or
+    * a mis-called base is enough to produce.  The extrapolation is exact whenever the region between the two
+    * alignments is ungapped, and it is continuous with the query-space count above, which reduces to exactly this
+    * expression as the shared span shrinks to a single reference position.
+    *
+    * The reference/query conflation that the query-space count exists to avoid cannot bite here: neither alignment
+    * places a base between the two, so neither can place an indel between them either.  An indel in the sample that
+    * falls in that gap is invisible to both reads and shifts the estimate by its own length, in either direction; the
+    * result is capped by the record's soft clipping regardless, so no aligned base is ever removed on the strength of
+    * it.
+    *
+    * A record whose 3' end points away from its mate — a positive strand record lying entirely to the right of its
+    * mate, or a negative strand record entirely to its left — describes an outward-facing template rather than
+    * read-through, and nothing is clipped.
+    *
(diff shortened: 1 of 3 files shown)


nh13:
Force-pushed an amendment. The query-space count itself is unchanged; what changed is the disjoint-alignment branch.

The previous revision answered **0** when the two alignments shared no reference position, on the reasoning that without an overlap neither read can have read through into the other's adapter. That reasoning is wrong. Both alignments lie *within* the insert, so a pair that read through can still leave them disjoint — it happens whenever each alignment stops short of the other's, which soft clipping at a variant or a mis-called base is enough to produce. Two 100-base reads over a 39-base insert, `20M80S` at 1000 against `80S20M` at 1019, share exactly one reference position and give up 61 bases each; moving the mate one base right removes the shared position and the correct answer is 60, not 0. The soft-clip extrapolation is the analytic continuation of the query-space count, not a separate heuristic, and #842 added it deliberately for this reason: without it, "the `numBasesExtendingPastMate` function always returned 0 [...] generally resulting in reads now failing when they used to pass" in `FilterConsensusReads`.

The randomized sweep in the original description could not see this, because it generated only overlapping pairs. On a new set of 20,000 pairs that spans overlapping, touching and disjoint geometries, the 0-answering revision differed from the pre-fix formula on **28 indel-free evaluations** — a fix for a reference/query conflation must be inert on indel-free input, and it was not. The amended version differs on 662 of 40,000 evaluations, every one of them on a pair carrying an indel, with all 9,210 indel-free evaluations unchanged.

The conflation this PR fixes provably cannot arise in that branch: neither alignment places a base between the two, so neither can place an indel between them. An indel in the *sample* that falls in the gap is invisible to both reads and moves the estimate by its own length, but the result is capped by the record's own soft clipping, so no aligned base is removed on the strength of it.

Two things are new in the diff:

- `numBasesExtendingPastDisjointMate`, carrying the #842 extrapolation, plus an explicit `0` for a record whose 3' end points *away* from a mate it does not overlap — an outward-facing template rather than read-through. The mate's alignment is a parameter, so it can contradict what the record's own flags say about the pair.
- Two tests: the 61/60 continuity pair (both strands, with a hard-clip assertion pinning that no aligned base is taken), and the outward-facing guard.

The same 20,000 pairs were also run against [fulcrumgenomics/fgumi#759](https://github.com/fulcrumgenomics/fgumi/pull/759), an independent rework of this arithmetic in Rust that had reached the opposite conclusion on this one case. The two implementations now agree on all 40,000 evaluations.

#1174 has been rebased onto the amended commit; its diff is unchanged and the full suite is green on both (1622 and 1625 tests).

PR #1176, opened 2026-08-19
feat: preserve raw alignment-family counts in consensus reads

fitzlane86:
Addresses #1177 

Consensus formation discards minority alignment families before emitting a consensus read, so cD/aD/bD describe only the family that survived selection. That hides a real signal: a consensus built from a raw bundle with substantial alignment disagreement is indistinguishable from one built from a clean bundle of the same surviving size. Such cases are a source of spurious variant calls and require a return to the pre-consensus BAM files to debug. 

This PR adds six per-read tags recording the size of the input family as it was grouped, before any consensus-family selection or downsampling: `cT`/`cR` (raw template/record counts for the whole family) and `aT`/`bT`/`aR`/`bR` (per-strand duplex equivalents). Counts are computed only for families that go on to form a consensus, to avoid the cost on the (majority) rejected path. `CallCodecConsensusReads` deliberately does not set these tags, since CODEC's protocol puts both strands of a molecule in a single read pair and the raw template unit these tags assume has no settled translation there.


nh13:
Hi @fitzlane86 — really like this change. The silent 5→3 family collapse is a real gap: a consensus built from a noisy family looks identical to one from a clean family of the same final size, and I can see how that would bite you when a spurious variant sends you back to the pre-consensus BAM to figure out what happened. Tracking the pre-selection sizes closes that off nicely.

One design question on the per-strand tags (`aT`/`bT`/`aR`/`bR`). As written they're keyed on the `/A` vs `/B` MI suffix, while the sibling `aD`/`bD` depths are keyed on the `a`/`b` output slot — whichever single-strand consensus survives gets promoted into it. That means on a read where the lone surviving strand is the `/B` one, `aD` describes `/B` while `aT` still describes `/A`, so `aT − aD` isn't meaningful (your docs and tests call this out clearly, so I know it's deliberate).

I'd like to propose keying `aT`/`bT`/`aR`/`bR` on the same `a`/`b` slot as `aD`/`bD` instead. Then all the `a*` tags describe one strand, `aT ≥ aD` always holds, and `aT − aD` becomes exactly "raw reads dropped from that strand by family selection + downsampling" — the per-strand version of what `cT − cD` gives for the whole family. Nothing is lost: a strand that formed no consensus at all still shows up in `cT − aD − bD`. It also lets the MI-vs-slot caveat come out of the docs entirely.

I've prototyped it (raw counts carried on `DuplexConsensusRead`, attached in `duplexConsensus` where the slot is decided, written next to `aD` in `createSamRecord`; CODEC still emits none, guarded). Full `umi` suite is green. Happy to push it to this branch if you're on board, or open it as a separate PR against your branch / just hand you the diff — whatever you prefer. Let me know.

fitzlane86:
Hi @nh13. Appreciate the prompt review! I went back and forth on that particular decision, so happy to accept your version of it. As you say, no information is lost. Feel free to push those changes to this branch.

